PR #116, opened 2014-04-02
blazer_soi: blazer driver, serial port over tcp/ip

alezz:
This is a version of blazer that allows the use of a ethernet to serial device, also called YPort.
Such a device allows a serial port to be connected to an existing ethernet network, and uses tcp to access the serial port. An example of this device is provided by Ethersex Project (http://www.ethersex.de); this driver extension has been tested with it.

### DIFF
--- a/drivers/Makefile.am
+++ b/drivers/Makefile.am
@@ -134,6 +134,10 @@ victronups_SOURCES = victronups.c
 riello_ser_SOURCES = riello.c riello_ser.c
 riello_ser_LDADD = $(LDADD) -lm
 
+#alezz's blazer_soi
+blazer_soi_SOURCES = blazer.c blazer_soi.c
+blazer_soi_LDADD = $(LDADD) -lm
+
 # non-serial drivers: these use custom LDADD and/or CFLAGS
 
 # dummy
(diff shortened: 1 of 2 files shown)


jimklimov:
Hi all, given that this adds a standalone driver, there seems to at least be no code-wise conflict with this PR despite its age. But the points raised in review earlier about possibly sharing and unifying code with other NUT parts, and parsing "port" commonly as an "IP:PortNum" value in particular, make sense.

So I need some team consensus: is this good to merge "as is" to add a new driver and refine later when someone is annoyed (e.g. also regarding new `nutdrv_*` naming standard) or should issues raised above be resolved before or shortly after a merge? CC @clepple @aquette @zykh @alezz 

Also seems I can not update the source branch of this PR, to merge the current NUT master into it and have Travis CI take a bite.

clepple:
@jimklimov As you mentioned, the original PR adds a standalone driver, but it is resurrecting the blazer_* code that we were trying not to maintain in parallel with @zykh's nutdrv_qx codebase, and it does not have a man page to describe the differences between it and the original blazer_ser driver.

I think if someone is truly interested in supporting serial-to-TCP adapters, they need to be willing to start from more recent code (such as @zykh's netport branch), and then test against actual hardware. I have neither the time nor the hardware to do it myself, so maybe take my comments with a grain of salt.

PR #121, opened 2014-04-17
Add support for 0d9f:0001 (USB HID, Powercom)

zeezooz:
Tested with BNT-1000AР (old model). Confirmed values: battery.charge, input.frequency (0 on battery, 50 on AC), input.voltage, output.voltage, ups.load and likely output.frequency (always 50); statuses: online, cal (ups test), dischrg, trim.

Key changes:
* Read exactly 8 bytes on interrupt.
* Get values of input objects. Getting both will be better since there are some feature objects during init.
* Avoid comm_driver->get_report since it often returns corrupt data (0 values).

### DIFF
--- a/data/driver.list.in
+++ b/data/driver.list.in
@@ -760,6 +760,7 @@
 "Powercom"	"ups"	"4"	"(various)"	"USB (<= 2009 models, product id: 0002)"	"powercom (requires 'usbserial' kernel module)"
 "Powercom"	"ups"	"5"	"(various)"	"USB (2009 models, product id: 00a?)"	"usbhid-ups (experimental)"
 "Powercom"	"ups"	"5"	"BNT-xxxAP"	"USB (product id: 0004)"	"usbhid-ups (experimental)"
+"Powercom"	"ups"	"1"	"BNT-xxxAP"	"USB (product id: 0001)"	"usbhid-ups (experimental)"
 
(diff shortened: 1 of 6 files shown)


clepple:
@aquette have you had a chance to verify that this still works on other models? (I didn't know this was scheduled for 2.7.3) It introduces two warnings, although I think we can change the new variables to `unsigned int` to fix that:

http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-gcc/builds/293/steps/compile/logs/warnings%20%282%29

../../drivers/libhid.c:486:80: warning: signed and unsigned type in conditional expression [-Wsign-compare]
../../drivers/libhid.c:486:80: warning: signed and unsigned type in conditional expression [-Wsign-compare]

aquette:
@clepple you're right, I forgot to mention that I made some quick non-regression testing on Eaton units (5SC, Protection Station and Ellipse ECO). No issues found, as expected considering that changes were isolated.

clepple:
@zeezooz can you please check that the following pull request still works on your hardware? https://github.com/networkupstools/nut/pull/213

I had to change around some of the logic when reading from the interrupt pipe for Tripp Lite hardware, and I don't have access to Powercom hardware to test.

zeezooz:
@clepple Built from the branch `usbhid_ups_input_vs_feature` still works.

clepple:
Thanks for testing!

PR #158, opened 2014-10-10
Powerware BCMXCP advanced features

gavrilov-i:
Added more variables (battery packs count, sleep mode, beeper control etc.).
Added function which request list of settable variables.
Added indication of ABM status for internal batteries.
Added outlet control (immidiate shutdown, turn on)/
outlet.n.shutdown.return now wait shutdown_delay time, not outlet.n.delay.shutdown because not all firmware accept it, so shutdown time will be -1 which is wrong.

Tested on pw 5125.


gavrilov-i:
Successfully tested on Eaton 9130 1000VA, with RS232 connection.

clepple:
@aquette I think this branch should be in 2.7.3, but I would like to squash the incremental commits (rebase the branch) before merging. Do you think you might be able to test this on some other PW units?

gavrilov-i:
>Do you think you might be able to test this on some other PW units?

All I could get is 5115,5125,9125,9130. On all UPS driver work as expected/

aquette:
@clepple : sorry for the lag on this point, I'm checking what I have underhand.
@gavrilov-i : would be great if you could also generate some .dev for the DDL. That would also help to check for non-regression, while supporting other objectives.
A script is available to help: https://github.com/networkupstools/nut/commit/ca9f6019a7d05ea23762610c55c935db50ee0c88 
thx and cheers

gavrilov-i:
Done:
PW5125 http://pastebin.com/J3GAU3QS 
PW9130 http://pastebin.com/yjQhnZRk

clepple:
@gavrilov-i thanks for the DDL info!

@aquette I squashed the branch into a single commit (00b8c83) but it seems as though your new Eaton energysave variable names are slightly different than those in this patch. Thoughts before we merge this?

aquette:
@clepple : on energysave, indeed, just saw that while reviewing the PR.
I've been hesitating to use the exact same varname as @gavrilov-i , since it was tied to the ups.load and not battery load (which may be confusing...). I need to consider some upcoming similar features, prior to being able to finally decide.

aquette:
@clepple & @gavrilov-i : beside from the cmdvartab changes, there is also a need to update docs/nut-names.txt, as per the new variables additions

aquette:
@clepple & @gavrilov-i : ok, so after quickly checking, I see no issue in aligning my eaton-energysave branch varnames with @gavrilov-i . However, I want to ensure that the future new ones (related to ups.realpower, not ups.load) is aligned and not confusing.

Additional EnergySaving data, beside from the existing ones (HID data provided for disambiguation):

| HID data | small desc | proposed NUT variable |
|-------------|--------------|-------------------------------|
| UPS.PowerConverter.Input[3].ConfigActivePower | threshold in Watts (related to ups.realpower) | battery.energysave.realpower? |
| UPS.PowerConverter.Input[3].ConfigPercentLoad | threshold in percent (related to ups.load) | battery.energysave.load |
| UPS.PowerConverter.Input[3].EnergySaving.ShutdownTimer | ... | battery.energysave.delay |

aquette:
pushed by @clepple  into branch pull_158_bcmxcp_rebased.
Amended by @aquette for the missing nut-names.txt updates
Good to merged (after a last rebase on master

aquette:
(comment replayed from the NUT developer mailing list)
part of the 2.7.3 release, we are about to merge the below branch / PR for the bcmxcp drivers family:
https://github.com/networkupstools/nut/pull/158
https://github.com/networkupstools/nut/tree/pull_158_bcmxcp_rebased

It has introduced the battery.charger.status variable, which is something I wanted to implement since around 2007.

However, while considering a similar implementation for usbhid-ups, and thinking a bit more broadly, this new variable tends to collide with the CHRG / DISCHRG status bits from ups.status, and adding 2 more (floating and resting).

There, we have 2 options:
1) Keep the legacy CHRG / DISCHRG status bits for ups.status, along with the complementary ones for battery.charger.status. And advocate (document) for the use of / switch to the latter, that is more suitable for publishing such information. All that with an impact on all the NUT driver, and a transition period to address that cleanly.
2) Add the 2 other status bits (FLOATING and RESTING) for ups.status, still along with the same new ones for battery.charger.status

I'm personally more in favor of (1), but the long history of NUT makes that (2) would probably be more suitable. An hybrid approach could also be: apply (2) (new status bits), and apply the transition of (1).

Feedback and comments warmly welcome.

aquette:
As per agreement on the [mailing list](http://thread.gmane.org/gmane.comp.monitoring.nut.devel/6904), choice (2) was selected.

aquette:
Matching branch pull_158_bcmxcp_rebased merged into master.
Closing this PR.
@gavrilov-i thanks for your much appreciated contribution :)

gavrilov-i:
I'm glad I was able to participate and help. =)

PR #179, opened 2015-01-11
Undo ${systemdsystemunitdir} mangling.

ravenexp:
Running sed 's/\/lib/\${libdir}/' destroys any ${systemdsystemunitdir}
values that don't start with '/lib' e.g. '/usr/lib64/systemd/system'
becomes '/usr/usr/lib6464/systemd/system'.

If a local installation prefix is needed use appropriately prefixed
--with-systemdsystemunitdir='' parameter instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1063,7 +1063,6 @@ dnl files will try to get intalled to the actual system directories
 if test -n "${systemdsystemunitdir}"; then
 	systemdsystemshutdowndir="${libdir}/systemd/system-shutdown"
 	AC_MSG_RESULT(using ${systemdsystemunitdir})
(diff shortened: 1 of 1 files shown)


clepple:
@bigon I am inclined to merge this, since it removes some "magic". Any objections? @aquette had a comment in there about 'make distcheck', but we can also add another line to DISTCHECK_CONFIGURE_FLAGS in Makefile.am to compensate: https://github.com/networkupstools/nut/blob/master/Makefile.am#L23

bigon:
Good for me

clepple:
@ravenexp thanks!

PR #194, opened 2015-03-14
nut-voltronic: minimal driver cleanup

nickma82:
Do you like a little bit of cleanup/refactoring, than I'm gonna invest a bit of work?

### DIFF
--- a/drivers/nutdrv_qx.h
+++ b/drivers/nutdrv_qx.h
@@ -116,6 +116,7 @@ typedef struct item_t {
 
 #define MAXTRIES		3	/* Max number of retries */
 
+#ifdef TESTING
 /* Testing struct */
 typedef struct {
 	const char	*cmd;			/* Command to match */
@@ -125,6 +126,7 @@ typedef struct {
 						 * - if set to -1 -> auto calculate answer length (treat 'answer' as a null-terminated string)
 						 * - otherwise -> use the provided length (if reasonable) and preserve inner '\0's (treat 'answer' as a sequence of bytes till the item_t->preprocess_answer() function gets called) */
 } testing_t;
+#endif /* TESTING */
 
 /* Subdriver interface */
 typedef struct {
--- a/drivers/nutdrv_qx_voltronic.c
+++ b/drivers/nutdrv_qx_voltronic.c
@@ -131,28 +131,19 @@ static info_rw_t	voltronic_r_batt_low[] = {
 /* Preprocess range value for battery low voltage */
 static int	voltronic_batt_low(char *value, size_t len)
 {
-	int		val = strtol(value, NULL, 10);
-	const char	*ovn = dstate_getinfo("output.voltage.nominal"),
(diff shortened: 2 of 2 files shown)


zykh:
ahem, sorry for my delay and thanks for your interest.
Quick notes (just my opinion):
* nickma82@1a174f7b3c74a92cd1595786e7dbb68398c37936:
 * ungrouping of variable definitions: I prefer and find more readable the currently used style;
 * #defined returns: actually most of the functions return 0 or -1/1 as a boolean thing: I think that only if there were a wider range of possible returned values, then #defined returns would make the code (and the 'API') more readable and usable (and probably one should also take care of all other functions to align them all with the same standard) - plus it doesn't make sense to hide the returned value behind a #define if you don't change all occurences of calls to that function to check against it;
 * added variables that are used only once (wattsNominal): I can't see the advantage over the status quo;
* nickma82@b1a076719740f26860b62a3c89edcec59a2773f1 &
  nickma82@e0df185b6130f2e29132a8ee25598e905bfb6f14:
  rather than simplify, it makes things harder to read and to use (and - as a consequence of that? - there's a comparison between a char and a char*);
* Indentation: nutdrv_qx* files are optimized for 8 chars tabs;
* Most importantly: when something is changed in nutdrv_qx.{c,h} *all* subdrivers (and nutdrv_qx 'API' reference in docs) **must** be updated accordingly.

Having said that, I hope I didn't dampen your enthusiasm and that you're still with us.

nickma82:
Hi @zykh,
> Having said that, I hope I didn't dampen your enthusiasm and that you're still with us.

That's ok, its for me too to spot which changes you guys like/or not...

> added variables that are used only once (wattsNominal): I can't see the advantage over the status quo

IMHO the advantage is that the content (even if used once) is already parsed and easyer readable for one who wants to understand the code. From an optimisation point of few it doesnt matter at all, compiler optimisation is going to get rid of it.

I agree with most of your other concerns. I'm going to change the pull request.
Thanks, Nick

clepple:
@nickma82 Stepping back for a minute, what is the overall goal here?

A lot of your recommendations are valid (e.g. enums over #defines over magic numbers), but with an established project like NUT, even these sorts of well-intentioned changes have the potential to introduce mistakes. If nothing else, inconsistency between drivers can be distracting when trying to address problems across the whole codebase.

If I might make an observation about changes to NUT code: typically, the person who writes or rewrites the majority of a driver gets to determine the coding style. @zykh overhauled the blazer_* driver after that driver's author retired from NUT. The apcsmart driver has similarly been rewritten several times over the years. If you have a suggestion for a similarly large code change to NUT, a better process would be to propose the high-level idea on the nut-upsdev mailing list, and then start in on the code.

nickma82:
> ven these sorts of well-intentioned changes have the potential to introduce mistakes

that's why I didn't change anything within apis. Or types.

My goal is to implement new features in the voltronic driver. I'm going to leave the rest, sorry for your time.

PR #207, opened 2015-05-06
Docu and fixes

nickma82:
added some inline documentation and fixes, please see commits

### DIFF
--- a/drivers/libhid.c
+++ b/drivers/libhid.c
@@ -429,7 +429,7 @@ char *HIDGetItemString(hid_dev_handle_t udev, const char *hidpath, char *buf, si
  */
 int HIDSetDataValue(hid_dev_handle_t udev, HIDData_t *hiddata, double Value)
 {
-	int	i, r;
+	int	r;
 	long	hValue;
 
 	if (hiddata == NULL) {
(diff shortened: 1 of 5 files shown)


drivers/libhid.c
```diff
@@ -456,9 +456,7 @@ int HIDSetDataValue(hid_dev_handle_t udev, HIDData_t *hiddata, double Value)
 	}
 
 	/* flush the report buffer (data may have changed) */
-	for (i=0; i<256; i++) {
-		reportbuf->ts[i] = 0;
-	}
+	memset(reportbuf->ts, '\0', sizeof(reportbuf->ts)/sizeof(reportbuf->ts[0]));
```
clepple: Dividing by `sizeof(...ts[0])` doesn't seem right - each element of `ts` is a `time_t`, and the len parameter of memset() is in bytes (despite the value being an int).

nickma82: You are absolutely right, I fixed that.

clepple:
@zykh, any thoughts on the rest of this pull request?

zykh:
I merged most of this pull request (apart from what follows, I just fixed a couple of typos here and there), a few notes:
- upsdebug* functions take, at the moment, an int level (not syslog #defined priorities), see #211;
- comments are, at the moment, just plain text, see #212;
- when you change source code, you should bump versions;
- if you feel the need to update comments, probably docs are to update, too - just keep them aligned;
- (ok, nitpicking here) I believe imperative is the tense people uses in commit messages.

For the moment, I left out the unittest-related things.

PR #213, opened 2015-06-12
usbhid-ups input vs feature

clepple:
See http://thread.gmane.org/gmane.comp.monitoring.nut.devel/6961 for details.

Basically, `FindObject_with_Path()` was returning NULL on events from the interrupt pipe, and `find_hid_info()` was matching the first entry in `tripplite_hid2nut[]` (which was not present on the Tripp Lite UPS in question).

This patch allows the interrupt code to match either `TYPE_INPUT` or `TYPE_FEATURE` items.

### DIFF
--- a/drivers/tripplite-hid.c
+++ b/drivers/tripplite-hid.c
@@ -29,7 +29,7 @@
 #include "tripplite-hid.h"
 #include "usb-common.h"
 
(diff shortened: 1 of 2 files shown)


aquette:
seems good to me for merging.
beware that I've merged a PR that will create a minor conflict (at least on DRIVER_VERSION), so you'll need to rebase first.

PR #215, opened 2015-07-13
Update driver.list.in

Thermionix:
As per: https://bugs.freenas.org/issues/9256
Add "apcupsd-ups" to driver.list as has been confirmed working by several people

### DIFF
--- a/data/driver.list.in
+++ b/data/driver.list.in
@@ -74,6 +74,7 @@
 "APC"	"pdu"	"1"	"AP9210"	"8 outlets"	"powerman-pdu (experimental)"
 "APC"	"pdu"	"1"	"AP79xx"	"8 to 24 outlets"	"powerman-pdu (experimental)"
 "APC"	"pdu"	"3"	"(various)"	"SNMP monitoring card"	"snmp-ups (experimental)"
+"APC" "ups" "3" "Connect to APCUPSD daemon"  ""  "apcupsd-ups" 
 
 "Aphel"	"pdu"	"3"	"various PDU"	""	"snmp-ups (experimental)"
 


clepple:
@Thermionix thanks!

@aquette and @zykh: I have no strong opinions either way (so I am inclined to merge this as-is), but did you have any thoughts on how to handle this, plus the other "(various)" entries in the HCL?

Thermionix:
hmm, maybe something along the lines of;

support-level : 3
device-type : ups
manufacturer : apc
model : APC UPS Daemon
connection : network

zykh:
(those spaces...)

By the way, to keep it along the same lines of other similar entries, shouldn't it be something like the following?

````
"APC" "ups" "3" "(various)" "APCUPSD-controlled devices" "apcupsd-ups"
````
(with tabs..)

clepple:
@zykh I agree with the "(various)" part, just wasn't sure if we wanted to pick a new way of representing that in the HCL.

Also, while I like the idea of tab-separated entries in the driver.list file, I think we should explicitly check for that as part of the build process (probably when we generate the HTML/JS files for the HCL page). Checking for tabs vs spaces is not a job that a human should do :-)

aquette:
Fully agree with you @clepple , and that's something we've been discussing for quite some time already...

Storing memories before they go away with new years eve:
I've pushed a proposal in the drvlist-format branch:
https://github.com/networkupstools/nut/commit/f77974cc13106116b3293ff5e493881ab555c1cd

Tested with some tab replaced by single and multiple spaces, along with added EOL spaces.

zykh:
Added to HCL (39bb64f), using the formula of comment https://github.com/networkupstools/nut/pull/215#issuecomment-122718306

PR #220, opened 2015-07-25
tripplitesu: Fix initialization after UART power cycle when tripplite firmware is buggy

nkaminski:
With some Tripplite SU1000RT2U (and possibly more) UPS's, a firmware bug causes a malformed response to the very first command that is sent after the serial port is opened following a warm or cold boot of the system. My theory is that this related to either the RS232 data lines or handshaking lines being pulled high once the server's UART is powered however I have not determined precisely if this is related to the data lines or the handshaking lines. However, I have been able to consistently reproduce the issue where the tripplitesu driver fails to start on the first attempt after a cold/warm boot across 3 different machines and 2 SU1000RT2U UPS's. To workaround this, the initial enumeration is repeated a 2nd time after 300ms(to allow all garbage data to arrive) if the first attempt fails, which allows the driver to consistently startup successfully on the 1st attempt.

### DIFF
--- a/drivers/tripplitesu.c
+++ b/drivers/tripplitesu.c
@@ -125,7 +125,7 @@
 #include "serial.h"
 
 #define DRIVER_NAME		"Tripp Lite SmartOnline driver"
-#define DRIVER_VERSION	"0.03"
+#define DRIVER_VERSION	"0.04"
 
 /* driver description structure */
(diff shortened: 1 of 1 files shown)


clepple:
Just reformatted the commit message and fixed the C++-style comment. Thanks!

PR #221, opened 2015-07-25
Spelling fixes

tas50:
Spelling fixes and capitalization of SUSE


clepple:
@tas50 Thanks! I was curious as to why `make spellcheck` didn't catch `preferred` but it seems that it only covers `docs/`.

Issue #211, opened 2015-06-04
Abstract/align debug levels


zykh:
At the moment, we have:
* upslog* functions [1], which take as first argument one of syslog #defined priorities [2];
* upsdebug* functions [3], which take as first argument an arbitrary debug level.

That arbitrariness means we have, mainly in drivers, but in other places too, wildly differentiated debug levels.
As long as you are familiar with/remember the source, all's well, but when you venture in someone else's code or just cannot check source, this can become a problem.
Plus, the upslog* vs upsdebug* discrepancy seems to cause some confusion in newcomers.

So.. what about abstracting and(/or, at least,) aligning debug levels to something common (ok, that's not gonna be easy)?
e.g.:
* `DEBUG__VANILLA`, every kid loves it [level -> 1; e.g. can't find something];
* `DEBUG__CINNAMOM`, for that christmas flavour [level -> 2; e.g. can't do something else];
* ...
* `DEBUG__MILK_AND_MINT_SYRUP`, not to everyone's taste (how?? how is that even possible?) [level -> 7; e.g. developer errors, better sweep them under the carpet];
* ...
* `DEBUG__PEPERONATA_FOR_BREAKFAST`, you better not be here [level -> 15; e.g. every single thing that happens].

(..and I need to eat something)

--
1. https://github.com/networkupstools/nut/blob/master/common/common.c#L387:L405
2. https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sys/syslog.h#l51
3. https://github.com/networkupstools/nut/blob/master/common/common.c#L407:L515

clepple:
@zykh hopefully you've had a chance to eat since you first posted this ;-)

Seriously, though - I think the general idea is good, but if we are going to make sweeping changes, it might be good to standardize on some debug codes that represent what we are trying to debug. For instance, a lot of our serial drivers deal with issues at the byte level, as well as at the packet level. Then, there is the notion of logging normal events (on battery, on line) at the driver level to get better timing resolution.

I think the upslog* vs upsdebug* issue can be addressed with documentation, but we could also provide some recommendations on the various debug levels for new drivers.

PR #223, opened 2015-07-28
Add/consolidate string-related functions

zykh:
Lately I'm toying a bit with the idea of splitting common in smaller coherent chunks (strings, log/debug - see also #211, ...), either as simple files or libraries.
Here's a first step (files, but potentially library) with a couple of new functions for string operations.

### DIFF
--- a/common/Makefile.am
+++ b/common/Makefile.am
@@ -9,8 +9,8 @@ libparseconf_la_SOURCES = parseconf.c
 # 'dist', and is only required for actual build, in which case
 # BUILT_SOURCES (in ../include) will ensure nut_version.h will
 # be built before anything else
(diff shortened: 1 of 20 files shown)


clepple:
@zykh seems logical to me.

If you do decide to split things into libraries, we might need to update the shared library version numbers for things like upsclient. This might depend on how they are linked.

zykh:
@clepple, I think I'll merge this as is: the transition to separate libs can wait till all the bits and pieces (and docs-from-Doxygen) are ready.

My choice has absolutely nothing to do with my renowned laziness and with the fact that this version was already ready. I swear. Seriously.

@aquette, anything against it?

zykh:
Merged as is.

PR #225, opened 2015-08-03
nutclient library socket reading

thalman:
Problem: nutclient library sometimes reads socket closed by server. Causes 100% CPU utilization.
Solution: proper ::read() return value evaluation

read returns -1 only if socket option O_NONBLOCK is set. But this option is explicitly avoided in connect method. When there is reading problem 0 is returned.

### DIFF
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -374,6 +374,11 @@ std::string Socket::read()throw(nut::IOException)
 
 		// Read new buffer
 		size_t sz = read(&buff, 256);
+		if(sz==0)
+		{
+			disconnect();
+			throw nut::IOException("Server closed connection unexpectedly");
+		}
 		_buffer.assign(buff, sz);
 	}
 }


clepple:
This patch doesn't seem quite right. IIRC, read() returns -1 on error, and in blocking mode, 0 at the end of the stream. So calling disconnect() in either case might work.

What does the code that calls nutclient look like?

Also, I haven't used nutclient directly myself, but it would seem that the calling code would need to wrap the read() method with a loop to catch short reads. That would be the first place I would look if something was using 100% CPU.

aquette:
@thalman : I agree with @clepple. Though your fix may work in a specific situation, this seems not to be the most broad nor optimal one.
Could you please send in a code snippet and some details to reproduce the issue, and check for the best fix?
@balooloo : I would welcome your lights here. 
Thanks to all

thalman:
@clepple oh yes, you are right, the call might return -1. I will make better proposal.

the calling code is quite simple. Just nutclient.connect(), getDevice(), disconnect(); this is done from time to time again and again.

The problem is caused by stopping nut-server in the middle of reading from client.

I attached gdb on the running process and I have seen that read() returns 0, errno is not set.
    std::string Socket::read()throw(nut::IOException)
is to blame.
    size_t sz = read(&buff, 256);
returns 0, _buffer is empty and read is called again.

clepple:
@thalman the second commit does fix the error case, but now we have two commits that overlap each other, and the second doesn't have a good one-line description. Please use `git rebase -i ...` to squash them, and use `git push -f ...` to update the pull request: http://www.networkupstools.org/docs/developer-guide.chunked/ar01s03.html#_repository_etiquette_and_quality_assurance

@balooloo is there public example code for nutclient? It would be useful to convert it into a unit test for verifying changes like this.

thalman:
@clepple rebase done

balooloo:
Hello all,
This fix sounds good for me.
Probably write function as similar problem (not closing the Socket object).
@clepple Sadly, I have no code to use with nutclient. But I can looking for.

PR #235, opened 2015-09-19
patch for Microsol Back-Ups BZ1200-BR

bsalvador:
This driver has a problem to identify the model of UPS. 
The first byte from the RecPack[0] indicates the model.
The problem is that the driver categorize different models, and this causes conflicts on model identification.
As I own a BZ1200-BR (internal name STAY1200_USB), I could only update this model.
I did the code in a way to not affect the old code.

### DIFF
--- a/drivers/solis.h
+++ b/drivers/solis.h
@@ -22,7 +22,8 @@
    2004/10/30 - Version 0.40 - add model data structs
    2004/11/22 - Version 0.50 - add internal e external shutdown programming
    2005/06/16 - Version 0.60 - save external shutdown programming to ups,
- 			       support new cables and Solaris compilation
+ 			       support new cables and Solaris compilation   
+   2015/09/19 - Version 0.63 - patch for correct reading for Microsol Back-Ups BZ1200-BR
 
 */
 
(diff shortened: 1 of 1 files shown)


clepple:
@bsalvador, I merged both these commits into the solis_debug branch. There are a few new warnings. In particular, `corrente_saida_F1_MI` and `corrente_saida_F2_MI` are unused, and recent clang points out that with float and double arguments, you probably want `fabs()` rather than `abs()`.

More build results here: http://buildbot.networkupstools.org/public/nut/waterfall

Thanks!

bsalvador:
Hello clepple,
I have added the constants corrente_* but did not implemented and I'm not willing to implement now.
How to proceed then? Should I update and remove the constant and fix the abs()?
I'm not used to work in SW development, so I need some guidance.

clepple:
I think everything is fixed in the solis_debug branch now. (We try to keep the number of warnings to a minimum, but I left the extra constants in a comment.)

If you want to try out the whole solis_debug branch, you can grab a tarball from here in a few minutes: http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-gcc/builds/437

clepple:
correction: latest is here: http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-gcc/builds/439

bsalvador:
OK, i'll try this out tonight!

PR #237, opened 2015-09-21
General improvements for ePDU support

aquette:
This includes NUT variables namespace extension and SNMP driver
improvements, for now, related to Eaton ePDU (G2 and G3), with many
other ePDU support scheduled. This includes:
* Extension of NUT variables namespace for:
** ambient status / thresholds and dry contacts
** input status (voltage, current, frequency) and thresholds (voltage
and current) 
** outlets status / thresholds (voltage, current) 
** outlets groups, and status / thresholds (voltage, current)
** 3phase status / thresholds (voltage, current, frequency) 

* Improvement of snmp-ups driver to handle:
** Outlet groups
** Alarm generation, in ups.alarm, for ambient / input / outlet /
outlet.group and 1phase / 3phase

* Implementation of all the above for Eaton ePDU G2 and G3


aquette:
@clepple , @zykh , @balooloo : please review, comment and ack if possible...

clepple:
Since I don't use a SNMP UPS or PDU, it's hard to comment on the technical content of this branch.

Two style items:
* Almost none of the commits mention which driver is being changed. I guess it is obvious for those familiar with ePDUs. 
* There are a lot of merges from master with no explanation why. http://www.networkupstools.org/docs/developer-guide.chunked/ar01s03.html#_repository_etiquette_and_quality_assurance (GitHub automatically tries the "test merge" mentioned in that section, but if you want to compile, having a separate, temporary integration branch while developing the feature branch is a more thorough test.)

data/cmdvartab
```diff
@@ -47,6 +47,11 @@ VARDESC input.voltage "Input voltage (V)"
 VARDESC input.voltage.extended "Extended input voltage range"
 VARDESC input.voltage.maximum "Maximum incoming voltage seen (V)"
 VARDESC input.voltage.minimum "Minimum incoming voltage seen (V)"
+VARDESC input.voltage.status "Voltage status relative to the thresholds"
```
jimklimov: Should the string descriptions be followed by "(V)" like other lines around here?

data/cmdvartab
```diff
@@ -106,14 +117,31 @@ VARDESC battery.charger.status "Battery charger status"
 
 VARDESC ambient.temperature "Ambient temperature (degrees C)"
 VARDESC ambient.temperature.alarm "Ambient temperature alarm is active"
+VARDESC ambient.temperature.status "Ambient temperature status relative to the configured thresholds"
 VARDESC ambient.temperature.alarm.maximum "Maximum allowed ambient temperature"
 VARDESC ambient.temperature.alarm.minimum "Minimum allowed ambient temperature"
 VARDESC ambient.temperature.alarm.enable "Enable ambient temperature alarm"
+VARDESC ambient.temperature.low "Temperature threshold low"
```
jimklimov: Should the string descriptions be followed by "(degrees C)" like some other lines around here?

data/cmdvartab
```diff
@@ -106,14 +117,31 @@ VARDESC battery.charger.status "Battery charger status"
...
+VARDESC ambient.temperature.high.warning "Temperature threshold high warning"
+VARDESC ambient.temperature.high.critical "Temperature threshold high critical"
 VARDESC ambient.humidity "Ambient humidity"
 VARDESC ambient.humidity.alarm "Ambient humidity alarm is active"
+VARDESC ambient.humidity.status "Ambient humidity status relative to the configured thresholds"
 VARDESC ambient.humidity.alarm.maximum "Maximum allowed ambient humidity"
 VARDESC ambient.humidity.alarm.minimum "Minimum allowed ambient humidity"
 VARDESC ambient.humidity.alarm.enable "Enable ambient humidity alarm"
+VARDESC ambient.humidity.low "Ambient humidity threshold low"
+VARDESC ambient.humidity.low.warning "Ambient humidity threshold low warning"
```
jimklimov: Is there a common unit of measurement here (e.g. "%hum"?) that can be appended to relevant lines?

drivers/eaton-mib.c
```diff
@@ -237,10 +245,111 @@ static info_lkp_t outlet_switchability_info[] = {
...
+	{ -1, "unknown" },
+	{ 0, "no" },  /* disconnected */
+	{ 1, "yes" }, /* connected */
+	{ 0, NULL }
+};
+
+static info_lkp_t marlin_threshold_status_info[] = {
+	{ 0, "good" },          /* No threshold trigged */
+	{ 1, "warning-low" },   /* Warning low threshold trigged */
+	{ 2, "cricital-low" },  /* Critical low threshold trigged */
```
jimklimov: "criCiTal"? Where is the typo - here in code or in some external counteragent which we must match?

drivers/eaton-mib.c
```diff
@@ -237,10 +245,111 @@ static info_lkp_t outlet_switchability_info[] = {
...
 
+static info_lkp_t marlin_ambient_presence_info[] = {
+	{ -1, "unknown" },
+	{ 0, "no" },  /* disconnected */
+	{ 1, "yes" }, /* connected */
+	{ 0, NULL }
+};
+
+static info_lkp_t marlin_threshold_status_info[] = {
+	{ 0, "good" },          /* No threshold trigged */
```
jimklimov: "trigged"? Perhaps "triggered" or "tripped"?

drivers/eaton-mib.c
```diff
@@ -237,10 +245,111 @@ static info_lkp_t outlet_switchability_info[] = {
...
+	{ 1, "warning-low" },   /* Warning low threshold trigged */
+	{ 2, "cricital-low" },  /* Critical low threshold trigged */
+	{ 3, "warning-high" },  /* Warning high threshold trigged */
+	{ 4, "cricital-high" }, /* Critical high threshold trigged */
+	{ 0, NULL }
+};
+
+static info_lkp_t marlin_threshold_frequency_status_info[] = {
+	{ 0, "good" },          /* No threshold trigged */
+	{ 1, "out of range" },  /* Frequency out of range trigged */
```
jimklimov: Are spaces acceptable, or can consumers expect singular tokens when parsing output etc.?

drivers/eaton-mib.c
```diff
@@ -237,10 +245,111 @@ static info_lkp_t outlet_switchability_info[] = {
...
+	{ -1, "unknown" },
+	{ 0, "open" },
+	{ 1, "closed" },
+	{ 0, NULL }
+};
+
+static info_lkp_t marlin_threshold_voltage_alarms_info[] = {
+	{ 0, "" },                       /* No threshold trigged */
+	{ 1, "low voltage warning!" },   /* Warning low threshold trigged */
+	{ 2, "low voltage cricital!" },  /* Critical low threshold trigged */
```
jimklimov: "cricital" and "trigged" here and below on a number of occasions

drivers/snmp-ups.c
```diff
@@ -878,9 +940,10 @@ void su_setinfo(snmp_info_t *su_info_p, const char *value)
 	if (SU_TYPE(su_info_p) == SU_TYPE_CMD)
 		return;
 
-	/* ups.status and ups.alarm have special handling, not here! */
+	/* ups.status and {ups, Lx, outlet, outlet.group}.alarm have special
+	 * handling, not here! */
 	if ((strcasecmp(su_info_p->info_type, "ups.status"))
-		&& (strcasecmp(su_info_p->info_type, "ups.alarm")))
+		&& (strcasecmp(strrchr(su_info_p->info_type, '.'), ".alarm")))
```
jimklimov: Do we have a guarantee (check somewhere) that "su_info_p->info_type" does not contain a "." other than expected separation of ".alarm", ".status" etc?

drivers/snmp-ups.c
```diff
@@ -914,12 +977,40 @@ void su_status_set(snmp_info_t *su_info_p, long value)
...
+					(su_info_p->flags & SU_OUTLET_GROUP)?"outlet group ":"outlet ",
+					item_number);
+				info_value = &alarm_info_value[0];
+			}
+			/* Special handling for phase alarms
+			 * Note that SU_*PHASE flags are cleared, so match the 'Lx'
+			 * start of path */
+			if (su_info_p->info_type[0] == 'L') {
+				/* Extract phase number */
+				item_number = atoi(strchr(su_info_p->info_type, 'L')+1);
```
jimklimov: Efficiency vs. readability: fromt the "if" above, we know that "su_info_p->info_type[0] == 'L'", so for `atoi` we can pass something like "su_info_p->info_type+1"?

drivers/snmp-ups.c
```diff
@@ -1211,10 +1311,138 @@ static int guestimate_outlet_count(const char *OID_template)
...
+{
+	/* Check if it is outlet / outlet.group */
+	if (!strncmp(varname, "outlet.group", 12)) {
+		return SU_OUTLET_GROUP;
+	}
+	else if (!strncmp(varname, "outlet", 6)) {
+		return SU_OUTLET_GROUP;
+	}
+	else {
+		upsdebugx(2, "Unknow template type: %s", varname);
```
jimklimov: unknow*N*

data/cmdvartab
aquette: fixed, thanks

data/cmdvartab
aquette: not for voltage.status, but indeed for the thresholds (low, high).
Also applicable to docs/nut-names.txt.
Fixed in both files, thanks

data/cmdvartab
aquette: indeed, as mentioned in docs/nut-names.txt, the unit is "(percent)".
Fixed, thanks

drivers/eaton-mib.c
aquette: indeed, big typo from /me.
By chance, the SNMP agent only send the numeric value, so the typo is limited to the above NUT value-lookup structure, and generally all the thresholds related ones, since I copy/pasted from this original one.
Fixed, thanks

drivers/eaton-mib.c
aquette: indeed, triggered and not trigged. Not sure where I inherited that one, but definitely wrong!
Fixed, thanks.

drivers/eaton-mib.c
aquette: theoretically, there is no standard defined for such things. Historically, we used to have single-word values for such things. Practically, the other threshold status being dash-separated, I've also applied the same to the "out-of-range".
Fixed, thanks

drivers/eaton-mib.c
aquette: both fixed everywhere in the previous commit

drivers/snmp-ups.c
aquette: not sure to fully get your comment here: we just want to process anything that is not "ups.status" or "*.alarm" (i.e. mapping ending with ".alarm, since it's strRchr()), that is "{ups, Lx, outlet, outlet.group}.alarm".
These are not processed (published) using dstate_setinfo(), but using {status,alarm}_{init,set,commit} 

=> grep "\.alarm\"" drivers/*mib*

jimklimov: Right, i missed that it is strRchr and so the last hit. No-op on this comment then ;)

drivers/snmp-ups.c
aquette: indeed, fixed, thanks.

drivers/snmp-ups.c
aquette: No other occurrence detected.
Fixed, thanks.

aquette:
@clepple: you're fully right, I've violated our rules many times here. All apologies, will try harder to conform for the upcoming commits

zykh:
As ever, I'm dead last...
@aquette, I cannot test it with real hardware, so here's just a few notes/questions in no particular order - otherwise everything seems OK.
- `su_alarm_set()` [d8e7ce24bc1581cbe12248f980fa5ea91f144f89, 0b1f116ca5566e4d39dffe743a7b2d50b570b8c7]: given that the printf format for outlets/groups/phases depends on what is in this function, instead of specifying it in OID/NUT lookup table, I think it could be better to simply leave it to this function by, for example, prepending the context to the string, something like:
  ```C
  if (
      (info_value = su_find_infoval(...)) != NULL &&
      info_value[0] != 0
  ) {
      ...
      if (
          su_info_p->flags & SU_OUTLET ||
          su_info_p->flags & SU_OUTLET_GROUP
      ) {
          item_number = ...;
          snprintf(
              alarm_info_value,
              sizeof(alarm_info_value),
              "outlet%s %i %s",
              (su_info_p->flags & SU_OUTLET_GROUP) ? " group" : "",
              item_number,
              info_value
          );
          info_value = alarm_info_value;
      } else if (su_info_p->info_type[0] == 'L') {
          item_number = ...;
          snprintf(
              alarm_info_value,
              sizeof(alarm_info_value),
              "phase L%i %s",
              item_number,
              info_value
          );
          info_value = alarm_info_value;
      }
      ...
  }

  ```
  This has also the added bonus of removing duplicates like `marlin_threshold_voltage_alarms_info[]` / `marlin_threshold_other_voltage_alarms_info[]`, `marlin_threshold_current_alarms_info[]` / `marlin_threshold_other_current_alarms_info[]` [3ba90efe040ce22810fad317867ad6bf81726e33, 0145c03d3c88926306fc0e7ae11dee4f178c4909].
- Don't alarms for phases (L*.alarm) [caf14dbf6b74e884437262ff53a702133c004885, 0b1f116ca5566e4d39dffe743a7b2d50b570b8c7] lack 'DOMAIN' and hence -if I read that correctly- don't comply to our 'DOMAIN.CONTEXT.SPEC' spec?
- Since we already have at our disposal `snprintf()`, shouldn't we use it in place of `sprintf()`? [d8e7ce24bc1581cbe12248f980fa5ea91f144f89, 0b1f116ca5566e4d39dffe743a7b2d50b570b8c7, 0237094d932622bacf06e141d5a69f8ae262385b, 296d50a5533d8ef116d62167e5ee0cb8a679179b, 94d9a0b7e4f537e6418b477dd8170878e32e2a16, a7bd2ae837ee27877985642aaedd57dd64b89af6, c47ca0e9da1a842a2fb85f406d0e1a815df7a0b0 and in other -untouched by commits in this PR- parts of snmp-ups.c]
- 66fc5535b39aa09c587c2f8eae6bbcd1b63dafd3: `nut_snmp_get_oid()` always returns `FALSE`. Shouldn't `ret` be set to `TRUE` when type is `ASN_OBJECT_ID`?
- 53884d49134ae9b33f760b10406082c73f2e4bf8: we already use `__func__` in other parts of the code.
- 296d50a5533d8ef116d62167e5ee0cb8a679179b and a7bd2ae837ee27877985642aaedd57dd64b89af6: `total_items = atoi(dstate_getinfo("outlet.group.count"));` / `total_items = atoi(dstate_getinfo("outlet.count"));`, shouldn't they make sure that `dstate_getinfo()` doesn't return `NULL`?
- The author of commit 932ed70fd529de0972cbe399fcf2099cdac94e56 doesn't seem quite right.
- I think that fixup commits (like 7c3abe133f8e2a18aa60018124dfdcb71858bb52, 26d115c5be5f7a41e718bf12e6bcd2bb1442a35b) could be merged with the commits they fix.
- I second Charles: I'd do a rebase on master and prefix commits with driver name, if any, to give them some context.
- What happens when there's no number in outlet/outlet.group (e.g. outlet.shutdown.return, with the 0, to represent the whole set, omitted)?

Side note: I think we should merge data/cmdvartab and docs/nut-names.txt to ease maintenance.

aquette:
hi @zykh 
thanks for your review, very much appreciated!

* su_alarm_set() and snprintf: indeed, should have applied that part of the improvements. It's now done.
* alarms for phases (L*.alarm): the variable name are indeed not complying to the standard (missing DOMAIN), but as mentioned, these are not published as is. The result is pushed in ups.alarm
* sprintf Vs snprintf: indeed, I've also replaced all calls to sprintf by snprintf, wiser like that.
* nut_snmp_get_oid(): return TRUE ;)
* 53884d4: now use __func__everywhere possible
* total_items: theoretically, yes we should check for NULL. However, process_template() already ensured that "outlet.[group.]count" is valued, either explicitly or using guestimate_template_count()
* 932ed70: indeed, stacked commands (add & push) during some testing...

* What happens when there's no number in outlet/outlet.group (e.g. outlet.shutdown.return, with the 0, to represent the whole set, omitted)? On these unit, there is no ".0" to target the whole unit (all outlets or all outlet groups). Data and commands are either per outlet or per outlet groups.

* merge data/cmdvartab and docs/nut-names.txt: I've indeed been thinking again at that, with the numerous mods of the present PR. The issue still being that one is for program processing and the other for human/asciidoc processing (more format). If you have a proposal, as usual, more than welcome...

I'm now going to merge the PR...

PR #247, opened 2015-12-08
Fix Powerware usb driver

LynxChaus:
Without this patch current driver unable to establish communication with Eaton 9125-5000g UPS.


aquette:
tested in non-regression with a 3105 (don't have that many XCP devices nowadays).
Works OK

PR #250, opened 2015-12-20
Avoiding conditional directives that split up statements.

RomeroMalaquias:


clepple:
This patch, while helpful for readability in some respects, trades your pet peeve for mine[*]: function calls that don't look like functions.

[*] I realize this is not documented.

I will try to get to this next week.

PR #253, opened 2016-01-05
First stab at checking driver.list.in format

aquette:


aquette:
@clepple: going through a PR to get a reviewer and broader tests ;)

clepple:
Hmm, I'm not sure I like the idea of the configure script modifying the source tree. We already have that problem with the USB scripts.

Maybe this is a separate issue, but I guess I was thinking that we could improve the logic in whatever parses the driver.list file (and check for the proper number of columns, etc). It could also be a simple script that can be invoked to check the file. IMHO, either way would be better than having it buried in the autoconf scripts.

aquette:
@clepple although I get your point, and I'm also not that happy with that, option (1) (improve the apps parsing driver.list) is almost impossible, if we consider the 3rd party that may use it ; and option (2) means that we rely on human execution... And human always fail ;)
The best is still to ensure that the file we generate and distribute, and that we use to generate other things (such as the web HCL) is well formatted.
We can at best execute the sed substitution on the final driver.list and not the template, so moving it to a make step. But we would miss the source fix (i.e. fixing the .in that is in the repository).
Modifying the template .in, as for the udev file, still has the bonus of alerting us through "git status"...
So...?!

aquette:
@clepple so when do you see this script being called? Again, the idea would be to catch formatting issues early, so that these are fixed in the repository, and that the generated / distributed file is clean...

clepple:
I don't know - I was going to take a look over the weekend when I have more time.

aquette:
perfect, thx @clepple

aquette:
@clepple : post week-end ping... ;)

clepple:
I still think this should be in a script, possibly invoked as a `make distcheck` rule.

aquette:
I've moved the code into a script, called at make dist time, to ensure that the distributed file is clean.
It is called from dist-hook, but can also be called manually (upon addition in driver.list.in for example) and will also try to fix both driver.list.in and driver.list, just in case.
Gonna merge the PR now...

PR #259, opened 2016-02-01
nutdrv_qx: TS Shara

zykh:
Let's recap:
- some time ago, TS Shara provided us with a patched version of blazer_usb, which added support for their devices,
- we then moved that code (after fixing a couple of minor bugs) to nutdrv_qx,
- over the past year, I kept it updated (meanwhile fixing another couple of small bugs).

Now, I got one user to test it: everything works as expected, but.. the shutdown/restore sequence.
What he described me (the shutdown works, but when mains returns the UPS keeps turning itself off and then on endlessly, unti he unplugs the USB cable) makes me think of a problem in the test procedure (he manually sets the FSD flag -which cannot be cleared unless you restart upsd: he told me the system running upsd is turned off, but maybe it's not really turned off but only suspended to disk - I have to investigate), but I can't tell for sure until I see the logs.

So, I'm opening this pull request but I will merge it only after that user reports back with some log: if the problem doesn't lay in the test procedure and I can't fix it, I'll disable all the `Sn`/`SnRm` commands.

Additional question: given that TS Shara provided us the code (and they even sent me a small part of their firmware, which implements the megatec commands), what support level should we use in the HCL?
- 2 -> based on fragments of publicly available protocol (which holds true for the megatec protocol), or
- 4 -> vendor provided protocol (which, conversely, holds true for the USB subdriver)?

### DIFF
--- a/data/driver.list.in
+++ b/data/driver.list.in
@@ -1098,6 +1098,8 @@
 "Trust"	"ups"	"2"	"UPS 1200VA Management PW-4120M"	""	"blazer_ser"
 "Trust"	"ups"	"2"	"UPS 1300VA Management PW-4130M"	""	"blazer_ser"
 
+"TS Shara"	"ups"	"4"	"(various)"	""	"nutdrv_qx"
+
 "UNITEK"	"ups"	"2"	"ALPHA 500 IC"	""	"blazer_ser"
 "UNITEK"	"ups"	"2"	"Alpha 1000is"	""	"blazer_ser"
 "UNITEK"	"ups"	"2"	"Alpha 500"	""	"blazer_ser"
--- a/docs/man/nutdrv_qx.txt
+++ b/docs/man/nutdrv_qx.txt
@@ -322,7 +322,7 @@ The argument is a regular expression that must match the bus name where the UPS 
 
 *subdriver =* 'string'::
 Select a serial-over-USB subdriver to use.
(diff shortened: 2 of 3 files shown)


clepple:
@zykh The ratings are approximate at best, so it's your call. Sounds like a 4 to me, given the assistance from the vendor that you describe, but I think it's fair to take into account how much work you ended up doing on your own.

aquette:
@zykh anything remaining on this PR?
Otherwise, please merge as I'm working on 2.7.4 release... thx and cheers

zykh:
@aquette: I'm still waiting for a reply from the user mentioned in the PR. Again, I think the problem is in the test procedure, so I'm inclined to merge this as is: anything against it?
@clepple: thanks, I agree with you and made it a 4.

aquette:
@zykh : for the test procedure, it may be easier for the user to test with a release than testing a GH branch?! So I'm ok for the merge

clepple:
> it may be easier for the user to test with a release than testing a GH branch?!

@aquette we still automatically generate snapshots for branches on Buildbot, as long as a branch has been created on the main Github repository. (It's as simple as `git branch github/pull/259 pull_259 && git push github pull_259`)

@zykh no objections to merging, though, since it is limited to a single sub-driver.

aquette:
@clepple : true, my memory tends to fails easily these days ;)

zykh:
merged - I'll update this discussion if/when that user (or someone else) reports back.

PR #263, opened 2016-02-26
Fix docs location of upssched as sbin

idodeclare:


clepple:
Cherry-picked as 2ae6792. Thanks!

PR #275, opened 2016-03-23
Fix "--with-doc"(=auto|yes) to actually succeed if some parsers are not available

jimklimov:
configure.ac : when using "--with-doc" do actually proceed if some supported methods are found and some are not... build what we can on this platform

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -630,10 +630,13 @@ dnl is set to 'no', we may still want to build some doc targets manually
 NUT_CHECK_ASCIIDOC
 
 case "${nut_with_doc}" in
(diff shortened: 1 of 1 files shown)


clepple:
I don't mind doing this while merging, but for future reference:

http://networkupstools.org/docs/developer-guide.chunked/ar01s03.html#_commit_message_formatting <http://networkupstools.org/docs/developer-guide.chunked/ar01s03.html#_commit_message_formatting>

The title of the pull request is pretty close.

jimklimov:
1sec...

aquette:
perfect Jim, thx!

PR #276, opened 2016-03-24
SNMP: Power on and power off delay time writing; timeticks support;

uprt:
- Added a feature of power on and power off delay time writing to UPS configuration (like in "blazer", "nutdrv_qx", "powerp" and "safenet" drivers) in SNMP driver.
  When ondelay= and offdelay= params are configured in ups.conf, snmp driver will try to write this values to UPS during initialization (to make user able to watch current status and be sure if the write was successful) and before shutdown procedure (if the first init write was unsuccesfull for some reasons).
- Added support of writing "timeticks" values to su_setvar() routine (because if we will try to write integer value to timeticks-type OID it will cause an error)
- Also there is "ups.delay.shutdown" and "ups.delay.shutdown" parameters fix in "apc-mib": they used to have ST_FLAG_STRING mark, but in fact they are not strings at all
  (they are "timeticks", so we also have to use SU_TYPE_TIME flag to perform writing routine correctly).
  This OIDs has the same STRING flag in some other drivers (compaq-mib, mge-mib), perhaps they need such fixes too.
  In SNMP driver, SU_FLAG_STRING is supposted to be used only in writing routines, and because of writing of such paramtres was not implemented at all in the driver, nobody cared about this.

Ondelay and offdelay writing as well as automatic shutdown procedure were successfully tested on APC/Schneider Smart-UPS RT 2000 RM XL with  AP9630 SNMP Board on CentOS 7.1


aquette:
hi @uprt , sorry for the eternity taken to consider your PR.
As per your comments, apc-mib.c should also be modified for ups.delay.{start,shutdown} to include SU_TYPE_TIME, which is missing from your PR.
As for the similar ones:
* the ones in mge-mib.c are from the IETF MIB (RFC1628) is an integer, not a timeticks
* the compaq ones are also integers, not timeticks

Could you please update your PR against the latest master and complete apc-mib.c?
thanks and cheers

uprt:
Done.

jimklimov:
There is a conflict with change made to master branch around (maybe also before) PR #687 :

https://github.com/networkupstools/nut/pull/276/files#diff-9f64ddc4e9473c799bd598ed80fa6109R2723-R2728

that seems to also add the same decision but at a different indentation level (not only in this "if" block but outside it too). I am inclined to trust that reviewed PR and will amend this one to resync with master.

jimklimov:
TODO: Bump resulting snmp-ups driver version after merge.

PR #279, opened 2016-04-06
Problem: need to create DMF

vyskocilm:
Solution: add jsonify-mib.py to dump snmp_info_t and info_lkp_t
structures for snmp subdrivers

Add a --test pass, which generate C arrays back and a small testing programs, which compares all items of original and generated arrays. So far the tool is correct on all -mib.c files from master.

@aquette


aquette:
good progress, thanks @vyskocilm 
Beside from the values lookup (INFO) and data mapping, the matching info (mib2nut, which includes the sysOID and more) is missing.

vyskocilm:
Hi,

1. missing mib2nut_info_t - good point, I'll do it
2. flags are numeric for a reason - this is by design just intermediate format, which is easier to deal with than Abstract Syntax Tree of C. I try to avoid coding de-dacto DMF in this PR, so the result is unfriendly.
3. I'll make different PR for netvision?

vyskocilm:
The missing mib2nut_info_t and alarms_info_t were added - I've changed the JSON structure, so now everything belongs to its own JSON object.

aquette:
I've just pushed a DMF branch into our repo: https://github.com/networkupstools/nut/tree/DMF

For now, this is not the thing we (NUT) want to have in the long run, but that is already useful enough to us (Eaton) and will allow to update the spec and implementation through some trial testing on this.

Thanks to our Eaton fellows for working on this long standing REQ!

clepple:
A few build/portability issues:
 * [valgrind needs an autoconf test, and circular make dependencies should be addressed](http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-gcc/builds/571/steps/compile/logs/stdio)
 * [Python modules should be checked at configure time](http://buildbot.networkupstools.org/public/nut/builders/MacOSX-10.11-x64/builds/1283/steps/compile/logs/stdio)
 * [C compiler name should not be hardcoded - use $(CC)/$(CXX)](http://buildbot.networkupstools.org/public/nut/builders/Ubuntu-trusty-x64-clang/builds/90/steps/compile/logs/stdio)
 * [Neon LDFLAGS need to be used](http://buildbot.networkupstools.org/public/nut/builders/OpenBSD-x64/builds/401/steps/compile/logs/stdio)

clepple:
This should not be a fatal configuration error:

```
checking for a Lua interpreter with version >= 5.1... none
configure: error: cannot find suitable Lua interpreter
```

jimklimov:
I suggest that this PR can be closed to reduce confusion, as PR #305 includes this code and a lot more (and now includes a fix to get the best of `Sxx` style naming and ability to read all filenames from a directory - the user just has to provide a directory with alphabetically sorted symlinks to his choice of original DMF files, and packaging delivers both originals and symlinks for the ~17 DMFs we manage in the codebase).

vyskocilm:
As you wish @jimklov

PR #280, opened 2016-04-07
Allow to customize path to gdlib-config and net-snmp-config programs

jimklimov:
Current m4 scripts assume a fixed program name to be in PATH. This is not always the case, e.g. in Solarish oi-userland we have /usr/bin/net-snmp-config-{32,64} and /usr/bin{,/$(MACH64)}/gdlib-config


aquette:
hi @jimklimov , the patch seems fine, though I looked quickly and not tested.
Could you please also complete the related documentation (docs/configure.txt).
thx and cheers

jimklimov:
Ack, forgot to add that :)
Done now

PR #281, opened 2016-04-08
Configure : allow to fully toggle generation of manpages among other docs, and do not ALWAYS blindly attempt systemd setup

jimklimov:
Rationale: on a build-host that prepares packaging for multiple architectures in empty workspaces, it suffices to build the docs just once, including manpages (should the buildhost recipe define so). Now this can be done indeed.

Note: disabled manpage generation explicitly causes "make dist" to fail, in order to avoid dist'able tarballs not conforming to spec and expectations.

Also this PR reduces noise about systemd on OSes or builds that are not relevant for it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1053,13 +1053,13 @@ fi
 AM_CONDITIONAL(WITH_PKG_CONFIG, test -n "${pkgconfigdir}")
 
 PKG_PROG_PKG_CONFIG
-systemdsystemunitdir=`$PKG_CONFIG --variable=systemdsystemunitdir systemd`
 AC_MSG_CHECKING(whether to install systemd files)
 AC_ARG_WITH([systemdsystemunitdir],
 	AS_HELP_STRING([--with-systemdsystemunitdir=DIR], [Directory for systemd service files (auto)]),
 [
 	case "${withval}" in
-	yes|auto)
+	yes|auto|"")
+		systemdsystemunitdir=`$PKG_CONFIG --variable=systemdsystemunitdir systemd`
 		;;
 	no)
 		systemdsystemunitdir=""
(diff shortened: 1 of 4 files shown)


jimklimov:
Per offline discussion with @aquette , given the previous default of "--with-doc=no" and a separately solved support for always-building manpages, the reverse-compatible default would be "--with-doc=man(=yes)"  to not regress on `make dist`. So there is a bit of work to do on this PR ;)

jimklimov:
Now this comes from default `configure` without parameters:

````
build and install documentation: man=yes 
build specific documentation format(s): yes  man
build and install the development files: no 
````

The `make dist` call also succeeds and the resulting tarball includes both text and `.digit` manpages.
The manpages are rebuilt after I explicitly do `rm docs/man/*.{1,3,5,8} ; ./configure; make dist`.

Note that `make distcheck` explicitly requires `--with-all` in the un-tarballed workspace, so I am now installing `*-dev` dependencies it requires iteration after iteration ;)

jimklimov:
UPDATE: `make distcheck` succeeded with this changeset for me.

clepple:
Hmm, error building fe17861 : [fail log](http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-gcc/builds/551/steps/compile/logs/stdio) versus [last known good](http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-gcc/builds/550/steps/compile/logs/stdio) seems to indicate that despite the warning, asciidoc does need one of those commits to write to the `_build` directory rather than the source directory.

clepple:
Corresponding log for 5e592d7 is [here](http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-gcc/builds/552/steps/compile/logs/stdio). Take my conclusions above with a grain of salt. Now I think `--destination-dir=.` might only be needed in certain cases, but I don't have time to verify that right now.

jimklimov:
Discussing with @aquette - seems to be an asciidoc bug (https://github.com/asciidoc/asciidoc/issues/44) so I'll roll back this particular patch.

clepple:
@aquette in the future, it would be clearer to merge from prior to the last commit, rather than reverting it (since it hadn't been merged to master yet). The extra commits will make bisecting to locate errors more tedious.

Issue #232, opened 2015-09-03
nutclient: stringset_to_strarr() and stringvector_to_strarr() only work for one element


clepple:
reported by Paul Vermeer: http://article.gmane.org/gmane.comp.monitoring.nut.user/9331

> Both functions stringset_to_strarr and stringvector_to_strarr have a similar bug in play. Although trivial in code, the coded function does not operate as intended.
> The pointer increment is missing (see added line marked yellow below [pstr++;]), which results in the same element being updated over and over again, and then only leaving the last element of the list in the array.

aquette:
@balooloo : if you have 5 mn to take a look at that one also, would be great. Please self assign if you can, thx.

aquette:
Closed through PR #284

PR #284, opened 2016-04-09
Fix bug #232

balooloo:
Fix bug #232


aquette:
thx @balooloo 
do you have some remaining commits on this one, and should we be waiting a bit more before merging that PR?

balooloo:
Hello,

AFAIK, no need to wait for something. It can be merged.

PR #285, opened 2016-04-11
Problem: netvision uses 0 as dfl

vyskocilm:
Solution: use NULL as other subdrivers

cc @aquette 

fixes comment in https://github.com/networkupstools/nut/pull/279


clepple:
From the [developer guide](http://networkupstools.org/docs/developer-guide.chunked/ar01s03.html#_commit_message_formatting):

> If your commit is just a change to one component, such as the HCL, upsd or a specific driver, prefix your commit message in a way that matches similar commits. This helps when searching the repository or tracking down a regression.

The "Problem:" / "Solution:" format is probably more appropriate for an issue than for a commit message.

In this case, please prefix the commits with something like `snmp-ups: ` or `netvision-mib: `

vyskocilm:
@clepple  renamed commit message

PR #289, opened 2016-05-03
Follow-up to NUT PR#281 (asciidoc dest. dir.) - notice for the future

jimklimov:


aquette:
with regard to https://github.com/asciidoc/asciidoc/issues/44

jimklimov:
and with regard to NUT PR #281 so we have the links ;)

PR #297, opened 2016-06-23
Add support for SU3000LCD2UHV

madscientist159:
Note that protocol 1330 does not respond to requests for
UPS.OutletSystem.Outlet.DelayBeforeReboot, therefore those
requests needed to be bypassed in the driver.

Reported values appear correct, no further testing performed.

Signed-off-by: Timothy Pearson <kb9vqf@pearsoncomputing.net>

### DIFF
--- a/drivers/libhid.c
+++ b/drivers/libhid.c
@@ -276,6 +276,7 @@ void HIDDumpTree(hid_dev_handle_t udev, usage_tables_t *utab)
 #ifndef SHUT_MODE
 	/* extract the VendorId for further testing */
 	int vendorID = usb_device((struct usb_dev_handle *)udev)->descriptor.idVendor;
+	int productID = usb_device((struct usb_dev_handle *)udev)->descriptor.idProduct;
 #endif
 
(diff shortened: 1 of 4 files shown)


clepple:
Thanks.

I'll look closer later, but should the productId/vendorId check be "&&" instead of "||" ?

madscientist159:
Yes, definitely!  My apologies for the copypasta...

clepple:
No worries. In case this gets fixed later, do you have any info about the firmware version? Some TL units report "ups.firmware", but maybe the "protocol" is good enough.

Also, we don't have any of the UHV models listed here: http://networkupstools.org/ddl/Tripp_Lite/index.html <http://networkupstools.org/ddl/Tripp_Lite/index.html> - is your output similar to the other LCD units?

madscientist159:
From nut: "ups.firmware	:	160"

Yes, aside from the input / output voltage ranges the output looks similar and sane.  You typically only find 208V / 240V equipment in datacenters, which is probably why this support hasn't been added yet.

PR #304, opened 2016-08-15
Problem: can't use nut-scanner and get dmf driver

vyskocilm:
Solution: fix it + small debug messages


vyskocilm:
Hi - without this patch dmf driver fails to work even in my simple testing scenario ;-)

I'll talk to @jimklimov tomorow to change the build somehow, OK?

PR #314, opened 2016-09-13
Give the Tripplite SNMP MIB a name, do not duplicate "ietf" entries

jimklimov:


aquette:
would also need to be listed in docs/man/snmp-ups.txt, but that's not the only one missing there

jimklimov:
Added a line about tripplite. For more, see a separate new feature PR #318 :)

jimklimov:
@aquette : is this good to be merged? I guess I'll need to port it into the DMF branch afterwards, so would prefer this to not be delayed by excessively long.

Issue #311, opened 2016-08-25
Test documentation tools such as dblatex in configure script


clepple:
Currently we check for tools such as `asciidoc` and `dblatex`, but we do not verify that they produce output.

Reference: https://github.com/networkupstools/nut/pull/305#issuecomment-242352514

Vaguely related: #162 (indicate required versions)

jimklimov:
Example failed build log: http://buildbot.networkupstools.org/public/nut/builders/FreeBSD-x86/builds/529/steps/compile/logs/stdio

````
...
asciidoc: include: /usr/home/_buildbot/nut-slave/FreeBSD-x86/build/nut-v2.7.4-18-ga12f3a8/docs/history.txt
asciidoc: user-manual.txt: line 187: reading: /usr/home/_buildbot/nut-slave/FreeBSD-x86/build/nut-v2.7.4-18-ga12f3a8/docs/history.txt

a2x: executing: "xmllint" --nonet --noout --valid "/usr/home/_buildbot/nut-slave/FreeBSD-x86/build/nut-v2.7.4-18-ga12f3a8/_build/sub/docs/user-manual.xml"


a2x: executing: "dblatex" -t pdf -p "/usr/local/etc/asciidoc/dblatex/asciidoc-dblatex.xsl" -s "/usr/local/etc/asciidoc/dblatex/asciidoc-dblatex.sty"  -V  "/usr/home/_buildbot/nut-slave/FreeBSD-x86/build/nut-v2.7.4-18-ga12f3a8/_build/sub/docs/user-manual.xml"

Build the book set list...
xsltproc -o /tmp/tmpTEw1YW/doclist.txt --xinclude --xincludestyle doclist.xsl /usr/home/_buildbot/nut-slave/FreeBSD-x86/build/nut-v2.7.4-18-ga12f3a8/_build/sub/docs/user-manual.xml
Build the listings...
xsltproc -o /tmp/tmpTEw1YW/listings.xml --xinclude --xincludestyle --param current.dir '/usr/home/_buildbot/nut-slave/FreeBSD-x86/build/nut-v2.7.4-18-ga12f3a8/_build/sub/docs' /usr/local/share/dblatex/xsl/common/mklistings.xsl /usr/home/_buildbot/nut-slave/FreeBSD-x86/build/nut-v2.7.4-18-ga12f3a8/_build/sub/docs/user-manual.xml
[...]

PR #315, opened 2016-09-14
Fix docgen ability (verify we actually CAN generate chosen doc types with our tools)

jimklimov:
This should take care of issue https://github.com/networkupstools/nut/issues/311

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -796,7 +796,12 @@ dnl not fail if we have no tools to generate it.
 
 	man*)
 		AC_MSG_CHECKING([if we can build ${nut_doc_build_target_base}])
+		can_build_doc_man=no
 		if test "${nut_have_asciidoc}" = yes ; then
+			( cd "docs/man" && ${A2X} --format manpage --destination-dir=. --xsltproc-opts "--nonet" snmp-ups.txt && test -s snmp-ups.8 ) && can_build_doc_man=yes
+			rm -f docs/man/snmp-ups.8
+		fi
+		if test "${can_build_doc_man}" = yes ; then
 			AC_MSG_RESULT(yes)
 			DOC_BUILD_LIST="${DOC_BUILD_LIST} ${nut_doc_build_target_base}"
 		else
--- a/configure.ac
+++ b/configure.ac
@@ -717,9 +717,19 @@ esac
 
 dnl Note: Do not cover ${nut_doc_build_list} in braces or quotes here,
 dnl to ensure that it is processed as several space-separated tokens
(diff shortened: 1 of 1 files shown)


clepple:
@jimklimov I haven't had a chance to look into the code itself, but `configure` shows the following in the summary:

```build and install documentation: auto```

Shouldn't it resolve "auto" into a boolean by that point?

jimklimov:
Note that e.g. the FreeBSD builder fails due to just this issue - http://buildbot.networkupstools.org/public/nut/builders/FreeBSD-x64/builds/429/steps/compile/logs/stdio

(and does not fail for PRed branches with this fix in place)

jimklimov:
@clepple : the line you quoted reports the value received (explicitly or by default) from the command line (`--with-doc=<receivedValue>`). Then the code goes on to interpret it by components (known doc types, with enablement explicitly specified or defaulted from the common setting - explicit or the "top-level default"), which was merged earlier in the year - so you can have e.g. `--with-doc=man=yes,pdf=no,html-single=auto`.

Then the configure script prints out what it detected of a) user's desire for particular doc type, and b) system's ability to build that, if the desire is not "no" or "skip". And finally if we can build all "yes" required doctypes, the script proceeds (and also enables those "auto"s that it can build, or skips those "auto"s that it cannot build).

For example see http://buildbot.networkupstools.org/public/nut/builders/FreeBSD-x64/builds/430/steps/configure/logs/stdio :
````
checking whether to build and install documentation... auto 
checking desire and ability to build man documentation... auto
checking if we can build man... a2x: WARNING: --destination-dir option is only applicable to HTML based outputs
yes
checking desire and ability to build html-single documentation... auto
checking if asciidoc version can build html-single (minimum required 8.6.3)... no
checking desire and ability to build html-chunked documentation... auto
checking if a2x version can build html-chunked (minimum required 8.6.3)... no
checking desire and ability to build pdf documentation... auto
checking if dblatex version can build pdf (minimum required 0.2.5)... a2x: ERROR: "dblatex" -t pdf -p "/usr/local/etc/asciidoc/dblatex/asciidoc-dblatex.xsl" -s "/usr/local/etc/asciidoc/dblatex/asciidoc-dblatex.sty"   "/usr/home/_buildbot/nut-slave/freebsd-x64-full/build/docs/asciidoc.xml" returned non-zero exit status 1
no
checking whether to build specific documentation format(s)... yes  man
...

Configuration summary:
======================
...
build and install documentation: auto 
build specific documentation format(s): yes  man
````

PR #318, opened 2016-09-15
Add mib2nut listing (this should help set up snmp-ups driver)

jimklimov:
Review on PR #314 mentioned that the snmp-ups manpage does not reflect all built-in MIBs. Indeed... and I found no easy way to list them in a pretty manner. So now here is one ;)


aquette:
@jimklimov : merged with a minor adjustment: simply call exit(EXIT_SUCCESS) instead of fatalx(EXIT_FAILURE, ...) to be coherent with similar code (help in main.c for example)

PR #320, opened 2016-09-16
snmp-ups: add sysContact and sysLocation support

aquette:
The generic MIB-2 provides system contact and location information that are now
publish respectively as device.contact and device.location

### DIFF
--- a/drivers/snmp-ups.c
+++ b/drivers/snmp-ups.c
@@ -3138,6 +3138,7 @@ bool_t su_ups_get(snmp_info_t *su_info_p)
 	alarms_info_t * alarms;
 	int index = 0;
 	char *format_char = NULL;
+	int saved_current_device_number = -1;
 	snmp_info_t *tmp_info_p = NULL;
 
 	upsdebugx(2, "%s: %s %s", __func__, su_info_p->info_type, su_info_p->OID);
@@ -3192,6 +3193,41 @@ bool_t su_ups_get(snmp_info_t *su_info_p)
 		else {
 			upsdebugx(2, "%s: can't instantiate template", __func__);
 			return FALSE;
+		}
+	}
+	else {
+		/* Non-templated OID, still may be aimed at a
+		 * daisy-chained device (master of the chain
+		 * makes sense for IETF device.contact etc.).
+		 * BUT: It could be a direct request for earlier
+		 * resolved OID. So check for "device.N." too.
+		 */
+		if (daisychain_enabled == TRUE
+		&&  devices_count > 1
+		&&  current_device_number > 0
+		) {
+			/* So we had a literal OID string, originally
+			 * Check for "device.N." in the string: */
+			char * varname = su_info_p->info_type;
+			if (!strncmp(varname, "device.", 7)
+			&&  (varname[7] >= '0' && varname[7] <= '9')
+			) {
+				upsdebugx(2, "%s: keeping original "
+					"current device == %d for "
+					"non-templated daisy value",
+					__func__, current_device_number);
+			} else {
+				upsdebugx(2, "%s: would fake "
+					"current device == 1 for "
+					"non-templated daisy value "
+					"instead of %d",
+					__func__, current_device_number);
+				saved_current_device_number = current_device_number;
+			}
+			/* At this point we applied no hacks yet,
+			 * just stashed a non-negative value into
+			 * saved_current_device_number
+			 */
 		}
 	}
 
@@ -3385,7 +3421,13 @@ bool_t su_ups_get(snmp_info_t *su_info_p)
 	}
 
 	if (status == TRUE) {
+		if (saved_current_device_number >= 0) {
+			current_device_number = 1;
+		}
 		su_setinfo(su_info_p, buf);
+		if (saved_current_device_number >= 0) {
+			current_device_number = saved_current_device_number;
+		}
 		upsdebugx(2, "=> value: %s", buf);
 	}
 	else {
--- a/drivers/snmp-ups.c
+++ b/drivers/snmp-ups.c
@@ -680,7 +680,9 @@ void upsdrv_initups(void)
 	 * tables (note: for lack of better knowledge, defined as read-only
 	 * entries here, and also read-once - not updated during driver uptime) */
 
-	if (NULL == dstate_getinfo("device.description")) {
+	if (NULL == dstate_getinfo("device.description")
+	&&  NULL == dstate_getinfo("device.1.description")
+	) {
 		/* sysDescr.0 */
 		if (nut_snmp_get_str(".1.3.6.1.2.1.1.1.0", model, sizeof(model), NULL) == TRUE) {
 			upsdebugx(2, "Using IETF-MIB default to get and publish sysDescr for device.description (once)");
@@ -690,7 +692,9 @@ void upsdrv_initups(void)
 		}
 	}
 
-	if (NULL == dstate_getinfo("device.contact")) {
+	if (NULL == dstate_getinfo("device.contact")
+	&&  NULL == dstate_getinfo("device.1.contact")
+	) {
 		/* sysContact.0 */
 		if (nut_snmp_get_str(".1.3.6.1.2.1.1.4.0", model, sizeof(model), NULL) == TRUE) {
 			upsdebugx(2, "Using IETF-MIB default to get and publish sysContact for device.contact (once)");
@@ -700,7 +704,9 @@ void upsdrv_initups(void)
 		}
 	}
 
-	if (NULL == dstate_getinfo("device.location")) {
+	if (NULL == dstate_getinfo("device.location")
+	&&  NULL == dstate_getinfo("device.1.location")
+	) {
 		/* sysLocation.0 */
 		if (nut_snmp_get_str(".1.3.6.1.2.1.1.6.0", model, sizeof(model), NULL) == TRUE) {
 			upsdebugx(2, "Using IETF-MIB default to get and publish sysLocation for device.location (once)");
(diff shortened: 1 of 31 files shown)


drivers/snmp-ups.c
```diff
@@ -392,6 +392,19 @@ void upsdrv_initups(void)
 		dstate_addcmd("shutdown.return");
 		dstate_addcmd("shutdown.stayoff");
 	}
+
+	/* Publish sysContact and sysLocation for all subdrivers */
+	/* sysContact.0 */
+	if (nut_snmp_get_str(".1.3.6.1.2.1.1.4.0", model, sizeof(model), NULL) == TRUE)
```
jimklimov: Should the OID strings be explicit here, or defined as macros (perhaps in common snmp-ups.h)? Or even (potentially) individualized in mappings if e.g. a particular verndor supports some extended attribute for this (or plain different, ignoring the standards) and NUT might honor it?

aquette: not sure for macros, since I've tried to get rid of this in implementation files (*-mib.c).
When doing the present PR, I was also hesitating between the central point (as done), since it's worth for all subdrivers, Vs adding to each subdriver and get that extracted for DMFs.
Each has pros and cons.
That said, it would be good for DMF to be able to have some base.dmf that could then be included in others, but that implies other things (not all .dmf to be parsed, ...)
As for extended attributes, these specific OIDs have no particular default values and are only useful for customers / users to push their own info into it.

aquette: Complementary thought:
while thinking back about that, and how to finalize implementation, I realized that:
1) these OIDs are read / write (some able to modify through upsrw, as long as you have write access to the device)
2) this implies that these OIDs must be resolvable (when issuing upsrw, the driver will resolve the NUT variable name into its OID), which implies that it has to be in MIB2NUT 
3) hence (2) implies that we should add it to all mapping tables
Or
Put in place a mechanism to add dynamically OIDs (or have some #include mechanism for DMFs).

Putting on hold for now

aquette:
removing the milestone, since the tech solution is not clear enough for a short term implementation

jimklimov:
PR updated to address comments above: now each existing `*-mib.c` file defines `device.contact`, `device.location` and `device.description` (currently initialized to IETF OIDs for those, can be tuned to vendor values if ever needed) as RW STRING entries. Tested locally that writes (`upsrw` to an Eaton ePDU) actually work.

As a fallback, if a (future/custom) MIB table does not define any of those names, the `snmp-ups.c` would patch those entries (as read-only) similarly to original proposal. This was also verified, by commenting away the 3 lines from eaton mib.c file.

While testing, found a bug in `su_setOID()` which got it to assume anything starting with "device" is part of a daisy-chain (the tested ePDU in fact was a "slave" part of a chain, but unlike many other places the "device" string is handled, here we did not test for template-related variables and would mis-classify the settings in any case).

drivers/snmp-ups.c
jimklimov: Finally addressed a few years later ;)

jimklimov:
LIMITATIONS FOUND:
* Verified that by accessing the daisy-chain master I can query entries like `device.X.description` (and see them in upsc listing), but it seems that they return data for the head node regardless of `X` value.
* Also I can not set these values with `upsrw` on a daisy-chain master (like I can on a standalone/slave device): the un-indexed name like `device.contact` is not writeable (ERR READONLY, probably falls under the default initialization with the snmp-ups.c fallback noted above), the `device.0.contact` is not resolved (ERR VAR-NOT-SUPPORTED), and `device.1.contact` etc. are not handled (ERR UNKNOWN) around `[D3] su_find_info: unknown info type (contact)`

While more investigation and enhancements are desirable for this daisychain-master use-case to make sense, the PR already brings value for single-device users, and this situation is not a regression compared to original content of the PR from 2016. Just some unfinished business...

jimklimov:
Running offline, leaving a note @aquette : Seems a change like this is on the right track:
````
--- a/drivers/snmp-ups.c
+++ b/drivers/snmp-ups.c
@@@ -1536,7 -1536,7 +1536,12 @@@ void su_setinfo(snmp_info_t *su_info_p
        /* pre-fill with the device name for checking */
        snprintf(info_type, 128, "device.%i", current_device_number);

--      if ((daisychain_enabled == TRUE) && (devices_count > 1)) {
++      /* Daisy-chain template magic should only apply to defaulted
++       * entries (oid==null) or templated entries (contains .%i)
++       */
++      if ((daisychain_enabled == TRUE) && (devices_count > 1)
++      &&  (su_info_p->OID == NULL || strstr(su_info_p->OID, ".%i") != NULL)
++      ) {
                /* Only append "device.X" for master and slaves, if not already done! */
````

At least, this way a daisy-chain master exposes "device.contact" and "device.macaddr" once, but e.g. "device.X.type: pdu" (where OID==null) all 3 times for the 3-piece daisy chain. But the "device.model" that has an `.%i` templated OID also has one instance this way :\ And the "device.contact" may not be written (INVALID-ARGUMENT) on the daisy master, while still can be on a single/slave device.

jimklimov:
Back to drawing board :\ 

While testing, I checked that `device.mfr` was successfully exposed for the many devices in a chain, but missed that it was a NULL OID too. A templated value like `device.model` seems exposed once (like `device.contact`) although the OID string has `.%i`...

jimklimov:
Good news (to post a bit later) is that we can discern in `su_ups_get()` of all places, whether the original OID was templated (methods like `su_setinfo()` in fact get the specific string from there). This allows to short-circuit assignment of non-templated "device.*" entries in a daisychain to e.g. "device.1.*" specifically, and so lose meaningless dupes (contact, location, macaddr...) attributed to all devices but taken from the first in reality.

Bad news is that setting the value (upsrw) becomes very problematic, because `su_find_info()` can't locate e.g. `device.1.contact` that is not actually in the `snmp_info` table.

For the location/contact/description in particular, daisychain seems to preclude detection of the values at "root" level (device.contact etc.) so it is populated by fallback (read-only IETF) for the master PDU device.

jimklimov:
For some reason, getting ALL bits to work, or be usable (e.g. both able to set daisy master and daisy slave and standalone devices) is a fine balancing act. What's more odd, at some point snmp-ups started for daisy devices (master or slave) with `debug_min 1` or unspecified, fails to send data to upsd; with debug level 2 or more, or for a stand-alone UPS, it works; USB driver showed no such problem. Why didn't I take the other pill?

jimklimov:
With recent commits, behaviors can be deemed achieving the feature (somewhat clumsily but logically) unless proven otherwise, and are as follows:
* single non-chained device exposes (upsc) and can set (upsrw) the device.contact, device.location and device.description
* single directly-accessible daisy chain slave device also exposes (upsc) and can set (upsrw) the device.contact, device.location and device.description (but currently seems to suffer with lowered debugging levels; maybe an issue on recent master branch though)
* the daisy-chain master exposes two sets of entries:
** device.contact, device.location and device.description are "fallback IETF-MIB" (read-only, read-once)
** indexed device.1.contact, device.1.location and device.1.description are readable and writeable, pointing to the daisy master device
** spurious device.{2,3...}.contact, etc. entries backed by non-templated OID definitions and in fact showing the master-device info (so also device.n.macaddr coming from IETF subtree) are not exposed anymore
** device.0.fieldname can not be used to read/set these entries

Along this journey, added `./tools/nut-dumpdiff.sh` to help compare upsc or driver-discovery outputs (discarding variations in numeric measurements), e.g.:
````
# ./tools/nut-dumpdiff.sh /tmp/daisyMaster-master.txt /tmp/daisyMaster-pr4.txt
=== ./tools/nut-dumpdiff.sh: comparing '/tmp/daisyMaster-master.txt' (-) vs '/tmp/daisyMaster-pr4.txt' (+)
+Network UPS Tools - Generic SNMP UPS driver 1.19 (2.7.4-4727-ga21f32f)
+device.1.contact: 42ITy Team
+device.1.description: EPDU MA 0U (C14 10A 1P)16XC13
+device.1.location: SRV Room, 42ITy Rack
-device.2.macaddr: 00 20 85 F4 61 BE
-device.3.macaddr: 00 20 85 F4 61 BE
+device.contact: 42ITy Team
+device.description: EPDU MA 0U (C14 10A 1P)16XC13
+device.location: SRV Room, 42ITy Rack
-driver.version.data: eaton_epdu MIB 0.58
-driver.version.internal: 1.18
-driver.version: 2.7.4-4696-g763e03c
+driver.version.data: eaton_epdu MIB 0.59
+driver.version.internal: 1.19
+driver.version: 2.7.4-4727-ga21f32f
````

jimklimov:
Presumably, for the data points in original question (contact, location, description) the quest is solved :)

The more generic issue of non-templated OIDs used in daisy-chain context for many other if-else code branches seen but not trampled (outlets, alarms, *.alarm, commands, etc.) probably is not consistently addressed, but I have no good ways or ideas to even test for that.

The best I can say about it currently is that "symptomatically" by the posted PR codebase state (took a week to get here with all the back and forth), there were no issues seen like lost data points or spurious new ones, when tested against a single UPS with 4 MIBs, a daisy master ePDU and a directly addressable daisy slave ePDU of same model; but I have no firm idea if there are things I missed going just by symptoms.

aquette (commented):
scripts/subdriver/gen-snmp-subdriver.sh should be amended too

jimklimov:
At least, as of https://ci.networkupstools.org/job/nut/job/nut/job/PR-320/47/ the NUT CI farm liked this codebase per se, across the board :)

jimklimov:
Great catch, apparently I kept forgetting about the snmp/usb generator scripts in a few other PRs and had to post-process later :\

jimklimov:
After discussion with @aquette this is good to merge.

Experimentally the PR did not add regressions and did add features; the rest can be iterated later if found lacking: e.g. where I "fake" a `device.1` for non-templated stuff, so far that's for generic int/str data - can be done similarly (but is not currently), for earlier bifurcations about alarms etc.

PR #324, opened 2016-09-19
scan_xml_http.c : do not suggest NetXML support for ePDUs

jimklimov:
(almost exclusively they are XMLv4 and do not work with current NUT)

### DIFF
--- a/tools/nut-scanner/scan_xml_http.c
+++ b/tools/nut-scanner/scan_xml_http.c
@@ -270,7 +270,7 @@ nutscan_device_t * nutscan_scan_xml_http(long usec_timeout)
 				}
 				else
 				{
-					fprintf(stderr,"Device replied with NetXML but was not deemed compatible\n");
+					fprintf(stderr,"Device at IP %s replied with NetXML but was not deemed compatible with 'netxml-ups' driver (unsupported protocol version, etc.)\n", string);
 					continue; // skip this device; note that for broadcast scan there may be more in the loop's queue
 				}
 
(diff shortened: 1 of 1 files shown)


tools/nut-scanner/scan_xml_http.c
```diff
@@ -221,14 +257,22 @@ nutscan_device_t * nutscan_scan_xml_http(long usec_timeout)
...
+					nut_dev->port = strdup(buf);
 
-				current_nut_dev = nutscan_add_device_to_device(
+					current_nut_dev = nutscan_add_device_to_device(
 						current_nut_dev,nut_dev);
+				}
+				else
+				{
+					fprintf(stderr,"Device replied with NetXML but was not deemed compatible\n");
+					return NULL;
```
aquette: the above "return" statement aborts the XML scan at the first discovered ePDU and breaks the upstream scan (not unitary) and MUST be removed. Otherwise, we end up with a big regression.
I'm also in favor of replacing the fprintf with another upsdebugx, since it's not an error in essence but a known limitation.

jimklimov:
Good point, TODO ;)

--

19.09.2016 16:48 пользователь "Arnaud Quette" <notifications@github.com>
написал:

tools/nut-scanner/scan_xml_http.c
jimklimov: Should be fixed in https://github.com/networkupstools/nut/pull/324/commits/0535c01ef55eca3d35c16b5dab2c1baf3eb4b25f and improved in https://github.com/networkupstools/nut/pull/324/commits/1ea2f99a226629a21f55802e1333ab02e7a34e5b

jimklimov:
Note: PR #328 with further improvements to nut-scanner includes a more tested variant of this fix, so perhaps this PR #324 can be closed in favor of merging that one.

aquette:
Closing since #328 includes the present #324, as per @jimklimov comment

PR #342, opened 2016-11-21
Fix docgen ability - small fixes

jimklimov:
While setting up a local farm to build the code in different modes and OSes, stumbled upon a few complaints in the configure/make logs. These commits should fix them.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -724,7 +724,7 @@ if test -z "${abs_srcdir}" ; then
 		*)  abs_srcdir="$(cd "${srcdir}" && pwd)" ;;
 	esac
 fi
-DOCTESTDIR="$(mktemp -d --tmpdir="`pwd`" configure-test.docbuild.$$.XXXXXXX)" && \
+DOCTESTDIR="$(mktemp -d configure-test.docbuild.$$.XXXXXXX)" && \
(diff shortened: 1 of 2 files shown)


aquette:
@jimklimov instead of removing "--tmpdir="`pwd`"", it would be better to use "-p "`pwd`"

clepple:
@aquette see https://www.freebsd.org/cgi/man.cgi?query=mktemp&apropos=0&sektion=1&manpath=FreeBSD+11.0-stable&arch=default&format=html (similar on OS X)

aquette:
@jimklimov thanks for the pointer, I only looking at OpenBSD and Oracle, which support "-p"
http://docs.oracle.com/cd/E23824_01/html/821-1461/mktemp-1.html

still, for keeping the same behavior as currently, that form would be preferable:
mktemp -d "`pwd`/configure-test.docbuild.$$.XXXXXXX"

jimklimov:
Well, this is equivalent in practice - by default `mktemp` bases in current directory (everywhere I tried, and it was the botnet builds that uncovered this unportability issue). Even if some system uses e.g. `/tmp/` as the base location for such items, it is good enough for us anyway, as long as it would echo the full path to created objects.

PR #346, opened 2016-11-29
Make gen-nutupsconf-aug.py work with Python 3

### LINKED ISSUE #345: Python 2/3
Portions of NUT rely on the first `python` command in `$PATH` being Python 2. (Fortunately, this tends to only affects those building from Git, not from a source tarball.)

Before we jump in and upgrade everything to Python 3, let's summarize what the various distributions ship by default:

* Debian 8 (jessie): minimal install does not provide Python
  * `python-minimal`, which provides symlinks `/usr/bin/python2` and `/usr/bin/python`, depends on `python2.7-minimal` package, which provides `/usr/bin/python2.7`  
* Ubuntu 12.04 (precise): same as Debian, but `python-minimal` is installed by default
* Ubuntu 14.04 (trusty): same as Debian, but `python-minimal` is installed by default
* Ubuntu 16.04 (xenial): same as Debian (none installed by default)
* Arch (prior to 2010-10-18): no Python in 'base' group; `/usr/bin/python` pointed to python 2 (at that time `/usr/bin/python2.6`, not 2.7), now to python 3 (package 'python')
* Arch (2010-10-18 through at least 2016-11-28): no Python in 'base' group; python 2 can be installed with 'python2' package, which provides `/usr/bin/python2` symlink pointing to `/usr/bin/python2.7`.
* Mac OS X 10.11: `/usr/bin/python` points to `python2.7`; **no `python2` symlink**
  * Fink provides a `python2` symlink as part of the `python` meta-package

The short answer is that `python2.7` seems to be the only name that won't cause confusion.

We could probably run 2to3 and/or use `six` to be able to run on whatever version is available. Thoughts?

(If you have proposed changes, please commit them to a branch other than `master` first, thanks!)

**Edit:** incorporated https://github.com/networkupstools/nut/issues/345#issuecomment-263298858 and https://github.com/networkupstools/nut/issues/345#issuecomment-263409611

The-Compiler:
This makes the script work with Python 2.6+.
Fixes #345.

cc @zman0900

### DIFF
--- a/scripts/augeas/gen-nutupsconf-aug.py
+++ b/scripts/augeas/gen-nutupsconf-aug.py
@@ -18,6 +18,8 @@
 # This program extracts all drivers specific variables, declared
 # using 'addvar()' and output a complete ups.conf lens for Augeas
 
+from __future__ import print_function
+
 import sys
 import re
 import glob
@@ -69,7 +71,7 @@ if __name__ == '__main__':
 
 	if (len(sys.argv) == 2):
 		dirPrefix = sys.argv[1]
-		print dirPrefix
+		print(dirPrefix)
 
 	# 1/ Extract all specific drivers parameters, in a sorted list with unique entries
 	# 1.1/ List all drivers implementation files
--- a/tools/gitlog2changelog.py
+++ b/tools/gitlog2changelog.py
@@ -80,7 +80,7 @@ for line in fin:
             else:
                 message = message + " " + line.strip()
     # If this line is hit all of the files have been stored for this commit
-    elif re.search('files? changed', line) >= 0:
+    elif re.search('files? changed', line):
         filesFound = True
         continue
     # Collect the files for this commit. FIXME: Still need to add +/- to files


clepple:
Thanks! (If only my other 2-to-3 conversion projects were that simple...)

Since this will automatically close the parent issue when merged, I think we need to check and make sure that all of the other instances of scripts invoking `python` are fixed (hence the WIP tag). Have you had a chance to compare the rest of the output of the build process?

 * `scripts/python/module/PyNUT.py` (not build-related, but also invokes `python`)
 * `tools/gitlog2changelog.py`
 * `tools/nut-snmpinfo.py`

The-Compiler:
- `scripts/python/module/PyNUT.py` seems to run (and the changelog claims it supports python 3)
- `tools/gitlog2changelog.py` has been fixed in 7bf209a - the output is line-wrapped slightly differently though, but I haven't tracked this down (I'd guess it's fine?)
- `tools/nut-snmpinfo.py` seems to work fine

What about those?

- `scripts/Solaris/precheck.py` (broken)
- `scripts/Solaris/prepackage.py` (broken)
- `scripts/python/module/test_nutclient.py`(seems to work)

aquette:
@jimklimov could you please check the scripts/Solaris/pre{check,package}.py as mentioned above by @The-Compiler ? you've probably been looking at those already, or otherwise we may want to remove these and get from a new ground (still tied to you ;)). Thanks!

jimklimov:
Just in case, what and where should I check? :)

The first part seems easier: `git checkout The-Compiler:augeas-py3` and try to configure and run it on a bunch of machines?

As far as relevant platforms go, there is Solaris 10u8 (last released by Sun on more or less free-to-use conditions). then there are later releases of Oracle Solaris 10u*, 11.{0,1,2,3} and some beta of 12, and various open descendants of OpenSolaris, the popular general-purpose ones being OpenIndiana Hipster (bleeding-edge rolling-release) and OmniOS (several releases). Some customers may quite run Solaris 8 or 9 still.

Oh, and also we have an AIX 6.1 and 7.1 to try and feed this to ;)

Solaris 10 has Python 2.4, illumos opensourced ones have 2.6/2.7 (in progress of eradicating 2.6, but not all finished yet), and may have 3.4 or so as an optional package with which not everything works...

The-Compiler:
The main question is whether those need to work on Python 3 - I guess not?

jimklimov:
Well, it depends. Many of them do not have python at all, or not python3. Those that do, do not have a `python2` symlink out of the box, but a `/usr/bin/python` pointing right to `./python2.7` or whichever the chosen implementation is. On OI/Hipster at least there is a pre-packaged `/usr/bin/python3` symlink however.

So I guess the main goals are to:

* work where python3 is default
* work where python2 is default
* work where python is not available

Maybe different levels of "work" are achievable given the (lack of) tools, so perhaps some prepared files should be available in tarballs and/or git, so these other systems can still compile NUT as well - just not recompile everything from origins, It may well be up to `autogen.sh` to detect the situation and call the proper python interpreter perhaps for a customized version of the script, or something like that.

Or abandon python and rewrite stuff in POSIX shell, sed, grep... =D

The-Compiler:
FWIW those scripts would be trivial to reimplement as a shellscript (I'm not volunteering :laughing:), but at the moment I think it's fine to keep them as-is (with a `python` shebang and python2-only) as we know it'll only run on Solaris (right?).

This should be ready to merge from my POV then.

clepple:
@The-Compiler meant to merge this earlier - thanks again! (We can open a new issue for Solaris if needed.)

PR #352, opened 2017-01-04
Travis CI matrix builds

jimklimov:
This extends the basics of PR #351 and defines several build scenarios that would run in parallel - and provides a template for future similar customized tests. As detailed in issue #338 this adds several build scenarios to quickly see if a coding/Makefile error breaks the repo:
* a "default-nodoc" build takes about a minute (just build and install - no docs, no distcheck)
* a "default" build (just manpages and a distcheck) is over 2 minutes
* a "default-withdoc" (all formats and distcheck) is over 5 minutes
* a "default-alldrv" (no doc => no distcheck, but otherwise `--with-all`) is just a bit longer than "default-nodoc"

These times might be improved by integrating the changesets (PR #353) which added support for `--with-doc=skip` so that a (limited) `make distcheck` can still be done and uncommenting the prepared lines; some of the benefit will come from not building even the manpages, and some will come from not installing the doc-processing packages in builds that don't care about them.

After this is integrated, development in branches will have an easy way to define tests for features brought by the branch (e.g. some custom scenarios for DMF that are not relevant in codebases that do not have it yet).

PR #305, opened 2016-08-15
DMF code and recipes rearranged to compile cleaner as part of NUT, again

jimklimov:
Major change - LUA code and compiler flags are now compartmentalized, so the project can again be built both with and without LUA capabilities, and successfully builds to the end and regenerates DMF files from scratch without unexpected differences from the Git-saved reference copies. Many temporary or example files from development process are now removed or relocated.

There is still a lot to clean up and revise, but this set of issues was a major roadblock before upstreaming can be discussed :)

(NOTE: This PR duplicates https://github.com/eaton-bob/nut/pull/117 in the original DMF team development repo)

PR #353, opened 2017-01-04
configure.ac, docs/man/Makefile.am - support SKIP_MANS as an explicit non-fatal situation

jimklimov:
Extracted from DMF PR #305 as beneficial in particular for improvements after PR #352 gets accepted.


jimklimov:
Closed automatically as the PR #354 was merged and includes this commit.

PR #354, opened 2017-01-04
Travis matrix builds with docskip (PRs 352+353 synergized together)

jimklimov:
Combining the features of PR #352 and #353, this one offers builds that are faster thanks to not installing the doc-processing packages and not generating any docs in most cases; it does however enable distcheck for the `default-alldrv` case so it became a bit slower than in PR #352 by doing twice the work (but being on a safer side now).

Compare the results at https://travis-ci.org/jimklimov/nut/builds/188979713 (`--with-doc=skip`) and https://travis-ci.org/jimklimov/nut/builds/188980602 (without skipping).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.l[oa]
 .deps/
 .libs/
+.inst/
 Makefile
 Makefile.in
 ## Parent directory only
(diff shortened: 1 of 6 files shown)


jimklimov:
Enabled monitoring of `networkupstools/nut` in Travis CI, now merging the PR to have it roll.

jimklimov:
Ok, and it is live (and green) at https://travis-ci.org/networkupstools/nut ! :)

PR #357, opened 2017-01-06
Refine ST_FLAG_NUMBER-related behaviours

zykh:
Follow-up to #265/#267:
- explicitly setting `ST_FLAG_NUMBER` for ranges through `dstate_setflags()` causes the elimination of previously-set flags, inducing some side effects (e.g. if a var is first set as r/w and then its range is added -> the var loses its r/w status and becomes unchangeable): this PR introduces a support function to add flags (and its counterpart to remove flags) without losing the existing ones.
- also, a couple of remaining bits were not updated for `ST_FLAG_NUMBER`.

### DIFF
--- a/common/state.c
+++ b/common/state.c
@@ -424,6 +424,11 @@ void state_setflags(st_tree_t *root, const char *var, int numflags, char **flag)
 			continue;
 		}
 
+		if (!strcasecmp(flag[i], "NUMBER")) {
+			sttmp->flags |= ST_FLAG_NUMBER;
+			continue;
+		}
+
 		upsdebugx(2, "Unrecognized flag [%s]", flag[i]);
 	}
 }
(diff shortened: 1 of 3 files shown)


clepple:
I will admit that I have not been following this very closely, and I think that eventually we should have some test cases that exercise this code. But given the general lack of C unit-test infrastructure, don't let that stop you from merging.

@aquette, any thoughts?

aquette:
@clepple fully agree with the need to improve tests in this area (was thinking the exact same yesterday).
I've logged an issue to track that.
As for the current issue, I was not able to test, so @zykh if you made the necessary test (and I'm sure you did), I trust you, so don't hesitate to go ahead and merge.

zykh:
Thanks to all, rebased and merged.

PR #360, opened 2017-01-11
Resynced the upstream/DMF with news from upstream/master as of 2017-01-11

jimklimov:
Just a bump to facilitate subsequent development and review (e.g. daisychain code merged earlier as part of DMF too is obsolete == different in master now, so was fixed here...)

NOTE: not sure this state of "DMF" code will actually work, but at least unlike PR #305 this here is a minor change to what was accepted and merged as `upstream/DMF` before.

FWIW, a snapshot of current state of `upstream/DMF` is preserved in https://github.com/jimklimov/nut/tree/DMF-master-20160715 - and note that unlike the recently carefully re-merged PR #305, this code does not seem to apply cleanly to current `upstream/master`; this is a bump over `upstream/DMF` specifically (at least as far as Github web interface reports...)


jimklimov:
Too many fixes needed to just get the "upstream/DMF" branch to build (and it is even not the state-of-the-art code, but something that the green and tested PR #305 should merge into) :\

This work helped fix a few wrinkles on main codebase though...

Issue #338, opened 2016-11-04
Continuous Integration: setup Travis through Github for NUT


aquette:
Beside for our historic Buildbot, Github provides 3rd party integration with some apps, including Travis.

Some tests that can be part of the scope:
* [ ] QRT scripts: test the whole NUT framework, including protection / shutdown, using dummy-ups, (need integration, see #3)
* [ ] memcheck target (needs creation) using valgrind, on a few components (I know, the best would be through endurance testing, so some iterations may be needed to improve this)
* [ ] "check" target: still limited currently to libnutclient unit test, using cppunit
* [ ] distcheck / distcheck-light
* [ ] [SonarQube](https://sonarqube.com/): subject to another upcoming ticket (which triggered the present ticket creation...). Would be better than the closed-source Coverity Scan 
* [ ] spellcheck
* [ ] (...) anything else that can be tested and would be worth to test

A potential point could be to create a [dedicated CI user](https://docs.travis-ci.com/user/private-dependencies/#Dedicated-User-Account) for automated reports and bugs filling. See also http://docs.sonarqube.org/display/PLUG/GitHub+Plugin?preview=/5311422/5636115/PullRequestAnalysis.png

@clepple especially, I'd like your feedback / approval prior to moving on that topic. I don't want you to perceive this as killing buildbot and all the energy you put into it... But the present approach would share actions with the team and allows us to give more public visibility while making it easier to increase our CI scope.

clepple:
TL;DR: don't let me stop you.

As has been pointed out several times on the Buildbot mailing lists, there are things that should be done in the code with scripts and makefile targets, and there are things that the CI tool should do. If there is overlap, the CI tool should invoke the script or call make. (It's sort of like separating content from presentation in HTML, or the MVC layers in applications.)

For NUT, I think this means that about half of the suggestions you mentioned are applicable to any CI system (or more accurately, can be invoked by any CI system). More importantly, any user can invoke just the scripts they need while developing on their own system. I use Buildbot for other projects because it is self-hosting and does not depend on an external service, so I don't see myself doing much with Travis - there is too much overlap for little gain.

There is also a [buildbot_travis](https://github.com/buildbot/buildbot_travis) bridge that I haven't looked into much, but seems to build based off of the same `.travis.yml` file that travis-ci.org uses. There is some value to keeping the build configuration in the repository itself, but we have been doing that for years with `autogen.sh` and `configure.ac`.

[...]

PR #363, opened 2017-01-19
Travis spellcheck and cppunit with valgrind

jimklimov:
For issue #338 

This enables Travis tests for cppunit (to whatever extent was available in codebase, which was not much), and fixes spellchecking to do its work (and also fixes a number of discovered spelling errors).

Also adds configure capability to find and optionally enable valgrind, but currently does not go beyond that (e.g. enable debug builds and/or run any special checks under valgrind... not sure how/what to test here - start every binary to display help text and see that nothing went wrong? the commented-away code for travis enablement was slightly improved but is still commented away). This piece is ported from DMF branch which does have a target to make a custom build of everything it needed to test that the parser does not leak.

UPDATE: For lack of better ideas at the moment, added a distcheck taryet with valgrind support that runs the cppunit test programs with leak inspection. This approach can be replicated to add other types of 'make check' as they come.


jimklimov:
Ok, so after some debugging to drop later - the mismatch is in aspell shipped dictionaries on travis hosts == more exceptions needed in nut.dict.

aquette:
Cool, so that worked with modifying the first line of nut.dict with utf8 ;)
Thx for all your work on NUT @jimklimov
Happy to finally see the actual use of all these personal nights of
works!!!!
I owe you some beers ... or vodka ;)

jimklimov:
I did not yet experiment with utf8 header nor other options, so contributor
names are still a problem :(

--

19.01.2017 23:11 пользователь "Arnaud Quette" <notifications@github.com>
написал:

jimklimov:
Seems finally fixed for "UTF8" input and dicts (just one Pawel with polish L got missed by aspell and is now seen as Pawe with a skipped next character that is otherwise deemed invalid in a word), so un-WIPing the PR.

jimklimov:
@aquette : PS: I am not quite a vodka guy, but I get the drift ;)
Kudos to you too for getting a solid foundation (and hints) without which I wouldn't get so far with the spells.

docs/Makefile.am
```diff
@@ -125,17 +125,75 @@ SPELLCHECK_SRC = $(ALL_TXT_SRC) ../README ../INSTALL.nut ../UPGRADING  ../NEWS \
...
+# international committers). The grep tests below look if the aspell output
+# contained something other than the OK lines (tagged with asterisk) and
+# aspell's version (tagged with @) and if it did - those lines must be the
+# spellcheck complaints. Empty OUT is ok.
+# We also must indent the input, because certain piped-in characters are
+# interpreted as commands, and seems this feature can not be turned off.
+# See also http://aspell.net/man-html/Through-A-Pipe.html
+# TODO: Is "grep -a" or "grep -b" (treat input as ascii/bin) portable enough?
+# Is "egrep == grep -E" always valid? (maybe all a job for configure.ac)
+# TODO: Define a custom language in-repo or during build, that is english
```
aquette: is this and the below still valid?

docs/Makefile.am
```diff
@@ -125,17 +125,75 @@ SPELLCHECK_SRC = $(ALL_TXT_SRC) ../README ../INSTALL.nut ../UPGRADING  ../NEWS \
 	../TODO ../scripts/ufw/README ../scripts/augeas/README ../lib/README \
 	../tools/nut-scanner/README
```
aquette: add daisychain.txt to ALL_TXT_SRC since it's missing

aquette:
@jimklimov great work! and thx for the thank ;)
on the valgrind, we can improve that when integrating QRT and use it (similar approach, with dummy-ups+upsd+upsmon+upsc/rw) to check for leaks on these. Nearly impossible for the drivers... at least for now.

docs/Makefile.am
jimklimov: Custom language - probably not anymore, at least not until we hit a platform with sufficiently different or old aspell.

Concerns of portable `egrep` and `-a/-b` flags - these may still be valid (buildbots might show what works where). If the issues do pop up, the flag may be removed at expense of warnings that unexpected binary input was detected - and those might need to be grepped out so as to not pollute logged reports and/or interpretation of aspell's "$OUT".

docs/Makefile.am
jimklimov: done

jimklimov:
Good point on manpages - done. No copy-paste involved ;)

jimklimov:
For reviewers : so far my fantasy is exhausted as far as "small initial steps" go, and this PR looks solid enough for merging, as a foundation for possible (or inevitable) improvements. I'd like to avoid it growing into a monster like DMF PRs that take months to review, and to begin bringing benefits to subsequent codebase improvements ASAP ;)

aquette:
@jimklimov thx, you've addressed all the points from my review. Good to go, and as you said, this is already improving the situation, along with being a solid foundation to build upon.
Now merging / closing the PR

PR #371, opened 2017-01-27
Spellcheck tech improvements

jimklimov:


README
```diff
@@ -1,6 +1,8 @@
 Network UPS Tools Overview
 ===========================
 
+image:https://travis-ci.org/networkupstools/nut.svg?branch=master["Build Status", link="https://travis-ci.org/networkupstools/nut"]
```
clepple: At the moment, I don't think this will render on GitHub. See discussion as to why we haven't renamed it: https://github.com/networkupstools/nut/pull/226

Also, this ties the image to the master branch, which would be misleading for feature branches or released documentation (As I recall, README is included in the user manual.)

jimklimov: Interesting point... ok, I guess I'll drop this one ;)

jimklimov:
@clepple, Was the banner in README the only nit? Can the rest be merged? :)

clepple:
It would be nice to have time to look over the rest in more detail, but I suppose I can always do that later. Nothing else stands out.

PR #376, opened 2017-01-28
Document and pre-set the "maxretry" option in ups.conf.sample

jimklimov:
Upstreaming fixes applied to our project at work

By: ???

### DIFF
--- a/conf/ups.conf.sample
+++ b/conf/ups.conf.sample
@@ -40,6 +40,24 @@
 #
 # Configuration directives
 # ------------------------
+#
+# These directives are used by upsdrvctl only and should be specified outside
+# of a driver definition:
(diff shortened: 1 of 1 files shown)


conf/ups.conf.sample
```diff
@@ -117,3 +135,6 @@
 #
 # To find out if your driver supports any extra settings, start it with
 # the -h option and/or read the driver's documentation.
+
+# Set maxretry to 3 by default, this should mitigate race with slow devices:
+maxretry = 3
```
clepple: I would recommend putting the global `maxretry` before the sample UPS configuration sections.

jimklimov: done

PR #390, opened 2017-02-09
mge-xml.c : calculate (input|input.bypass|output).phases number (1 or 3)…

jimklimov:
…based on available non-zero measurements

NOTE: The NetXML specs for XMLv3 and XMLv4 do not mention a separate value for "phases" and devices do just return the per-line values (non-zero voltages compared to neutral and among lines, as well as an overall input Voltage value), and currents (value depends on actually connected loads).

Comments and questions to reviewers and testers:

As seen in the patch, the added logic is to wait until we've completed parsing a `GET_OBJECT` response (where we have dstate's set for all discovered values), and set states for the `(input|bypass|output).phases` based on querying the per-line data and getting non-zero numeric replies (and in fact any replies at all).

In effect this can mean that a device claiming (or really having) 3 phases but running with input or load connected on just one phase vs. neutral at the moment of driver startup, would be considered and remembered as a 1-phase device until the driver restart. On the other hand, some UPSes might support both 1- and 3-phase connectivity, so their firmware would expose 3 but practical usage will be for 1...

TODO: Maybe this should be occasionally reconsidered during driver run-time (e.g. if we remember 1 phase, still do check if we get data proving that we became 3-phase)?

Also, no definite dstate value is published for these items until we have definitely decided that at some point in time there was proof of either 1- or 3-phase usecase, This is a conscious choice - as it is better so than publishing e.g. hardcoded 1 phase from the start, and then connecting to the actual device and changing the mind (and reported value) after a minute or so.

Tested with both 1- and 3-phase devices:

````
$ make && sudo ./drivers/netxml-ups -a 93pm1 -u root -DDDDD 2>&1 | ggrep 'calculated non-XML value for NUT '
...
   0.747425	[D3] -> calculated non-XML value for NUT variable input.phases was set to 3
   0.757701	[D3] -> calculated non-XML value for NUT variable input.bypass.phases was set to 3
   0.747523	[D3] -> calculated non-XML value for NUT variable output.phases was set to 3
````

````
$ make && sudo ./drivers/netxml-ups -a 9px -u root -DDDDD 2>&1 | ggrep 'calculated non-XML value for NUT '
...
   1.045194	[D3] -> calculated non-XML value for NUT variable input.phases was set to 1
   1.045274	[D3] -> calculated non-XML value for NUT variable input.bypass.phases was set to 1
   1.046301	[D3] -> calculated non-XML value for NUT variable output.phases was set to 1
````

Another note: I saw that many drivers implement something of this sort, often with quirks like "this additional logic applies to this device model". So, refactoring to share some of the code and logic makes sense, but is out of scope for this PR :)

### DIFF
--- a/drivers/dstate.c
+++ b/drivers/dstate.c
@@ -1058,7 +1058,7 @@ int dstate_detect_phasecount(
 		           *v1n, *v2n, *v3n,
 		           *v12, *v23, *v31,
 		           *c1,  *c2,  *c3,  *c0;
-		char buf[80]; /* For concatenation of "xput_prefix" with items we want to query */
+		char buf[MAX_STRING_SIZE]; /* For concatenation of "xput_prefix" with items we want to query */
 		size_t xput_prefix_len;
 		int bufrw_max;
 		char *bufrw_ptr = NULL;
@@ -1080,6 +1080,7 @@ int dstate_detect_phasecount(
 			upsdebugx(0, "%s(): Bad xput_prefix was passed: it is too long - function skipped", __func__);
 			return -1;
 		}
+		memset(buf, 0, sizeof(buf));
 		strncpy(buf, xput_prefix, sizeof(buf));
 		bufrw_ptr = buf + xput_prefix_len ;
 
--- a/drivers/dstate.c
+++ b/drivers/dstate.c
@@ -1048,25 +1048,29 @@ int dstate_detect_phasecount(
 		           *v12, *v23, *v31,
 		           *c1,  *c2,  *c3,  *c0;
 		char buf[80]; /* For concatenation of "xput_prefix" with items we want to query */
+		size_t xput_prefix_len;
+		int bufrw_max;
+		char *bufrw_ptr = NULL;
+
 		if (!xput_prefix) {
 			upsdebugx(0, "%s(): Bad xput_prefix was passed: it is NULL - function skipped", __func__);
 			return -1;
 		}
 
-		size_t xput_prefix_len = strlen(xput_prefix);
+		xput_prefix_len = strlen(xput_prefix);
 		if (xput_prefix_len < 1) {
 			upsdebugx(0, "%s(): Bad xput_prefix was passed: it is empty - function skipped", __func__);
 			return -1;
 		}
 
-		int bufrw_max = sizeof(buf) - xput_prefix_len;
+		bufrw_max = sizeof(buf) - xput_prefix_len;
 		if (bufrw_max <= 15) {
 			/* We need to append max ~13 chars per below, so far */
 			upsdebugx(0, "%s(): Bad xput_prefix was passed: it is too long - function skipped", __func__);
 			return -1;
 		}
 		strncpy(buf, xput_prefix, sizeof(buf));
-		char *bufrw_ptr = buf + xput_prefix_len ;
+		bufrw_ptr = buf + xput_prefix_len ;
 
 		/* We either have defined and non-zero (numeric) values below, or NULLs */
 		strncpy(bufrw_ptr, "L1.voltage", bufrw_max);
--- a/drivers/dstate.c
+++ b/drivers/dstate.c
@@ -1077,34 +1077,30 @@ int dstate_detect_phasecount(
 		bufrw_ptr = buf + xput_prefix_len ;
 
 		/* We either have defined and non-zero (numeric) values below, or NULLs */
-		strncpy(bufrw_ptr, "L1.voltage", bufrw_max);
-		if ((v1 = dstate_getinfo(buf)))    { if (v1[0]  == '0') { v1  = NULL; } }
-		strncpy(bufrw_ptr, "L2.voltage", bufrw_max);
-		if ((v2 = dstate_getinfo(buf)))    { if (v2[0]  == '0') { v2  = NULL; } }
-		strncpy(bufrw_ptr, "L3.voltage", bufrw_max);
-		if ((v3 = dstate_getinfo(buf)))    { if (v3[0]  == '0') { v3  = NULL; } }
-		strncpy(bufrw_ptr, "L1-N.voltage", bufrw_max);
-		if ((v1n = dstate_getinfo(buf)))   { if (v1n[0] == '0') { v1n = NULL; } }
-		strncpy(bufrw_ptr, "L2-N.voltage", bufrw_max);
-		if ((v2n = dstate_getinfo(buf)))   { if (v2n[0] == '0') { v2n = NULL; } }
-		strncpy(bufrw_ptr, "L3-N.voltage", bufrw_max);
-		if ((v3n = dstate_getinfo(buf)))   { if (v3n[0] == '0') { v3n = NULL; } }
-		strncpy(bufrw_ptr, "L1-L2.voltage", bufrw_max);
-		if ((v12 = dstate_getinfo(buf)))   { if (v12[0] == '0') { v12 = NULL; } }
-		strncpy(bufrw_ptr, "L2-L3.voltage", bufrw_max);
-		if ((v23 = dstate_getinfo(buf)))   { if (v23[0] == '0') { v23 = NULL; } }
-		strncpy(bufrw_ptr, "L3-L1.voltage", bufrw_max);
-		if ((v31 = dstate_getinfo(buf)))   { if (v31[0] == '0') { v31 = NULL; } }
-		strncpy(bufrw_ptr, "L1.current", bufrw_max);
-		if ((c1 = dstate_getinfo(buf)))    { if (c1[0]  == '0') { c1  = NULL; } }
-		strncpy(bufrw_ptr, "L2.current", bufrw_max);
-		if ((c2 = dstate_getinfo(buf)))    { if (c2[0]  == '0') { c2  = NULL; } }
-		strncpy(bufrw_ptr, "L3.current", bufrw_max);
-		if ((c3 = dstate_getinfo(buf)))    { if (c3[0]  == '0') { c3  = NULL; } }
-		strncpy(bufrw_ptr, "voltage", bufrw_max);
-		if ((v0 = dstate_getinfo(buf)))    { if (v0[0]  == '0') { v0  = NULL; } }
-		strncpy(bufrw_ptr, "current", bufrw_max);
-		if ((c0 = dstate_getinfo(buf)))    { if (c0[0]  == '0') { c0  = NULL; } }
+#define dstate_getinfo_nonzero(var, suffix) \
+		{ strncpy(bufrw_ptr, suffix, bufrw_max); \
+		  if ( (var = dstate_getinfo(buf)) ) { \
+		    if (var[0] == '0') { \
+		      var = NULL; \
+		    } \
+		  } \
+		} ;
+
+		dstate_getinfo_nonzero(v1,  "L1.voltage");
+		dstate_getinfo_nonzero(v2,  "L2.voltage");
+		dstate_getinfo_nonzero(v3,  "L3.voltage");
+		dstate_getinfo_nonzero(v1n, "L1-N.voltage");
+		dstate_getinfo_nonzero(v2n, "L2-N.voltage");
+		dstate_getinfo_nonzero(v3n, "L3-N.voltage");
+		dstate_getinfo_nonzero(v1n, "L1-N.voltage");
+		dstate_getinfo_nonzero(v12, "L1-L2.voltage");
+		dstate_getinfo_nonzero(v23, "L2-L3.voltage");
+		dstate_getinfo_nonzero(v31, "L3-L1.voltage");
+		dstate_getinfo_nonzero(c1,  "L1.current");
+		dstate_getinfo_nonzero(c2,  "L2.current");
+		dstate_getinfo_nonzero(c3,  "L3.current");
+		dstate_getinfo_nonzero(v0,  "voltage");
+		dstate_getinfo_nonzero(c0,  "current");
 
 		if ( (v1 && v2 && v3) ||
 		     (v1n && v2n && v3n) ||
--- a/drivers/dstate.c
+++ b/drivers/dstate.c
@@ -1076,11 +1076,21 @@ int dstate_detect_phasecount(
 		strncpy(buf, xput_prefix, sizeof(buf));
 		bufrw_ptr = buf + xput_prefix_len ;
 
-		/* We either have defined and non-zero (numeric) values below, or NULLs */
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by jimklimov** [comment]
Ok, so a bit of refactoring happened, and now the bulk of the code is part of dstate.c/.h and can be used by other consumers, including a readonly mode (not setting dstate even if there's an urge to do so) and stateless mode (do not necessarily retain/flag the successful detection into caller's state vars).

Hope the cellnet will soon allow to push the update to PR...

File: drivers/dstate.c [hunk 1]
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+				}
+				return 1;
+			}
+		}
+
+		upsdebugx(5, "set_phasecount(): Nothing changed: could not determine a value");
+		return 0;
+	}
+
+	upsdebugx(5, "set_phasecount(): Nothing changed, with a valid reason; already inited");
```
zykh: Why not `__func__`?

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+					/* Technically, a bit has changed: we have set the flag which may have been missing before */
+					upsdebugx(5, "set_phasecount(): Nothing changed, with a valid reason; dstate already published with the same value: %s=%s (detected %d)", buf, oldphases, *num_phases);
+					return 3;
+				}
+			}
+
+			if ( (*num_phases != old_num_phases) || (!oldphases) ) {
+				if (may_change_dstate) {
+					dstate_setinfo(buf, "%d", *num_phases);
+					upsdebugx(3, "-> calculated non-XML value for NUT variable %s was set to %d",
```
zykh: Maybe it would be better to always add the function name... -- in future, we may decide to change the debug level and end up with 2 different debug levels for the `Entering ...` message and the other debug messages.

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+			/* We need to append max ~13 chars per below, so far */
+			upsdebugx(0, "set_phasecount(): Bad xput_prefix was passed: it is too long - function skipped");
+			return -1;
+		}
+		strncpy(buf, xput_prefix, sizeof(buf));
+		char *bufrw_ptr = buf + xput_prefix_len ;
+
+		/* We either have defined and non-zero (numeric) values below, or NULLs */
+		strncpy(bufrw_ptr, "L1.voltage", bufrw_max);
+		if ((v1 = dstate_getinfo(buf)))    { if (v1[0]  == '0') { v1  = NULL; } }
```
zykh: I'm the first one not to follow our coding style guide blindly, so I'm not entitled at all to speak about this topic.
But, hey, were I to nitpick, then... If I recall correctly our guidelines recommend against multiple statements per line. But I see what you're doing, so, if you want to retain these oneliners, perhaps, something like `if (!(var = dstate_getinfo(...)) || !*var) var = NULL;` could be slightly better, at least it avoids the nested `if`.

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+			return -1;
+		}
+
+		size_t xput_prefix_len = strlen(xput_prefix);
+		if (xput_prefix_len < 1) {
+			upsdebugx(0, "set_phasecount(): Bad xput_prefix was passed: it is empty - function skipped");
+			return -1;
+		}
+
+		int bufrw_max = sizeof(buf) - xput_prefix_len;
```
zykh: Still nitpicking on the guidelines... keep the declaration of vars at the beggining of the scope (`size_t xput_prefix_len`, `int bufrw_max` and `char *bufrw_ptr`, I'm looking at you).

drivers/dstate.c
jimklimov: Yes, especially in original form (without refactoring to use buf) these were one-liners and so much more manageable as single lines regarding copypaste etc.

Also note that the check above is for character zero, so `!*var` (numeric zero) will not quite fit. And I'm not sure if it is safe regarding older/vendor compilers to not get a NULL pointer dereference with this collapse to a single clause.

Looking at the repetitions here, might as well macro this line, though - and write that macro formatted to style ;)

drivers/dstate.c
jimklimov: Hm... I had it at some point, must have fallen out during branch-switch.

drivers/dstate.c
jimklimov: At least in this particular case, formatting fir the report came from copying in the mge-xml.c loops.

jimklimov: At least in this particular case, formatting for the report came from copying in the mge-xml.c loops.

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+		     (v0 && !v1 && !v2 && !v3) ||
+		     (c1 && !c2 && !c3) ||
+		     (!c1 && c2 && !c3) ||
+		     (!c1 && !c2 && c3) ||
+		     (v1 && !v2 && !v3) ||
+		     (!v1 && v2 && !v3) ||
+		     (!v1 && !v2 && v3) ||
+		     (v1n && !v2n && !v3n) ||
+		     (!v1n && v2n && !v3n) ||
+		     (!v1n && !v2n && v3n) ) ) {
```
jimklimov: Interesting point... initially (in earlier commits) the logic was all around voltages. From what I gather, on a 3-phase capable device with just 1 phase connected to active load (others not connected or loads not powered on), there is still a voltage reading but negligible/zero current (and realpower, probably).

So my checks revolved around devices reporting non-zero value on just one of the line-metrics (since a single-phase load might be wired to any one phase).

The tests above for `!c0 && !v0` are not correct however, because in the XML for a 3-phase capable device I see both an overall metric and per-phase values - so for the disbalanced usecase (or one reported by a general form of 3-phase markup) we'd see both `v0` and one of those other voltages.

Otherwise, your form (and comment) gives good food for thought, maybe I'll get back to think it over after some rest (e.g. adopt this solution, or really prove it wrong), or maybe someone else will :)

drivers/dstate.c
jimklimov: done

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+			*num_phases = 1;
+			*inited_phaseinfo = 1;
+			detected_phaseinfo = 1;
+			upsdebugx(5, "set_phasecount(): determined a 1-phase case");
+		} else{
```
jimklimov: done

drivers/dstate.c
jimklimov: done

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
 		dstate_delinfo(info_name);
 	}
 }
+
+/* For devices where we do not have phase-count info (no mapping provided
+ * above), nor in the XML, we can guesstimate and report a value. This may
```
jimklimov: done

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+		if ((v2n = dstate_getinfo(buf)))   { if (v2n[0] == '0') { v2n = NULL; } }
+		strncpy(bufrw_ptr, "L3-N.voltage", bufrw_max);
+		if ((v3n = dstate_getinfo(buf)))   { if (v3n[0] == '0') { v3n = NULL; } }
+		strncpy(bufrw_ptr, "L1-L2.voltage", bufrw_max);
+		if ((v12 = dstate_getinfo(buf)))   { if (v12[0] == '0') { v12 = NULL; } }
+		strncpy(bufrw_ptr, "L2-L3.voltage", bufrw_max);
+		if ((v23 = dstate_getinfo(buf)))   { if (v23[0] == '0') { v23 = NULL; } }
+		strncpy(bufrw_ptr, "L3-L1.voltage", bufrw_max);
+		if ((v31 = dstate_getinfo(buf)))   { if (v31[0] == '0') { v31 = NULL; } }
+		strncpy(bufrw_ptr, "L3.current", bufrw_max);
```
jimklimov: Thanks! :) done

drivers/dstate.c
jimklimov: Refactored, looks better and still does get the job done, thanks!

drivers/dstate.c
```diff
@@ -1002,3 +1002,163 @@ void device_alarm_commit(const int device_number)
...
+		strncpy(bufrw_ptr, "voltage", bufrw_max);
+		if ((v0 = dstate_getinfo(buf)))    { if (v0[0]  == '0') { v0  = NULL; } }
+		strncpy(bufrw_ptr, "current", bufrw_max);
+		if ((c0 = dstate_getinfo(buf)))    { if (c0[0]  == '0') { c0  = NULL; } }
+
+		if ( (v1 && v2 && v3) ||
+		     (v1n && v2n && v3n) ||
+		     (c1 && (c2 || c3)) ||
+		     (c2 && (c1 || c3)) ||
+		     (c3 && (c1 || c2)) ||
```
jimklimov: Yes, mathematically your version is correct and equivalent :)

But I feel mine is clearer to read. And for review (or authorship) any aid to avoid a stupid mistake and not outsmart oneself or others is useful :)

jimklimov:
Thanks @zykh for the review, much appreciated, most items already fixed :)

Possibly more attention is welcome at the logic where we determine that the device has certainly a single-phase connection/load.

drivers/dstate.c
jimklimov: Fixed to match other reports in this function - this was a local sample of formatting from mge-xml.c not seen elsewhere in the codebase.

drivers/dstate.c
zykh: Oops, I didn't notice the quotes.

I didn't even consider the scenario of a driver/device reporting a `0` (or near zero) value that has to be read as 'not-used', but now, if I understand what you're doing, I have to raise another concern of mine: some drivers use what they get from the device verbatim and some devices report values padded with zeroes or spaces (you may get wonders like `0200` or `<one or more spaces>0`), so a simple check for the first character to be not `0` is not enough to guarantee that the value is not zero.

Maybe we should convert what we get to a number and see if it is (near) zero (and, given the increased complexity, perhaps a dedicated function would be better than a macro).

jimklimov:
I'd first say if drivers do that, it is a bug there that should get fixed
there.

On second thought, we expect numbers in this context, right? And
technically `[+-][0\.][0-9]` are also still valid numerals, that might
resolve to zero (exact or within an epsilon)... But still, i'd wager that
numbers should be normalized, so a zero string is '0' '\0' - for the sake
of consumers like this one (e.g. some scripts may think that leading zeroes
mean octal notation, etc.)

--

10.02.2017 22:51 пользователь "Daniele Pezzini" <notifications@github.com>
написал:

jimklimov:
Also note that here we process strings saved into dstate's, not raw driver
responses, so arguably NUT should have taken care of normalization (or is
in good position to do so via conversion functions).

--

10.02.2017 22:51 пользователь "Daniele Pezzini" <notifications@github.com>
написал:

zykh:
> I'd first say if drivers do that, it is a bug there that should get fixed there.

@jimklimov, I agree with you that, ideally, drivers should not do that, but NUT has been around for several years and some drivers have not been touched for a while and I'm pretty sure that there are drivers that do exactly that.

> On second thought, we expect numbers in this context, right? And technically `[+-][0\.][0-9]` are also still valid numerals, that might resolve to zero (exact or within an epsilon)... But still, i'd wager that numbers should be normalized, so a zero string is '0' '\0' - for the sake of consumers like this one (e.g. some scripts may think that leading zeroes mean octal notation, etc.)

> Also note that here we process strings saved into dstate's, not raw driver responses, so arguably NUT should have taken care of normalization (or is in good position to do so via conversion functions).

OK, but this is not the case for now: at the moment, if my memory serves me well, dstate / state are quite permissive and no checks are operated on values when you add a var or change it (if I recall correctly, you can even use an empty value).
Recently a flag to explicitly indentify numbers was added and numbers are now considered (at least in our wishes... aehm, protocol specs) to be expressed in a specific way -- we can consider this like a first small step in the right direction, but a lot of things still need to be addressed.

So, at least for now, we can't assume that, if `dstate_getinfo()` returns a non-null value, it is a string containing a number in a specific format, even if the var is explicitly flagged as a number (or it is not flagged as a string), let alone if we only wish/expect it to be a number (for now, it can be empty, it can be a number expressed in some strange format, it can be a string).

jimklimov:
Ok, so for now I slightly improved the check to verify that we have (and ignore) an empty string or an `'0' '\0'` string, and commented about somewhat hypothetical ways that other cases may happen and can break - so when and if that happens, future-us (or others) can fix it. Otherwise it seems like a bit of wasted upfront investment to get regexps or `sprintf`s into the loop to fix breakage might never happen with bandaid that belongs elsewhere anyway.

jimklimov:
@aquette @zykh @clepple : any objections to merging this (so eventually other 1-vs-3 phase aware drivers can be refactored to use this routine plus maybe their existing magic based on device model etc. - or just stick with common default logic)?

jimklimov:
Just in case TLDR : the changes in dstate.c do not modify existing code,
but add a function intended to be used later by several drivers as a common
bit of logic to guess how many phases a device has based on other data it
provides (for devices which do not serve a "*.phases" value directly).
Currently it is used by mge-xml.c, and I've spotted several other drivers
who do similar decisions based on more or less clues - IMHO haphazardly at
times.

2017-02-15 16:09 GMT+01:00 Charles Lepple <notifications@github.com>:

aquette:
@jimklimov looks good to me too, relying on @zykh more advanced look at the code (I only check over 1mn). snmp-ups may indeed also benefit from that...

jimklimov:
One more thought came up - we had recent discussions that apparently 2-phase devices do also exist on the market. All the NUT code I've seen so far to guess phase counts assumes there's either 1 or 3...

aquette:
@jimklimov seems I never commented on the 2ph thread: this indeed exist for US (120/208V) and we call that "split phase" at Eaton (and also MGE time). I've never been able to actually work on that, and never had request for that, but here is an example: http://www.powerprosinc.com/Eaton-9PX-Split-Phase/
What I'm not sure is the exact behavior of that, i.e. exposing 2 input/output or just one with tolerance and default voltage...

Beside from that, we will also have the dual-input ePDU soon...

PR #401, opened 2017-03-04
Try to adapt paths in documentation

aquette:
Documentation still widely uses the default paths source when compiled from
source. While it is feasible to adapt some at compilation time, it is not
desirable since, for example, the online documentation on networkupstools.org
has to be generic. So try to make these paths more generic when possible, and
otherwise make a note


docs/scheduling.txt
```diff
@@ -245,7 +245,11 @@ make upsmon set the "forced shutdown" (FSD) flag on the upsd so your
 slave systems shut down early too.  Just do something like this in your
 CMDSCRIPT:
 
-	/usr/local/ups/sbin/upsmon -c fsd
+	/sbin/upsmon -c fsd
+
+NOTE: the path to upsmon must be provided. The default for a source installation
+is /usr/local/ups (so /usr/local/ups/sbin/upsmon), while packaged installation
+will generally comply to FHS (so /sbin/upsmon).
```
jimklimov:
````
600 FAILED : Aspell reported errors here:
601 ----- vvv
602 3310:& FHS 23 26: FHA, FH, HS, FS, F'S, FMS, FTS, HHS, NHS, VHS, FDS, FPS, OHS, H'S, HF'S, FE'S, FA'S, PH'S, FM'S, FR'S, CH'S, RH'S, TH'S
603 ----- ^^^
````

clepple: Also, given that the indented blocks will be fixed-width, pathnames in the text should probably be in `backticks` for consistency.

PR #403, opened 2017-03-13
Docs whitespace, paragraphs and dashes

jimklimov:
These changes might impact rendering of bullet-lists, so it is preferable to build and revise (visually) the various final formats of the docs. Also, and out of scope ATM, many texts use different style regarding filenames, varnames etc. (italics, backticks, etc.) which should at some point in time become more uniform.

Main purpose of this PR is to make those texts more editable and readable in the console, without making one's eyes hurt ;)


zykh:
Emerging from my hibernation just to express my humble opinion on this one...

nutdrv_qx* docs were not wrapped at 80 on purpose, since I'm more for a 'one sentence per line' style, and, if noone has anything against it, I would prefer them to stay that way, so, please, consider partially revert these commits.

Except for that, I'm not against this PR, since the other affected docs were already mostly wrapped at 80, but, were I the one who decides and had I enough time to do so, I would remove all the hard breaks at 80 from all docs and move them to a sentence-per-line style.

jimklimov:
The thing with ultra-long lines is that it is difficult to view and/or edit in things like LESS, VI or MCEDIT, and the github, bitbucket etc. web interfaces for change-comparisons also does not wrap them. So it becomes a game of lots of horizontal scrolling just to edit a line or see what changed in a diff, or reading with effective line-breaks inside words at terminal width.

Granted, wrapped lines suffer it differently if you toss the words around the line-break; but this is at least visible and reviewable without scrolling :)

In either case, tools like asciidoc and the chain beyond it (which convert the source texts to final HTML/PDF/MAN/...) are okay with such short linebreaks (where the blank line separates paragraphs). And the text files should IMHO be treated as sources, with style convenient for developers and reviewers :)

zykh:
Yes, there are pros and cons for both (I'm not going to list them here, people way smarter than me already have done that in the past), but, as I said, I'm more for a sentence-per-line style and this is not going to change anytime soon, given that I'm old and my confirmation bias (and my stupidity) would surely work against any change.
I don't want to start a style war or waste your time, so, back to this PR, point is, nutdrv_qx* files are not wrapped at 80, and, given that I maintain them, I would prefer them to stay that way, if this doesn't annoy anyone.
Mine is just a humble request, so if we end up dictating wraps at 80 (but, at the moment, what I see doesn't seem, at least to me, to point in that direction), then, although not happily, I will comply.

clepple:
While I agree that the web interfaces don't handle the longer lines well, I think that only comes into play when making changes to those files. (And part of the reason why I usually take a few days to even look at pull requests is because I intend to look at them in a 132-column terminal emulator with a comfortably-sized font, rather than fight with a web UI.)

Aside from zykh's request, the rest of 12eb3be seems to be correcting lines that were very close to 80 characters already, so I also recommend dropping that commit.

Overall, I think we need to take the coding guidelines in context: if someone drops off the code for a driver and never touches it again, the rest of the NUT maintainers are going to be working with that code further down the road, so a one-time code reformatting request makes sense. Going back and re-imposing those coding guidelines on the entire tree without a discussion first is wasted time, IMHO. It's one of the "features" of pull requests that irks me. We have a developer mailing list to sort out these discussions before changing the code (and potentially basing another PR on the changed code). At the end of the day, I think we have more important things to handle than whether the text wraps at 79 or 81 columns. I imagine that some of those lines were long because of a one-word replacement earlier in the paragraph, and I personally think that a cascading reformat obscures the intent of the original change.

It's like warnings: code should not introduce new warnings, but driving the warning count to zero has other consequences.

jimklimov:
Well, one initial cause for this PR to appear was the revision started in other PRs to backtick pathnames. In lines that have several files or command names listed, a couple of backticks per each often pushes them beyond the 80-column limit.

Also, given that many terminal programs I've used default to 80 char displays, the comfortable editing size is even less - about 76-77 chars max so I see there is no trailing text outside my screen :) Also I tend to hop from laptop to PC to phone over the course of the day, and they all have different physical resolutions - so sticking to the least common denominator is easier than switching habits all the time, and in at least one case the 80-char limit is more or less a hard one, physically.

jimklimov:
Ok, so the most contended questions are about paragraph reformatting. I guess I can split this commit over several, to pick or cancel specific changes.

Can the first two commits of the PR be merged in the meantime (whitespace and dashes)?

jimklimov:
FWIW, the following files under `docs/` alone contain "this many" lines that are 79 or more chars long. There's of course even more hits if we'd desire lower limits (e.g. 76 chars as is common elsewhere), and not all of these are easy or reasonable to fix at all (tables, URLs) to make it easier for users to read and for future devs to update and maintain... but still:

````
docs$ for F in *.txt ; do cat "$F" | cut -c 79 | egrep -vc '^$' && echo "^^^ $F"; done | egrep -v '^0$'
8
^^^ acknowledgements.txt
1
^^^ cables.txt
14
^^^ config-notes.txt
1
^^^ configure.txt
1
^^^ developer-guide.txt
11
^^^ developers.txt
27
^^^ documentation.txt
19
^^^ download.txt
44
^^^ FAQ.txt
6
^^^ features.txt
4
^^^ hid-subdrivers.txt
22
^^^ history.txt
4
^^^ maintainer-guide.txt
3
^^^ net-protocol.txt
10
^^^ new-drivers.txt
50
^^^ nut-names.txt
29
^^^ nut-qa.txt
128
^^^ nutdrv_qx-subdrivers.txt
1
^^^ outlets.txt
4
^^^ packager-guide.txt
8
^^^ scheduling.txt
13
^^^ security.txt
10
^^^ snmp-subdrivers.txt
7
^^^ support.txt
````

jimklimov:
Housekeeping: Bumped the PR with current master branch contents, to reduce deviation over time.

PR #404, opened 2017-03-14
Phoenixcontact modbus driver

sivann:
This is a modbus driver (the first?) for the Phoenix Contact QUINT-UPS model 2320461. The driver has been tested and works as expected. 
The custom IFS-RS-232 modbus cable is required (about 72$!), and UPS must be setup via its windows software to work correctly with NUT. All that is documented on the manual page of the driver.

This includes all skeleton modbus driver changes for building with libmodbus.

The ups on the web: https://www.phoenixcontact.com/online/portal/us?uri=pxc-oc-itemdetail:pid=2320461

Best Regards,
-Spiros

### DIFF
--- a/docs/man/phoenixcontact_modbus.txt
+++ b/docs/man/phoenixcontact_modbus.txt
@@ -30,7 +30,7 @@ How to configure the UPS
 
 Note: this UPS and its manual refers to Low-Batt as "Shutdown Event".
 
-You need the IFS-RS232-DATACABLE to communicate with the UPS in linux as the IFS-USB cable doesn't seem to be supported.
+You need the IFS-RS232-DATACABLE to communicate with the UPS in linux as the IFS-USB cable doesn't seem to be supported. FYI communication parameters are: 115200,E,8,1.
 You also need the UPS-CONF windows software (free, download from their site), to configure the UPS signals and timers. 
 1) Run the UPS-CONF
 2) Go to Settings->Time Setting
--- a/docs/man/phoenixcontact_modbus.txt
+++ b/docs/man/phoenixcontact_modbus.txt
@@ -25,6 +25,29 @@ More information about this UPS can be found here: https://www.phoenixcontact.co
 
 phoenixcontact_modbus uses the libmodbus project, for Modbus implementation.
 
+How to configure the UPS
+------------------------
+
+Note: this UPS and its manual refers to Low-Batt as "Shutdown Event".
+
+You need the IFS-RS232-DATACABLE to communicate with the UPS in linux as the IFS-USB cable doesn't seem to be supported.
+You also need the UPS-CONF windows software (free, download from their site), to configure the UPS signals and timers. 
+1) Run the UPS-CONF
+2) Go to Settings->Time Setting
+3) Choose "state of charge shutdown delay"
+4) Choose Remote starts PC-Shutdown in Mains and Battery mode
+5) On the PC-Shutdown enter the maximum value (5 minutes)
+6) On the PC-Restart delay enter the time you want the ups to leave the output power off before restarting (e.g. 60 seconds), after mains power is restored.
+7) On the UPS, turn the screw to the "PC-MODE" position
+
+Configuring the above way ensures that:
+* When power is lost ups constantly calculates remaining battery time
+* When remaining bat. time is <= 5 minutes (PC-Shutdown setting), it will raise the "Shutdown" event (seen as LOW-BATT in NUT)
+* From that moment even if input power is restored, the UPS will cut the output power to its load after 5 minutes
+* When the input power is restored, the ups will restore output power after 60 seconds (PC-RESTART delay setting).
+
+Unfortunately 5 minutes is the max, and 60 seconds also the max setting it can accept. It would be beneficial to raise PC-Restart to say 5 minutes to allow the UPS battery to charge.
+
 
 EXTRA ARGUMENTS
 ---------------
--- a/docs/man/phoenixcontact_modbus.txt
+++ b/docs/man/phoenixcontact_modbus.txt
@@ -1,9 +1,8 @@
 PHOENIXCONTACT_MODBUS(8)
-================
+========================
 
 NAME
 ----
-
 phoenixcontact_modbus - Driver for Phoenix Contact
 
 SYNOPSIS
@@ -32,21 +31,24 @@ Note: this UPS and its manual refers to Low-Batt as "Shutdown Event".
 
 You need the IFS-RS232-DATACABLE to communicate with the UPS in linux as the IFS-USB cable doesn't seem to be supported. FYI communication parameters are: 115200,E,8,1.
 You also need the UPS-CONF windows software (free, download from their site), to configure the UPS signals and timers. 
-1) Run the UPS-CONF
-2) Go to Settings->Time Setting
-3) Choose "state of charge shutdown delay"
-4) Choose Remote starts PC-Shutdown in Mains and Battery mode
-5) On the PC-Shutdown enter the maximum value (5 minutes)
-6) On the PC-Restart delay enter the time you want the ups to leave the output power off before restarting (e.g. 60 seconds), after mains power is restored.
-7) On the UPS, turn the screw to the "PC-MODE" position
 
-Configuring the above way ensures that:
-* When power is lost ups constantly calculates remaining battery time
-* When remaining bat. time is <= 5 minutes (PC-Shutdown setting), it will raise the "Shutdown" event (seen as LOW-BATT in NUT)
-* From that moment even if input power is restored, the UPS will cut the output power to its load after 5 minutes
-* When the input power is restored, the ups will restore output power after 60 seconds (PC-RESTART delay setting).
+1. Run the UPS-CONF
+2. Go to Settings->Time Setting
+3. Choose "state of charge shutdown delay"
+4. Choose Remote starts PC-Shutdown in Mains and Battery mode
+5. On the PC-Shutdown enter the maximum value (5 minutes)
+6. On the PC-Restart delay enter the time you want the ups to leave the output power off before restarting (e.g. 60 seconds), after mains power is restored.
+7. On the UPS, turn the screw to the "PC-MODE" position
 
-Unfortunately 5 minutes is the max, and 60 seconds also the max setting it can accept. It would be beneficial to raise PC-Restart to say 5 minutes to allow the UPS battery to charge.
+.Configuring the above way ensures that:
+ * When power is lost ups constantly calculates remaining battery time
+ * When remaining bat. time is <= 5 minutes (PC-Shutdown setting), it will raise the "Shutdown" event (seen as LOW-BATT in NUT)
+ * From that moment even if input power is restored, the UPS will cut the output power to its load after 5 minutes
+ * When the input power is restored, the ups will restore output power after 60 seconds (PC-RESTART delay setting).
+
+.Meaning of settings:
+ * PC-Shutdown: How long before output cutoff the ups will raise the "shutdown event" signal. Max value for this is 5 minutes. So PC should be able to shutdown within 5 minutes.
+ * PC-Restart: How long to delay output power after power is restored. Max is 60 seconds.
 
 
 EXTRA ARGUMENTS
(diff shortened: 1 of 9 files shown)


sivann:
you need to install libmodbus-devel on your build-machine for the checks to succeed probably.

jimklimov:
Hi, the "build machine" in the failure above is a cloud Travis CI instance. Can you please find the package name(s) relevant for "Ubuntu 14.04.5 LTS (trusty)" and list them in `.travis.yml` of your PR, under `packages: &deps_driverlibs` listing, so this change is tested and ultimately merged along with the code by the same PR? Thanks.

sivann:
@jimklimov thanks, it worked.

aquette:
hi @sivann 
thanks for moving on this topic, this helps me in getting more visibility on the modbus topic... :)
FYI, I'm willing to integrate your driver into the more generic nutdrv_modbus (both RTU and TCP), and benefit from the DMF - Data Mapping Files work that we're integrating too for SNMP (allows to have txt file that describes the data mapping loaded dynamically, instead of C structs)... I would really appreciate your support when working on this.
For now, I'll merge your driver, once you've made a minor addition: an entry in the HCL (data/driver.list.in)
thx and cheers!

clepple:
@aquette I haven't had a chance to check over the code, but I think we could use a bit of clarification in the configure scripts that this is not general Modbus support. Also, we've gone back and forth about driver names - this one is on the long side (`phoenixcontact_modbus`). Thoughts?

@sivann I noticed a few instances of whitespace at the end of lines, which `git diff` should show in red. Still checking the code.

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+	DRV_EXPERIMENTAL,
+	{ NULL }
+};
+
+void upsdrv_initinfo(void)
+{
+	upsdebugx(2, "upsdrv_initinfo");
+
+	dstate_setinfo("device.mfr", "Phoenix Contact");
+	dstate_setinfo("device.model", "QUINT-UPS 2320461");
```
clepple: Do we know this is the only supported model? Is it possible to read a Modbus register to determine this?

If the model number is needed in status displays, it can always be set with `override.device.model` in `ups.conf`.

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+
+	/* upsh.instcmd = instcmd; */
+	/* upsh.setvar = setvar; */
+}
+
+void upsdrv_updateinfo(void)
+{
+	upsdebugx(2, "upsdrv_updateinfo");
+    
+    uint16_t tab_reg[64];
```
clepple: Hard tabs should be used per the [coding guidelines](http://networkupstools.org/docs/developer-guide.chunked/ar01s03.html#_miscellaneous_coding_style_tools) but if nothing else, they should be consistent within a file.

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+
+    mrir(ctx, 29697, 3, tab_reg);
```
clepple: Please move magic numbers to `#defines` and/or document them.

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+
+void upsdrv_cleanup(void)
+{
+	if (ctx != NULL) {
+		modbus_close(ctx);
+		modbus_free(ctx);
+	}
+}
+
+int mrir(modbus_t *ctx, int addr, int nb, uint16_t *dest) {
```
clepple: Non-public functions should be declared `static`.

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+
+    mrir(ctx, 29745, 1, tab_reg);
+    dstate_setinfo("output.voltage","%d",(int)(tab_reg[0]/1000));
+
+    mrir(ctx,29749,5,tab_reg);
+    dstate_setinfo("battery.charge","%d",tab_reg[0]);
+    //dstate_setinfo("battery.runtime",tab_reg[1]*60);
+
+    mrir(ctx,29792,10,tab_reg);
+    dstate_setinfo("battery.voltage","%f",(float)(tab_reg[0])/1000.0);
```
clepple: The cast should be a `(double)`, or omitted (`1000.0` defaults to a double).

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+    dstate_setinfo("battery.temperature","%d",tab_reg[1]-273);
+    dstate_setinfo("battery.runtime","%d",tab_reg[3]);
+    dstate_setinfo("battery.capacity","%d",tab_reg[8]*10);
+    dstate_setinfo("output.current","%f",(float)(tab_reg[6])/1000.0);
```
clepple: Same with this cast.

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+        dstate_dataok();
+
+}
+
+void upsdrv_shutdown(void)
+{
+	fatalx(EXIT_FAILURE, "shutdown not supported");
+}
+
+/*
```
clepple: Unused code should be removed - if it is added later, please consult `skel.c`.

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+void upsdrv_initups(void)
+{
+    int r;
+	upsdebugx(2, "upsdrv_initups");
+
+    ctx = modbus_new_rtu(device_path, 115200, 'E', 8, 1);
+    if (ctx == NULL)
+        fatalx(EXIT_FAILURE, "Unable to create the libmodbus context");
+
+    r = modbus_set_slave(ctx, 192);
```
clepple: Same "magic number" comment applies here. Is there a chance that a unit will be configured with a different ID?

drivers/phoenixcontact_modbus.c
```diff
@@ -0,0 +1,200 @@
...
+	/* allow '-x foo=<some value>' */
+	/* addvar(VAR_VALUE, "foo", "Override foo setting"); */
+}
+
+void upsdrv_initups(void)
+{
+    int r;
+	upsdebugx(2, "upsdrv_initups");
+
+    ctx = modbus_new_rtu(device_path, 115200, 'E', 8, 1);
```
clepple: It's a good idea to mention the baud rate in the manual, near the cable description.

sivann:
I attach the modbus manual for this vendor. There are 4 supported sets of modbus addresses for phoenixcontact models. This implementation is about the 1st and 2nd ones (chapter 4.1, 4.2). The difference is the one I have (4.1) supports two different output voltages (e.g. both 24V and 12V), but didn't find a way to report this to NUT (not critical either). 

Our company has extensive experience for various modbus devices, and the input from our main modbus programmer is that you can not do auto-detection. 

Modbus is a very old and simple protocol, just 
1) register addresses to read from and write to, 
2) a slave ID for the slaves to "filter" requests.
3) function codes to determine data blocks (addresses overlap on data blocks). There are 4 codes and types of registers, type 0x4 supported by this UPS is called "input registers" and those registers are read-only (always).

Those addresses and slave IDs have no consistency, they can be anything.
Have in mind that in this UPS, reading addresses not defined in the documentation can also return data. Those data may be undocumented or garbage, I have no idea.

So I do not believe it is prudent to try and detect the UPS model, even from phoenixcontact models, by reading various addresses. And there is no vendor string anywhere defined unfortunately.

I will reply inline to @clepple comments.

[modbus reference phoenixcontact-105360_en_02.pdf](https://github.com/networkupstools/nut/files/843897/modbus.reference.phoenixcontact-105360_en_02.pdf)

drivers/phoenixcontact_modbus.c
sivann: A more generic model would be just QUINT-UPS/24DC, but 2320461 is the actual model tested with. You cannot detect the actual model, I leave up to you to choose device.model.

drivers/phoenixcontact_modbus.c
sivann: missed that, will fix.

drivers/phoenixcontact_modbus.c
sivann: These are not hacks, these are the modbus register addresses, chosen by the vendor (probably at random). They can be found in the vendor's modbus documentation (attached on a comment below). Is there any point in including this on the manual?

drivers/phoenixcontact_modbus.c
sivann: This is the slaveID. Hard coded on the device, and chosen by the vendor. Cannot be changed.

drivers/phoenixcontact_modbus.c
jimklimov: BTW, should it (and other bits of the setting) be configurable? Either for this device in particular (does it accept different settings) or for generalization of the driver at a later stage?

Anyhow, this bit to interact more with NUT settings/config engine can be added at a later stage, as another PR.

drivers/phoenixcontact_modbus.c
jimklimov: Even if not configurable, it is preferable to set magic numbers as macros with descriptive names (and/or comments nearby), especially to avoid questions a year later like "does this 128 mean the same as that 128"? ;)

sivann: I can add 192 as SLAVE_ID but what do you propose as naming for the register addresses? It's just the first non-consecutive address of an address block.

clepple:
Also, before this can be merged, someone will need to add the man page to docs/man/Makefile.am

docs/man/phoenixcontact_modbus.txt
```diff
@@ -0,0 +1,89 @@
...
+6) On the PC-Restart delay enter the time you want the ups to leave the output power off before restarting (e.g. 60 seconds), after mains power is restored.
+7) On the UPS, turn the screw to the "PC-MODE" position
+
+Configuring the above way ensures that:
+* When power is lost ups constantly calculates remaining battery time
+* When remaining bat. time is <= 5 minutes (PC-Shutdown setting), it will raise the "Shutdown" event (seen as LOW-BATT in NUT)
+* From that moment even if input power is restored, the UPS will cut the output power to its load after 5 minutes
+* When the input power is restored, the ups will restore output power after 60 seconds (PC-RESTART delay setting).
+
+Unfortunately 5 minutes is the max, and 60 seconds also the max setting it can accept. It would be beneficial to raise PC-Restart to say 5 minutes to allow the UPS battery to charge.
```
clepple: This part is unclear. I'm guessing the first max refers to PC-Shutdown?

sivann: PC-Shutdown: How long before output cutoff the ups will raise the "shutdown event" signal. Max value for this is 5 minutes. So PC should be able to shutdown within 5 minutes.
PC-Restart: How long to delay output power after power is restored. Max is 60 seconds.

sivann:
cleared that up, and fixed the asciidoc format.

docs/man/phoenixcontact_modbus.txt
clepple: okay, thanks.

clepple:
@aquette still looking for feedback on the driver name. Also, I didn't find an example in the nutdrv-modbus branch - should the man page be conditionally generated based on whether libmodbus was found (as is done for the USB drivers, etc)? If so, I have a commit (9f3d349df) on the pull_404 branch which will do that.

I think this branch should probably be rebased/squashed instead of merging directly. Tagging as WIP for that reason.

sivann:
Please I would like some input the aspell assertion, I cant get it to work or train.

aquette:
@clepple: indeed for the manpages.
As for the driver name, it's a specific one and it mentions both the mfr (which has a very long name) and the protocol. Nothing much we can do apart from truncating the mfr name. However, as said, I take it more as a transient driver / first effort, that will get merged into the generic nutdrv_modbus. So no big deal.

PR #406, opened 2017-03-17
Re-sort nut.dict

clepple:
@jimklimov I noticed a few of the words (ID's, Javadoc) jumped around in the sort order. Is that a fluke, or should we force `LANG=C` for sort in docs/Makefile.am?

It is also possible that I messed something else up, hence the PR rather than committing directly.


jimklimov:
I think the LANG is already forced there, at least I intended to do that... if not (or done wrongly) this should indeed be fixed :)

PR #412, opened 2017-03-30
liebert-esp2 changes: Correct battery V scaling, update docs, implement split-phase unit support

nkaminski:
• Refactor output format of liebert-esp2 driver to follow that of other drivers to fix the nut-cgi interface as well as for overall consistency. 

• Implement proper support for "2 phase" i.e. split phase input Liebert GXT2 series units. Existing driver treats these units as 3 phase which results in a large number of unpopulated/zero values being returned by the driver.

• Correct battery voltage scaling for GXT2 series units.

• Include notes in liebert-esp2 man page that the DB9 cable used to interface a GXT2 series UPS to its host system must not connect and handshaking/flow control lines since these are used for other features such as non "smart" shutdown and power loss notification. Using a DB9 cable with full handshaking with a GXT2-6000RT208 UPS is confirmed to cause an unintended shutdown due to the RTS line being connected to the "shutdown on battery" input of the UPS.

### DIFF
--- a/docs/man/liebert-esp2.txt
+++ b/docs/man/liebert-esp2.txt
@@ -41,7 +41,7 @@ Set the speed of the serial connection - 1200, 2400 (default), 4800, 9600 or 192
 
 AUTHOR
 ------
-Richard Gregory <R.Gregory at liverpool.ac.uk>, Arjen de Korte <adkorte-guest at alioth.debian.org>
+Richard Gregory <R.Gregory at liverpool.ac.uk>, Arjen de Korte <adkorte-guest at alioth.debian.org>, Nash Kaminski <nashkaminski at kaminski.io>
 
 SEE ALSO
 --------
--- a/docs/man/liebert-esp2.txt
+++ b/docs/man/liebert-esp2.txt
@@ -12,9 +12,24 @@ This man page only documents the hardware-specific features of the
 liebert-esp2 driver.  For information about the core driver, see  
 linkman:nutupsdrv[8].
 
+SPECIAL CABLING NOTE
+--------------------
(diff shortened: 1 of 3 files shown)


jimklimov:
LGTM looking at the code changes, nothing fatal pops out ;)

@clepple @zykh @aquette ?

aquette:
2nding @jimklimov, through a quick review (on a smartphone) nothing hurts... apart  possibly from the move from ups.Lx.load to output.Lx.power.percent (a point of diff between 1ph and 3ph that we have to clarify)

nkaminski:
My rationale for following the convention of the APC driver as well as other stable drivers and what the default template provided with nut-cgi expects, is because using the previous release of the driver (using ups.Lx.load) results in a broken interface in not only nut-cgi but also the interface of the network UPS tools package provided with pfSense, since both interfaces expect output.Lx.power.percent for polyphase units and ups.load otherwise. If there is a better/more preferred way of accomplishing this I'm more than willing to change this if desired. 

Could also add a note in the documentation that single phase units provide ups.load and split phase/"2 phase" as well as three-phase units follow the output.Lx.power.percent convention.

Let me know if there are any further questions or concerns.

On March 30, 2017 4:55:40 AM CDT, Arnaud Quette <notifications@github.com> wrote:
>2nding @jimklimov, through a quick review (on a smartphone) nothing
>hurts... apart  possibly from the move from ups.Lx.load to
>output.Lx.power.percent (a point of diff between 1ph and 3ph that we
>have to clarify)

-- 
Sent from my Android device with K-9 Mail.

aquette:
No worries @nkaminski, my comment was more related to the NUT namespace than the driver ;)

PR #413, opened 2017-03-30
Problem: NUT tests in Travis CI often wait for default-withdoc

jimklimov:
Solution in .travis.yml : since Travis cloud starts the jobs as soon as it has resources, move the bulk of quicker matrix jobs to start after the longest withdocs build. Experimental note: we still need one job defined in the simple matrix for default environment, but can keep others in a single explicitly ordered list. Now the "Running for" time got cut from 8-9 to 5.5-6 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,26 +53,23 @@ addons:
     - libxml2-utils
     - libmodbus-dev
 
+# Define one job in the matrix, to avoid a failing job
+# with "no environment variables set"; but others are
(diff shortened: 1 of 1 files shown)


aquette:
Very nice @jimklimov, kudos!
Anything preventing you from self-merging?

jimklimov:
I guess not, seems to work (although the effect is rather lost in larger builds like DMF). Just the habit of wanting another pair of eyes when the fix is not urgent :)

PR #421, opened 2017-04-10
Some systems detect man-pages as "ASCII text" - please them

jimklimov:


jimklimov:
Bump buildbot http://buildbot.networkupstools.org/public/nut/builders/OpenBSD-x64/

PR #422, opened 2017-04-12
 snmp-ups: add support for outlet.group phase

aquette:
Outlets groups can now publish the electrical phase to which the physical
outlets group (Gang) is connected to. This is useful for 3ph ePDU

### DIFF
--- a/docs/nut-names.txt
+++ b/docs/nut-names.txt
@@ -558,6 +558,9 @@ Some specific data to outlet groups exists:
 | Name                           | Description                    | Example value
 | outlet.group.n.type            | Type of outlet group (OPAQUE)  | outlet-section
 | outlet.group.n.count           | Number of outlets in the group | 12
+| outlet.group.n.phase           | Electrical phase to which the
+                                   physical outlet group (Gang) is
+                                   connected to                   | L1
 |=================================================================================
 
 Example:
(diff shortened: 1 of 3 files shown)


aquette:
@clepple , @zykh , @jimklimov : calling for reviewer since there is a new variable addition...

clepple:
> @clepple , @zykh , @jimklimov : calling for reviewer since there is a new variable addition...
> 

new variable discussion should really go to nut-upsdev to make sure it doesn't get missed. Will look later.

jimklimov:
Maybe stress in the doc the difference between `.phase` (name of the used phase) and `.phases` (amount available) as suffixed to different data points? :)

aquette:
@clepple you're right, I've posted on -upsdev a RFC
@jimklimov do you mean something like "This variable should not be confused with {input[.bypass,output}.phases which count the number of available phases"? The variable name seems different and specific enough to me. Or did I got you wrong?

jimklimov:
Yes, sort of that. I wondered for a few seconds, until I bit into the
context, if it is the same old song or a new one... so I thought a
clarification is in order ;)

aquette:
No comment on the ML on the RFC for this new variable. Gonna merge now.
@jimklimov : I've not added the disambiguation you mentioned. Feel free to PR ;)

PR #423, opened 2017-04-18
Add cmdline option to run drivers not based on ups.conf.

EatonEKia:
Hello,

Sometimes, some specific scenarios need to be able to run some driver instances not based on ups.conf.
For example, we could want to temporary launch a driver with specific configuration without having to modify the ups.conf file and discuss directly to it, without passing by upsd. This can be useful to validate existence of some devices, temporarily testing credentials, acquire some data, all of that without polluting upsd data.

This patch only add a specific driver option "-s" similar to existing "-a" but without reading ups.conf file.

Best regards,

Emilien

### DIFF
--- a/docs/man/nutupsdrv.txt
+++ b/docs/man/nutupsdrv.txt
@@ -1,7 +1,7 @@
 NUTUPSDRV(8)
 ============
 
-NAME  
+NAME
 ----
 
 nutupsdrv - generic manual for unified NUT drivers
@@ -54,6 +54,14 @@ driver's author may have provided.
 Autoconfigure this driver using the 'id' section of linkman:ups.conf[5].
 *This argument is mandatory when calling the driver directly.*
 
+*-s* 'id'::
+Configure this driver only with command line arguments instead of reading
+linkman:ups.conf[5]. To be used instead of *-a* option when need to run
+a driver not present in driver configuration file. Instead, driver
+configuration have to be set with *-x* options directly in the command line.
+As the driver instance cannot be controlled by linkman:upsdrvctl[8],
+this option should be used for specific needs only.
+
 *-D*::
 Raise the debugging level.  Use this multiple times to see more details.
 Running a driver in debug mode will prevent it from backgrounding after
@@ -70,7 +78,7 @@ either too limited or too verbose to be of any use.
 Raise log level threshold.  Use this multiple times to log more details.
 +
 The debugging comment above also applies here.
- 
+
 *-i* 'interval'::
 Set the poll interval for the device.  The default value is 2 (in seconds).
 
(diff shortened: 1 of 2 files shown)


jimklimov:
+1 from me, we discussed an "explore" option like this with @aquette earlier.

One downside that he noted was that such exploration may be harmful (e.g. sending random stuff on serial cable that the UPS may misinterpret while you try to discover it), so was sort of against such mode in nut-scanner. But at least for the manual shoot-yourself-in-the-foot risk I longed for such option since the introduction of `-a` and removal of direct driver program invokation ;)

EatonEKia:
One general question: if someone want to

EatonEKia:
Excuse me for the pollution (bad key shortcut) !!

I have a general question: if someone want to start a driver in "standalone" mode, do we really want to create the pid file and unix socket in the same repository than for other "standard" drivers (run with upsdrvctl) ? Should it be interesting to specify an alternate location by cmdline ?

jimklimov:
IMHO, the answer depends on whether we can have this driver visible through `upsd` (e.g. to inspect locally with `upsc`, dump data for HCL database, etc.), so if presence of an entry in the socket-directory suffices for that - I'd want it to be there. I *think* it worked like this in pre-`-a` times :)

Not so sure about pid-files, but for commonality and to not complexify code I'd leave them there as well, at least by default. Options for alternate location(s) also make sense though, e.g. so that testing does not interfere with production deployment.

EatonEKia:
Currently, to be visible in upsd, drivers have to be present in ups.conf so if a driver is run in "standalone" mode, it will not be visible in it (and it is the primary goal of this mode).

What we can easily do is:
- to set an alternative directory path for pidfile and unix socket by setting a driver specific property like "driver.statepath", filenames keeping the "<drvtype>-<id>" pattern
- or to set alternative pidfile and unix socket file path as driver specific properties like "driver.pidfile" and "driver.socketfile"

aquette:
jumping into the discussion, as we discussed all that a lot over :coffee: with @balooloo 

To clarify a bit, there are 2 main use cases to running the drivers without conf:

1)  Discovery / inventory

Run the driver and get a one shot upsc output
This is very useful to complement the nut-scanner in a discovery / inventory approach.
i.e. basic discovery of available devices with nut-scanner, then data collection with the specific pointed driver(s) & info.

Note that I have stagging PR (data-dump) on top of the present PR which implements this "-s <id> -d" use case.

2) One shot / dedicated monitoring

Run the driver and get data through the state socket, not upsd.
Possibly beside of a production NUT system (so, must not interfere with it), hence the no ups.conf (to not be visible by upsd). The same device may already be configured and we may need to use other runtime params.
The latency is a main point when using direct connection to drivers.

This is the subject of the present PR.


@clepple @zykh any comment before I merge the present one and push mine (use case 1)?

PR #427, opened 2017-04-21
gitlog2changelog.py : do not explode on non-Gitlog lines with a "Date:"

jimklimov:
Solves this:

````
./tools/gitlog2changelog.py v2.6.0 || \
	echo "gitlog2changelog.py failed to generate the ChangeLog. See https://github.com/networkupstools/nut/commits/master" > ChangeLog
  File "./tools/gitlog2changelog.py", line 55, in <module>
    date = dateList[1]
IndexError: list index out of range
````

that happens after (I) recently posted patches with their full historical email-style headers, including a different `Date:` pattern. In hindsight, maybe it is THAT date that should be in changelog; or maybe not - since the sources changed on the commit/merge date...


clepple:
superseded by #447

PR #440, opened 2017-06-06
Ve-direct driver

pkubanek:
Working release of Ve-direct driver for Victron Energy UPS/solar panels monitoring.


pkubanek:
Sorry for failed CI, fixed now.

drivers/ve-direct.c
```diff
@@ -0,0 +1,298 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ *
+ */
+
+#include "main.h"
+#include "serial.h"
+
+#define DRIVER_NAME	"Victron Energy Direct UPS and solar controller driver"
+#define DRIVER_VERSION	"0.20"
```
jimklimov: GiHub web interface doesn't quite let decipher - is there a whitespace between macro name and value (of driver version)?

drivers/ve-direct.c
```diff
@@ -0,0 +1,298 @@
...
+		}
+
+		break;
+	}
+
+	dstate_dataok();
+}
+
+void upsdrv_shutdown(void)
+{
```
jimklimov: Is there certainly no session teardown?

drivers/ve-direct.c
```diff
@@ -0,0 +1,298 @@
...
+}
+
+void upsdrv_help(void)
+{
+	printf("Read The Fine Manual ('man 8 ve-direct')\n");
+}
+
+/* list flags and values that you want to receive via -x */
+void upsdrv_makevartable(void)
+{
```
jimklimov: Aren't there variables to set up or modify like the port, maybe baud speed (e.g. use a default 19200 but reserve ability to probe related units with other hardware abilities)?

drivers/ve-direct.c
```diff
@@ -0,0 +1,298 @@
...
+	dstate_dataok();
+}
+
+void upsdrv_shutdown(void)
+{
+}
+
+void upsdrv_help(void)
+{
+	printf("Read The Fine Manual ('man 8 ve-direct')\n");
```
jimklimov: Related to question on configurable `-x` flags - maybe a bit more is to be said here. Maybe at least the standard wall of text with `-a`, `-D` and other standard flags.

drivers/ve-direct.c
```diff
@@ -0,0 +1,298 @@
...
+		*v_value = '\0';
+		v_value++;
+		char *end = v_value;
+		while (*end != '\r' && *end != '\0')
+			end++;
+		if (*end != '\r')
+			break;
+		*end = '\0';
+
+		snprintf(vn, sizeof(vn), "ve-direct.%s", v_name);
```
clepple: "ve-direct" is not a valid top-level NUT variable name. Please use the standard names defined, and discuss on nut-upsdev if you need anything not covered here: http://networkupstools.org/docs/developer-guide.chunked/apas01.html

drivers/ve-direct.c
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/* list flags and values that you want to receive via -x */
+void upsdrv_makevartable(void)
+{
+}
+
+void upsdrv_initups(void)
+{
+#ifndef TESTING
```
clepple: The `TESTING` #define is not handled elsewhere in NUT, but if you include it, I would recommend adding a case to parse a few static text buffers. This will help other developers who might have to maintain the code later down the road.

clepple:
I would prefer that these commits be rebased/squashed rather than merged as-is. The individual commits do not stand on their own.

https://github.com/networkupstools/nut/blob/master/docs/developers.txt#L327

drivers/ve-direct.c
pkubanek: Yes, there is a tab (whitespace of a sort), as is case in other drivers

drivers/ve-direct.c
```diff
@@ -0,0 +1,298 @@
...
+/* driver description structure */
+upsdrv_info_t upsdrv_info = {
+	DRIVER_NAME,
+	DRIVER_VERSION,
+	"Petr Kubanek <kubanek@fzu.cz>",
+	DRV_EXPERIMENTAL,
+	{ NULL }
+};
+
+char line[500];
```
jimklimov, line 39: Is the size fixed in device protocol etc? We have macros like `SMALLBUF` or `LARGEBUF` that might be nicer to use commonly.

jimklimov:
Was this work done by/for/on-behalf of a vendor (e.g. Victron)? If so, please update the docs/acknowledgements.txt file as well.

jimklimov:
Looking at previous comments, it looks like some more work is needed on driver variable naming, at least. If this is still the case, it is pretty much a blocker until someone solves it.

jimklimov:

drivers/ve-direct.c
jimklimov: Centralized in NUT code as of 2.8.3.

drivers/ve-direct.c
jimklimov: There's actually a lot of it in blazer and nutdrv_qx code...

drivers/ve-direct.c
jimklimov: "Fixed" by moving to `experimental.*` namespace. Generally we would need a PV-oriented namespace, I suppose (#1407 also stumbles on that).

jimklimov:
TODO: Man page and similar docs...

jimklimov:
Well, there are a few rough edges remaining, but I think the PR became good enough to merge (and then maybe chisel to perfection as part of common NUT code base - e.g. document the parameters which the INSTCMD can get/set).

Thanks @pkubanek !

PR #446, opened 2017-06-25
gitlog2changelog: Date header only occurs at start of line

clepple:
Also check for Author at the beginning of the line (since the rest of the
commit message is indented by four spaces).

Alternative to https://github.com/networkupstools/nut/pull/427


clepple:
never mind, the Author change should be reverted.

PR #447, opened 2017-06-25
gitlog2changelog: Date header only occurs at start of line

clepple:
Alternative to https://github.com/networkupstools/nut/pull/427

### DIFF
--- a/tools/gitlog2changelog.py
+++ b/tools/gitlog2changelog.py
@@ -47,15 +47,22 @@ for line in fin:
     # (Don't use startswith to allow Author override inside commit message.)
     elif 'Author:' in line:
         authorList = re.split(': ', line, 1)
-        author = authorList[1]
-        author = author[0:len(author)-1]
-        authorFound = True
+        try:
+            author = authorList[1]
+            author = author[0:len(author)-1]
+            authorFound = True
+        except:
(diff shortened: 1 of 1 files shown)


clepple:
@jimklimov this version only considers dates in the commit metadata, rather than in the body of the commit. Otherwise, some of the larger commits get the dates stripped out of the constituent patches.

jimklimov:
I'm not sure about the expected correct input format, so perhaps your solution is better. Still, maybe there is value in merging both - mine as a fallback (to log a problem and not abort).

clepple:
Agreed, since with the original code, an exception would get logged, but the build would continue.

We are no longer looking for `Date:` headers inside the body of the commit message, so the chance of that try/except block being used is slim (would mean that the `git log` format had changed). But it's there for completeness, along with a new check for the `Author:` header.

jimklimov:
Ok, thanks, LGTM now :)

Did you test this on potentially offending commits (e.g. with patch-email "Date:" indented headers in the commit comment, as seen in `git log`... possibly some are now in the master branch's log)?

clepple:
I've been meaning to look for other examples not on master, but then again, I am pretty sure that only `git log` metadata will match the `startswith('Date:')` - commit messages will be indented.

Issue #448, opened 2017-06-25
Install pre-generated man pages from tarball


clepple:
If asciidoc is not installed when configuring NUT from a `make dist` tarball, the included pre-generated man pages do not get installed.

Reference: https://lists.alioth.debian.org/pipermail/nut-upsuser/2017-June/010786.html

jimklimov:
Looking at the mailing list message, the `configure` arguments include `--with-doc` which assumes a `yes` value for all formats (IIRC, can't look into code tonight). So if tools are not available to build man, html, html-chunked and pdf - the explicitly requested action can not be fulfilled, so the configuration fails.

It may be possible to relax the check for "man" type of documents and indeed distribute pre-generated versions from the tarball (e.g. add another value for fallback, like `--with-doc=man=fallback == --with-doc=man`), but this would not be possible for html and pdf docs which we do not pre-package. Then it may also be possible to change default `--with-doc` to only process manpages and not all doc types (IF this is the case now), or automate the guesswork for building or skipping the other types. I think I did something like that, but not sure at the moment if it has already landed into `upstream/master`, nor if what I did fully implements the graceful "build or skip" for any doc types.

Finally note that specifically due to `make dist` constraints, the Makefile rules for manpages vs. other doc types are somewhat different (man's are required unless explicitly skipped e.g. to speed up CI builds).

clepple:
I will admit that I got distracted while looking into this before, so I do not have a good bound for when things broke relative to previous releases. I suspect that there may be some related unresolved pull requests around when I was on vacation earlier this year. If you run across them before I do, feel free to link them here.

Also, I am fine with documenting around the problem in the short term, if something like `--with-doc=man` would work for Wolfy's RPM spec file. However, we can also take the opportunity to untangle some of the options for future releases.

jimklimov:
Just in case, current state should be documented : https://github.com/networkupstools/nut/blob/master/docs/configure.txt#L89

In the unlikely but possible case that (subsequent) changes in code diverged functionally from the description here, e.g changed or added some behaviors, this divergence can be treated as a bug and likely I am to blame for that discrepancy, then ;)

jimklimov:
> If the "--with-doc" argument is passed without a list, or lists just "=yes" 
> or "=all", it enables all supported formats with a "=yes". 
> 
> An (explicit!) "--with-doc=auto" argument tries to enable all supported 
> formats with an "=auto". 

[...]

PR #451, opened 2017-06-26
Support builds and installs on systems without suitable asciidoc (fix issue #448)

jimklimov:
As reaction to issue #448 this PR changes handling of `--with-doc=man=auto` (which is the default behavior if no `--with-doc` is specified on command-line) to fall back to installation of pre-built docs (currently just manpages) that can be available in either build or source dir, for tarball distributions.

Tested with `./configure --with-all=no && make distcheck-light` by requiring a non-existently large minimal version of ASCIIDOC in m4 in one case, and leaving the correct version in the other - so both codepaths seem to install the manpages correctly.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -742,9 +742,9 @@ dnl to detect if we can build the wanted documentation format and yet
 dnl not fail if we have no tools to generate it (so add to SKIP list).
 
 	html-single*)
-		AC_MSG_CHECKING([if asciidoc version can build ${nut_doc_build_target_base} (minimum required 8.6.3)])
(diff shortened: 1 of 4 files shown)


jimklimov:
UPDATE: it seems there is a sort of conflict for the distchecks where the original source directory contains pre-built manpages: the resulting test-build tree seems to be satisfied by the existing groff files and does not rebuild them. Probably should revise something to ensure building of docs if requested to.

jimklimov:
Latest commits should fix the issue of not-rebuilding docs if requested to. Maybe not the cleanest solution (as commented in code), but tested to work and should be more portable than my idea of a cleaner one.

jimklimov:
@clepple - is this good to go as the solution to the issue?

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #461, opened 2017-08-07
Fix issue #460 (rectify apcupsd-ups FSD)

### LINKED ISSUE #460: No FSD when using driver apcupsd-ups
When using the driver apcupsd-ups then the shutdown condition is configured in /etc/apcupsd/apcupsd.conf (BATTERYLEVEL & MINUTES = remaining battery percentage & runtime when shutdown will be initiated).
Then the shutdown has to be disabled in apcupsd or just this one machine with apcupsd will get shut down without informing others via nut. This can be done for example by editing /etc/apcupsd/apccontrol and change "${SHUTDOWN} -h" to "${SHUTDOWN} -k".

Now nut should be usable with the apcupsd-ups-driver. But nut will never initiate a shutdown (FSD), because apcupsd changes the ups-status from "ONBATT" to "SHUTTING DOWN" when it's working on battery and the battery (or runtime) is below the limits, while nut interprets "SHUTTING DOWN" in the same way as a loss in connection.

This is wrong. "SHUTTING DOWN" should be the same as "LOWBATT". I will post a fix shortly.

tobby88:
Fix #460 so nut with apcupsd-ups-driver will correctly initiate a shut down on low battery.

### DIFF
--- a/drivers/apcupsd-ups.c
+++ b/drivers/apcupsd-ups.c
@@ -54,9 +54,8 @@ static void process(char *item,char *data)
 	{
 	case DU_FLAG_STATUS:
 		status_init();
-		if(!strcmp(data,"COMMLOST")||!strcmp(data,"SHUTTING DOWN")||
-			!strcmp(data,"NETWORK ERROR")||!strcmp(data,"ERROR"))
-				status_set("OFF");
+		if(!strcmp(data,"COMMLOST")||!strcmp(data,"NETWORK ERROR")||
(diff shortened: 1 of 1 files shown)


jimklimov:
On the logical side, looks good to me. But I don't have APC hardware ATM to actually test this. :+1:

clepple:
I'd like to look at this a little more tonight or later this week, but looks good at first glance.

Issue #332, opened 2016-10-11
export or deprecate `timehead.h`


clepple:
Depending on what version of POSIX we are assuming, `timehead.h` may be unnecessary. Per issue #331, NUT already exports a header with a `struct timeval` parameter, but `timehead.h` is not exported for users of the NUT client interface.

PR #466, opened 2017-08-24
A few small fixes for cleaner codebase and compilations

jimklimov:


jimklimov:
> I am concerned that the CFLAGS (should be CXXFLAGS, but that's beside the point) are being hardcoded here. What is the reasoning?

Building the full distcheck suite on an older debian release, with gcc-4.8 as default, I found that the cppunit tests complained that C++11 is required for the stuff they do. Based on experience with Travis CI environments, I know that their gcc-4.6 does not have C++11 at all, 4.9 and newer have it as a first-class citizen, and for 4.8 it must be enabled via command-line, with default being an older standard.

The `CFLAGS` (vs `CXXFLAGS`) notion came from an existing variable in the file... I started out tweaking its value, than added a separate lever to pull (or undefine altogether) easily, and overlooked the naming.

jimklimov:
As for `timehead.h` - systems do vary with their paths, and some distros may be sloppier than others. I added my comment to #332 on this part now :)

clepple:
> Building the full distcheck suite on an older debian release, with gcc-4.8 as default, I found that the cppunit tests complained that C++11 is required for the stuff they do. Based on experience with Travis CI environments, I know that their gcc-4.6 does not have C++11 at all, 4.9 and newer have it as a first-class citizen, and for 4.8 it must be enabled via command-line, with default being an older standard.
> 
It seems like it should be an autoconf test. If nothing else, it would make it easier to tell the difference between a compiler not interpreting that flag, and a test suite compilation failure.

I haven't had time to look at everything here, but I have a vague concern that building only part of the code with one compiler flag is problematic.

Issue #472, opened 2017-09-03
Wanted: autoconf tests for C++11 abilities and flags


jimklimov:
Following from #466 discussion, there are parts of the codebase (e.g. cppunit tests) that require C++11 support. Depending on compilers, it may be available by default, or require certain `CXXFLAGS` tuning, or be missing altogether (so that part of code should not be built). The `configure` script and `Makefile`s should be amended to figure this out properly.

clepple:
Related: #366

Issue #366, opened 2017-01-24
Gather notes on compiler flags for various platforms


clepple:
This is a bit more open-ended than most of our issues, but I think we need a place to track things like C89/C99 and POSIX with respect to the platforms where NUT is being used.

In particular, we reimplement some functions which have become more commonplace in the C standard library, such as `snprintf()`. It looks like Solaris-derived systems may need additional compiler flags to reveal the standard library version: https://github.com/networkupstools/nut/pull/343#issuecomment-262402045

Also, we have not been very specific about versions of POSIX.

PR #480, opened 2017-09-17
Improve cppunit and C++11 support via configure.ac

jimklimov:
See also #472 and #366 and #466

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -104,10 +104,12 @@ AC_CONFIG_LIBOBJ_DIR([common])
 dnl +-------------------------------------------------------------------
 
 AC_PROG_CC
+AC_PROG_CC_C99
 dnl Needed for per-target flags
 AM_PROG_CC_C_O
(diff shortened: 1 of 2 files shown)


jimklimov:
Note; newer autoconf includes macros for this, but systems I've played with lack them. So iterated back to stone ax which just works ;)

jimklimov:
Fixed a few wrinkles in the PR - now it does actually build and run the `cppunittest` where compiler is able to, including gcc-4.9.4 which needs an explicit `-std=c++11` to do so. Hopefully the `configure.ac` code structure is sufficiently flexible to add support for other compilers should the need arise. 

Alternately, if the script determines that the test code is not buildable, it won't try to do so and thus blow up the build.

This PR is somewhat orthogonal to the embryotic state that the actual C++ test code is in.

clepple:
I would also agree that merging this should close #472.

Issue #336, opened 2016-10-28
ups.conf: reword "user" directive section


clepple:
> **user**
> 
> Optional. If started as root, the driver will setuid(2) to the user id associated with username.

This should be reworded to indicate that NUT calls setuid() by default, but that this overrides the username.

clepple:
Logic is in [common/common.c](https://github.com/networkupstools/nut/blob/v2.7.4/common/common.c#L171-L189).

Move text to ups.conf doc, and reference that from nutupsdrv.

PR #493, opened 2017-10-29
Reword user directive section to resolve issue #336

nkaminski:


jimklimov:
@aquette @clepple : bump on this one? Rewording makes sense to me... except, can a non-root user `setuid` to someone else easily? Possibly, in privileges-based systems (SELinux, Solaris RBAC, etc.) and the privilege for `setuid` given to a non-`root` user, this nuance can be important.

aquette:
in theory, a non-root user can indeed setuid() to someone else. That just depends on the OS setup.
But then, there is also the privileges requirements to access config files (ups.conf) prior to becoming that user.
I'll reword to the following if that suits you all:
> The driver will setuid(2) to the user id associated with 'username',
> as opposed to its current effective user id.
> Note that drivers are generally started as root to satisfy privileges to access resources such ups.conf(5)

aquette:
closing in favor of #671
Please comment there if needed, otherwise I'll merge it very soon. thx

PR #494, opened 2017-10-30
configure.ac: Fix ${systemdsystemshutdowndir} detection

michal42:
On non-Debian systems, ${libdir} is /usr/lib64, but systemd files are
installed below /usr/lib/systemd. Instead of assuming that the shutdown
directory is below ${libdir}, assume that it has the same parent
directory as ${systemdsystemunitdir}, which we already detect properly
using pkg-config. The proper fix would be a --with option for the
shutdown directory, but that would be an overkill.

Signed-off-by: Michal Marek <MichalMarek1@eaton.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1197,11 +1197,8 @@ AC_ARG_WITH([systemdsystemunitdir],
 		;;
 	esac
 ], [])
-dnl Override installation directory, with the local installation
-dnl prefix. This is needed for 'distcheck*' targets, otherwise
-dnl files will try to get intalled to the actual system directories
 if test -n "${systemdsystemunitdir}"; then
-	systemdsystemshutdowndir="${libdir}/systemd/system-shutdown"
+	systemdsystemshutdowndir=`dirname "${systemdsystemunitdir}"`/system-shutdown
 	AC_MSG_RESULT(using ${systemdsystemunitdir})
 else
 	AC_MSG_RESULT(no)


configure.ac
```diff
@@ -1197,11 +1197,8 @@ AC_ARG_WITH([systemdsystemunitdir],
 		;;
 	esac
 ], [])
-dnl Override installation directory, with the local installation
-dnl prefix. This is needed for 'distcheck*' targets, otherwise
-dnl files will try to get intalled to the actual system directories
 if test -n "${systemdsystemunitdir}"; then
-	systemdsystemshutdowndir="${libdir}/systemd/system-shutdown"
+	systemdsystemshutdowndir=`dirname "${systemdsystemunitdir}"`/system-shutdown
```
jimklimov: Better still quote this, to rule out issues with systems that might have spaces embedded into paths that `dirname` can cause a choke upon, producing two or more tokens and assigning just the first in some shells.

configure.ac
```diff
@@ -1197,11 +1197,8 @@ AC_ARG_WITH([systemdsystemunitdir],
 		;;
 	esac
 ], [])
-dnl Override installation directory, with the local installation
```
jimklimov: I am not sure removal of the comment is proper: we do still override the value if set, can keep the explanation why...

configure.ac
jimklimov: Also, did you test that on other distros with possibly different path layouts, this `systemdsystemunitdir` points where we expect it to?

configure.ac
michal42: That comment referred to code removed by f65dc913d862 ("Undo ${systemdsystemunitdir} mangling.").

configure.ac
michal42: Variable assignment is not subject to work splitting. Re different systemd path layouts, indeed systemd does not guarantee these two directories to have a common parent. The unit directory is ${prefix}/lib/systemd/system, while the shutdown directory is ${libexecdir}/system-shutdown (see https://github.com/systemd/systemd/blob/master/meson.build#L105). Not sure what to do, distros typically place both into the same directory and introducing another --with option would require adapting various packaging recipes. :-/

michal42:
I will change the wording of the chagelog about /usr/lib64

michal42:
I will think about it and send a v2.

jimklimov:
Succeeded by #496

Issue #500, opened 2017-11-26
nut-scanner: nutscan_init() called before nut_debug_level is set; library scan messages are hidden


clepple:
`nut_debug_level` starts out as 0, but the `-D` flags are not parsed until after `nutscan_init()` is called. Therefore, it is not possible to see the debug messages when searching for libraries.

Tested with https://github.com/networkupstools/nut/tree/libusb-1.0%2B0.1 @ 80bc452.

Noticed while debugging #499.

bigon:
https://github.com/networkupstools/nut/pull/502 is merged, shouldn't this bug be closed?

PR #502, opened 2017-11-27
Problem: nutscan_init() called before debug is enabled

jimklimov:
Solution: Move the call to this stateless function to the point after we parse arguments and just before we'd start the actual scans. Later this can be extended to also consider which scans were requested and so which libs to load.

Thanks to @clepple for research in https://github.com/networkupstools/nut/issues/500


zykh:
ahem... @jimklimov, what about the various places (when parsing opts or displaying help) that rely on a `nutscan_avail_*` var already initialized by `nutscan_init()`?

PR #504, opened 2017-11-27
Add support for openssl-1.1.0

arjendekorte:
Tumbleweed (openSUSE) is switching to OpenSSL-1.1.0 shortly. Apply changes to server & client code and AutoConf magic to make this happen.

### DIFF
--- a/clients/upsclient.c
+++ b/clients/upsclient.c
@@ -745,7 +745,7 @@ static int upscli_sslinit(UPSCONN_t *ups, int verifycert)
 	switch(res)
 	{
 	case 1:
-		upsdebugx(3, "SSL connected");
+		upsdebugx(3, "SSL connected (%s)", SSL_get_version(ups->ssl));
 		break;
 	case 0:
 		upslog_with_errno(1, "SSL_connect do not accept handshake.");
--- a/clients/upsclient.c
+++ b/clients/upsclient.c
@@ -311,15 +311,12 @@ int upscli_init(int certverify, const char *certpath,
 	
 #ifdef WITH_OPENSSL
 
-	SSL_load_error_strings();
-	
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
+	SSL_load_error_strings();
 	SSL_library_init();
 
 	ssl_ctx = SSL_CTX_new(SSLv23_client_method());
 #else
-	OPENSSL_init_ssl(0, NULL);
-
 	ssl_ctx = SSL_CTX_new(TLS_client_method());
 #endif
 
--- a/clients/upsclient.c
+++ b/clients/upsclient.c
@@ -299,11 +299,6 @@ int upscli_init(int certverify, const char *certpath,
 {
 #ifdef WITH_OPENSSL
 	int ret, ssl_mode = SSL_VERIFY_NONE;
-#if OPENSSL_VERSION_NUMBER >= 0x10000000L
-	const SSL_METHOD	*ssl_method;
-#else
-	SSL_METHOD	*ssl_method;
-#endif
 #elif defined(WITH_NSS) /* WITH_OPENSSL */
 	SECStatus	status;
 #endif /* WITH_OPENSSL | WITH_NSS */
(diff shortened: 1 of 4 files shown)


arjendekorte:
Fix  a deprecated warning for openssl-1.1.0. Also allow TLSv1 and above (not just TLSv1).

arjendekorte:
I amended this pull request a couple of times, to add proper support for protocols > TLSv1.

The OpenSSL code was developed against openssl < 1.0.1. At that time, specifying TLSv1_client_method and TLSv1_server_method, was enough to disable the use of SSLv2 and SSLv3 and automatically the highest (being the only remaining one) was used. With TLSv1.3 just around the corned and in some industries, TLSv1 is nearing its end-of-life.

Rather than hardcoding the protocol to TLSv1, this pull requests let the client and server figure out the highest mutual understood version and use that. In order to facilitate debugging which one is actually used, I just added the reporting of the version in debug mode. I doubt it is worth reporting to the user.

clepple:
Hi @arjendekorte, thanks for putting this together. I agree that the code should negotiate the newest TLS version, and it sounds like the newer OpenSSL APIs make this easier than the last time we looked at that.

If I am not mistaken, the negotiated TLS version should be visible in Wireshark (which would seem to me to be a good way to verify that no NUT traffic is in cleartext), so debug logging is probably sufficient.

At the risk of adding too much to the TODO list for 2.7.5, I think this is distinct enough from the other post-2.7.4 changes that we should be able to merge this without any trouble. I would like to do a bit of regression testing (probably this weekend) against packaged versions of NUT beforehand.

arjendekorte:
Well, I guess we never looked at supporting newer TLS versions. The last time I looked at that piece of code (Nov 29, 2010) openSSL didn't support TLSv1.1 and TLSv1.2 (these were added in openSSL 1.0.1, released Mar 14, 2012). At that time supporting everything better than SSLv3, effectively meant TLSv1 only and since then, nobody bothered to add better handshakes when they became available.

For the record, I already tested interoperability pre- and post changes. With both server and client using the changed openSSL initialization code, TLSv1.2 is agreed (for both openSSL-1.0.2 and -1.1.0). If either one is 'old', TLSv1 is used. I did not check against NSS by the way. From looking at that code, it is not clear to me if it supports anything better than TLSv1, it seems to me it might suffer from the same in this aspect.

arjendekorte:
As a side note, I'm working (low profile) on supporting the mbedTLS library as well. I'm loosely involved in OpenWRT/LEDE where this is the default SSL library. Due to constraints on size, both openSSL and NSS libraries are too bloated for a lot of hardware. This would make the mess the upsclient.c and netssl.c are already in even worse, so my proposal would be to remove the #ifdef hell we're in by splitting these in seprate modules:

net-nossl.c
net-openssl.c
net-nss.c
net-mbedtls.c

These would all expose the same functions and since there is very little common code between the different SSL libraries, picking the right one at link time makes more sense. From a maintenance point of view also preferred, since changes like the one above would only touch one SSL implementation, so the risk of accidentally breaking support for other SSL libraries will be lower.

dmacks:
@clepple, I'm tracking openssl-1.1.0 status of fink packages; I'll tag nut as "pending". Let me know if you need help testing in the fink environment.

jimklimov:
Just in case, was it tested against different versions of openssl, to at least compile with them all? :)

Now that we have travis testing, we can add test suites that would request different ubuntu versions (12, 14, now 16 recently) and/or point to specific openssl package versions to install in the testing buildenv.

arjendekorte:
As written above, this was interoperability tested with both OpenSSL 1.0.2 and 1.1.0, the only supported versions nowadays. It should be fine with 1.0.1 though.

clients/upsclient.c
```diff
@@ -315,22 +310,35 @@ int upscli_init(int certverify, const char *certpath,
 	}
 	
 #ifdef WITH_OPENSSL
+
+	SSL_load_error_strings();
```
cotequeiroz: This is done automatically, and was deprecated, starting with openssl 1.1.0; my suggestion would be to move it inside `OPENSSL_VERSION_NUMBER < 0x10100000L`

server/netssl.c
```diff
@@ -388,17 +382,31 @@ void ssl_init(void)
 #ifdef WITH_OPENSSL
 
 	SSL_load_error_strings();
```
cotequeiroz: Same as above, use only when < 1.1.0

m4/nut_check_libopenssl.m4
```diff
@@ -57,8 +57,9 @@ if test -z "${nut_have_libopenssl_seen}"; then
 	AC_MSG_RESULT([${LIBS}])
 
 	dnl check if openssl is usable
-	AC_CHECK_HEADERS(openssl/ssl.h, [nut_have_openssl=yes], [nut_have_openssl=no], [AC_INCLUDES_DEFAULT])
-	AC_CHECK_FUNCS(SSL_library_init, [], [nut_have_openssl=no])
```
cotequeiroz: Checking for `SSL_CTX_new` would be enough for what we need.

clients/upsclient.c
```diff
@@ -315,22 +310,35 @@ int upscli_init(int certverify, const char *certpath,
...
 	
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 	SSL_library_init();
-	SSL_load_error_strings();
 
-	ssl_method = TLSv1_client_method();
+	ssl_ctx = SSL_CTX_new(SSLv23_client_method());
+#else
+	OPENSSL_init_ssl(0, NULL);
```
cotequeiroz: This call is not necessary, as library initialization is done automatically now.

server/netssl.c
```diff
@@ -388,17 +382,31 @@ void ssl_init(void)
...
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 	SSL_library_init();
 
-	if ((ssl_method = TLSv1_server_method()) == NULL) {
+	ssl_ctx = SSL_CTX_new(SSLv23_server_method());
+#else
+	OPENSSL_init_ssl(0, NULL);
```
cotequeiroz: Not necessary, same as above.

m4/nut_check_libopenssl.m4
arjendekorte: Agree

clients/upsclient.c
arjendekorte: Agree

server/netssl.c
arjendekorte: Agree

PR #512, opened 2018-01-03
PEP8 cleanup.  

Cabalist:
PEP8.   Updated code to be more compliant with Python3.  Renamed a couple things that were colliding with reserved kws.  Use "in" instead of ".has_key()"

The app has dependencies on an older GTK library that will take some time to replace.  I am submitting this pull request as laying the groundwork for that cleanup.   Thanks!

### DIFF
--- a/scripts/python/app/NUT-Monitor
+++ b/scripts/python/app/NUT-Monitor
@@ -28,269 +28,274 @@
 #            Corrected unsafe permissions on ~/.nut-monitor (Debian #777706)
 
 
-import gtk, gtk.glade, gobject
-import sys
 import base64
-import os, os.path
-import stat
-import platform
-import time
-import threading
-import optparse
-import ConfigParser
-import locale
 import gettext
-import PyNUT
+import optparse
+import os
+import os.path
+import platform
+import stat
+import sys
+import threading
+import time
 
+try:
+    from configparser import ConfigParser  # Py3
+except ImportError:
+    from Configparser import ConfigParser  # Py2
+
+import PyNUT
+import gobject
+import gtk
+import gtk.glade
 
 # Activate threadings on glib
 gobject.threads_init()
 
-class interface :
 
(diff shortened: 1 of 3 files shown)


clepple:
Sorry I haven't had a chance to look at this until now. I just checked out this branch on Debian stretch, and did a quick regression test using Python 2.7. Relative to the NUT-Monitor script in stretch (from NUT 2.7.4), your changes do not seem to allow saving favorites. Let me know if you need an exact traceback, but I think it had a problem trying to save the `False` boolean value to "auth" - something about values only being strings.

jimklimov:
A few merge conflicts cropped up :(

jimklimov:
I fixed the merge conflicts to make this PR applicable to master, to the best of my understanding of the script (in particular, there was a bit about renaming the variables to avoid reserved key words).

Still, marked as "Needs testing" to have someone check the GUI tool changes before merging.

CC @melenki as a recent contributor to NUT-Monitor : Alexey, would you please have a chance to revise this PR and test its version of the scripts on a GUI workstation to make sure stuff still works? ;)

melenki:
[...] 
> CC @melenki as a recent contributor to NUT-Monitor area: Alexey, would you please have a chance to revise this PR and test its version of the scripts on a GUI workstation to make sure stuff still works? ;)

Since the patch was published, the Python version has changed, and so has the ABI.
NUT-Monitor does not start after applying the patch. A number of language constructs for Python versions 3.6/3.8 require refactoring. For example, the method 'glade' is not present in the Gtk module for Python3.

To fully meet the requirements of Python 3.x, a patch needs to be reworking.

jimklimov:
Note regarding comment for Python 3: it is a different language than Python 2 :)

That said, several layers of PRs were made to make some code dually supported on both, and reimplement the GUI for the new tech, e.g. #1310

PR #528, opened 2018-02-20
Sort input file list

bmwiedemann:
so that nut builds in a reproducible way
in spite of indeterministic filesystem readdir order
and http://bugs.python.org/issue30461

See https://reproducible-builds.org/ for why this is good.

### DIFF
--- a/tools/nut-snmpinfo.py
+++ b/tools/nut-snmpinfo.py
@@ -80,7 +80,7 @@ output_file.write( "\n" )
 output_file.write( "/* SNMP IDs device table */\n" )
 output_file.write( "static snmp_device_id_t snmp_device_table[] = {\n" )
 
-for filename in glob.glob('../drivers/*-mib.c'):
+for filename in sorted(glob.glob('../drivers/*-mib.c')):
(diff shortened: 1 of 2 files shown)


bmwiedemann:
Can I do something to help get this merged?

jimklimov:
Poke more? :)
Looks good to me, though I think the sorting here would be based on LANG or LC_ALL active during a build (not sure if that impacts the ASCII filenames anyhow).

bmwiedemann:
I think that is more a concern for the standalone sort(1) program and less so for the builtin perl/python sort functions.

https://stackoverflow.com/questions/8776807/is-pythons-sort-function-the-same-as-linuxs-sort-with-lc-all-c seems to suggest that it is not using locales by default.

jimklimov:
@clepple : was there a particular reason you self-assigned this PR and it got on hold over time, e.g. do you have something in mind to experiment with / make sure of? Or can it indeed be merged?

aquette:
LGTM
as for Charles' self assignment, I probably asked him to take over...

aquette:
@bmwiedemann thanks for your contribution

PR #536, opened 2018-03-09
main.c upsdrvctl.c : make debug messages a bit more useful

jimklimov:
Tracing issues with service development, I found these messages are less than useful, and in case of `upsdrvctl stop` - even misleading (why does it make a pidfile with device name but look for the port name? oh, now we know).


jimklimov:
Thanks for the kind words ;)

PR #539, opened 2018-03-13
doc: Update installation instructions for FreeBSD

georgefx:
Details usb device permissions. Also pkg and new port config style.


clepple:
@georgefx thanks!

PR #545, opened 2018-03-23
siemens-sitop: new driver for Siemens SITOP UPS500-series UPSes

mtenberge:
Please consider this new driver for inclusion in NUT.

Siemens has a line of small UPSes, specially designed for use in machine control cabinets with embedded industrial PCs: they operate at 24V DC instead of AC line voltages, and are suited for DIN-rail mounting.
Machines are mostly switched off by simply interrupting the mains voltage, so with the growing popularity of embedded Linux systems in industry, I expect these UPSes to be used more and more.
There are models with lead acid batteries, but also maintenance-free models with supercaps. Some models have serial ports, others have USB interfaces.

Siemens only provides a Windows application to interface with these UPSes, but they have published the protocol in the device manual.
The protocol is not compatible with existing drivers in NUT, so I have written a new driver and corresponding documentation.
Theoretically, all models from the series (both serial and USB) should work, but as mentioned in the docs, I have only tested the driver with only 1 model variant with USB port.

### DIFF
--- a/docs/man/siemens-sitop.txt
+++ b/docs/man/siemens-sitop.txt
@@ -20,7 +20,8 @@ so as far as this driver is concerned, all models are serial models.
 This driver should work with all models in the SITOP UPS500 series,
 as long as your kernel has support for the serial port device
 (see section *USB driver* below).
-Note: this driver has only been tested with the SITOP UPS500S-2.5
+
+NOTE: This driver has only been tested with the SITOP UPS500S-2.5
 with USB port (Siemens product number 6EP1933-2EC41).
 
 DEVICE SETTINGS
@@ -32,8 +33,8 @@ to the following:
 *switch 1-4*::
 Choose whatever suits your situation. Any combination will work with NUT.
 
-*switch 5* ('=>' / 't')::
-Set to OFF ('t'). This ensures that the UPS will not cut power unless NUT
+*switch 5* (=> / t)::
+Set to OFF (t). This ensures that the UPS will not cut power unless NUT
 tells it to do so (or unless the batteries are exhausted).
 
 *switch 6-10* (delay)::
@@ -54,26 +55,33 @@ supply power from its batteries.
 USB driver
 ----------
 The USB-versions of the UPS contain an FTDI USB-to-serial converter chip.
-They are programmed with a non-standard product ID (for example 0403:e0e3),
+It is programmed with a non-standard product ID (for example _0403:e0e3_),
 but can still be used with the normal ftdi_sio driver.
 
-Note: the following hints may be specific to GNU/Linux.
+NOTE: The following hints may be specific to GNU/Linux.
+
 Use *lsusb* to figure out which product ID is used in your model, and
-replace all occurrences of e0e3 in the following examples with the actual
+replace all occurrences of _e0e3_ in the following examples with the actual
 Product ID.
 
-# modprobe ftdi_sio
-# echo 0403 e0e3 > /sys/bus/usb-serial/drivers/ftdi_sio/new_id
+....
+modprobe ftdi_sio
+echo 0403 e0e3 > /sys/bus/usb-serial/drivers/ftdi_sio/new_id
+....
 
 If your system uses *udev*, this can be automated via a udev rule:
+
+----
 ACTION=="add", ATTRS{idVendor}=="0403", ATTRS{idProduct}=="e0e3", \
   RUN+="/sbin/modprobe ftdi_sio", \
   RUN+="/bin/sh -c 'echo 0403 e0e3 > /sys/bus/usb-serial/drivers/ftdi_sio/new_id'"
+----
 
 You can use the following udev rule to obtain a predictable device name,
-for example /dev/ttyUPS:
+for example _/dev/ttyUPS_:
+----
 SUBSYSTEM=="tty" ATTRS{idVendor}=="0403", ATTRS{idProduct}=="e0e3" SYMLINK+="ttyUPS"
-
+----
 
 EXTRA ARGUMENTS
 ---------------
@@ -81,9 +89,9 @@ This driver supports the following optional settings:
 
 *max_polls_without_data*='num'::
 According to the specifications, the UPS should emit a new status update
-every 75 msec. The serial port is polled every second for new data. If there
-is no valid new data after 'num' polls, it is assumed that communication with
-the UPS is lost.
+every 75 msec. The serial port is polled periodically for new data (recommended
+setting: every second). If there is no valid new data after 'num' polls, it is
+assumed that communication with the UPS is lost.
 The default value is 2. Lower values may cause spurious 'Data stale' messages,
 especially at startup.
 
@@ -92,19 +100,21 @@ INSTANT COMMANDS
 
 *shutdown.return*::
 The behavior of this command depends on the line state:
-*on line*: after 5 seconds (or longer, if DIP switches 6-10 are not OFF),
+
+* *on line*: after 5 seconds (or longer, if DIP switches 6-10 are not OFF),
 the UPS will shut off its output. After another 5 seconds, the output is
 activated again.
-*on battery*: after 5 seconds (or longer, if DIP switches 6-10 are not OFF),
+* *on battery*: after 5 seconds (or longer, if DIP switches 6-10 are not OFF),
 the UPS will shut off its output. The output will stay off, until the
 line voltage has returned.
 
 *shutdown.stayoff*::
 The behavior of this command depends on the line state:
-*on line*: after 5 seconds (or longer, if DIP switches 6-10 are not OFF),
+
+* *on line*: after 5 seconds (or longer, if DIP switches 6-10 are not OFF),
 the UPS will shut off its output. The output stays off, until the line voltage
 has been removed for at least 1 second, and has been re-applied.
-*on battery*: this command behaves the same as *shutdown.return*.
+* *on battery*: this command behaves the same as *shutdown.return*.
 
 INSTALLATION
 ------------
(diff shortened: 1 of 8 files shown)


drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+		if (check_for_new_data())
+			found = 1;
+		else
+			sleep(1); /* Sleep a while, then try again */
+		max_attempts--;
+	}
+	if (!found)
+		fatalx(EXIT_FAILURE, "No data received from the UPS");
+	
+	dstate_setinfo("ups.mfr", "Siemens");
```
zykh: I'd remove these redundant 'ups.*' items.

drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+	}
+	
+	if (nr_polls_without_data > max_polls_without_data) {
+		/* data is stale */
+		dstate_datastale();
+		return;
+	}
+	
+	/* This is all we know about the charge level... */
+	dstate_setinfo("battery.charge.indication",
```
zykh: This is a blocker. If you add a new var you are supposed to:
- document it in nut-names.txt / cmdvartab
- get it approved.

drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+	dstate_dataok();
+	ser_comm_good();
+
+	/*
+	 * Fast polling is preferred, because
+	 * A) the UPS spits out new data every 75 msec,
+	 * B) some models in this SITOP series have a _very_ small capacity
+	 *    (< 10sec runtime), so every second might count.
+	 */
+	poll_interval = 1;
```
zykh: hmm... I'm not so sure this is the best thing to do (without even documenting it in the man page): people will complain that the driver doesn't respect their settings.

drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+}
+
+void upsdrv_shutdown(void) {
+	/* tell the UPS to shut down, then return - DO NOT SLEEP HERE */
+	upslogx(LOG_NOTICE, "instcmd: sending command R");
+
+	/* Note: the UPS does not really like to receive data.
```
zykh: Why not moving these remarks in `instcmd()` and simply call `instcmd("shutdown.return", NULL)` here?

drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+	int parsed;
+	
+	upsfd = ser_open(device_path);
+	ser_set_speed(upsfd, device_path, B9600);
+
+	/* option max_polls_without_data: */
+	max_polls_without_data = 2;
+	maxPollsString = getval("max_polls_without_data");
+	if (maxPollsString) {
+		if (sscanf(maxPollsString, "%d", &parsed) == 1) {
```
zykh: What about using our `str_to_uint()` (the sign is ignored)?

docs/man/siemens-sitop.txt
```diff
@@ -0,0 +1,164 @@
...
+SUPPORTED HARDWARE
+------------------
+*siemens-sitop* supports Siemens UPS models from the SITOP UPS500 series.
+Some models have a serial port, others have a USB port.
+The models with USB port actually contain a serial-over-USB chip,
+so as far as this driver is concerned, all models are serial models.
+This driver should work with all models in the SITOP UPS500 series,
+as long as your kernel has support for the serial port device
+(see section *USB driver* below).
+Note: this driver has only been tested with the SITOP UPS500S-2.5
```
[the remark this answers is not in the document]
mtenberge: Thanks for your review and comments.
I'm new to writing manpages, and just found out that 'AsciiDoc' is the magic word to google... ;-)
I reformatted the text in various locations.

drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+static int check_for_new_data() {
+	int new_data_received = 0;
+	int done = 0;
+	int num_received;
+	
+	while (!done) {
+		/* Get new data from the serial port.
+		 * No extra delay, just get the chars that were already buffered.
+		 */
+		num_received = ser_get_buf(upsfd, rx_buffer + rx_count, RX_BUFFER_SIZE - rx_count, 0, 0);
```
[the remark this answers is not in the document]
mtenberge: For normal situations it is indeed intentional, because it cannot be guaranteed that a string arrives at once. When it is split up by the transport layer, it will be re-assembled automatically by the buffer append.
It doesn't hurt if stale data is lying around in the input buffer, because it will eventually be removed by the 'shift unusable data out of the buffer' action.
For comm failure, I have added a 'flush buffer' action in the next commit, because it is very easy and readable to implement. For timeout situations, the added code complexity does not weigh up against the simplicity of shifting out the stale data, so I left this as it is.

drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+			ser_set_speed_nf(upsfd, device_path, B9600);
+			done = 1;
+		} else if (num_received == 0) {
+			/* no (more) new data */
+			done = 1;
+		} else {
+			rx_count += num_received;
+			
+			/* parse received input data: */
+			while (rx_count >= (5 + NR_DELIMITER_CHARS)) { /* all input data are strings of 5 characters plus 1 or more delimiter characters */
```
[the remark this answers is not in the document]
mtenberge: Good point. The delimiter characters are not mentioned in the protocol description, so by not relying on it will make the driver more robust. I changed this in the next commit.

drivers/siemens-sitop.c
```diff
@@ -0,0 +1,314 @@
...
+				if (strncmp(rx_buffer, "BUFRD", 5) == 0) {
+					current_ups_status.battery_alarm = 0;
+					rm_buffer_head(5 + NR_DELIMITER_CHARS);
```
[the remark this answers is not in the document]
mtenberge: Done

drivers/siemens-sitop.c
mtenberge: I got those from the skeleton driver skel.c, so that one might needs a revision as well.

drivers/siemens-sitop.c
mtenberge: I have considered to use the existing variable battery.charge, but people (and code?) would probably assume that it contains an integer (percentage), not a string ("<85%" or ">85%").
As far as I could see, the list of existing variables does not contain a variable that suits this purpose, or does it?

drivers/siemens-sitop.c
mtenberge: You are right. Instead, I added a log message if the interval is set higher than 5.

drivers/siemens-sitop.c
mtenberge: Yes, better.

drivers/siemens-sitop.c
mtenberge: If it only would have been mentioned somewhere in the Developer Guide, I would have known about it ;-)

mtenberge:
One issue remaining: the naming of new drivers.
Looking in the source tree, there are only a few driver names that start with nutdrv_. Most of them are qx drivers, and they contain both dashes and underscores in their names (e.g. nutdrv_qx_voltronic-qs-hex).
Compilation results in e.g. nutdrv_qx-nutdrv_qx_voltronic-qs-hex.o. I'm not sure if that should be the desired result...
Apart from the qx-drivers, there is only nutdrv_atcl_usb (which apparently uses underscores). If this is the desired new naming convention, I will rename the siemens driver to follow this convention.
Please consider adding this information to Chapter 4 of the Developer Guide ('Creating a new driver'), as there is no use in having a naming convention if nobody knows about it...

zykh:
Thanks for addressing those issues.
And yes, you're right on the need to update `skel.c` and document `str_*()` functions and the new naming convention (speaking of the latter, the built drivers are `nutdrv_qx` and `nutdrv_atcl_usb`... and there are only these two simply because those are the only ones added after we decided to adopt the new convention): I'll either do that in the coming days or add issues here on GitHub so that we don't forget about that.

Two (three) things left:
- I agree with you on `battery.charge` (that's why my comment was about the things to do when adding a new var) -- I'm more for a name like `battery.charge.approximation`, though, and I'm not completely sure on what should be the format of its value, apart from the fact that it should not have the '%' explicitly expressed in the value to go along with other similar vars, but, probably, something like you did (`< 85` / `> 85`) should be just fine with me, or maybe we could use an interval (e.g. `85-100`), but I'm not so fond of it, either,
- in the manpage, I think it should be better to actually mention `pollinterval` (that is, something like: 'The serial port is polled periodically, every *pollinterval*, for new data (recommended setting: every second) ...').
- (my git complains about whitespace errors).

In my opinion this is (almost) ready, so I'm adding it to 2.7.5 milestone.

mtenberge:
About the charge level indication variable: I started a thread about this on the nut-upsdev list, but up to now it remains quiet there… (https://alioth-lists.debian.net/pipermail/nut-upsdev/2018-April/thread.html). For now, I removed the % characters from the indication values.

I added a separate section to the man page about polling, and fixed the line endings.

aquette:
@zykh and @mtenberge what's the status on this one?
I'm back on trying (again) to push 2.7.5...

zykh:
@aquette: code-wise, this should be ready (after a rebase).

As mentioned in the last comment, the only thing left is the publication of the (approximate) level of battery charge reported (these devices only report if they are below or above 85%) and the formalisation of the var used for that purpose and of its content (i.e. document it in `docs/nut-names.txt` and in `data/cmdvartab`): what do we want to do?

a. add a new variable for things like that?
b. or, don't use that data at all?

I'm more for the 'a' option, which is the one @mtenberge already implemented, but with a name of `battery.charge.approx` (in future, we could also consider to make it a general suffix and extend its use to other vars), instead of the implemented `battery.charge.indication`.
And, for the published content, the already implemented option (`<85` / `>85`) looks (not beautiful, but) good to me.

But I'm not good at naming things...

mtenberge:
I don't have a strong opinion on the variable naming, I'll leave that up to you core devs to decide.

I'll see if I can do a rebase this week.

zykh:
@mtenberge, thanks and sorry for the huge delay.
Since, meanwhile, no one objected, later today, I'll merge this and apply the change I suggested.
So.. last chance to shout at me...

PR #558, opened 2018-06-01
Add compatibility with openssl 1.1.0 - minimal changes

cotequeiroz:
Hello

I saw PR #504 about openssl 1.1.0, but I decided to add another one that does not touch code outside of openssl 1.1.0, so perhaps it could get merged sooner.  I do agree with most of the changes brought by @arjendekorte, by the way.

Signed-off-by: Eneas U de Queiroz <cote2004-github@yahoo.com>

### DIFF
--- a/m4/nut_check_libopenssl.m4
+++ b/m4/nut_check_libopenssl.m4
@@ -58,7 +58,7 @@ if test -z "${nut_have_libopenssl_seen}"; then
 
 	dnl check if openssl is usable
 	AC_CHECK_HEADERS(openssl/ssl.h, [nut_have_openssl=yes], [nut_have_openssl=no], [AC_INCLUDES_DEFAULT])
-	AC_CHECK_FUNCS(SSL_library_init, [], [nut_have_openssl=no])
+	AC_CHECK_FUNCS(SSL_CTX_new, [], [nut_have_openssl=no])
 
 	if test "${nut_have_openssl}" = "yes"; then
 		nut_with_ssl="yes"
--- a/clients/upsclient.c
+++ b/clients/upsclient.c
@@ -316,10 +316,13 @@ int upscli_init(int certverify, const char *certpath,
 	
 #ifdef WITH_OPENSSL
 	
+# if OPENSSL_VERSION_NUMBER < 0x10100000L
 	SSL_library_init();
 	SSL_load_error_strings();
+#  define TLS_client_method TLSv1_client_method
+# endif /* OPENSSL_VERSION_NUMBER < 0x10100000L */
 
-	ssl_method = TLSv1_client_method();
+	ssl_method = TLS_client_method();
 
 	if (!ssl_method) {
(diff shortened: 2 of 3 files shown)


m4/nut_check_libopenssl.m4
```diff
@@ -58,7 +58,7 @@ if test -z "${nut_have_libopenssl_seen}"; then
 
 	dnl check if openssl is usable
 	AC_CHECK_HEADERS(openssl/ssl.h, [nut_have_openssl=yes], [nut_have_openssl=no], [AC_INCLUDES_DEFAULT])
-	AC_CHECK_FUNCS(SSL_library_init, [], [nut_have_openssl=no])
+	AC_CHECK_FUNCS(SSL_CTX_new, [], [nut_have_openssl=no])
```
jimklimov: Just in case : Is this symbol available in all openssl versions?

If not, maybe its presence vs. the older symbol can be used to differentiate behaviors somehow (as a measure of codebase age)...

jimklimov:
Looks elegant, I like :)

m4/nut_check_libopenssl.m4
cotequeiroz: SSL_CTX_new is available since at least 0.9.6.  My goal was to find a common-ground function.

arjendekorte: It's been available since SSLeay 0.5.1 (the predecessor of OpenSSL), so it's available everywhere.

One should only look at OPENSSL_VERSION_NUMBER to determine if some function is available, the check for SSL_CTX_new is only there to make sure if the libraries are installed properly.

arjendekorte: Using the TLSv1_client_method was a quick-and-dirty way to disable SSLv2 and SSLv3 and really never the way to do this. Look at PR#504 for the proper way. You really shouldn't use TLSv1 if both server and client support TLSv1.2.

arjendekorte: Also, PR #504 also documents why we did this back then. The code has been slowly rotting in the tree until OpenSSL 1.1.0 came out, by which time I was no longer involved in active development.

cotequeiroz: I never meat this to replace #504, since it is more complete than this one.  I just saw it open for a while, and tried a slimmer version in the hope to get it to at least compile with openssl 1.1.x.  By all means, your idea is right on.

cotequeiroz: I'll close this, and add some comments on #504.  This is the only one change I think could be used the way I wrote, as it does its job and is simpler/faster.

arjendekorte: I agree, checking for SSL_CTX_new is probably less intrusive and should be enough to make sure that not only the headers are available, but one can actually link to OpenSSL as well. I remember this was a frequent source of problems.

Issue #441, opened 2017-06-09
nutdrv_qx: support replies not terminated by the required CR


zykh:
Apparently some devices don't close their replies to our commands/queries with the expected (and mandated by the specs) CR and, at the moment, this is not supported by nutdrv_qx:
http://lists.alioth.debian.org/pipermail/nut-upsuser/2017-June/010714.html
('krauler' USB subdriver)

As far as I can remember, this eventuality can only happen with certain USB subdrivers that only read a predetermined amount of data, read data from a given index or wait till no more data is available (if I recall correctly, serial communication and other USB subdrivers need the terminating CR in order to work properly), so, probably, the easiest fix should be to handle this situation in those USB subdrivers.

Also, this should not be considered an improvement, but a regression that needs to be solved, since blazer_* drivers, being less strict on the terminating CR of Q1 replies (3747fd5), should (sort of) support (some of) this kind of devices.

MAINTAINER NOTE: The null byte vs. `\r` at end of reads sounds reminiscent of work recently done by @blaa for the `nutdrv_qx` subdriver for `armac` in #2005; maybe this part should/can be generalized somehow to help with all devices that do not use CR/LF in protocol? It seems that the 1595a06501daa93e06035a861e3db7ccab2871dd commit which allegedly solved this issue previously added the fix for certain `*_command()` methods, and so is lacking in some others that appeared afterwards.

clepple:
@zykh is it okay if I move this to the 2.7.6 milestone?

zykh:
ahem, yes, @clepple... it turns out I completely forgot this one...

clepple:
no problem. (Hopefully the next release won't be a year from now :-) )

zykh:
moving back to 2.7.5, since a) there is a workaround (https://github.com/zykh/nut/commit/1595a06501daa93e06035a861e3db7ccab2871dd) that apparently works as expected, and b) this is needed for #616

trentbuck:
For the record,

I encountered this bug in nut 2.7.4 on Debian 11 with my UPS, model "UPSONIC IRT-3K 2U".  My UPS is about 10 years old.  The equivalent current model appears to be https://www.upsonic.com.au/upsonic-esart-rack-or-tower-series-ups/

https://bugs.debian.org/1019624

[...]

PR #559, opened 2018-06-05
Check for <CR> in status answer

dudanov:
This commit fix no check for trailing character <CR> in status answer.

### DIFF
--- a/drivers/blazer.c
+++ b/drivers/blazer.c
@@ -189,7 +189,7 @@ static int blazer_status(const char *cmd)
 	 *    01234567890123456789012345678901234567890123456
 	 *    0         1         2         3         4
 	 */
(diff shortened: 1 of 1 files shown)


zykh:
That was actually done intentionally (maybe in a perfectible way) to support those (USB) devices that don't terminate the reply with a CR (3747fd5).
Are you experiencing problems with it?

We (I) are waiting that the libusb-1.0 things get merged (in order to avoid conflicts) to add support for this kind of devices even in `nutdrv_qx` (#441), and knowing of any problem beforehand would surely be useful.

jimklimov:
Unfortunately the original fork this PR came from is no longer available, so it can not be bumped to current master. But the change in code is small and trivial; complexity is about checking the logic and/or testing against hardware regarding whether this change is correct in essence.

Issue #560, opened 2018-06-13
support for PowerWalker VFI 2000 TGS


Links2004:
Hi,
I have a PowerWalker VFI 2000 TGS UPS and I am able to get some basic values via `upsc`,
but many values are missing, like:

- Input Frequency
- Input Voltage
- Output Frequency
- Output Voltage
- correct battery Voltage (54.8V not 55V)

what is needed to do to map values?

```
# upsc PowerWalker-hid
Init SSL without certificate database
battery.charge: 92
battery.runtime: 31503
battery.type: PbAc
battery.voltage: 55
battery.voltage.nominal: 48
device.mfr: PPC
device.model: HID UPS
device.serial: 170909-54810007     
device.type: ups
driver.name: usbhid-ups
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.port: auto
driver.parameter.productid: ffff
driver.parameter.synchronous: no
driver.parameter.vendorid: 06da
driver.version: 2.7.4
driver.version.data: Phoenixtec/Liebert HID 0.3
driver.version.internal: 0.41
ups.load: 0
ups.mfr: PPC
ups.model: HID UPS
ups.productid: ffff
ups.serial: 170909-54810007     
ups.status: OL CHRG
ups.vendorid: 06da
```

```
# usbhid-ups -D -a PowerWalker-hid

Network UPS Tools - Generic HID driver 0.41 (2.7.4)
USB communication driver 0.33
   0.000000	debug level is '1'
   0.000309	upsdrv_initups...
   0.010152	Using subdriver: Phoenixtec/Liebert HID 0.3
   0.010160	144 HID objects found
   0.010346	Path: UPS.BatterySystem.Battery.BatteryID, Type: Feature, ReportID: 0x20, Offset: 0, Size: 8, Value: 1
   0.010578	Path: UPS.BatterySystem.Battery.ffff0090, Type: Feature, ReportID: 0x21, Offset: 0, Size: 8, Value: 1
   0.010841	Path: UPS.BatterySystem.Battery.PresentStatus.Present, Type: Feature, ReportID: 0x27, Offset: 0, Size: 8, Value: 1
   0.010866	Path: UPS.BatterySystem.Battery.PresentStatus.Present, Type: Input, ReportID: 0x27, Offset: 0, Size: 8, Value: 1
[...]

PR #564, opened 2018-06-19
support for PowerWalker VFI 2000 TGS by usbhid-ups

Links2004:
see #560


clepple:
Please squash these two commits together.

Links2004:
done

aquette:
trying to resurrect this PR:
* there is indeed an issue between liebert-hid.c and mge-hid.c, which both defines 0x06da / 0xffff due to the OEM behind (Phoenixtec)
* I need to test your present modification, but at first sight, there is some conflict between few vars.
* if the above issue is proven, I'd be in favor of putting the adapted content from mge-hid with your mods into liebert-hid.c

jimklimov:
Looking quickly into the vendorID clash, it seems that "Liebert/Phoenixtec HID" is tracked since at least 2012, and the same VID was added to mge-hid.c with commit a08a2292af in 2016 (perhaps after NUT 2.7.4 release, did not check dates closer), apparently because AEG PROTECT NAS began using that instead of an earlier VID recorded in the file.

I do not know if really those "newer" AEG devices use the MGE protocol as the "older" ones did, or also switched to a Liebert dialect along with the chip. Maybe the subdriver should check that this vendor id is accompanied by string containing "AEG" to keep processing with mge-hid and otherwise mark as unsupported so liebert-hid would take over? Or just drop that ID from mge-hid?

Maybe an extended subdriver matcher (or structure at least) like that used in `nutdrv_qx.c` to consult more fields than the basic two is a better solution for other drivers as well.

Either way, it does not seem that PowerWalker devices are particularly "married" to one or another HID Vendor, and Liebert HID was there first :)

jimklimov:
Transplanted the lines from mge-hid.c to liebert-hid.c so as to not break the former.

It seems that the mapping table for mge-hid has many more entries than liebert-hid. I have no idea if those other mapped entries are actually served by Liebert/Phoenixtec devices or come from other vendor subdrivers using that file. Original issue #560 has a data walk from this UPS which can help select the missing mappings for served data points (copy-paste from mge-hid?) but this is something better done with (and tested against) a real device.

jimklimov:
Looking at the usbhid-ups.c driver source, it seems that it goes through the subdrivers array and calls their `claim()` methods until someone says the device is handled. This way the mge subdriver probably grabs 0x06da and liebert subdriver never gets a chance. Posted PR #1357 to try addressing that gap.

PR #584, opened 2018-07-24
Updates for Tripp Lite 3005 protocol

mbastiaan:
Added battery.charge status for 3005 protocol. 
Tested on SMX500RT1U, charge level seems ok-ish considering how it's calculated and the voltage jumping up and down.


clepple:
Two small requests:
* The first line of the each of the commit messages is important for determining what changed. Please check the commit logs for the driver and driver.list.in files to see examples. This helps when searching logs for changes related to a given model.
* When changing the logic of a driver, it is helpful to increment the `DRIVER_VERSION` variable so that the difference shows up in `upsc` output.

A corollary of the last one: the output of `upsc` and `upsrw` would be useful as described here: https://networkupstools.org/stable-hcl.html#footnotes (the [last SMX entry](https://networkupstools.org/ddl/Tripp_Lite/SMX1500XLRT2U.html) we have is from a while ago, with a different protocol)

mbastiaan:
upsc returns the following info with the above commits, without those commits the battery.charge is not there and the driver version is 0.29:
battery.charge: 100
battery.voltage: 13.90
battery.voltage.nominal: 12
device.mfr: Tripp Lite
device.model: SMX500RT1U
device.type: ups
driver.name: tripplite_usb
driver.parameter.port: auto
driver.parameter.synchronous: no
driver.version: 2.7.4.1
driver.version.internal: 0.30
input.voltage: 234
input.voltage.maximum: 239
input.voltage.minimum: 229
input.voltage.nominal: 120
outlet.1.desc: Load 1
outlet.1.id: 1
outlet.1.switch: 1
outlet.1.switchable: 1
outlet.2.desc: Load 2
outlet.2.id: 2
outlet.2.switchable: 0
ups.debug.D: 00 ea 00 8b 0d 00 00 '.......'
ups.debug.L: 10 00 00 07 58 58 0d '....XX.'
ups.debug.load_banks: 1
ups.debug.M: 00 e5 00 ef 0d 00 00 '.......'
ups.debug.P: 30 30 35 30 30 58 0d '00500X.'
ups.debug.S: 01 04 00 00 64 00 0d '....d..'
ups.debug.T: 29 00 01 f3 00 58 0d '.....X.'
ups.debug.V: 06 00 0c 01 58 58 0d '....XX.'
ups.delay.shutdown: 64
ups.firmware: F334403
ups.firmware.aux: protocol 3005
ups.id: 0
ups.mfr: Tripp Lite
ups.model: SMX500RT1U
ups.power.nominal: 500
ups.productid: 0001
ups.status: OL
ups.temperature: 0
ups.vendorid: 09ae

upsrw returns:
[outlet.1.switch]
Outlet switch control
Type: STRING
Maximum length: 3
Value: 1

[ups.delay.shutdown]
Interval to wait after shutdown with delay command (seconds)
Type: STRING
Maximum length: 3
Value: 64

[ups.id]
UPS system identifier
Type: STRING
Maximum length: 5
Value: 0

clepple:
> 
> input.voltage: 234
> input.voltage.maximum: 239
> input.voltage.minimum: 229
> input.voltage.nominal: 120

From the product link, it looks like 120V is incorrect. Is this a universal (120V/240V) input device?

Also, have you had a chance to do any shutdown testing?

mbastiaan:
The 120V is indeed incorrect, it's a 230/240V only device.
I'll do some testing to see what is going on there, I'm expecting it to be the standard values set in the driver.

Shutdown isn't working as the status never gets set to low battery.

mbastiaan:
That commit raises the question if the other cases are correct or that the quotes needs to be removed?
And it also messes up the calculation of the input voltage as those are multiplied by the scaled voltage and then divided by 120. Is there a reason for this multiplication and then division? using the value as is seems perfectly fine for my ups?

mbastiaan:
I did some major work on the driver, please check this diff before I cut it into smaller commits and add them to this pull request. The indents are a bit messy due to copy/paste, but I'll clean that up when doing commits.

https://github.com/networkupstools/nut/compare/master...mbastiaan:tripplite_3005_2?expand=1#diff-47c4f39bba5fcd77234c60c1d9d322feL1

With all the changes all values as I mentioned in issue 585 are implemented for the 3005 protocol as well as some changes to make automatic shutdown work. The UPSs I have don't set LB themselves, so I tried different methods but simply tracking the lowest battery charge and setting status to LB when it drops below a certain value works best. I added 2 driver parameters to make this user manageable, 1 parameter to enable the overruling, 1 parameter to set the threshhold value. User will need to do testing to get the correct value for their situation.

I also moved decove_V to be handled the same as the other debug arrays.

clepple:
Thanks for cleaning up the whitespace (this driver predates the repository conversion to Git, so it wasn't as obvious).

I would actually prefer that more of the decoding logic get moved out into separate functions, if only to keep the average function size small enough to be viewed at once, and to avoid introducing errors related to reuse of variables. As you can see, the addition of other protocols has bloated the code somewhat. I know it's not great at the moment, but inlining decode_v() is a step in the wrong direction, IMHO.

Also, I'm not sure I follow the reasoning for removing the multiplication by input_voltage_scaled. Isn't this going to break 208V setups?

mbastiaan:
I can agree in the decode_v(), I'll make separate functions like decode_v() for all the decodes.

The reason for removing the multiplication by input_voltage_scaled and then divide by 120 is because this breaks the voltage measurements on my 230V UPSs.
My UPSs (I have multiple of these) all report the actual voltage measured, so there is no need for scaling.

See also these UPS reports, they return back hex values of 007# on D[1,2], which is in the 110~120 range.
https://networkupstools.org/ddl/Tripp_Lite/SMART500RT1U.html
MikeMark's data also reports back hex values of 007#.
https://github.com/networkupstools/nut/issues/585
My UPSs report hex values in the 00e#-range, which is around 230V.
These are all single phase measurements, but I have no reason to believe that for a 3 phase unit (208V is 3 phase 120V) suddenly scaling would be required.

I believe the reason nobody ever noticed or bothered to fix this is because everybody was either testing on 120V units or using the standard values set in the driver because case 6 was missing. All of which have a input_voltage_scaled of 120 so the only thing ever happening is multiply by 120, divide by 120 and you end up with the value you began with.

See line [268](https://github.com/networkupstools/nut/blob/d16b508a63660f308cbaf5badcb3d5043538350a/drivers/tripplite_usb.c#L268) of the current driver for the standard setting of input_voltage_scaled.
See line [452](https://github.com/networkupstools/nut/blob/d16b508a63660f308cbaf5badcb3d5043538350a/drivers/tripplite_usb.c#L452) of the current driver for the case statements that set input_voltage_scaled.
See line [1267](https://github.com/networkupstools/nut/blob/d16b508a63660f308cbaf5badcb3d5043538350a/drivers/tripplite_usb.c#L1267) of the current driver for the voltage calculations.

So let just go through all of the cases and see what happens:
- Case 0 sets input_voltage_scaled to 100, on line 1267 we then multiply the measured value (which I assume is the actual value on all UPSs since the only data we have available says so) by 100 and divide it by 120, meaning it'll display a value which is 20% low. I don't know a lot of people who would be running a 100V system so this case won't come up very often. And besides that, who would program a system to send out a value which is off by 20% from the actual measured value?
- Case 1 sets input_voltage_scaled to 120, on line 1267 we then multiply the measured value by 120 and divide it by 120, so we get the measured value without any changes.
- Case 2 sets input_voltage_scaled to 230, on line 1267 we then multiply the measured value by 230 and divide it by 120, so the value will end up roughly 2x too high.
- Case 2 sets input_voltage_scaled to 230 while it's a 208V system, on line 1267 we then multiply the measured value by 230 and divide it by 120, so the value will end up roughly 2x too high as well.

The OMNI's might need the scaled values, but for the 3005 protocol there is no need for it. I can create a new logic to calculate the values for 3005 without the scaled values.
Something like:
```
if( tl_model == TRIPP_LITE_SMART_3005) {
    set values without input_voltage_scaled
} else if (is_smart_protocol() ) {
    use original code with input_voltage_scaled
}
```
That way only the logic for the 3005 protocol is updated.

clepple:
> And besides that, who would program a system to send out a value which is off by 20% from the actual measured value?

If you think about the input stage as a circuit to adjust the AC voltage (which varies by region) to a range suitable for the DC charging circuitry (which is fixed for a given battery size), it makes more sense. There are probably several standard input transformers, and the voltage would be measured on the secondary side. On those cheaper units, the value returned is probably directly read from the A-to-D converter.

You are correct that all evidence points to the 3005 protocol being unscaled. Protocols 0002, 0004 and 0005 are specifically scaled to 120V, but there are different mappings for the nominal voltage cases. I am fine with your proposed "if/else" construct.

mbastiaan:
I updated my second branch to have all decodes in separate functions:
https://github.com/mbastiaan/nut/blob/tripplite_3005_2/drivers/tripplite_usb.c

Unfortunately I there is no usable diff, just about everything is red/green.

But all the code is based on the current 0.29 version without my previous changes. I created a new function to convert a string of hex values to ascii as I didn't know how to modify the function already there to get it to work. The code for this function is copied from the hexascdump function.

I choose to separate all the debug's into dedicated pieces of code, so there is no If model X then Debug A, B and C. It's all If model X then debug A. If model X then debug B. It creates a bit of unnecessary code but I think it'll be easier to maintain that way if we find out a certain protocol also needs a certain debug.

The same inside the decode functions, I've split the code into sections per value that is extracted and then per protocol. Most of the code was already like this but I cleaned it up a bit and separated it even further.

@clepple If you agree with this I can make small commits into the tripplite_3005 branch and than we  can discus each commit separately.

clepple:
@mbastiaan looks good overall. I don't think it needs to be split up into many commits - the note in the developer guide about splitting was more about changes that affect several areas of NUT (drivers, upsd, upsmon, etc) that might not be able to be merged all at once. I think we had discussed the first line of the commit messages, though - basically, I want to be able to see a quick summary of what changed when I run e.g. `git log --pretty=oneline -- drivers/tripplite_usb.c`, with more detail in the body of the commit message. No real preference if you want to squash the `driver.list.in` update with the driver code changes, or keep it separate.

Also, I think it would be useful to mention which features have been tested. I'm looking at `control_outlet()` and something looks slightly off for the binary handling: https://github.com/mbastiaan/nut/blob/tripplite_3005_2/drivers/tripplite_usb.c#L1210 (For certain combinations, `strlen()` is going to return a short value, and I think the `N` command takes two bytes). That code predates your changes, I think, so let's merge the code 3005 updates first, and then circle back to the other features.

Sorry for the delay in getting back to you - I was looking for changes in the `gihub/pull/584` reference, and I see now that you mentioned a second branch.

jimklimov:
Was this work done by/for/on-behalf of a vendor (e.g. Tripplite)? If so, please update the docs/acknowledgements.txt file as well.

drivers/tripplite_usb.c
```diff
@@ -467,6 +467,10 @@ void decode_v(const unsigned char *value)
 			  input_voltage_scaled  = 230;
 			  break;
 
+		case 6:   input_voltage_nominal =
```
jimklimov, line 470: Is this supposed to be a numeric `6` or a character `'6'`? Nearby precedent leaves me wondering :)

clepple: From earlier in the thread:
`ups.debug.V: 06 00 0c 01 58 58 0d '....XX.'`

so numeric `6` is likely correct (first byte of `V` response).

jimklimov:
Cheers, trying to wrap my head around this PR as it stands now: the historic discussion is about large work and decode routines, etc. - but the diff shown by Github is rather small and trivial to add support for a device (macro) and its voltage.

Is the big scary rest of solution offloaded to some other PR, and/or merged earlier in the past couple of years?

clepple:
> Cheers, trying to wrap my head around this PR as it stands now: the historic discussion is about large work and decode routines, etc. - but the diff shown by Github is rather small and trivial to add support for a device (macro) and its voltage.
> 
> Is the big scary rest of solution offloaded to some other PR, and/or merged earlier in the past couple of years?

My understanding is that the other changes are in https://github.com/mbastiaan/nut/blob/tripplite_3005_2/drivers/tripplite_usb.c and might not be in a PR yet.

jimklimov:
Also CCing @dzomaya for the Tripplite-related changes

drivers/tripplite_usb.c
```diff
@@ -507,6 +507,10 @@ static void decode_v(const unsigned char *value)
 			  input_voltage_scaled  = 230;
 			  break;
 
+		case 6: input_voltage_nominal =
```
jimklimov, line 510: Looking at context around, should this be `6` or `'6'`?

clepple: I think the majority of the Protocol 3005 responses are binary not ASCII, so `6` seems correct, and matches @mbastiaan 's debug log.

jimklimov:
At least this change seems small and not invasive so ok to merge, and the larger changes/refactorings discussed above can belong in another PR.

PR #592, opened 2018-08-14
Merge remote-tracking branch 'opensource/master' into DMF as of 2018-08-14 and 2018-09-03

jimklimov:


jimklimov:
Note: the failure is in test-case for missing pkgconfig, without which libneon can not be fount (while required for DMF).

m4/nut_check_libneon.m4
```diff
@@ -17,6 +17,7 @@ if test -z "${nut_have_neon_seen}"; then
 	AC_MSG_CHECKING(for libneon version via pkg-config (0.25.0 minimum required))
 	NEON_VERSION="`pkg-config --silence-errors --modversion neon 2>/dev/null`"
 	if test "$?" != "0" -o -z "${NEON_VERSION}"; then
+		AC_MSG_WARN([could not get pkg-config information for libneon version, using fallback defaults])
```
zykh: @jimklimov, if possible, don't print anything (non fatal) after an `AC_MSG_CHECKING()`, before its `AC_MSG_RESULT()` has been printed (like here, or below), or the result will end up in a new line with the additional message in-between (which is not exactly the clearest thing to read)

jimklimov:
Tested the pkgconfig/libneon fallback in Travis:

````
checking for usb_init... yes
checking for usb_detach_kernel_driver_np... yes
checking for libneon version via pkg-config (0.25.0 minimum required)... none found
configure: WARNING: could not get pkg-config information for libneon version, using fallback defaults
checking for libneon cflags... -I/usr/include/neon
configure: WARNING: could not get pkg-config information for libneon cflags, using fallback defaults
checking for libneon ldflags... -lneon
configure: WARNING: could not get pkg-config information for libneon libs, using fallback defaults
checking for ne_xmlreq.h... yes
````

The build passed overall for the DMF part which requires LibNEON.

jimklimov:
@zykh, Thanks for tge comment, hope I handled warnings better now. Note that equivalent PR for master branch is #596 - the DMF one just provides a good test case.

Issue #593, opened 2018-08-14
TODO: Fallback configure.ac detection of libneon without pkgconfig


jimklimov:
In travis use-case for `NO_PKG_CONFIG=true` avoid this:

````
checking for libneon version via pkg-config (0.25.0 minimum required)... none found
checking for libneon cflags... 
checking for libneon ldflags... 
...

libtool: compile:  /usr/lib/ccache/gcc -std=gnu99 -DHAVE_CONFIG_H -I. -I../../../tools/nut-scanner -I../../include -I/home/travis/build/networkupstools/nut/tmp/include -I../../../clients -I../../../include -I../../../drivers -DWITH_DMFMIB=0 -DWITH_DMF_LUA=0 -DNETSNMP_ENABLE_IPV6 -fno-strict-aliasing -g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -DNETSNMP_USE_INLINE -Ulinux -Dlinux=linux -D_REENTRANT -D_GNU_SOURCE -DDEBIAN -fstack-protector -fno-strict-aliasing -pipe -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -I/usr/lib/perl/5.18/CORE -D_FORTIFY_SOURCE=2 -I/usr/include -I/home/travis/build/networkupstools/nut/tmp/include -Wall -Wsign-compare -MT libnutscan_la-scan_snmp.lo -MD -MP -MF .deps/libnutscan_la-scan_snmp.Tpo -c ../../../tools/nut-scanner/scan_snmp.c  -fPIC -DPIC -o .libs/libnutscan_la-scan_snmp.o
In file included from ../../../include/dmf.h:34:0,
                 from ../../../tools/nut-scanner/scan_snmp.c:89:
../../../include/dmfcore.h:36:2: error: #error "LibNEON is required to build DMF"
 #error "LibNEON is required to build DMF"
  ^
../../../include/dmfcore.h:80:2: error: unknown type name 'ne_xml_startelm_cb'
  ne_xml_startelm_cb *xml_dict_start_cb;
  ^
../../../include/dmfcore.h:81:2: error: unknown type name 'ne_xml_cdata_cb'
  ne_xml_cdata_cb *xml_cdata_cb;
  ^
../../../include/dmfcore.h:82:2: error: unknown type name 'ne_xml_endelm_cb'
  ne_xml_endelm_cb *xml_end_cb;
[...]

PR #596, opened 2018-08-20
Use a fallback setup for libneon if pkg-config failed

jimklimov:
This should take care of issue #593


jimklimov:
See #592 for testing of this feature.

jimklimov:
I agree it would be nice to follow a consistent behavior pattern here, but it would belong in a different PR at least, and we need the pattern to replicate into other M4 files (e.g. this one if it has no complaints now) :)

I'll log an issue to that effect.

PR #610, opened 2018-09-07
Nut scanner deps

jimklimov:
Avoid rebuilding nut-scanner upon every invokation of `make` or after `make deps`.

### DIFF
--- a/tools/Makefile.am
+++ b/tools/Makefile.am
@@ -1,9 +1,3 @@
-# TODO: remove redundancies!
-
-# XXX this does not work with Automake!!!
-#
-# In fact the very concept is entirely antithetical to Automake.
-#
 # SUBDIRS are explicitly a listing of all the directories that make
 # must recurse into BEFORE processing the current directory.
 #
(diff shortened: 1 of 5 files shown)


jimklimov:
Seems there are many callers of this script and/or make target, and in many different contexts (`autogen`, `make all`, `make nut-scanner`, `make dist`, `make distcheck` ...) fixing one breaks another :\

It's complicated (C)

tools/Makefile.am
```diff
@@ -19,55 +19,85 @@ SUBDIRS = . nut-scanner
...
+
+# BSD devd output file
+GENERATED_USB_FILES += ../scripts/devd/nut-usb.conf.in
+
+# UPower output file
+GENERATED_USB_FILES += ../scripts/upower/95-upower-hid.rules
+
+all: nut-scanner-deps $(GENERATED_USB_FILES)
 
 # XXX these rules are all bogus!  They cause un-named target files to
```
clepple: it looks like this comment is out-of-date, at least for this particular rule.

clepple: The way this is rendered is a bit misleading - I meant that "XXX these rules are all bogus!" was out-of-date with respect to the rule below (which does have specific dependencies), not the green portion of the previous diff.

jimklimov: Thanks for pointing out that I never got to it, I hoped to but forgot ;)

PR #612, opened 2018-09-07
nut-driver-enumerator.sh.in updates

jimklimov:
Alleviate the case of empty list (in config or system) seeming like a one-liner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,8 +72,43 @@ env:
 # runs only a few sub-jobs in parallel, so we want the withdoc and
 # perhaps spellcheck jobs to start first, and while they are still in
 # progress, others are spawned and finished - reducing overall job TTL.
+# Note that the nut-driver-enumerator tests should be tried in at least
+# the shell interpreters reasonable for default setups of Solaris/illumos
+# (ksh) and Linux (bash, dash, etc.) common distros.
 matrix:
   include:
+  - env: BUILD_TYPE=nut-driver-enumerator-test SHELL_PROGS=bash
+    os: linux
+    services:
+        - docker
+    addons:
+      apt:
+        packages:
+        - bash
+  - env: BUILD_TYPE=nut-driver-enumerator-test SHELL_PROGS=ksh
+    os: linux
+    services:
+        - docker
+    addons:
+      apt:
+        packages:
+        - ksh
+  - env: BUILD_TYPE=nut-driver-enumerator-test SHELL_PROGS=dash
+    os: linux
+    services:
+        - docker
+    addons:
+      apt:
+        packages:
+        - dash
+  - env: BUILD_TYPE=nut-driver-enumerator-test SHELL_PROGS=ash
+    os: linux
+    services:
+        - docker
+    addons:
+      apt:
+        packages:
+        - ash
   - env: BUILD_TYPE=default-withdoc
     os: linux
     addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,36 @@ env:
 # (ksh) and Linux (bash, dash, etc.) common distros.
 matrix:
   include:
+  - env: BUILD_TYPE=default-withdoc
+    os: linux
+    addons:
+      apt:
+        packages: &deps_gendocs
+        - asciidoc
+        - xsltproc
+        - dblatex
+        - docbook-xsl
+        - docbook-xsl-ns
+        - source-highlight
+        - libxml2-utils
+  - env: BUILD_TYPE=default-spellcheck
+    os: linux
+    addons:
+      apt:
+        packages: &deps_aspell
+        - aspell
+        - aspell-en
+  - env: BUILD_TYPE=default-tgt:distcheck-valgrind
+    os: linux
+    sudo: false
+    services:
+        - docker
+    addons:
+      apt:
+        packages:
+        - *deps_driverlibs
+        - valgrind
+# Other quicker builds in standard env follow:
   - env: BUILD_TYPE=nut-driver-enumerator-test SHELL_PROGS=bash
     os: linux
     services:
@@ -109,36 +139,6 @@ matrix:
       apt:
         packages:
         - ash
-  - env: BUILD_TYPE=default-withdoc
-    os: linux
-    addons:
-      apt:
-        packages: &deps_gendocs
-        - asciidoc
-        - xsltproc
-        - dblatex
-        - docbook-xsl
-        - docbook-xsl-ns
-        - source-highlight
-        - libxml2-utils
-  - env: BUILD_TYPE=default-spellcheck
-    os: linux
-    addons:
-      apt:
-        packages: &deps_aspell
-        - aspell
-        - aspell-en
-  - env: BUILD_TYPE=default-tgt:distcheck-valgrind
-    os: linux
-    sudo: false
-    services:
-        - docker
-    addons:
-      apt:
-        packages:
-        - *deps_driverlibs
-        - valgrind
-# Other quicker builds in standard env follow:
   - env: BUILD_TYPE=default-nodoc
     os: linux
     addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,15 +57,17 @@ addons:
     - libxml2-utils
     - libmodbus-dev
 
-# Define one job in the matrix, to avoid a failing job
-# with "no environment variables set"; but others are
-# listed below to prioritize the longer jobs to start first.
+# Define at least two jobs in the matrix, to avoid a failing job
+# with "no environment variables set" and have them built (one
+# is too few); but others are listed below to prioritize the
+# longer jobs to start first as faster ones complete in parallel.
 env:
   global:
     - CI_TIME=true
     - CI_TRACE=false
   matrix:
     - BUILD_TYPE=default
+    - BUILD_TYPE=default-tgt:distcheck-light
 
 # Builds with customized setups
 # Note that doc-related builds take the longest, and Travis CI cloud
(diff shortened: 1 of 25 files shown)


zykh:
@jimklimov, since you're already touching this...

nut-driver-enumerator.sh.in:
- upsconf_getSection_content(): what if a section (`$1`) *is* named "other"?
  Also, there should not be two sections with the same name... and while reading we've *not* already displayed anything...
  In my opinion, something like the following should be enough:
  ```sh
  upsconf_getSection_content() {
  
  ### ...
  
      CURR_SECTION=""
      SECTION_CONTENT=""
      RES=1
      [ -n "$1" ] || RES=0
      while read LINE ; do
          case "$LINE" in
              \[*\])
                  if [ "$RES" = 0 ]; then
                      # We have already processed a section, here is a new one,
                      # and this routine only process one (TODO: toggle?)
                      break
                  fi
                  if [ "$LINE" = "[$1]" ]; then
                      SECTION_CONTENT="$LINE"
                      CURR_SECTION="$1"
                      RES=0
                  fi
                  ;;
              "") ;;
              *)  if [ "$CURR_SECTION" = "$1" ]; then
                      if [ -n "$SECTION_CONTENT" ]; then
                          SECTION_CONTENT="$SECTION_CONTENT
  $LINE"
                      else
                          SECTION_CONTENT="$LINE"
                      fi
                  fi
                  ;;
          esac
      done
  
  ### ...
  
  }
  
  ```
- should we also support flags (i.e. no "= value")?
- shouldn't we do something (e.g. reload the drivers) also whenever any global key changes?
- upsconf_getDriverMedia(): what about `nutdrv_qx` when used with USB devices?
  (hint: in that case, at the very least, `port` should be set to `auto`, and, possibly, other USB-only options could be set, e.g. `vendorid`, `productid`, ...)

upsdrvsvcctl.in:
- when parsing options, debug (unused) should be `-D`, not `-d`

Cosmetics (so feel free to ignore them):
- upsdrvsvcctl.in's  usage(): in my opinion, something like the following would be more clear (help message):
  ```
  usage: $0 [OPTIONS] COMMAND [<ups>]

  OPTIONS:
    -h                  display this help
    -t                  testing mode - prints actions without doing them
    -D                  raise debugging level

  COMMAND:
    start               start all UPS drivers in ups.conf
    start <ups>         only start driver for UPS <ups>
    stop                stop all UPS drivers in ups.conf
    stop <ups>          only stop driver for UPS <ups>
    shutdown            shutdown all UPS drivers in ups.conf
                        Note: the "shutdown" options from original upsdrvctl are
                        not currently directly supported by this service management
                        framework wrapper; instead they are passed to the native
                        upsdrvctl binary (your current user account should have
                        sufficient permissions to do that all).
    shutdown <ups>      only shutdown UPS <ups>
    resync              call $ENUMERATOR
                        to update the mapping of service instances for
                        NUT drivers to device sections in 'ups.conf'
    reconfigure         call $ENUMERATOR
                        to remove and re-create the mapping of all service
                        instances for NUT drivers to device sections in
                        'ups.conf' e.g. after a NUT package upgrade
    list                call $ENUMERATOR
                        to list the mapping of service instances to device sections
    list <ups>          (optionally return the service instance name for one device)
    show-config <ups>   output config section from ups.conf for device <ups>
    show-config         ...or all devices if no <ups> argument was passed
  ```
- even more cosmetic: there are some discrepancies between the actually parsed options and the documented ones (both in manpages and in `-h` help) of both upsdrvsvcctl and nut-driver-enumerator.sh... still a work in progress?

jimklimov:
Seems my reply a few days back did not get posted :(

So I still have yet to get back to processing the suggestions. Regarding the discrepancy of documented and implemented options, I guess it is an artefact of user-acceptance testing vs. myself - something shorter to type vs. something worded clearer to commit to as a stable CLI API :)

The section parser is what it is due to (currently not included) work on parsing all the sections into shell variables to cache ("bash" is not granted, so no arrays). This is done in another branch, but did not prove to be the faster option for common cases in my early testing, so was shelved so far. Still the comment about `other` is valid, I should look for a token that would not happen as a realistic section name... something with brackets maybe...

Global keys make sense, especially as the script now supports the global config as a fake empty-named section. There is still a question then where to record the checksum though :)

As for flags, I'm not sure. Initially at least, to allow for faster detection of media type, only the driver and port values were deemed valuable (QX may extend the list though, or cause special handling). For checksums, all assignments in the section are used... I'll review if there can be flags in that context, I thought there couldn't (may be relevant for the global "section" though)...

Thanks for the apparently thorough review :)

jimklimov:
@zykh : I've hopefully addressed most of the suggestions (and at least one bug) in commits added just above. This was quite tested to still work as expected under Linux systemd, though not yet with Solaris SMF (as far as storing and detecting global section changes goes).

jimklimov:
UPDATE: a little fix later, this also tested OK in Solarissh systems (OpenIndiana).

jimklimov:
Updated vs current master branch state.

aquette:
I've just reviewed the PR, though not tested. LGTM.
@zykh anything missing before merging in your opinion?

zykh:
@jimklimov, thanks for addressing those issues.
@aquette, a couple more things, actually:
- if I'm not mistaken, we have a problem with end-of-line comments: in eb6ea1f the last `egrep` call in `upslist_normalizeFile()` filters out any flag with spaces/tabs still present at that point in the line (e.g. `blabla # I'm a comment`), plus, end-of-line comments end up being considered part of the value (`port = /dev/ttyS1 # hmm... it was /dev/ttyS0`), and 'invalidate' sections (`[boing] # the boing ups`)... so, perhaps, we should simply remove/ignore any comment, and not just comment-only lines (e.g. adding something like `'s,\([^\]\)#.*$,\1,'` as the first `sed` script)...
- 1a8df91: maybe, we should try matching also the other things the driver considers (currently: https://github.com/networkupstools/nut/blob/master/drivers/nutdrv_qx.c#L1817-L1825)?

jimklimov:
@zykh : I too had some concerns about comments, in particular it does not seem invalid to have a hash character inside a string: `description = "UPS in rack #2"` which would not be a comment. So short of writing another JSON.sh (which is an option) I went the safer way of just considering those lines different.

The point about section names does remain however, good catch.

Also, if you fancy experimenting some strings, take a look at the scripts added to `tests/` to spell out the weird section

jimklimov:
Updated code and tests for a section line with comments beyond its brackets.

jimklimov:
Updated code and tests for a "commented flag" line with one token (no assignment), followed by an optional whitespace and some comments.

jimklimov:
I think for people trying to shoot themselves in the foot, we should provide a bed in the psych ward, not a bandaid :-)

Still, if those examples are valid (even the proper processing of e.g. backslash-escaped hash chars), it complicates things... Not sure about brackets with whitespace inside - technically filenames are allowed to have that, so why not. Non-whitespace before brackets (perhaps assignment of something with a bracket in value) may indeed be an issue, ok I'll not strip it :)

I think solving those niche nuances would really belong in a different PR, with two primary vectors OTOH being either indeed porting a shell parser line JSON.sh to this job, or maybe even better for consistent results WRT binary scanner - making a small binary consumer of the NUT parser code (or even an option in `upsd` since its presence on the system is essential for this script anyway) to normalize a config file (make all meaningful lines single-line with `\n` if needed, strip unneeded whitespace and comments) and relying on its output in the script to do its analysis. Effectively, the call to chain of sed's and grep's in `upslist_normalizeFile()` to get `UPSCONF_DATA="..."` contents would be replaced by a call to this tool, and it would be more guaranteed to give same (sane) results as relevant for other NUT consumers of this config.

I'd suggest to get the benefits of this PR quickly and then working on a better iteration as time permits :)

jimklimov:
Curious, I tested regarding your comments about brackets by adding a couple of lines to a test config section:
````
    desc="value with [brackets]"
    [brackets with spaces are not sections] # but rather an invalid mess as binary parser may think
```

and the scripted parser indeed "found" two more sections:
````
+[brackets]
+[brackets with spaces are not sections]
````
and even more as it iterated the devices to guess their media:
````
+INST: c647a4a59653d8e296090433da7e26cf~[brackets]: DRV='' PORT='' MEDIA='' SECTIONMD5='54f95311cf2ee3c7553ab416b41efb87'
+INST: c647a4a59653d8e296090433da7e26cf~[brackets]: DRV='' PORT='' MEDIA='' SECTIONMD5='54f95311cf2ee3c7553ab416b41efb87'
+INST: 81ab62a717e411ce854fb278c7776989~[with]: DRV='' PORT='' MEDIA='' SECTIONMD5='68b329da9893e34099c7d8ad5cb9c940'
+INST: c1f10cfd640618484a2a475c11410fd3~[spaces]: DRV='' PORT='' MEDIA='' SECTIONMD5='68b329da9893e34099c7d8ad5cb9c940'
+INST: eacbe4d1b2dee530eee7460477877c4d~[are]: DRV='' PORT='' MEDIA='' SECTIONMD5='68b329da9893e34099c7d8ad5cb9c940'
+INST: 61aa5b6c78fa4e3636069347ae39df10~[not]: DRV='' PORT='' MEDIA='' SECTIONMD5='68b329da9893e34099c7d8ad5cb9c940'
+INST: b4aef0e33c28557063958cd0883b21e9~[sections]: DRV='' PORT='' MEDIA='' SECTIONMD5='68b329da9893e34099c7d8ad5cb9c940'
````
(since it does a `for` and so is sensitive to whitespace). So I'll poke that a bit more.

zykh:
The sad thing is that they usually shoot anyone else in the room, before their feet...
Ok, ok, I'm done with my stupid things...

Let's merge this right away (I can't test this now, but the code looks good), we can always come back later to those small issues (I like your `upsd` idea...).

jimklimov:
So, merging a bunch of bug fixes as permitted by peer review :)

PR #611, opened 2018-09-07
libusb tracing debug text

jimklimov:
Help figure out faster which codepath to debug :)

PR #613, opened 2018-09-07
Libusb 1.0+0.1 debug

jimklimov:
See also #611 for similar change proposed in current master.


jimklimov:
Merging as approved by @aquette

PR #615, opened 2018-09-08
Libusb compat 1.0 debug

jimklimov:
Equivalent to #611 and #613 as much as that applies


jimklimov:
Sort of : as I was tracing what went wrong in my builds, I was not fully convinced my Make et al took the right path. So to be on the safer side, the enhanced debugs were welcome  to speed up the code-digging process, and I guessed others can be in similar situation sometimes.

jimklimov:
Merging as approved by @aquette

zykh:
Thanks @jimklimov, and sorry for the huge delay... (I wouldn't have merged #611, in order to avoid possible conflicts with the libusb branches, though... well, nevermind...)

PR #632, opened 2018-10-23
Added UPS load, battery charge, input voltage/frequency, and output 

CashAMatic:
voltage, MIB mappings to cyberpower-mib.c

Tested with an RM202 card.


clepple:
Excellent, thanks!

PR #638, opened 2018-12-01
Add support for Hunnox HNX-850

marianojan:
This PR adds support for the Hunnox HNX-850 UPS.

It's a special case for the `fabula` subdriver, with a required lockstep for the commands. For this UPS, if the commands are not sent in a specific order, the output status remains unchanged.

To make it work, my `ups.conf` file, has the following configuration:
```
[u1]
       driver = "nutdrv_qx"
       port = "auto"
       vendorid = "0001"
       productid = "0000"
       protocol = hunnox
       langid_fix = 0x0409
       novendor
       noscanlangid
```

Also I had to add a new flag, `noscanlangid` because this scan hangs the USB endpoint of the UPS.

I hope it can help someone else who has this brand of UPS.


zykh:
Thanks for the patch, @marianojan.
I'm not lucid enough and I've not enough time now for a complete review, but I'll come back to this in a few days...

So just one thing for now... basically the working sequence is:
- initialization (only once):
  - get string at index 0 twice,
  - then string at index 1,
  - then sleep for a while,
- update sequence (loop):
  - get string at index 0xd (`F` query),
  - get string at index 3 (`Q1` query),
  - get string at index 0xc (`I` query),

right?

marianojan:
Hello @zykh!

You are right, the working sequence is as you expressed it. 
The initialization steps are required only once, as far as I tested it.
The update sequence requires that before each `Q1` query, one of `F` or `I` query has to be done. I copied the steps from the Windows client, by capturing the requests with Wireshark.
Both the `I` and `F` queries return null data. (The windows client somehow receives a versión number, but when I tested in linux, I always got an empty string)

I don't have a great knowledge about the project, so if something is misplaced or has to be done in another way, please tell me.

Thank you for taking a look at it!

drivers/nutdrv_qx.c
```diff
@@ -790,7 +841,7 @@ static int	krauler_command(const char *cmd, char *buf, size_t buflen)
 }
 
 /* Fabula communication subdriver */
-static int	fabula_command(const char *cmd, char *buf, size_t buflen)
+static int	_fabula_command(const char *cmd, char *buf, size_t buflen, char hunnox_patch)
```
panoti: I think it is not worth adding a `hunnox_patch` parameter which is only for Hunox 850. A function should be use for other protocols. Do you have any other options?

drivers/Makefile.am
```diff
@@ -283,7 +284,7 @@ dist_noinst_HEADERS = apc-mib.h apc-hid.h baytech-mib.h bcmxcp.h	\
  nutdrv_qx_voltronic.h nutdrv_qx_voltronic-qs.h nutdrv_qx_voltronic-qs-hex.h nutdrv_qx_zinto.h \
  xppc-mib.h huawei-mib.h eaton-ats16-mib.h apc-ats-mib.h raritan-px2-mib.h eaton-ats30-mib.h \
  apc-pdu-mib.h eaton-pdu-genesis2-mib.h eaton-pdu-marlin-mib.h \
- eaton-pdu-pulizzi-mib.h eaton-pdu-revelation-mib.h emerson-avocent-pdu-mib.h
+ eaton-pdu-pulizzi-mib.h eaton-pdu-revelation-mib.h emerson-avocent-pdu-mib.h nutdrv_qx_hunnox.h
```
panoti: We will have a conflict at here.

drivers/nutdrv_qx.c
```diff
@@ -2017,7 +2132,7 @@ void	upsdrv_initups(void)
 		dstate_setinfo("ups.productid", "%04x", usbdevice.ProductID);
 
 		/* Check for language ID workaround (#2) */
-		if (langid_fix != -1) {
+		if ((langid_fix != -1) && (!getval("noscanlangid"))) {
```
panoti: You add AND comparison operator which may cause error for other protocols. Are you sure about that?

drivers/nutdrv_qx.c
```diff
@@ -859,14 +910,59 @@ static int	fabula_command(const char *cmd, char *buf, size_t buflen)
...
+		// Seems that if we inform a large buffer, the USB locks.
+		// This value was captured from the Windows "official" client.
+		if (buflen > 102) {
+			buflen = 102;
+		}
+	}
+
 	/* Send command/Read reply */
-	ret = usb_get_string_simple(udev, index, buf, buflen);
+	if (langid_fix != -1) {
```
panoti: It is important to determine whether additional conditions here do not affect other sub-drivers or this protocol.

drivers/nutdrv_qx.c
marianojan: I thought that duplicating the entire function was a bad idea. But maybe it's the safest way to avoid problems for other UPSs? What do you recommend?

knoid:
Thanks to this PR I was able to get data from my newly bought UPS!

I'm not sure which model it really is but it's sold as "Ups Energit Iron Guardian 800 Plcd Dispay Usb Monofasica" (with spelling errors and all) and comes with software branded as "UPSmart v1.5" for redhat and I 1) don't own a CD-R and 2) use Ubuntu.

microfrost1:
This worked perfectly for my "Powercool Smart UPS 1200VA".

drivers/nutdrv_qx_hunnox.c
```diff
@@ -0,0 +1,137 @@
+/* nutdrv_qx_zinto.c - Subdriver for Zinto protocol based UPSes
+ *
+ * Copyright (C)
+ *   2013 Daniele Pezzini <hyouko@gmail.com>
+ *
```
jimklimov, line 5: @marianojan : I think your copyright belongs here ;)

drivers/nutdrv_qx_hunnox.h
```diff
@@ -0,0 +1,29 @@
+/* nutdrv_qx_zinto.h - Subdriver for Zinto protocol based UPSes
+ *
+ * Copyright (C)
+ *   2013 Daniele Pezzini <hyouko@gmail.com>
+ *
```
jimklimov, line 5: @marianojan : I think your copyright belongs here ;)

jimklimov:
Was this work done by/for/on-behalf of a vendor (e.g. Hunnox)? If so, please update the docs/acknowledgements.txt file as well.

drivers/nutdrv_qx_hunnox.h
```diff
@@ -0,0 +1,29 @@
+/* nutdrv_qx_zinto.h - Subdriver for Zinto protocol based UPSes
```
jimklimov, line 1: @marianojan : if this was adapted from a subdriver for another UPS model, worth mentioning taht but the title here should point to Hunnox ;)

drivers/nutdrv_qx_hunnox.c
```diff
@@ -0,0 +1,137 @@
+/* nutdrv_qx_zinto.c - Subdriver for Zinto protocol based UPSes
```
jimklimov, line 1: @marianojan : if this was adapted from a subdriver for another UPS model, worth mentioning taht but the title here should point to Hunnox ;)

jimklimov (changes requested):

LukePrior:
Is this going to get merged?

drivers/nutdrv_qx.c
jimklimov: Seems to be okay: "if the user did not specially configure `noscanlangid` option, then..." - for majority of cases this is going to be true, unless users specially did something.

jimklimov (approved):

drivers/nutdrv_qx.c
jimklimov: Constrained to `&& hunnox_patch`

jimklimov:
Note for posterity: Compared to the original PR, and following concerns in comments, I added a constraint at https://github.com/networkupstools/nut/pull/638/files/3a5a758dbbfbb8bdf22a45cdde06fe10d4e0249b#diff-88ad31354671c9f25256baa8adf221a5fd2cf476R928 to limit the behavior change to `&& hunnox_patch` for the different way to `usb_get_string()`. 

Technically seems that this may be against the original PR intent for devices which are "krauler" (so have a "langid_fix", at least as documented) but are not "hunnox", if these sets even overlap in hardware and in codepaths.

drivers/nutdrv_qx.c
jimklimov: My stance on this wavered over time, but in the end I guess making an explicit "hunnox_command" that happens to repeat some of the other code can be more readable and maintainable, and would clearly "protect" support of different devices from these and later changes in not-their subdriver.

It may be or not be possible and practical that shared code can be de-duplicated into new routines called from various places, as a separate effort.

drivers/nutdrv_qx.c
jimklimov: UPDATE: Ended up separating original `fabula_command()` from `fabula_hunnox_command()` and dropping the `hunnox_patch` value support, so restoring the originally PRed clause for this choice point.

drivers/nutdrv_qx_hunnox.c
```diff
@@ -0,0 +1,138 @@
...
+	/*
+	 * > [FW?\r]
+	 * < [#-------------   ------     VT12046Q  \r]
+	 *    012345678901234567890123456789012345678
+	 *    0         1         2         3
+	 */
+
+	{ "device.mfr",			0,	NULL,	"FW?\r",	"",	39,	'#',	"",	1,	15,	"%s",	QX_FLAG_STATIC | QX_FLAG_TRIM,	NULL,	NULL,	NULL },
+	{ "device.model",		0,	NULL,	"FW?\r",	"",	39,	'#',	"",	17,	26,	"%s",	QX_FLAG_STATIC | QX_FLAG_TRIM,	NULL,	NULL,	NULL },
+	{ "ups.firmware",		0,	NULL,	"FW?\r",	"",	39,	'#',	"",	28,	37,	"%s",	QX_FLAG_STATIC | QX_FLAG_TRIM,	NULL,	NULL,	NULL },
```
jimklimov, line 78: @aquette: going over recent merges, and given your earlier comments about mfr/model/serial in https://github.com/networkupstools/nut/pull/807#discussion_r501496383 - should these fields here be removed also (does something else handle them by default)? And/or apply some same namespace (ups vs device) here?

JulyIghor:
Anyone knows what does mean '*' which sometimes comes instead of '('?
image

JulyIghor:
Sometimes it won't start
```
  0.118743	[D3] nut_usb_set_altinterface: skipped usb_set_altinterface(udev, 0)
   0.120429	[D3] send: Q1
   0.124323	[D3] read: Input/output error (-5)
   0.124355	[D2] qx_process_answer: short reply (input.voltage)
   0.124377	[D2] Skipping protocol Q1 0.07
   0.124385	Device not supported!
   0.124395	Device not supported!
```
And works again only after re-attaching usb cable

PR #639, opened 2018-12-04
Problem: nut-driver-enumerator can miss changes of ups.conf happening WHILE it runs 

jimklimov:
Apply https://github.com/42ity/nut/pull/69 to upstream/master

Minimize the chance that this happens by comparing the config file early in the startup of the script and just before it is done reconfiguring things. This way, while a window for missing a change still remains (and would get picked up by a re-run e.g. with daemonized processing or subsequent editions of the config file), it would be sub-second rather than a much longer configuration period we face now.

The technical issue with systemd codepath at least here seems to be that if the `ups.conf` file is edited for the first time, this tickles the `nut-driver-enumerator.path` unit whose purpose is to "start" the `nut-driver-enumerator.service` unit. That "start" activity takes some time, parsing the configs and determining changes and applying those to reconfigure services and start the drivers and connect to physical devices... and if during this timeframe the file is edited again, the `path` unit tries to "start" the `service` unit, and since that one is already running/not-exited, there is nothing to do (so it just logs "Started..." in the journal), and the second set of editions stays effectively unnoticed as far as service re-configurations are concerned.

This version should be more robust (than the originally integrated iteration) against editions of `ups.conf` coming in sooner than processing of previous change has completed, by keeping track of whether the file has changed during processing.

For single-run mode, this still leaves a miniscule window of opportunity for the config file to change and end up unprocessed between the near-end of script where this comparison is made, and the final change of systemd unit status change to not-active (so the state where systemd path-unit trigger would try to activate an already-active unit and so do nothing), so for systems under load or intensively changing the NUT configs, the daemonized mode may be preferable. Maybe the triggering unit can be made smarter later on, w.g. to wait until its victim is not active, or to just wait some arbitrary timeout :\

CC @bigon as the resident systemd expert :)


jimklimov:
Note: further work about daemonized service is now being triaged with https://github.com/42ity/nut/pull/70 and will be posted to NUT after some more field testing.

PR #642, opened 2018-12-08
Add FD_CLOEXEC flag to config files, pipes and sockets

sshambar:
File descriptors are leaking to processes spawned by upsmod and
upssched, leading to SELinux errors when (for example) sendmail
attempts to read from fd #4.


clepple:
LGTM, thanks!

PR #644, opened 2018-12-14
Readme broken url

prasket:
Noticed a broken link in README when I was trying to research NUT for my use case so I thought I would fix it. I also reworded just a bit as it looked odd just having [online]. at the end after the hyperlink.

### DIFF
--- a/README
+++ b/README
@@ -171,7 +171,7 @@ For example, it is available on Debian systems as:
 
 	/usr/share/nut/driver.list
 
-This table is also available link:http://www.networkupstools.org/stable-hcl.html[online].
+This table is also available link:http://www.networkupstools.org/stable-hcl.html [online].
 
 
 If your driver has vanished, see the link:FAQ.html[FAQ] and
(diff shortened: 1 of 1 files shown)


clepple:
Hmm, I don't remember Github rendering those links as HTML before.

The problem is that the README is also parsed as AsciiDoc, and is included as part of the user manual:

https://networkupstools.org/docs/user-manual.chunked/ar01s02.html#_hardware_compatibility_list

(Unlike Markdown, bare URLs are not automatically turned into hyperlinks with `asciidoc`.)

jimklimov:
I believe one way around this would be to `git mv README README.asciidoc` (Github honors the extensions in rendering), and update the Makefiles that reference it.

Maybe even generate a plaintext `README` with stripped markup elements, if needed (not sure on this one) :)

aquette:
This is a longstanding "issue"!
iirc, github honors both .asciidoc and .adoc extensions for AsciiDoc markup. I'll make something beside this PR (that won't be merged, but keeping it open as a reminder for now)
Staging work visible here: https://github.com/aquette/nut/tree/gh_adoc

aquette:
closing in favor of #669

PR #652, opened 2019-01-28
Add support for extra parameter for instant commands

aquette:
Instant commands were almost ready for supporting additional parameter.
However, it was not documented, and libnutclient was not supporting it yet

Signed-off-by: Arnaud Quette <arnaud.quette@free.fr>


jimklimov:
I am not fully sure about this change as it breaks an ABI for existing clients. Should there be a version of routines with old signatures that would call the new routine with e.g. `null` new parameter?

PR #654, opened 2019-01-29
libnutclient: introduce getDevicesVariableValues()

boricj:
A performance bottleneck was identified within the 42ity project, where querying all variable values from 300 ePDUs through upsd and libnutclient took 15 seconds to complete. The new method `TcpClient::getDevicesVariableValues()` allows querying device variable values in bulk, by sending all `LIST VAR` commands ahead of time and then parsing all the results at once. Throughput is vastly improved, with the mega-query completing about 15 times faster.

### DIFF
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -461,6 +461,18 @@ std::map<std::string,std::vector<std::string> > Client::getDeviceVariableValues(
   return res;
 }
 
+std::map<std::string,std::map<std::string,std::vector<std::string> > > Client::getDevicesVariableValues(const std::set<std::string>& devs)throw(NutException)
+{
+	std::map<std::string,std::map<std::string,std::vector<std::string> > > res;
+
+	for(std::set<std::string>::const_iterator it=devs.cbegin(); it!=devs.cend(); ++it)
+	{
+		res[*it] = getDeviceVariableValues(*it);
+	}
+
+	return res;
+}
+
 bool Client::hasDeviceCommand(const std::string& dev, const std::string& name)throw(NutException)
 {
   std::set<std::string> names = getDeviceCommandNames(dev);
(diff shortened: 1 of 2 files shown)


jimklimov:
Cursorily, LGTM. But better wait for review from a better C++ specialist ;)

clients/nutclient.cpp
```diff
@@ -461,6 +461,18 @@ std::map<std::string,std::vector<std::string> > Client::getDeviceVariableValues(
   return res;
 }
 
+std::map<std::string,std::map<std::string,std::vector<std::string> > > Client::getDevicesVariableValues(const std::set<std::string>& devs)throw(NutException)
+{
+	std::map<std::string,std::map<std::string,std::vector<std::string> > > res;
+
+	for(std::set<std::string>::const_iterator it=devs.cbegin(); it!=devs.cend(); ++it)
```
jimklimov: @boricj @aquette : how much required is `cbegin`/`cend` here? Some systems, such as the buildbot on OpenBSD, use a rather old gcc which does not support these : http://buildbot.networkupstools.org/public/nut/builders/OpenBSD-x64/builds/706/steps/compile/logs/stdio

Internet lore says these can be changed into `begin`/`end` at a loss of `const`ness of the iterator. Is it a critical moment here, or can the code be amended for slightly better portability?

Alternately, perhaps test in `configure` whether the chosen C++ compiler can build a sample program with these and... something: disable C++ builds, or enable some macro to rename `cend => end` on particular systems, or ...

boricj: After taking a good look at `std::set::begin()`, changing `cbegin`/`cend` to `begin`/`end` should still work with a `const` container in C++03 as it'll return implicitly a `const_iterator`. My knowledge of C++ is heavily biased towards C++11, so I'm not surprised that one slipped by...

...but I will celebrate the day when C++11 will finally reach OpenBSD base :-)

PR #655, opened 2019-01-31
[WIP - DO NOT MERGE] Multiple polling mechanisms support for upsd

boricj:
This is a PoC that allows upsd to use different polling mechanisms for its sockets at compile-time. It provides two polling implementations:
* `poll` for UNIX (current existing mechanism in upsd) ;
* `epoll` for Linux.

`poll` has scaling issues with large file descriptor sets, so it makes sense to use something better if possible. Using the `epoll` implementation on Linux reduces the sys load of upsd a fair bit when there are ~300 sockets to monitor, winnings are expected to scale with number of devices to monitor. Other polling mechanisms can be implemented as required.

This incomplete pull request is meant for @aquette. Notably, GNU autotools magic bits are missing.


aquette:
@boricj iirc, you told me you saw an issue with this WRT drivers load (and thus probably upsd being too fast to PING or something like that). Please give a bit of visibility / traces here, thx

boricj:
When tested on a 1200 dummy devices setup, the NUT drivers started to go haywire and consumed all the CPU time they could, possibly because they were flooded with PING requests. Not much more info than that for now.

PR #660, opened 2019-02-11
 INSTCMD and SET VAR status tracking implementation - bis

zykh:
Follow-up to #659 -- https://github.com/networkupstools/nut/pull/659#issuecomment-462202474


clepple:
Same comments as on #659. I do think the `TRACKING` name is preferable.

I have only had time to test the client-side error handling - I was hoping to do more with this over the weekend, but I was rewiring the network, and went down a rabbit hole trying to get a physical UPS running for testing. I might have some time later tonight to build and test the changes on the server side.

PR #662, opened 2019-02-15
Problem: 32-bit builds fail to load 64-bit libs

jimklimov:
Solution: Do our best to guess the bitness of a build,
and use it in search paths for get_libname()

### DIFF
--- a/common/common.c
+++ b/common/common.c
@@ -33,6 +33,15 @@
 #include "nut_version.h"
 const char *UPS_VERSION = NUT_VERSION_MACRO;
 
+#include <stdio.h>
+#include <limits.h>
+
+/* Know which bitness we were built for,
+ * to adjust the search paths for get_libname() */
+#if UINTPTR_MAX == 0xffffffffffffffffULL
+# define BUILD_64   1
+#endif
+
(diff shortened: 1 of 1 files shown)


aquette:
@jimklimov in that case, why not grouping 64 declaration and put the others in a else?

jimklimov:
Some OSes we use are smart enough to check also in e.g. `/lib/64` if you ask for `/lib` and are a 64-bit process; others are not.

So there should be some preference primarily sorted by root paths (package, local, distro?..) and maybe with respect to `LD_LIBRARY_PATH(_64)` even if modern docs disregard it as a debugging bandaid that should not be seen in production.

Some systems can also have different mechanisms to configure globally preferred/trusted paths for finding libs (e.g. Solarish `crle (-64)` which I am not sure how to interact with smoothly... maybe dlopen a library basename first (or after package path pass)?

In any case, even if these improvements are reasonable, they would belong in another PR for more extensive platform testing I guess.

jimklimov:
Further, possibly better, solutions can be triaged in another PR. This one does solve a practical problem however, so I merged it.

PR #663, opened 2019-02-17
Upssvc improvements upstreaming

jimklimov:
This version should be more robust (than the originally integrated iteration) against editions of `ups.conf` coming in sooner than processing of previous change has completed, by keeping track of whether the file has changed during processing. 

For single-run mode, this still leaves a miniscule window of opportunity for the config file to change and end up unprocessed between the near-end of script where this comparison is made, and the final change of systemd unit status change to not-active (so the state where systemd path-unit trigger would try to activate an already-active unit and so do nothing), so for systems under load or intensively changing the NUT configs, the daemonized mode may be preferable. Maybe the triggering unit can be made smarter later on, w.g. to wait until its victim is not active, or to just wait some arbitrary timeout :\

CC @bigon as the resident systemd expert :)

### DIFF
--- a/docs/nut.dict
+++ b/docs/nut.dict
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 2433 utf-8
+personal_ws-1.1 en 2435 utf-8
 AAS
 ACFAIL
 ACFREQ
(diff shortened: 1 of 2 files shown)


jimklimov:
Sorry, seems I've posted (most of) this before at #639 so closing this one as duplicate

PR #668, opened 2019-02-23
Fix builds against older GCC (no C++11 in the system)

jimklimov:
Hopefully would rectify http://buildbot.networkupstools.org/public/nut/builders/OpenBSD-x64/builds

CC : @boricj

### DIFF
--- a/clients/Makefile.am
+++ b/clients/Makefile.am
@@ -1,4 +1,5 @@
 # Network UPS Tools: clients
+EXTRA_DIST =
 
 # by default, link programs in this directory with libcommon.a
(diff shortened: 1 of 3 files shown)


jimklimov:
Updated the `pull_668_cxx11` branch with the updated commit for buildbot to pick it up, but now it replies HTTP 503... :\

clepple:
> but now it replies HTTP 503... :\

One of these days I should get that proxy to start on its own :)

jimklimov:
Thanks for this and other recent reviews :)

clepple:
No problem--sorry I'm terribly behind on this. New batteries are coming for my MGE UPS, though, so the network is finally getting put back together.

PR #669, opened 2019-02-25
[DO NOT MERGE YET] Try to reconcile README, Asciidoc and GitHub

aquette:
Signed-off-by: Arnaud Quette <arnaud.quette@free.fr>


aquette:
submitting as a PR to ensure I'm not missing anything. I've simply git mv the README to .adoc, to ensure GH interpretation, and symlinked it to the standard README for compatibility, including for automake and asciidoc inclusion

aquette:
found back what I was missing: #226 / https://github.com/zykh/nut/commits/rename_readme
it seems that @zykh made a more complex / better try at this. I don't have time to go deeper into for now, but definitely something we should finally address, since these are the topics that are back again and again

jimklimov:
This concern is expected to be resolved by completion and merge of #2049

PR #671, opened 2019-02-26
ups.conf/nutupsdrv: reword "user" directive section

aquette:
Thanks to Nash Kaminski for the base rewording

Reference: https://github.com/networkupstools/nut/pull/493
           https://github.com/networkupstools/nut/issues/336

### DIFF
--- a/docs/man/nutupsdrv.txt
+++ b/docs/man/nutupsdrv.txt
@@ -117,8 +117,10 @@ opened before the chroot call, so you do not need to create them inside
 the jail.  In fact, it is somewhat safer if you do not.
 
 *-u* 'username'::
-If started as root, the driver will setuid(2) to the user id
-associated with 'username'.
+The driver will setuid(2) to the user id associated with 'username',
+as opposed to its current effective user id.
+Note that drivers are generally started as root to satisfy privileges to access
+resources such ups.conf(5).
 +
 If you do not specify this value and start it as root, the driver will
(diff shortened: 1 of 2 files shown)


clepple:
The sentence about drivers starting as root is probably more confusing than helpful. While the driver starts as root, it changes to a less-privileged user ID by default, even if `-u` is not provided.

aquette:
@clepple: so do you think that the wording in this PR fix this?

clepple:
@aquette no, as mentioned [originally](
https://github.com/networkupstools/nut/issues/336#issue-185915778), I think it is important to point out that the `user` argument to the driver is optional, and overrides the built-in unprivileged user ID. I'll see if I can come up with something later.

clepple:
Still working on it, but I'm looking at something like this:
```
*-u* 'username'::
Override the unprivileged username that the driver may use after startup.
If started as root, the driver will look up 'username' in the `passwd` 
database, then change to the user and group identities associated with
'username'. (If started with a nonzero UID or effective UID, the driver will 
silently ignore this option.)
+
When compiling NUT from source, the default username is typically `nobody`, and 
this may cause permission errors when the driver opens the UPS device node. You 
can use this option to temporarily override the defaults. For testing purposes,
you can set this option to `root` to bypass permission errors, especially with
USB-based drivers. However, you will want to remove this option later in order
to avoid permission conflicts between the driver and the unprivileged copy of 
linkman:upsd[8].
```

Issue #656, opened 2019-02-01
RFC: INSTCMD and SET VAR status tracking


aquette:
Dear fellows,
I'm happy to finally be able to provide a RFC for this long awaited feature!
(esp. @clepple , @zykh , @jimklimov)

From docs/net-protocol.txt, Future ideas -> Command status:

    After sending an INSTCMD or SET, a client will eventually be able to poll to see whether it was completed successfully by the driver.

Proposal:

* Add a new "CMDSET_STATUS" command in the network protocol to enable/disable
the feature and provide status tracking information upon request:

  * docs/net-protocol.txt additions

    SET CMDSET_STATUS
    ~~~~~~~~~~~~~~~~~

    Form:

        SET CMDSET_STATUS "<value>"
        SET CMDSET_STATUS ON
        SET CMDSET_STATUS OFF

    Response:

        OK
        ERR INVALID-ARGUMENT (if "value" is not "ON" or "OFF")
        ERR USERNAME-REQUIRED (if not yet authenticated)
        ERR PASSWORD-REQUIRED (if not yet authenticated)


    GET CMDSET_STATUS
    ~~~~~~~~~~~~~~~~~

    Form:

        GET CMDSET_STATUS               (activation status of CMDSET_STATUS)
        GET CMDSET_STATUS [<status_id>] (execution status of a command / setvar) 
        GET CMDSET_STATUS 1bd31808-cb49-4aec-9d75-d056e6f018d2

    Response:

        PENDING              (command execution is pending)
        SUCCESS              (command was successfully executed)
        ERR UNKNOWN          (command execution failed with unknown error)
        ERR INVALID-ARGUMENT (command execution failed due to missing or invalid argument)
        ERR FAILED           (command execution failed)

* whenever CMDSET_STATUS is enabled, and a valid command / setvar is received by upsd,
upsd will allocate an ID (UUID v4 random, see below notes) and forward to driver
the request with the additional STATUS_ID tag and ID parameter

  * docs/sock-protocol.txt additions

    Commands sent by the server
    ---------------------------

    INSTCMD
    ~~~~~~~

        INSTCMD <cmdname> [cmdparam] [STATUS_ID <ID>]

        INSTCMD panel.test.start
        INSTCMD load.off 10 
        INSTCMD load.on 10 STATUS_ID 1bd31808-cb49-4aec-9d75-d056e6f018d2

  * NOTE:
    * cmdparam is an additional and optional parameter for the command,
[...]

PR #673, opened 2019-02-26
Implement tracking commands in libnutclient

boricj:
Extends libnutclient's C++ API to support tracking commands.

### DIFF
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -406,6 +406,8 @@ void Socket::write(const std::string& str)throw(nut::IOException)
  *
  */
 
+const Feature Client::TRACKING = "TRACKING";
+
 Client::Client()
 {
 }
@@ -479,6 +481,19 @@ bool Client::hasDeviceCommand(const std::string& dev, const std::string& name)th
   return names.find(name) != names.end();
 }
 
+bool Client::hasFeature(const Feature& feature)throw(NutException)
+{
+	try
+	{
+		// If feature is known, querying it won't throw an exception.
+		isFeatureEnabled(feature);
+		return true;
+	}
+	catch(...)
+	{
+		return false;
+	}
+}
 
 /*
  *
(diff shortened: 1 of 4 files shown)


clients/nutclient.cpp
```diff
@@ -753,6 +768,57 @@ int TcpClient::deviceGetNumLogins(const std::string& dev)throw(NutException)
...
+
+	if (result == "PENDING")
+	{
+		return AsyncResult::PENDING;
+	}
+	else if (result == "SUCCESS")
+	{
+		return AsyncResult::SUCCESS;
+	}
+	else if (result == "ERR FAILED")
```
aquette: As per https://github.com/networkupstools/nut/blob/master/docs/net-protocol.txt#L209
Result can also be:
* ERR UNKNOWN          (command execution failed with unknown error)
* ERR INVALID-ARGUMENT (command execution failed due to missing or invalid argument)
beside from the ERR FAILED (command execution failed)

clients/nutclient.h
```diff
@@ -294,6 +312,18 @@ class Client
 	virtual void deviceMaster(const std::string& dev)throw(NutException)=0;
 	virtual void deviceForcedShutdown(const std::string& dev)throw(NutException)=0;
 
+	/**
+	 * Redeem an async action result from an async ID.
```
aquette: not sure what you wanna mean by "redeem", but should be reworded.
Also async ID is to be worded "tracking ID", as elsewhere.

aquette:
part of #656

boricj:
Updated pull request, tested with sample program on both old and new upsd with no issues.

docs/new-clients.txt
```diff
@@ -52,27 +52,78 @@ and commands with an object-oriented API in C++ and C.
...
+  #include <unistd.h>
+  #include <stdlib.h>
+  
   #include <nutclient.h>
+  
   using namespace nut;
   using namespace std;
   
-  int main(int argc, char** argv)
+  int main(int argc, char **argv)
```
aquette: @boricj nitpicking somehow, but:
a comment here in the code would be helpful to depict that
// argv[1] is the mandatory NUT device name (@localhost), used to list variables from
// argv[2] is an optional command. When provided, it will be executed and possibly with status tracking enabled
I'll merge the PR and do it myself to avoid another 1d delay

PR #675, opened 2019-03-08
Introduce shell script syntax checks

jimklimov:
It is annoying to find foolish typos like quotation mark mismatch after spending time to build a package... so this should happen no longer :)

NOTE: This is in the end not part of common required `check-local` because tools for this activity (like `file`) may be missing on an arbitrary system and should not block builds there. Maybe this can be extended by a `configure` check later on.

For good output, see e.g. https://travis-ci.org/networkupstools/nut/jobs/503591138

And a staged example failure at https://travis-ci.org/jimklimov/nut/jobs/503593246#L924 :
````
...
+/bin/sh -n tests/nut-driver-enumerator-test.sh
tests/nut-driver-enumerator-test.sh: 318: tests/nut-driver-enumerator-test.sh: Syntax error: newline unexpected (expecting ")")
ERROR: Syntax check failed for script file: tests/nut-driver-enumerator-test.sh
make: *** [shellcheck] Error 2
Command exited with non-zero status 2
````

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -59,7 +59,7 @@ doc spellcheck-sortdict:
 
 # This target adds syntax-checking for committed shell script files,
 # to avoid surprises and delays in finding fatal typos after packaging
-check-scripts-syntax shellcheck:
+check-scripts-syntax:
 	@echo 'NOTE: modern bash complains about scripts using backticks (which we ignore for portability reasons): `...` obsolete, use $(...)'
 	for F in `git ls-files || find . -type f` ; do \
 	    case "`file "$$F"`" in \
@@ -68,6 +68,15 @@ check-scripts-syntax shellcheck:
 	    esac || { RES=$$? ; echo "ERROR: Syntax check failed for script file: $$F" >&2 ; exit $$RES ; } ; \
 	done
 	@echo 'SUCCESS: Shell scripts syntax is acceptable, no fatal issues were found'
+
+shellcheck-disclaimer:
+	@echo "==============================================================================="
+	@echo "NOTICE: 'make shellcheck' is currently an alias for 'make check-scripts-syntax'"
+	@echo "Later it may become a call to the real shellcheck tool (if available on the"
+	@echo "build system during the configure phase)"
+	@echo "==============================================================================="
+
+shellcheck: shellcheck-disclaimer check-scripts-syntax
 
 # ----------------------------------------------------------------------
 # Automatically generate the ChangeLog from Git logs:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Makefile.am [hunk 1]
```diff
@@ -57,6 +57,18 @@ spellcheck spellcheck-interactive:
...
 	cd $(srcdir)/docs && $(MAKE) $@
 
+# This target adds syntax-checking for committed shell script files,
+# to avoid surprises and delays in finding fatal typos after packaging
+check-scripts-syntax shellcheck:
+	@echo 'NOTE: modern bash complains about scripts using backticks (which we ignore for portability reasons): `...` obsolete, use $(...)'
+	for F in `git ls-files || find . -type f` ; do \
+	    case "`file "$$F"`" in \
+	        *"Bourne-Again shell script"*) ( set -x ; /bin/bash -n "$$F" ; ) ;; \
+	        *"POSIX shell script"*|*"shell script"*) ( set -x ; /bin/sh -n "$$F" ; ) ;; \
```
clepple: I don't mean to nit-pick, but it seems as though "shell script" is a superset of "POSIX shell script". Did you intend to have another option here?

jimklimov:
This was more or less a reaction to what `file` returns on my systems and
Travis (and so it might be adapted easily to other systems). I believe it
says POSIX shell if interpreter is /bin/sh or /sbin/sh.

The implied expectation here is that other shells are usually at least
posix compatible (while the `case` can be extended for those that are not).

In practice i've seen various versions of ksh, dash, bash and busybox sh in
this role. The target distro of a build can thus check the system scripts
with its system shell.

aquette:
while this improves a bit the situation, this won't cover much.
what about using something like shellcheck (https://github.com/koalaman/shellcheck)?
seems to be available as deb too for our travis targets.

jimklimov:
I'd leave `shellcheck` integration for next iterations (seeing it did inspire the recipe name though). I'd need time to get more acquainted with the tool, particularly to see how well it handles non-bash shells (and older bash versions) we can face in various target OS builds, and how well it handles questionable syntax that could be used on purpose, such as quoted variables where we intend exactly one token (empty, whitespaced, or not) vs. expanding unquoted vars into lists of tokens, and/or whether certain warning instances can be quiesced which would otherwise be always red in CI or cause code refactoring to somehow silence the tool.

The current proposal has a benefit of checking the syntax during builds (developers, CI, packaging...) against the actual shell intepreter which a script would run in, and not requiring extra tools to detect the most common errors/typos at least, such as mismatched quotation and parentheses symbols, missed whitespace in critical places, and `do/done` `if/fi` `case/esac` etc. blocks. And of course whatever else the shell interpreter would choose to explode against.

jimklimov:
PR updated, so the makefile and tests now stress the (potential) difference of shellcheck and check-scripts-syntax targets; however the former for now serves as an alias to the latter (with possibility to redefine later; even then maybe optionally - if `configure` finds the `shellcheck` toolkit in the system).

PR #677, opened 2019-03-13
Use str* API to check for valid numeric value

### LINKED ISSUE #676: use str* API in place of isdigit() in server/conf.c
`isdigit()` takes a char/int, not a char*.

[Originally mentioned here](
https://github.com/networkupstools/nut/commit/9a40084a5c756fa47c436819c2b15c41fa91d93a#commitcomment-32558936), but we need to fix this before the next release.

aquette:
Closes networkupstools/nut#676

### DIFF
--- a/server/conf.c
+++ b/server/conf.c
@@ -116,44 +116,40 @@ static void ups_update(const char *fn, const char *name, const char *desc)
 /* return 1 if usable, 0 if not */
 static int parse_upsd_conf_args(int numargs, char **arg)
 {
+	unsigned int	temp;
+
 	/* everything below here uses up through arg[1] */
 	if (numargs < 2)
 		return 0;
 
 	/* MAXAGE <seconds> */
 	if (!strcmp(arg[0], "MAXAGE")) {
-		if (isdigit(arg[1])) {
-			maxage = atoi(arg[1]);
+		if (str_to_uint(arg[1], &temp, 10)) {
+			maxage = temp;
 			return 1;
 		}
-		else {
-			upslogx(LOG_ERR, "MAXAGE has non numeric value (%s)!", arg[1]);
-			return 0;
-		}
+		upslogx(LOG_ERR, "Could not convert MAXAGE (%s) to an unsigned int (%s)!", arg[1], strerror(errno));
+		return 0;
 	}
 
 	/* TRACKINGDELAY <seconds> */
 	if (!strcmp(arg[0], "TRACKINGDELAY")) {
-		if (isdigit(arg[1])) {
-			tracking_delay = atoi(arg[1]);
+		if (str_to_uint(arg[1], &temp, 10)) {
+			tracking_delay = temp;
 			return 1;
 		}
-		else {
-			upslogx(LOG_ERR, "TRACKINGDELAY has non numeric value (%s)!", arg[1]);
-			return 0;
-		}
+		upslogx(LOG_ERR, "Could not convert TRACKINGDELAY (%s) to an unsigned int (%s)!", arg[1], strerror(errno));
+		return 0;
 	}
 
 	/* MAXCONN <connections> */
 	if (!strcmp(arg[0], "MAXCONN")) {
-		if (isdigit(arg[1])) {
-			maxconn = atoi(arg[1]);
+		if (str_to_uint(arg[1], &temp, 10)) {
+			maxconn = temp;
 			return 1;
 		}
-		else {
-			upslogx(LOG_ERR, "MAXCONN has non numeric value (%s)!", arg[1]);
-			return 0;
-		}
+		upslogx(LOG_ERR, "Could not convert MAXCONN (%s) to an unsigned int (%s)!", arg[1], strerror(errno));
+		return 0;
 	}
 
 	/* STATEPATH <dir> */
@@ -187,14 +183,12 @@ static int parse_upsd_conf_args(int numargs, char **arg)
 #ifdef WITH_CLIENT_CERTIFICATE_VALIDATION
 	/* CERTREQUEST (0 | 1 | 2) */
 	if (!strcmp(arg[0], "CERTREQUEST")) {
-		if (isdigit(arg[1])) {
-			certrequest = atoi(arg[1]);
+		if (str_to_uint(arg[1], &temp, 10)) {
+			certrequest = temp;
 			return 1;
 		}
-		else {
-			upslogx(LOG_ERR, "CERTREQUEST has non numeric value (%s)!", arg[1]);
-			return 0;
-		}
+		upslogx(LOG_ERR, "Could not convert CERTREQUEST (%s) to an unsigned int (%s)!", arg[1], strerror(errno));
+		return 0;
 	}
 #endif /* WITH_CLIENT_CERTIFICATE_VALIDATION */
 #endif /* WITH_OPENSSL | WITH_NSS */
--- a/server/conf.c
+++ b/server/conf.c
@@ -128,7 +128,7 @@ static int parse_upsd_conf_args(int numargs, char **arg)
 			maxage = temp;
 			return 1;
 		}
-		upslogx(LOG_ERR, "Could not convert MAXAGE (%s) to an unsigned int (%s)!", arg[1], strerror(errno));
+		upslogx(LOG_ERR, "Could not convert %s (%s) to an unsigned int (%s)!", arg[0], arg[1], strerror(errno));
 		return 0;
 	}
 
@@ -138,7 +138,7 @@ static int parse_upsd_conf_args(int numargs, char **arg)
 			tracking_delay = temp;
 			return 1;
 		}
-		upslogx(LOG_ERR, "Could not convert TRACKINGDELAY (%s) to an unsigned int (%s)!", arg[1], strerror(errno));
+		upslogx(LOG_ERR, "Could not convert %s (%s) to an unsigned int (%s)!", arg[0], arg[1], strerror(errno));
 		return 0;
 	}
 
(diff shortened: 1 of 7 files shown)


server/conf.c
```diff
@@ -116,44 +116,40 @@ static void ups_update(const char *fn, const char *name, const char *desc)
...
-			maxage = atoi(arg[1]);
+		if (str_to_uint(arg[1], &temp, 10)) {
+			maxage = temp;
 			return 1;
 		}
-		else {
-			upslogx(LOG_ERR, "MAXAGE has non numeric value (%s)!", arg[1]);
-			return 0;
-		}
+		upslogx(LOG_ERR, "Could not convert MAXAGE (%s) to an unsigned int (%s)!", arg[1], strerror(errno));
```
zykh: now that I see it again... perhaps, to avoid retyping the options, we could use something like this:
```
upslogx(LOG_ERR, "Could not convert %s (%s) to an unsigned int (%s)!", arg[0], arg[1], strerror(errno));
```

zykh (commented):
As previously mentioned, I still think that we should also make `maxage`, and friends, `unsigned int`.

aquette:
@zykh fixed the both

clepple (approved):
LGTM

zykh (approved):
Just a couple of things:
- printf formats:
  https://github.com/networkupstools/nut/blob/3d79006c2ec3b038f71e1d9441aaf73f744b77ab/server/upsd.c#L699-L702
- shouldn't the following be `if (certrequest < NETSSL_CERTREQ_NO || certrequest > NETSSL_CERTREQ_REQUIRE)` (or, given that now `certrequest` is unsigned and `NETSSL_CERTREQ_NO` is zero, just `if (certrequest > NETSSL_CERTREQ_REQUIRE)`) ?
  (i.e. `||` instead of `&&`, and `NETSSL_CERTREQ_REQUIRE`, which is 2, instead of `NETSSL_CERTREQ_REQUEST`, which is 1)
  https://github.com/networkupstools/nut/blob/3d79006c2ec3b038f71e1d9441aaf73f744b77ab/server/netssl.c#L481-L485
- how unlikely is that `sysconf(_SC_OPEN_MAX)` returns a) `-1` or b) a value that (as `long`) exceeds what can be stored in an `unsigned int`?
  here:
  https://github.com/networkupstools/nut/blob/3d79006c2ec3b038f71e1d9441aaf73f744b77ab/server/upsd.c#L696
  and here:
  https://github.com/networkupstools/nut/blob/3d79006c2ec3b038f71e1d9441aaf73f744b77ab/server/upsd.c#L1251

clepple (changes requested):
In light of @zykh's comments, I think the change to `unsigned` is hasty. As pointed out, the `sysconf()` call shouldn't return a negative number, but a check is not expensive.

jimklimov:
@clepple : a few merge conflicts cropped up over time :( Probably your versions should be trusted, but can you please revise that this is the case - and that master-branch changes so far did not change non-trivially some semantics etc.?

jimklimov:
Updated over recent changes on the master branch, including PR #3631. Compared to it:
* Use same C formatting strings for invalid values as much as possible, so the compiler has easier time with optimizing the copies away
* Convert `maxage` and `tracking_delay` to `unsigned int`, as proposed in this PR originally
* Range-check `CERTREQUEST` setting
* Adapt `test_upsdconf.c` to new data types

Earlier discussion concerns about `sysconf()` returning `-1` or not became (hopefully) obsoleted by upstream evolution about this - see #3303 et al

jimklimov:
@user01010111 : wanna take a bite at another low-hanging fruit? It seems we have some trouble with 32-bit builds here, probably related to change of these data points into `unsigned int`... maybe easier to revert that one and keep range-checks in `conf.c`...

```
FAILURE [#1]    MAXAGE '2147483648': accepted=1, value=2147483648; expected 0, 17
FAILURE [#2]    load_upsdconf(0), MAXAGE '2147483648': value=2147483648; expected 17
FAILURE [#3]    load_upsdconf(1), MAXAGE '2147483648': value=2147483648; expected 17
FAILURE [#4]    TRACKINGDELAY '2147483648': accepted=1, value=2147483648; expected 0, 17
FAILURE [#5]    load_upsdconf(0), TRACKINGDELAY '2147483648': value=2147483648; expected 17
FAILURE [#6]    load_upsdconf(1), TRACKINGDELAY '2147483648': value=2147483648; expected 17
```

My recent commits on the branch fixed same issues for default (64-bit here) builds. Going offline now, so if you (or your AI) has a moment to comment here with a suggested fix, that would be welcome :)
* Repro (on Ubuntu): `./configure CFLAGS="-m32" CXXFLAGS=-m32 LDFLAGS=-m32 --build=x86_64-pc-linux-gnu --host=i686-pc-linux-gnu --without-docs` (may also require `apt install gcc-multilib`).

user01010111:
Suggested test-only follow-up for `b79b8cb7e54f21401f78bfb01db7c8017c361919`.

The six failures come from the test still treating `LONG_MAX + 1` as invalid for every option. With 32-bit `long`, that value is 2147483648, which fits the new unsigned MAXAGE and TRACKINGDELAY types. The test now derives the expected result from each option's limit, preserving the unsigned types and the existing MAXCONN limit.

Enabling NSS client validation exposed another outdated part of the same test: its CERTREQUEST stub was still signed, and its expectations allowed values above the parser's new 0..2 limit. The patch aligns those test expectations with the current parser. It changes only `tests/test_upsdconf.c`.

I reproduced the six failures on the unchanged head, then ran the corrected test natively on i386 and x86-64: 263 checks passed on each. The NSS client-validation configuration passed 365 checks. Both default builds, all 13 tests in the x86-64 test directory, and the whitespace/non-ASCII checks passed.

Top-level `make check` and `make distcheck-light` initially stopped on the existing possessive-word spelling failures. Both now pass with spelling enabled after applying only the exact dictionary changes from pending #3629 (`5131bdb6`). That dependency is separate from this test patch. The distribution check completed archive build, tests, staged install, uninstall and cleanup; the archived test and dictionary match the validated source. Documentation rendering was skipped. These full-gate results therefore depend on integrating the #3629 dictionary changes.

AI assistance: OpenAI Codex, using gpt-5.6-sol for initial scoping and gpt-6-astra for review, implementation and validation. I have reviewed this change and accept responsibility for contributing it under the project's licence.

<details>
<summary>Patch against b79b8cb7</summary>

```diff
--- a/tests/test_upsdconf.c
+++ b/tests/test_upsdconf.c
@@ -24,7 +24,7 @@
 char *certfile = NULL, *certpath = NULL, *certname = NULL, *certpasswd = NULL;
 int disable_weak_ssl = 0;
 #ifdef WITH_CLIENT_CERTIFICATE_VALIDATION
-int certrequest = 0;
+unsigned int certrequest = 0;
 #endif
 
 /* These daemon operations must not be reached by the configuration tests. */
@@ -185,23 +185,25 @@
 		sizeof(int), sizeof(long), sizeof(nfds_t));
 	for (i = 0; i < sizeof(options) / sizeof(options[0]); i++) {
 		previous = !strcmp(options[i], "CERTREQUEST") ? 2 : 17;
-		check_value(options[i], "0", 1, 0);
-		check_value(options[i], "1", 1, 1);
-		check_value(options[i], "2", 1, 2);
-		check_value(options[i], "00015", 1, 15);
-		check_value(options[i], "32 \t", 1, 32);
-		check_value(options[i], "3600", 1, 3600);
-		for (j = 0; j < sizeof(invalid) / sizeof(invalid[0]); j++)
-			check_value(options[i], invalid[j], 0, previous);
-
-		limit = (uintmax_t)INT_MAX;
+		limit = (uintmax_t)NETSSL_CERTREQ_MAX;
 		if (!strcmp(options[i], "MAXAGE") || !strcmp(options[i], "TRACKINGDELAY")) {
 			limit = (uintmax_t)UINT_MAX;
 		} else
 		if (!strcmp(options[i], "MAXCONN")) {
 			limit = (uintmax_t)((nfds_t)-1);
-			if (limit > (uintmax_t)LONG_MAX) limit = (uintmax_t)LONG_MAX;
+			if (limit > (uintmax_t)LONG_MAX) {
+				limit = (uintmax_t)LONG_MAX;
+			}
 		}
+		check_value(options[i], "0", 1, 0);
+		check_value(options[i], "1", 1, 1);
+		check_value(options[i], "2", 1, 2);
+		check_value(options[i], "00015", limit >= 15, limit >= 15 ? 15 : previous);
+		check_value(options[i], "32 \t", limit >= 32, limit >= 32 ? 32 : previous);
+		check_value(options[i], "3600", limit >= 3600, limit >= 3600 ? 3600 : previous);
+		for (j = 0; j < sizeof(invalid) / sizeof(invalid[0]); j++)
+			check_value(options[i], invalid[j], 0, previous);
+
 		snprintf(number, sizeof(number), "%" PRIuMAX, limit - 1);
 		check_value(options[i], number, 1, limit - 1);
 		snprintf(number, sizeof(number), "%" PRIuMAX, limit);
@@ -209,15 +211,19 @@
 		snprintf(number, sizeof(number), "%" PRIuMAX, limit + 1);
 		check_value(options[i], number, 0, previous);
 		snprintf(number, sizeof(number), "%" PRIuMAX, (uintmax_t)LONG_MAX + 1);
-		check_value(options[i], number, 0, previous);
+		if ((uintmax_t)LONG_MAX < limit) {
+			check_value(options[i], number, 1, (uintmax_t)LONG_MAX + 1);
+		} else {
+			check_value(options[i], number, 0, previous);
+		}
 	}
 #if defined(WITH_SSL) && defined(WITH_CLIENT_CERTIFICATE_VALIDATION)
 	check_value("CERTREQUEST", "NO", 1, 0);
 	check_value("CERTREQUEST", "REQUEST", 1, 1);
 	check_value("CERTREQUEST", "REQUIRE", 1, 2);
 	check_value("CERTREQUEST", "require", 0, 2);
-	/* ssl_init still owns the 0..2 domain check and its fatal error policy. */
-	check_value("CERTREQUEST", "3", 1, 3);
+	/* Configuration parsing owns the 0..2 domain check. */
+	check_value("CERTREQUEST", "3", 0, 2);
 #endif
 
 	f = fopen(config_file, "w");
```

</details>

Signed-off-by: user01010111 <lapses.50.booster@icloud.com>

jimklimov:
@user01010111 : thanks for the research and confirmation! So after all it seems quite similar to what I did to this test program for #3640. I'll model the fix here on that code, after merging that PR then.

user01010111:
No worries, @jimklimov .. Nice work as usual! 🫡

PR #679, opened 2019-03-18
Fix SNMPv3 in nut-scanner

boricj:
Scanning isn't working when SNMPv3 authentication/privacy password is required. Tested with an Eaton ePDU.


aquette:
tests confirmed the issue and the fix, thx @boricj

PR #682, opened 2019-03-29
Nut driver enumerator optimize

jimklimov:
Initial testing showed that, not too surprisingly, the honest full config parsing to get the device name for service instance is an overkill, especially on systems that monitor many power devices and parse such data all at once during startup. This particular issue can be addressed by caching the original device name into service properties, which can be inspected quickly - and only fall back to the original slower logic if no hits happened. This block of commits also includes updating of existing configurations made with earlier nut-driver-enumerator versions (which do not have those device section names saved into service instance properties).

Another big change here is relocation of some functionality from the CLI processing loop into routines, some of which are now used by other code inside the script.

And a few smaller typos and issues found over time while this was all tested in real life deployments.

CC @geraldguillaume also.

UPDATE: Note to self: I am sure there was some further development about optimizing the script, maybe in 42ity fork, but can't find it at the moment. I think it revolved around detecting `BASH_VERSION` and using bash arrays or hash-maps to store data, for some 30% boost in parser performance on platforms where it is possible.


jimklimov:
Also adding the `nut-driver-enumerator-daemon` development from another branch, which changed nearby parts of code and so they conflicted. This feature solves its share of issues on busier systems with many power devices to manage, especially where writes into `ups.conf` come faster than the script processes a config reconciliation cycle - and subsequent service triggerings end up ignored.

Issue #123, opened 2014-04-22
upsmon child process PID stored in upsmon.pid


bigon:
Hello,

When using systemd, it complains about the PID stored in the .pid file:

```
nut-monitor.service: Supervising process XXXX which is not our child. We'll most likely not notice when it exits.
```

And indeed when looking in upsmon.pid, the PID stored there is the one from the grand-child (unprivileged process) of the process started by init. Shouldn't this be the PID of the direct forked process instead?

iva2k:
From the surface looking at it - sounds like there should be two different .pid files for `upsmon` run as `root` and `upsmon` run as `nut`/unprivileged user. If that's the case, it is a design bug. Should systemd take care of pid of `upsmon` run as `root`, and forked unprivileged pid of `usbmon` run as `nut` should be taken care of by the forking code?

Is unprivileged pid overwriting root pid and that breaks `systemd` and / or `upsmon` operation?

Even if that's not the case, the error messages in syslog do not look comforting and give no confidence that UPS shutdown will work correctly in all circumstances. Sounds like a critical issue. Can someone from NUT team chime in and triage this?

gwaitsi:
The below message re fopen also appears on freebsd variations i.e. freenas/truenas (although nut appears to work and shutdown while all related pids are created)

fopen /var/run/nut/upsmon.pid: No such file or directory
fopen /var/db/nut/upsd.pid No such file or directory

electrofloat:
So... what is the solution to this issue?

RJHsiao:
Hi there,
I get same message in my Ubuntu 20.04 LTS server, and no solution found.
Is somebody work on it? Or the solution(s) is/are exist that we can google it with the keyword I don't know?

jimklimov:
I keep meaning to take a closer look at this, but it keeps drowning in
priorities :(

jimklimov:
PR #683 (and #349 before it, now part of it) introduces a separation of debugging options vs. foreground/background running behavior, and in particular redefines the daemons under systemd units to run in foreground. Hopefully that change would alleviate this issue.

jimklimov:
Playing around with the daemons and service units, for issues/PRs linked above, found an interesting behavior here:

When `nut-monitor.service` is initially started (newly as a foregrounded process without the extra forking for systemd, but still with forking to privileged/unprivileged pair), the "Main PID" for systemd is that of the (root-owned) parent process:
````
[...]

PR #683, opened 2019-04-01
Support debug level passed via ups.conf

jimklimov:
There were previously attempts to arrange debugability of "production-mode" start-up of NUT daemons vs. their forced foregrounding in case of `-D` options present in command line. As I recently needed to inspect how driver services start and why they failed in my case, I took another attempt in this area, and came up with the proposed solution (just for drivers so far) to:

* separate the `-D` CLI option vs. explicit option to foreground `-F` or background `-B` the daemon; if neither of the latter options is set, defaults are as before (go background unless debugged);

  ** Names/shortcuts/help-text of the options can be subject to peer review, to have some values acceptable for consistently named options across the board, if we extend this approach to other NUT daemons;

* allow to set the default (driver) debugging level in either global or driver-level section of `ups.conf`; if both are set, the driver-level setting takes precedence; either of these may be set to a zero or positive integer;

  ** unlike command-line `-D` this setting does not require foregrounding (`-F` can be explicitly passed if desired) so the service daemons stay backgrounded by default as before;

  ** this allows to see debug messages in original service log, with the service units (access rights, environment, etc.) otherwise not modified - which often corrupts or complicates command-line troubleshooting;

  ** a more arguable corollary: since we do not currently have a way to disable debugging (set `0` level) I chose to require that the valid option in `ups.conf` (global section, or preferably driver-level value) specifies the *minimum* debugging level applied to the daemon -- if the command-line invocation specifies less `-D` flags, the practically applied level would be the one in the config.

  *** Per peer review, this design choice (and config option name) can be amended; at the very least with current implementation a CLI caller can easily set a minimum of `-D` (== `1`) level instead, and something in CLI processing would need to be invented to reset a `"-D == 0"` level.

(see also #123 and #821 and #349)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -28,6 +28,13 @@ Release notes for NUT 2.7.5 - what's new since 2.7.4:
    for them to complete initializing. This matters on systems that monitor
    from dozens to hundreds of devices.
 
+ - Daemons such as upsd, upsmon, upslog, and device drivers previously
+   implied that enabled debugging (or upslog to stdout) means foreground
+   running, otherwise the daemon was always sent to the background.
+   Now there are explicit options for this (`-F`/`-B`), although default
+   behavior is retained. This change is used for simplified service unit
+   definitions.
+
  - Improvements for device discovery or driver "lock-picking", including
    general support for:
    * "Standalone" mode (`-s` option), to monitor a device which is not
--- a/NEWS
+++ b/NEWS
@@ -22,6 +22,11 @@ Release notes for NUT 2.7.5 - what's new since 2.7.4:
    server initially without any device configurations and reloading it later
    to apply config changes on the fly [PR #766]
 
+ - Add support for `debug_min=NUM` setting (ups.conf, upsd.conf, upsmon.conf)
+   to specify the minimum debug verbosity for daemons. This allows "in-vivo"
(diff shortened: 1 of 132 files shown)


jimklimov:
Testing with `ups.conf` to the effect of:
````
debug_min = 6

[nutdev1]
        driver = "usbhid-ups"
        port = "auto"
        debug_min = 4
````

Here driver-level setting takes precedence over global-level one, and over the (lower) debug set on CLI (pass-through from `upsdrvctl` courtesy of #1287):
````
# ./drivers/upsdrvctl -DDDd start nutdev1
...
   0.006049     [D1] debug level is '4'
...
````

jimklimov:
Similarly, checked for other daemons, e.g. `upsmon`:
````
# /usr/local/ups/sbin/upsmon -F
Network UPS Tools upsmon 2.7.4-4660-gc31748b
fopen /var/run/upsmon.pid: No such file or directory
UPS: nutdev1 (primary) (power value 1)
Using power down flag file /etc/killpower
   0.000000     [D1] debug level is '3'
   0.000706     [D1] debug level is '3'
...
````
(second line probably comes from child process arrangements? though there shouldn't be a child in foreground mode? Just privilege drop - and maybe a re-parse?)

PR #685, opened 2019-04-01
Problem: upsdebug*() routines pass data and allocate vars always

jimklimov:
Solution: wrap old well-known API routine names into macros that check debugging level first, and only invest into calling routines and passing data later - if that would not be in vain.

This should make "production-mode" runs a bit more efficient in CPU churn. We have tens of thousands debugging calls that are typically ignored during the daemon lifetime.


jimklimov:
The results are not extremely conclusive due to fluctuation in numbers and other congestion in the test system.

With both older and newer binaries in place, difference between lack of debugging and full debug (6 `-D`s) is up to 1 second from about 8 it takes to initialize an ePDU daisy chain, tested like this:
````
$ time sh -c 'NUT_STATEPATH=/tmp  ./drivers/snmp-ups -s test -x port=${host_pdu} -d 1 -DDDDDD >/dev/null 2>&1'

$ time sh -c 'NUT_STATEPATH=/tmp  ./drivers/snmp-ups -s test -x port=${host_pdu} -d 1 >/dev/null 2>&1'
````

It seems that the version in PR adds a few fractions of a second benefit compared to "old" version, but this is the part hard to prove :)

With high debugging level in place, both old and new binaries return comparable amounts of lines (29020 vs 29023, when piped to `wc -l` instead of `>/dev/null`) so at least no regression here.

clepple (approved):
LGTM

PR #686, opened 2019-04-01
cyberpower-mib: Add shutdown capability and other instant commands

overhacked:
## Shutdown commands:
1. `shutdown.return`
1. `shutdown.stayoff`

### Related variables
1. `ups.delay.start`
1. `ups.delay.shutdown`
    * CyberPower's CPS-MIB provides different delay values for return vs stayoff. I am using the return (aka sleep) value, because that is the more common use case.
1. `ups.delay.reboot`
    * Not supported by CPS-MIB but the subdriver reports as 0, because that is the de facto default for CyberPower.

## Other Commands
1. `load.off/on`
1. `test.failure.start`
1. `test.panel.start`
1. `test.battery.start`
1. `calibrate.start/stop`

## Also fixed
* `sysOID` was CyberPower's enterprise OID, not what the UPS RMCARDxxx returns, which is `CPS-MIB::ups` (*CyberPower_enterprise_OID*.1.1.1)
* `ups.status` is more informative
    1. Reports BOOST
    1. Reports RB (replace battery)
    1. Reports CAL (calibration in progress)
* `battery.runtime` now returns the correct units for upsd (seconds not minutes)

### DIFF
--- a/drivers/cyberpower-mib.c
+++ b/drivers/cyberpower-mib.c
@@ -111,6 +111,25 @@ static snmp_info_t cyberpower_mib[] = {
 	{ "output.voltage", 0, 0.1, ".1.3.6.1.4.1.3808.1.1.1.4.2.1.0", "",
 		0, NULL },
 
+	/* instant commands. */
+	/* upsAdvanceControlUpsOff */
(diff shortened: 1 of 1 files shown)


jimklimov (approved):

aquette:
Don't lag on me, if it seems good to you guys!

PR #687, opened 2019-04-02
snmp-ups: add ability to set TimeTicks correctly

overhacked:
Driver had been setting read/write parameters as
either Integer or String, but OIDs that expect
TimeTicks cannot be set using an Integer.

snmp-ups already supported getting TimeTicks
correctly and had helper functions for setting
TimeTicks, but logic had not been implemented in
su_setOID().

### DIFF
--- a/drivers/snmp-ups.c
+++ b/drivers/snmp-ups.c
@@ -2801,7 +2801,7 @@ int su_setOID(int mode, const char *varname, const char *val)
 	}
 	else {
 		if (mode==SU_MODE_INSTCMD) {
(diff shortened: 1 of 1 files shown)


jimklimov:
Generally seems ok, I just wonder if NUT's `str_to_long{_strict} ()` might be a preferred choice to raw `atol()` etc. calls?

overhacked:
I was unaware of the safer functions in `common/str.c` and kept with the existing usage of `atoi()` and `atol()`. I've pushed a commit that makes the change.

PR #690, opened 2019-04-05
Problem: isdigit() receives a char* instead of char

jimklimov:
Solution: as a quick fix, pick the first char of the string we pass.
The proper fix would be to use the str_*() API for safe conversions,
in a later iteration.

See-also: https://github.com/networkupstools/nut/issues/676


jimklimov:
I think they are different :)

This one is a quick fix to address a specific problem (we certainly do not need the value of a byte from the string pointer itself) with otherwise assumed relatively sane inputs, and that PR goes for safer conversions of user-inputs which can never be trusted enough :)

PR #692, opened 2019-04-25
Fix for Ferrups FE7000 "off" command

Freet128:
"off" command for FE7000 required at least 5 seconds for the command to be accepted.  

Also should match the off with autostart or "o" with "a" to be consistent with the manual: https://www.eaton.com/content/dam/eaton/products/backup-power-ups-surge-it-power-distribution/backup-power-ups/eaton-ferrups-tower-ups/eaton-ferrups-FE-QFE-userguide-.pdf  (o [time] a is how the manual references the "short" command)

### DIFF
--- a/drivers/bestfcom.c
+++ b/drivers/bestfcom.c
@@ -444,7 +444,7 @@ void upsdrv_shutdown(void)
 {
 	/* NB: hard-wired password */
 	ser_send(upsfd, "pw377\r");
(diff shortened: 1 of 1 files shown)


aquette:
@Freet128 thanks for your contribution.
I'm not familiar with these unit, but if you're interested in, instant commands and settings are not implemented.
There seem to be quite some of these in the pointed manual, Table 9. and page

Issue #709, opened 2019-07-09
Please use Requires field instead of Libs/Cflafgs in lib/libupsclient.pc.in


bigon:
Hello,

In lib/libupsclient.pc.in you are explicitly setting Libs and Cflags.

IMVHO, it would be better to use the Requires field and let pkg-config populate these

PR #711, opened 2019-07-09
Use pkg-config module names instead of listing the libraries

bigon:
Fixes: #709


bigon:
Can this be merged then?

Issue #713, opened 2019-07-09
Please only export functions that are supposed to be public


bigon:
Hello,

ATM, I've the feeling that the different libraries are exporting all their symbols.

It would be nice if they were only exporting the one that are part of the public API

Thanks

PR #714, opened 2019-07-10
libupsclient, libnutscan: Only export symbols part of the public API

bigon:
Fixes: #713


bigon:
This works on my machine, but I'm not sure that `-export-symbols-regex` is available on all platform

Not sure how to do that for C++ library

bigon:
Any comments on this?

zykh:
Perhaps, we should also bump the various versions... once per release should be fine, later today, I'll add a note so that we don't forget to do that (if we ever get to the next release).

Note to future self: fix the conflict in the `libusb-compat-1.0` branch (I used that option, too, but just to filter-out the libusb_* symbols in libnutscan, no longer necessary with this PR).

(yeah, I'm not dead yet - but I'm working on it)

zykh:
...so, @aquette, @jimklimov, nihil obstat?

aquette:
@zykh I let you manage this one, thanks

jimklimov (approved):

PR #715, opened 2019-07-11
Problem: Attempts to start a nut-driver@bogusname actually run …

jimklimov:
…upsdrvctl on everything

Solution: First check if service instance name is valid, then use it
Should fix https://github.com/42ity/nut/issues/87 (same problem applies to upstream NUT too)


jimklimov:
Tested locally on Debian and OmniOS, to both prove having had an issue before and restricting the impact of such misconfiguration after the fixes are applied.

PR #717, opened 2019-07-12
Report load for Tripp Lite USB 3005 protocol

sshimko:
The debug output for a device using the Tripp Lite
09ae 3005 protocol seems to show that the
ups.debug.L is the load as noted on mailing
list posts. Decode it and report it as ups.load.

Tested on TRIPP LITE SMART500RT1U.


clepple:
Thanks, and sorry for the delay in getting this merged.

PR #720, opened 2019-07-23
avoid index out of bounds

rpvelloso:
the parameter passed to autonomy_calc() indexes an array of 5 position and variable 'im' can be equal to '4' at line 460, leading to an index out of bounds.


zykh:
(rebased to add a bit of context to the commit message, and to bump the version of the driver, and) merged in c0a0e56, thanks @rpvelloso!

PR #729, opened 2019-08-18
nut-monitor: Add scalable and 256x256 icon for nut-monitor

bigon:
FTR, this icon comes from gnome-icon-theme 2.28


zykh:
@bigon, is it worth importing the icon in the other available sizes?
Also, I think we should drop the ".png" extension from the `Icon` field of the .desktop file.

zykh:
sorry, @bigon, I didn't notice the force-push (the next time, perhaps, throw something at me)... merging, thanks!

PR #761, opened 2020-02-19
Hotfix nut-driver-enumerator.sh.in (upstream NUT)

jimklimov:
Applies https://github.com/42ity/nut/pull/100 to upstream

Note: the 42ity version has some more pending differences, due to refactoring slowly in progress.


jimklimov:
Caveat emptor: the intensive `systemctl daemon-reload` may be not something intended as a mode of operation by systemd itself. If other "systemd jobs" like starting or stopping services execute at this time, journal gets messages about a disappearing command for this or that unrelated unit. Seems that the state change during daemon-reload is not an atomic transaction, and not transparent to units other than the one really changed. I still do not know if there is any practical effect however (e.g. aborted operations?) beside the worrysome message in logs.

PR #793, opened 2020-06-11
[Needs Work] Support CyberPower RMCARD205

ericzinnikas:
I have a CyberPower OL1500RTXL2UN with an RMCARD205 network card. These changes were necessary for nut to retrieve details via SNMP. All changes except `battery.runtime`, `battery.voltage` and `battery.current` are likely applicable to older models (judging by similar OIDs only).

What's the best way to add support for this newer card, while keeping older ones supported as well? The OID/sysOID are the same and RMCARD205 isn't exposed anywhere (only the actual UPS model is), so I'm not sure what to use. Possibly firmware revision, but I don't know how that changes.

Before/After:
```
+battery.charge: 100
+battery.current: 2.90
-battery.runtime: 93600.00
-battery.runtime.elapsed: 0.00
+battery.runtime: 93600
+battery.runtime.elapsed: 0
+battery.voltage: 40.70
 device.mfr: CYBERPOWER
 device.model: OL1500RTXL2UN
 device.serial: WERHX2000566
 device.type: ups
 driver.name: snmp-ups
+driver.parameter.mibs: cyberpower
 driver.parameter.pollfreq: 15
 driver.parameter.port: 192.168.86.207
 driver.parameter.synchronous: no
-driver.version: 2.7.4
-driver.version.data: cyberpower MIB 0.1
-driver.version.internal: 0.97
+driver.version: 2.7.4-747-g3d88a972
+driver.version.data: cyberpower MIB 0.2
+driver.version.internal: 1.12
+input.frequency: 59.90
+input.voltage: 115
+output.voltage: 120.20
 ups.firmware: Sv3A02
+ups.load: 26
 ups.mfr: CYBERPOWER
 ups.model: OL1500RTXL2UN
 ups.serial: WERHX2000566
```


ericzinnikas:
Before this gets merged, we need to consider that `battery.voltage` has a different OID on newer devices (i.e. RMCARD205) than on old devices. I'm not sure how to differentiate an RMCARD205 device vs older devices.

aquette:
The different oids on different card is not an issue. I've made this engine so that it can handle this, along with multiple oids avail, in high case there is the flag unique, and you should declare by order of pref. Check marlin mib for a ref implem

KenwoodFox:
I have the same card, i hope to see this in a release sometime!

jimklimov:
Note there was also PR #686 recently improving some support for these devices.

I have hopefully fixed the MIB-to-NUT mapping in this PR to allow the driver to query different OIDs as defined by different generations of MIBs, returning the value which replies first (if any).

PR #796, opened 2020-06-15
Do not call `dstate_delinfo()` for LB-condition variables

flobernd:
This commit fixes the `ignorelb` flag for the `apcupsd-ups` driver. Previusly "battery.charge" and "battery.runtime" was cleared before invoking `status_commit()` which caused the LB-condition to always be `false`.


jimklimov (approved):
Seems reasonable, hope it worked in HW testing :)

flobernd:
Thanks for merging. Yes, I tested extensively with my local APC device. Should be fine!

PR #800, opened 2020-06-18
Add info about CHRG status, which seems to not been documented yet

gnudiff:
One of my UPSes gives `ups.status = OL` , the other `ups.status = OL CHRG` when queried by `upsc`

Well, I don't like guessing, even if I am almost sure, so I would like to document this. The only place where CHRG abbreviation is mentioned is in 2007 nut-upsdev list, when it was introduced: https://alioth-lists.debian.net/pipermail/nut-upsdev/2007-June/002310.html

And in Peter Selinger's webpage on upshid-ups https://www.mathstat.dal.ca/~selinger/ups/backups.html

There doesn't seem to be documentation for CHRG anywhere on NUT website.


jimklimov:
Searching in github for the project, the keyword is hit quite a lot (mostly in sources), but there are also a few reveals in texts:
* Seems the legacy flags were obsoleted (though not yet dropped) in favor of a separate attribute: https://github.com/networkupstools/nut/blob/14301bdbdf586d31c3e80a97a3e0eb04c69326b5/NEWS#L13-L15 and https://github.com/networkupstools/nut/blob/9eb16cc462bdabb60f51de3ab2915f6373057620/docs/nut-names.txt#L402-L411
* https://github.com/networkupstools/nut/blob/14301bdbdf586d31c3e80a97a3e0eb04c69326b5/docs/new-drivers.txt#L215-L231
* Many drivers list a lot of items they (I assume) support which is usually a subset of the pool of keywords and related metrics, and a few barely mentioned elsewhere and/or named differently than elsewhere: e.g. https://github.com/networkupstools/nut/blob/14301bdbdf586d31c3e80a97a3e0eb04c69326b5/drivers/belkinunv.c#L76-L89 and https://github.com/networkupstools/nut/blob/14301bdbdf586d31c3e80a97a3e0eb04c69326b5/drivers/dummy-ups.h#L68-L82 vs. https://github.com/networkupstools/nut/blob/790f376917f2f688dc5e257b3effe57a3dfb1772/drivers/usbhid-ups.h#L124-L151 and https://github.com/networkupstools/nut/blob/e503b7bf018b37a5d0c8681a5e2d493e3a3ec53a/drivers/netxml-ups.h#L47-L71 shows "OVER" vs "OVERLOAD"

jimklimov:
It also seems the PR is made against a `genericups` driver manpage, whose sources do not mention it knowing or reporting about `CHRG` at all (note: this is not a document about common UPS nuances, it is a "generic" driver for very dumb devices that can communicate a handful of states by shorting or opening contacts, similar to what we call GPIO nowadays).

https://github.com/networkupstools/nut/blob/master/drivers/genericups.c#L318

jimklimov:
Per docs,

> the `CHRG` and `DISCHRG` flags are being replaced with `battery.charger.status`.  See the linkdoc:user-manual[NUT command and variable naming scheme,nut-names] for more information.

PR #810, opened 2020-08-08
WIP: Fixed various warnings and errors identified by cppcheck

seanm:
Specifically:

arrayIndexThenCheck,drivers/nutdrv_qx_voltronic-qs-hex.c:222,style,Array index 'i' is used before limits check.
arrayIndexThenCheck,drivers/oneac.c:176,style,Array index 'i' is used before limits check.
clarifyCalculation,drivers/belkinunv.c:1063,style,Clarify calculation precedence for '&' and '?'.
clarifyCalculation,drivers/belkinunv.c:922,style,Clarify calculation precedence for '&' and '?'.
clarifyCalculation,drivers/powercom.c:677,style,Clarify calculation precedence for '&' and '?'.
clarifyCalculation,drivers/powercom.c:732,style,Clarify calculation precedence for '&' and '?'.
clarifyCondition,drivers/asem.c:248,style,Boolean result is used in bitwise operation. Clarify expression with parentheses.
duplicateConditionalAssign,drivers/solis.c:725,style,The statement 'if (DaysOnWeek!=DaysOffWeek) DaysOnWeek=DaysOffWeek' is logically equivalent to 'DaysOnWeek=DaysOffWeek'.
duplicateExpression,clients/cgilib.c:49,style,Same expression on both sides of '||'.
duplicateExpression,drivers/nutdrv_qx_voltronic.c:270,style,Same expression on both sides of '||'.
duplicateExpression,drivers/nutdrv_qx_voltronic.c:303,style,Same expression on both sides of '||'.
duplicateExpression,drivers/nutdrv_qx_voltronic.c:397,style,Same expression on both sides of '||'.
identicalInnerCondition,drivers/tripplite_usb.c:579,warning,Identical inner 'if' condition is always true.
incorrectLogicOperator,drivers/dstate.c:1170,warning,Logical conjunction always evaluates to false: c2 && !c2.
incorrectLogicOperator,drivers/rhino.c:190,warning,Logical disjunction always evaluates to true: BattVoltage > 129 || BattVoltage < 144.
incorrectStringBooleanError,drivers/mge-utalk.c:899,warning,Conversion of string literal "\r\n" to bool always evaluates to true.
invalidPrintfArgType_sint,clients/upsset.c:393,warning,%d in format string (no. 1) requires 'int' but the argument type is 'unsigned int'.
invalidPrintfArgType_sint,clients/upsset.c:676,warning,%d in format string (no. 1) requires 'int' but the argument type is 'unsigned int'.
invalidPrintfArgType_sint,drivers/isbmex.c:178,portability,%d in format string (no. 1) requires 'int' but the argument type is 'ssize_t {aka signed long}'.
invalidPrintfArgType_sint,drivers/tripplitesu.c:435,warning,%d in format string (no. 1) requires 'int' but the argument type is 'unsigned int'.
knownConditionTrueFalse,clients/upsmon.c:919,style,Condition 'un' is always true
knownConditionTrueFalse,drivers/gamatronic.c:111,style,Condition 'ret>=0' is always true
knownConditionTrueFalse,drivers/powercom.c:696,style,Condition 'battval>bat0' is always true
knownConditionTrueFalse,drivers/solis.c:485,style,Condition 'AppPower==0' is always false
knownConditionTrueFalse,tools/nut-scanner/scan_nut.c:180,style,Condition 'numa>=3' is always true
memleak,clients/upsset.c:664,error,Memory leak: val
nullPointerRedundantCheck,drivers/snmp-ups.c:1554,warning,Either the condition 'info_template==NULL' is redundant or there is possible null pointer dereference: info_template.
pointerLessThanZero,drivers/nut-libfreeipmi.c:864,style,A pointer can not be negative so it is either pointless or an error to check if it is.
pointerLessThanZero,drivers/nut-libfreeipmi.c:884,style,A pointer can not be negative so it is either pointless or an error to check if it is.
postfixOperator,clients/nutclient.cpp:896,performance,Prefer prefix ++/-- operators for non-primitive types.
redundantInitialization,drivers/al175.c:548,style,Redundant initialization for 'reply'. The initialized value is overwritten before it is read.
uninitMemberVar,clients/nutclient.cpp:504,warning,Member variable 'TcpClient::_timeout' is not initialized in the constructor.
uninitMemberVar,clients/nutclient.cpp:513,warning,Member variable 'TcpClient::_timeout' is not initialized in the constructor.
uninitvar,server/netssl.c:546,error,Uninitialized variable: ret
unsignedLessThanZero,drivers/apcsmart-old.c:928,style,Checking if unsigned expression 'sdtype' is less than zero.
unsignedPositive,drivers/bcmxcp.c:785,style,Unsigned expression 'commandByte' can't be negative so it is unnecessary to test it.
uselessAssignmentPtrArg,drivers/riello_ser.c:105,warning,Assignment of function parameter has no effect outside the function. Did you forget dereferencing it?


seanm:
I don't know this codebase at all, but I'm a proficient C programmer, and am pretty sure of most of these changes. A few have a temporary "Sean" comment that should especially be double-checked.

drivers/isbmex.c
```diff
@@ -175,7 +175,7 @@ static const char *getpacket(int *we_know){
   }
 
   r=read(upsfd,buf,255);
-  D(printf("%d bytes read: ",r);)
+  D(printf("%ld bytes read: ",r);)
```
jimklimov, line 178: There is no big guarantee about `(s)size_t` and `(long) int` byte sizes (I believe it is "just some integer type"), so it should be printed with `%zd` (`ssize_t`) or `%zu` (`size_t`) from C99 onwards, or cast into an `intmax_t` and `%jd`.

See e.g. https://stackoverflow.com/a/5251918 for details.

- [ ] Fix fixed printing of ssize_t

drivers/mge-utalk.c
```diff
@@ -896,19 +896,17 @@ static int mge_command(char *reply, int replylen, const char *fmt, ...)
 	}
 
 	/* send terminating string */
-	if (MGE_COMMAND_ENDCHAR) {
-		for (p = MGE_COMMAND_ENDCHAR; *p; p++) {
-			if ( isprint(*p & 0xFF) )
-				upsdebugx(4, "mge_command: sending [%c]", *p);
-			else
-				upsdebugx(4, "mge_command: sending [%02X]", *p);
+	for (p = MGE_COMMAND_ENDCHAR; *p; p++) {
```
jimklimov, line 899: @aquette @clepple @zykh : experts wanted :) In current codebase, `MGE_COMMAND_ENDCHAR` is `#define`d above as `"\r\n"`. Are there variations of the codebase with a `NULL` instead? Is there some experimental, device-dependent, etc. need for such cases? The original (removed by PR) check seems to relate to that - "if we have a string, send it" - and with the `if` removed we'd get into NULL-pointer dereferencing by `for ... *p`.

Probably both goals could be achieved by `for (p = MGE_COMMAND_ENDCHAR; p && *p; p++) {...}` if there is a chance for builds with a `MGE_COMMAND_ENDCHAR==NULL` ?

- [ ] Revise dropping of NULLness check

drivers/nut-libfreeipmi.c
```diff
@@ -861,7 +861,7 @@ int nut_ipmi_get_sensors_status(IPMIDevice_t *ipmi_dev)
 		 * if ((sensor_state = ipmi_monitoring_sensor_read_sensor_state (mon_ctx)) < 0)
 		 * ... */
 
-		if ((sensor_reading = ipmi_monitoring_sensor_read_sensor_reading (mon_ctx)) < 0)
+		if ((sensor_reading = ipmi_monitoring_sensor_read_sensor_reading (mon_ctx)) == NULL)
```
jimklimov, line 864: @aquette @clepple @zykh : do you know if there were any API changes in FreeIPMI?

The code here looks like https://fossies.org/diffs/freeipmi/1.6.2_vs_1.6.3/contrib/libipmimonitoring/ipmimonitoring-sensors.c-diff.html (and in this file a dozen lines above) where they check for `ipmi_monitoring_sensor_read_sensor_reading_type() < 0` - but indeed a pointer returned from `ipmi_monitoring_sensor_read_sensor_reading()` should not be numerically `< 0` ;)

drivers/nutdrv_qx_voltronic.c
```diff
@@ -267,7 +267,7 @@ static int	voltronic_bypass_volt_max(char *value, const size_t len)
 		if (ivn >= 200)
 			return -1;
 
-		if (protocol == 2 || protocol == 2 || protocol == 10 || protocol == 13 || protocol == 14)
+		if (protocol == 2 || protocol == 3 || protocol == 10 || protocol == 13 || protocol == 14)  // •••Sean••• not sure... a guess from the comment above
```
jimklimov, line 270: @aquette @zykh @clepple : experts wanted here :)

drivers/nutdrv_qx_voltronic.c
```diff
@@ -300,7 +300,7 @@ static int	voltronic_bypass_volt_max(char *value, const size_t len)
 		if (ivn < 200)
 			return -1;
 
-		if (protocol == 2 || protocol == 2 || protocol == 10 || protocol == 13 || protocol == 14 || protocol == 99)
+		if (protocol == 2 || protocol == 3 || protocol == 10 || protocol == 13 || protocol == 14 || protocol == 99)
```
jimklimov, line 303: @aquette @zykh @clepple : experts wanted here :)

drivers/nutdrv_qx_voltronic.c
```diff
@@ -394,7 +394,7 @@ static int	voltronic_bypass_volt_min(char *value, const size_t len)
 		if (ivn >= 200)
 			return -1;
 
-		if (protocol == 2 || protocol == 3 || protocol == 10 || protocol == 10 || protocol == 13 || protocol == 14)
+		if (protocol == 2 || protocol == 3 || protocol == 10 || protocol == 13 || protocol == 14)
```
jimklimov, line 397: @aquette @zykh @clepple : experts wanted here :)

drivers/riello_ser.c
```diff
@@ -102,7 +102,6 @@ static int char_read (char *bytes, int size, int read_timeout)
 			return -1;
 		}
 		else {
-			bytes += now;
```
jimklimov, line 105: I think this got inherited from some loop where the `bytes` pointer got incremented so we read next portions into the buffer while we can. The `read(..., size - readen)` above points to this too - with the current code, `readen` is always zero until we increment it with `now`.

Probably the current routine can be simplified even further to drop `readen` altogether, `return now` here and `return 0` in default exit below. But I do not know the legacy context behind it... @aquette @zykh @clepple : any memories on this one?

drivers/solis.c
```diff
@@ -479,7 +479,7 @@ static void scan_received_pack(void) {
 	if (!SourceFail && InversorOn)
 		OutFreq = InFreq;
 
-	if (AppPower <= 0) /* charge pf */
+	if (AppPower < 0) /* charge pf */  // •••Sean••• not sure...
```
jimklimov, line 482: Makes sense with `if( AppPower == 0 )` just a bit below.

Still...  @aquette @zykh @clepple : any memories on this one?

drivers/tripplite_usb.c
```diff
@@ -576,7 +576,7 @@ static int send_cmd(const unsigned char *msg, size_t msg_len, unsigned char *rep
 		}
 
 #if ! defined(__FreeBSD__)
-		if(!done) { usleep(1000*100); /* TODO: nanosleep */ }
+		usleep(1000*100); /* TODO: nanosleep */
```
jimklimov, line 579: Why this change? Originally we slept if not "done", now we'd delay for sleep even if "done"... any physical reason for that?

- [ ] Not sure about dropped condition check

tools/nut-scanner/scan_nut.c
```diff
@@ -177,24 +177,22 @@ static void * list_nut_devices(void * arg)
 		/* FIXME:
 		 * - also print answer[2] if != "Unavailable"?
 		 * - for upsmon.conf or ups.conf (using dummy-ups)? */
-		if (numa >= 3) {
```
jimklimov, line 180: I wonder if this originally intended to check for other constraints on `numa` (e.g. less than some limit?) and behave differently than for values too small as done above...

@aquette @zykh @clepple : any memories on this one?

jimklimov (changes requested):
Thank you very much for this PR, caught a lot of curious nuances. I marked as "Requested changes" mostly to have someone take a second look at some commented points, and possibly change or revert something in those. But generally, very much LGTM :)

clients/upsmon.c
```diff
@@ -916,10 +916,7 @@ static void addups(int reloading, const char *sys, const char *pvs,
 	/* build this up so the user doesn't run with bad settings */
 	totalpv += tmp->pv;
 
-	if (un)
-		tmp->un = xstrdup(un);
-	else
-		tmp->un = NULL;
+	tmp->un = xstrdup(un);
```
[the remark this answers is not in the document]
seanm: Indeed, as per my commit message, the warning here was:

 knownConditionTrueFalse,clients/upsmon.c:919,style,Condition 'un' is always true

drivers/gamatronic.c
```diff
@@ -108,10 +108,9 @@ int sec_cmd(const char mode, const char *command, char *msgbuf, int *buflen)
 
     if (ret < 0) return -1;
 
-    if (ret >= 0) {
 	strncpy(msgbuf, msg, ret);
 	upsdebugx(1, "UPS<--PC: \"%s\"",msg);
-    }
+
```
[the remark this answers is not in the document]
seanm: Good catch, i'll amend my commit.

drivers/isbmex.c
seanm: What is the minimum C dialect required?  i.e. can >= C99 be assumed?  I don't find any other "%z" in the codebase currently.

drivers/tripplite_usb.c
seanm: `!done` is always true here because it's a condition of the `for` loop, at line 567.

identicalInnerCondition,drivers/tripplite_usb.c:579,warning,Identical inner 'if' condition is always true.

clients/nutclient.cpp
```diff
@@ -893,7 +895,7 @@ std::string TcpClient::sendQuery(const std::string& req)throw(IOException)
 
 void TcpClient::sendAsyncQueries(const std::vector<std::string>& req)throw(IOException)
 {
-	for (std::vector<std::string>::const_iterator it = req.cbegin(); it != req.cend(); it++)
+	for (std::vector<std::string>::const_iterator it = req.cbegin(); it != req.cend(); ++it)
```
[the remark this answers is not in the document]
seanm: It's pretty idiomatic in C++, ex:

https://releases.llvm.org/10.0.0/docs/CodingStandards.html#prefer-preincrement

drivers/dstate.c
```diff
@@ -1167,7 +1167,7 @@ int dstate_detect_phasecount(
 
 		if ( (v1 && v2 && !v3) ||
 		     (v1n && v2n && !v3n) ||
-		     (c1 && c2 && !c2) ||
+		     (c1 && c2 && !c3) ||  // •••Sean••• not sure...
```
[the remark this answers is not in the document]
seanm: yes, those are bullets. the comment is not intended to ever get committed though.

drivers/isbmex.c
jimklimov: Good question. I did not find a direct statement on that. In code I see a couple of commits from 2008-2010 allowing to build on compilers with standards before C99, of those fixes at least one is still in codebase:
https://github.com/networkupstools/nut/blob/master/configure.ac#L233

But then from 2011 up to some of my 2017 commits, there were a few fixes to get this working specifically with C99+ and at least for mine nobody complained or "broke back" AFAIK.

clients/nutclient.cpp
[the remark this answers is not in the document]
jimklimov: ack, thanks

drivers/dstate.c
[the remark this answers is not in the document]
aquette: @jimklimov aren't you the author of that code?! ;-p
confirming that @seanm is right, for dual/split phase UPS, we have x1 && x2 (x being voltage, current, for phase 1 and 2), and not x3 for phase 3, otherwise, it would be a 3 phase unit

drivers/isbmex.c
aquette: C99 is the last REQ we expressed
not sure where we said so, probably a volatile mail on the ML, plus the configure check AC_PROG_CC_C99.
That relates to old Unix systems, that we still (?) support. A refreshed audit of the situation here would be good

drivers/mge-utalk.c
aquette: good to me.
Not sure why we put that?! part of the very early work, back around 2001

drivers/nut-libfreeipmi.c
aquette: never looked back at it, since I did this PoC 10 years ago, or so. I'm even unsure about the interest to keep it, like some of my others exploratory attempts (HAL support).
To answer still: indeed, there seems to be a "int ipmi_monitoring_sensor_read_sensor_reading_type()" and a "void *ipmi_monitoring_sensor_read_sensor_reading()"
good for me

drivers/nutdrv_qx_voltronic.c
aquette: @seanm guess seems good. @zykh  would be the best judge here, as Qx expert

drivers/nutdrv_qx_voltronic.c
aquette: ditto above @zykh

drivers/nutdrv_qx_voltronic.c
aquette: ditto above @zykh

drivers/solis.c
aquette: never worked on solis

aquette (approved):

drivers/nut-libfreeipmi.c
jimklimov: FYI : As found during experiments about C standard support, we also have a similar failure nearby:

````
/usr/lib/ccache/gcc -DHAVE_CONFIG_H -I. -I../include    -I../include  -I/usr/include/neon   -I/usr/include/modbus -Wall -Werror -pedantic -std=gnu99 -Wall -Wsign-compare -MT nut-libfreeipmi.o -MD -MP -MF $depbase.Tpo -c -o nut-libfreeipmi.o nut-libfreeipmi.c &&\
mv -f $depbase.Tpo $depbase.Po
nut-libfreeipmi.c: In function 'nut_ipmi_get_sensors_status':
nut-libfreeipmi.c:864:79: error: ordered comparison of pointer with integer zero [-Werror=pedantic]
   if ((sensor_reading = ipmi_monitoring_sensor_read_sensor_reading (mon_ctx)) < 0)
                                                                               ^
nut-libfreeipmi.c:884:95: error: ordered comparison of pointer with integer zero [-Werror=pedantic]
   if ((sensor_bitmask_strings = ipmi_monitoring_sensor_read_sensor_bitmask_strings (mon_ctx)) < 0)
                                                                                               ^
Makefile:1545: recipe for target 'nut-libfreeipmi.o' failed
make[1]: *** [nut-libfreeipmi.o] Error 1
````

drivers/isbmex.c
jimklimov: I played on my workstation with Debian Linux (started life as v8, but updated since so mixed-version) and tried different `-std=...` flags to GCC, like this:

````
:; make distclean ; PATH=/usr/lib/ccache:$PATH CC=/usr/lib/ccache/gcc CXX=/usr/lib/ccache/g++ \
    ./configure --with-cgi CFLAGS="-std=gnu99" CXXFLAGS="-std=gnu++99" && \
    (make -j8 all -k 2>/dev/null 1>&2) ; echo ======================================== ; make all -k
````

It failed with all strictly-C flags (`-ansi` aka `-stc=c89`, c90, c99, c11) mostly due to Linux system headers not really working with that, missing some basic routine definitions or having confusing parameter expectations as a result.

It fared better for GNU C dialects, though in pedantic mode exposed some issues like those you fixed and more (printing into buffer too small, variadic macros requiring at least one argument, unused vars, possibly botched logical conditions where parentheses may be due...) with all dialects, so we should fix those eventually. Beside that...

For `-std=gnu90` it reported a few offenses seemingly easy to fix:
* declarations inside `for` like `for (int i=0; ...)` in solis.c
* BUT quite a few warnings so binary code might not work right in the end:
** format string mismatches that your PR probably fixes
** format chars introduced in later standard versions
** initializer element is not computable at load time
** ISO C90 forbids mixed declarations and code
** undefined `__func__` interpreted as `int not %s` for logging => probably would segfault at runtime
** C++ style comments with `//` in a few spots
** (pedantic) comma at end of enumerator list

...so while restoring compatibility with C90 (GNU dialect at least) may be possible, it is a considerable effort and not sure it is worth doing unless someone actually suffers from the lack. So far nobody complained AFAIK.

For `-std=gnu++11` there is a deprecation warning for
````
In file included from nutclient.cpp:20:0:
nutclient.h:154:79: error: dynamic exception specifications are deprecated in C++11 [-Werror=deprecated]
  virtual void authenticate(const std::string& user, const std::string& passwd)throw(NutException)=0;
````


````
:; make distclean ; PATH=/usr/lib/ccache:$PATH CC=/usr/lib/ccache/gcc CXX=/usr/lib/ccache/g++ \
    ./configure --with-cgi CFLAGS="-Wall -Werror -pedantic -std=gnu99" CXXFLAGS="-Wall -Werror -std=gnu++99" && \
    (make -j8 all -k 2>/dev/null 1>&2) ; echo ======================================== ; make all -k
````

seanm: OK, since >= C99 is required I'll switch to %z.

drivers/dstate.c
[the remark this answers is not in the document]
seanm: OK, with that confirmation, I'll remove the temporary comment.

drivers/riello_ser.c
seanm: Indeed this looks leftover from some old refactoring.  I'm inclined not to attempt any simplifications here.

drivers/isbmex.c
jimklimov: With #823 / #824 the current NUT master should test a dozen more combos (relatively arbitrarily chosen) on Travis to aid in code cleanup efforts like this. Thanks @seanm for triggering this ;)

seanm:
Friendly ping...

tools/nut-scanner/scan_nut.c
jimklimov: Anyway, this change does simplify code and does not change its practical behavior, so okay :)

drivers/nutdrv_qx_voltronic.c
jimklimov: At least, the original line was apparently bogus :) So if there was some error introduced here, at least @zykh was pinged and would know where to look :-p

seanm:
@jimklimov I could always remove the changes that haven't been approved so that we could at least merge some of this. perhaps I shouldn't have added so much to one PR...

jimklimov:
Nay, let's leave them be - and thanks a lot! I revisited this PR today, the changes around remaining questions looked reasonable after some more review of the context, and/or the original state was too bogus to keep. So on my side nothing blocks it at the moment.

This also spurred the completion of CI setup to test a lot of compilers, OSes and standards to catch more warnings coming from different perspectives, so thanks for that too :)

jimklimov:
A version of this PR's changes (plus some from "fightbugs" branch) looks decent also on the BuildBot farm covering more Linux distros and BSDs. There is work left, but at least builds pass green and I think it warns a bit less than before :)

seanm:
Oh, I really didn't mean/want for this to be merged as-in, hence the "WIP" in the title.  I had lots of temporary comments in there for discussion purposes only.

seanm:
Can you revert the merge?

jimklimov:
Hm, probably could do, but maybe better now would be to overlay with next fixes and/or comment removals?

drivers/nut-libfreeipmi.c
jimklimov: Ah, that one is fixed in this PR too, thanks :)

Issue #439, opened 2017-06-04
CPS: patch HID Report Descriptor to fix "output.voltage"


clepple:
Observed in issue https://github.com/networkupstools/nut/issues/337#issuecomment-257284089
and on [nut-upsuser](http://lists.alioth.debian.org/pipermail/nut-upsuser/2017-June/010685.html).

Report ID 18 seems to be inheriting Logical Minimum/Logical Maximum from Report ID 16:
```
Collection(Physical)
    Report ID(16)
        Usage(LowVoltageTransfer)
        Logical Minimum(97)
        Logical Maximum(103)
        Feature(Data,Variable,Absolute)

        Usage(LowVoltageTransfer)
        Input(Constant,Variable,Absolute)

        Usage(HighVoltageTransfer)
        Logical Minimum(136)
        Logical Maximum(142)
        Feature(Data,Variable,Absolute)

        Usage(HighVoltageTransfer)
        Input(Constant,Variable,Absolute)

    Report ID(18)
        Usage(Voltage)
        Feature(Constant,Variable,Absolute)

        Report Size(8)
        Logical Minimum(0)
        Logical Maximum(255)
        Unit(SI Linear: ...)
        Unit Exponent(0)
```

I'm pretty sure this is not a bug in the NUT HID parser, but feel free to correct me. (There are other places in the CPS HID Report Descriptor that seem to mix up the locations of the parser tokens relative to Input/Output/Feature.)

clepple:
Original HID Report Descriptor patcher issue is https://github.com/networkupstools/nut/issues/169

csurf:
what's the status of this issue? I'm experiencing the same thing for a CyberPower CP1000PFCLCD UPS. Both the output voltage & battery voltage are incorrect.

CorvetteCole:
I would also love to hear if anyone has looked in to this further

serfriz:
Any update on this? My CyberPower CP1500PFCLCD shows an output.voltage of 142.0

arlenarlenarlen:
👋🏼  having the same issue here with my CP1500PFCLCD
`Path: UPS.Output.Voltage, Type: Feature, ReportID: 0x12, Offset: 0, Size: 16, Value: 136`

slyticoon:
Having this issue with CP1500AVRLCD. Any updates or workarounds?

ArcticZero:
[...]

PR #811, opened 2020-08-11
CyberPower UT series do not seem to pass correct 'OB' flag, try to accommodate this with new "onlinedischarge" config flag

kgizdov:
As the title says. I have `CPS UT1500E` and unfortunately whatever I do, I cannot get it to show correct online status. I have been in contact with support and they say the UPS is operating as it should. Thus, I have arrived at this workaround.

The issue is as follows, when the UPS loses mains power, it goes from `OL` into `OL DISCHRG` mode. Yes, it does not change it's input mode. However, it correctly shows input voltage to be `0`. The only way to circumvent this I have found to be encapsulated in this PR.

I am, however, open to suggestions on how to make it better as it clearly only applies to certain types of UPSes. I was thinking to filter on `VRANGE`, but that somehow seemed inappropriate to me.

Let me know if you require further info or testing and/or logs. Maybe we can find the actual underlying reason this happens and fix it at the core. Cheers.


jimklimov:
There is a bit of code-style issue in the initial commit (indentation should start with TABs until the level of `if` on the previous line).

Given that this misbehavior may be not present on all devices supported by the driver, I think it makes sense to define (and document!) a driver option for the "quirk" similar to what was done in https://github.com/networkupstools/nut/pull/788/files - then it is up to end-users suffering same as you to toggle this for their deployment or stay with current behavior. 

It may make sense to log or even post ("wall?") the situation, like "We are in a strange OL DISCHRG situation at the moment, do you need the XXX quirk in your driver to handle this as an outage?"

Thinking out loud, an UPS calibration event might also seem like this validly - the device still has wall power, but is intentionally discharging the battery to see how long it lasts with your load.

drivers/usbhid-ups.c
```diff
@@ -1422,7 +1422,8 @@ static void ups_status_set(void)
 		dstate_delinfo("input.transfer.reason");
 	}
 
-	if (ups_status & STATUS(ONLINE)) {
+	if ((ups_status & STATUS(ONLINE)) &&
+                !(ups_status & STATUS(DISCHRG))) {
```
jimklimov, line 1426: As mentioned by PR author and in later comments, this code change *probably* should not impact everyone silently, so needs a config toggle as a quirk for certain devices, to be set at users' discretion.

jimklimov (changes requested):

kgizdov:
I agree with everything said. I will fix the indentation and will try to refactor as requested. For your last comment, yes, we should detect the UPS calibration and if we cannot, we should issue a warning like "you are in this weird state, if you are not in a calibration mode, maybe you need this setting: bla". How's that sound?

jimklimov:
Sounds right. See also #817 merged recently that dealt with similar question (whether calibration is an outage).

kgizdov:
obviously not ready, but is this something in the right direction?

drivers/usbhid-ups.c
```diff
@@ -1422,8 +1432,22 @@ static void ups_status_set(void)
...
-		status_set("OL");		/* on line */
+
+	if ((ups_status & STATUS(ONLINE)) && (ups_status & STATUS(DISCHRG))) {
+		/* warning or CyberPower UT quirk */
+		if (cputquirk) {
+			status_set("OB");
+		} else if ((ups_status & STATUS(CAL))) {
+			status_set("OL");
+		} else {
+			upslogx(LOG_WARNING, "%s: seems that UPS [%s] is OL+DISCHRG state now. "
```
jimklimov: English: ups "has" or "is in" some state

drivers/usbhid-ups.c
```diff
@@ -1422,8 +1432,22 @@ static void ups_status_set(void)
 		dstate_delinfo("input.transfer.reason");
 	}
 
-	if (ups_status & STATUS(ONLINE)) {
-		status_set("OL");		/* on line */
+
+	if ((ups_status & STATUS(ONLINE)) && (ups_status & STATUS(DISCHRG))) {
```
jimklimov: Given that below you check for part of the same condition, maybe fuse them together?
````
if (ups_status & STATUS(ONLINE)) {
    if (ups_status & STATUS(DISCHRG)) {
        ... lots of new logic about quirk, calibration, message ...
    } else {
        status_set("OL");
    }
} else {
    status_set("OB");
}
````

File: drivers/usbhid-ups.c [hunk 4]
```diff
@@ -114,6 +114,13 @@ bool_t use_interrupt_pipe = FALSE;
 #endif
 static time_t lastpoll; /* Timestamp the last polling */
 hid_dev_handle_t udev;
+/**
+ * CyberPower UT series sometime need a bit of help deciding their online status.
+ * This quirk is to enable the special handling of OL & DISCHRG at the same time
+ * as being OB (on battery power/no mains power)
+ */
+#define DEFAULT_CPUTQUIRK 0
+static int cputquirk = DEFAULT_CPUTQUIRK;
```
jimklimov, line 123: I do not like the name, primarily as it may be misunderstood as some CPU-related quirk: "What is a CPU T? Temperature?"

And also it is a sort of finger-pointing to a "person" of vendor/model rather than the action this quirk does (and which might apply to other vendors/models eventually, if the problem proves to be more widespread it could become a more global option across drivers, potentially).

Thinking of it, maybe not a boolean but something descriptively named and extensible like `onlinedischarge="[OL|OB]"` might fit better?

jimklimov (changes requested):
General direction looks good, some comments to code posted.
Also lacks a change to manpage, to document the new driver option for the quirk.

drivers/usbhid-ups.c
kgizdov: typo, fixed

drivers/usbhid-ups.c
kgizdov: fixed

drivers/usbhid-ups.c
kgizdov: I changed the name, but I'm not exactly sure about how to use the framework to make a proper named option. Could you point me at an example somewhere? Thanks.

kgizdov:
@jimklimov where would it be appropriate to explain in the manual structure. I can see the general config notes as well as the general hid-driver man. Not sure if either is a great place as this seems specific. Thoughts?

drivers/usbhid-ups.c
```diff
@@ -705,6 +712,10 @@ void upsdrv_makevartable(void)
 
 	addvar(VAR_FLAG, "pollonly", "Don't use interrupt pipe, only use polling");
 
+	snprintf(temp, sizeof(temp), "Treat discharging while online as being offline (default=%s)",
+		DEFAULT_ONLINEDISCHARGE);
```
jimklimov, line 716: Per https://travis-ci.org/github/networkupstools/nut/jobs/734837970 :
````
usbhid-ups.c: In function 'upsdrv_makevartable':

usbhid-ups.c:715:31: warning: format '%s' expects argument of type 'char *', but argument 4 has type 'int' [-Wformat=]

  snprintf(temp, sizeof(temp), "Treat discharging while online as being offline (default=%s)",
                               ^

usbhid-ups.c:715:31: warning: format '%s' expects argument of type 'char *', but argument 4 has type 'int' [-Wformat=]
````

File: drivers/usbhid-ups.c [hunk 6]
```diff
@@ -1422,10 +1433,25 @@ static void ups_status_set(void)
...
+		if (onlinedischarge) {
+			/* if we treat OL+DISCHRG as being offline */
+			status_set("OB");	/* on battery */
+		} else {
+			if (!(ups_status & STATUS(CAL))) {
+				/* if in OL+DISCHRG unknowingly, warn user */
+				upslogx(LOG_WARNING, "%s: seems that UPS [%s] is in OL+DISCHRG state now. "
+				"Is it calibrating or do you perhaps want to set 'onlinedischarge' option? "
+				"Some UPS models (e.g. CyberPower UT series) emit OL+DISCHRG when offline.",
+				  __func__, ups->upsname)
```
jimklimov, line 1450: Per https://travis-ci.org/github/networkupstools/nut/jobs/734837970 :
````
usbhid-ups.c: In function 'ups_status_set':

usbhid-ups.c:1450:17: error: 'ups' undeclared (first use in this function)
       __func__, ups->upsname)
                 ^

usbhid-ups.c:1450:17: note: each undeclared identifier is reported only once for each function it appears in

usbhid-ups.c:1451:4: error: expected ';' before '}' token

    }
    ^
````

jimklimov (changes requested):
A couple of blocking build issues appeared in recent iterations, commented.

jimklimov:
As for a good place to document the new driver option (added after I initially posted the reference to #788), I think a driver manpage could be good. In case of that PR, its origin could not be modified so a later commit was pushed to master adding the manpage changes for that: https://github.com/networkupstools/nut/commit/799fd5a3a05e5ffe71f652d29134262f5c49193e

For this PR, the suitable manpage at this time would probably be the USB family's common location: https://github.com/networkupstools/nut/blob/master/docs/man/usbhid-ups.txt

jimklimov:
@kgizdov : Just in case: is this PR's problem area related to #439 (as it evolved in the end) and #1245 as the generic solution to allow hotfixing known invalid mappings?

kgizdov:
oh, sorry, this MR has completely gone out of my schedule somehow as I jumped into some other work. Allow me a couple of week to have a look. But I think if you've got a more general solution to this then by all means that's the end goal. Let me know if there's anything specific (code-wise) and should be aware of to incorporate or completely refactor the changes.

jimklimov:
Hi, thanks for the reply. I think, since #1245 was by now merged, it would help if you review its change so that we can decide if your original fix is needed or your goal is achieved by that one, and/or if your solution can be adapted to use that mechanism.

jimklimov:
Brushed up the PR, it seems to have merit regardless of "general hotfix for invalid mappings". At worst, we would have two solutions for one of the problems :)

jimklimov:
Thanks for the contribution!

PR #817, opened 2020-09-22
Do not shutdown if UPS is performing calibration

overhacked:
Some UPS calibration procedures (e.g. CyberPower) involve
discharging the battery below the LB level, but we don't
want systems to shut down if the UPS is only on battery
because it is calibrating.

### DIFF
--- a/clients/upsmon.c
+++ b/clients/upsmon.c
@@ -658,8 +658,17 @@ static int is_ups_critical(utype_t *ups)
 	/* must be OB+LB now */
 
 	/* if UPS is calibrating, don't declare it critical */
-	if (flag_isset(ups->status, ST_CAL))
+	/* FIXME: Consider UPSes where we can know if they have other power
+	 * circuits (bypass, etc.) and whether those do currently provide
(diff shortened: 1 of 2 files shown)


clients/upsmon.c
```diff
@@ -657,6 +657,10 @@ static int is_ups_critical(utype_t *ups)
 
 	/* must be OB+LB now */
 
+	/* if UPS is calibrating, don't declare it critical */
+	if (flag_isset(ups->status, ST_CAL))
+		return 0;
```
jimklimov, line 662: Would it make sense to log this situation, especially to help troubleshoot when/if this backfires (e.g. we had an authentic outage during calibration, really ran out of power and crashed ungracefully - and here is the log message why NUT did not shut down the system).

Are there more factors to check (e.g. does an UPS have a bypass, is it live or un-powered during the calibration)?

jimklimov: Added the logging bit myself :)

jimklimov (approved):

PR #820, opened 2020-10-06
Driver datadump improvements

aquette:
Do not switch to statepath, nor write pidfile, when using the
datadump (-s -d) mode

### DIFF
--- a/drivers/main.c
+++ b/drivers/main.c
@@ -607,9 +607,9 @@ int main(int argc, char **argv)
 
 	become_user(new_uid);
 
-	/* Only switch to statepath if we're not powering off */
+	/* Only switch to statepath if we're not powering off or just dumping data, for discovery */
 	/* This avoid case where ie /var is umounted */
-	if ((!do_forceshutdown) && (chdir(dflt_statepath())))
+	if ((!do_forceshutdown) && (dump_data < 0) && (chdir(dflt_statepath())))
 		fatal_with_errno(EXIT_FAILURE, "Can't chdir to %s", dflt_statepath());
 
 	/* Setup signals to communicate with driver once backgrounded. */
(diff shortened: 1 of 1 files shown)


jimklimov:
Comments near writepid say about "if we are not backgrounding" - is the avoided dstate_init() "to not serve" below this very backgrounding? Or should there be a fix elsewhere for the ddumping-only mode?

aquette:
@jimklimov this patch allows to not have to launch drivers in datadump mode with root privileges, conversely to when starting drivers for monitoring through eg systemd.
dstate_init() and writepid() require that the statepath is accessible RW (to create the communication socket with upsd and pidfile) which will not be when launching as simple user.

jimklimov:
Well, the primary intent of my question was I guess "is there any un-handled loophole" and "did you test it in practice" (and with the details you added later - running as a non-root user, assuming that it has access if needed to the media device node in the OS for physical connections).

Out of curiosity and for sanity-check - this instance of the driver process is not visible to, and does not interact with, `upsd` on the system? 
Does it "background" or stays in the user's foreground console to dump the data it sees? 
Does it loop until killed, or collects data once (or N times) and exits by itself? 
Is there any room for conflict and surprises if it runs along a "production" instance of the same driver (probably there will be problems on physical medias since only one of these would get the bits from the wire; so the question is more about conflict-less running of two copies of same driver against same networked UPSes/ePDUs).

If it is safe on your second thought about it, then LGTM here :)

aquette:
this is what we use for advanced discovery, with fty-disco*
do x update loops, then print out upsc-like output then exit:
https://github.com/networkupstools/nut/blob/73c8644d53d7dbaae46ec9d3b6bbee9ad23f9f70/drivers/main.c#L741

tested with root and non root, variations of calls (like -a -d, dump & exits), good to me.

PR #824, opened 2020-10-08
#823 : Set up Travis CI testing for more compiler implementations and…

jimklimov:
… C/C++ standards

Issue #823, opened 2020-10-08
Set up Travis CI testing for more compiler implementations and C/C++ standards


jimklimov:
As discussed in https://github.com/networkupstools/nut/pull/810#discussion_r501679692 we have a number of coding issues hidden into warnings, ultimately different ones if we variate the requested C/C++ standard implementation. Currently we build on Travis CI with whatever compiler and its options are the default for distribution Travis offers at the time of build, and without hiccups on warnings.

The idea proposed here is to utilize Travis CI ability to run jobs allowed to fail (so no direct impact on passing-or-not CI tests) with our matrix of `CC`, `CXX`, `CFLAGS` and `CXXFLAGS` definitions to test with different versions of gcc, clang, maybe others and with different C standards inside that, with and without required failure on warnings.

When a target combo becomes green, we can mark it as no longer allowed to fail, to ensure that future contributions do not break it again (non-regrassion).

Overall this should make it easier for developers to update the NUT codebase and see potential issues introduced by their changes, without noise from already existing problems that this effort aims to ultimately eradicate. This would allow us as a community to be more confident about the quality of code even before we get it to testing against real hardware devices, media, protocols and third-party libraries and OS code involved. We'd be doing the best we can, on our side of equation :)

jimklimov:
CI Enhanced.

jimklimov:
For posterity, the mechanism added with PR #824 allows to track on Travis CI that new contributions do not break existing working platform cases and whether they fix cases currently broken (then we can comment away the test setup's line in `allowed_failure` block).

For developers with environment set up to compile NUT, the `env` block definitions of the case can be copy-pasted to terminal (and adapted for different compiler flags or versions) so they can check on their workstation what works and what does not, using the same `ci_build.sh` script, e.g.:
````
:; BUILD_TYPE='default-all-errors' CFLAGS="-std=gnu11 -Werror -Wall -pedantic" CC=clang ./ci_build.sh
````

This would help extend the non-regression platform support as more compilers are getting tested. One obvious and useful case not covered now could be adding builds on MacOS.

Per recent discussions, the codebase should work in a C99+ capable compiler, and there may have been some project communication in the past about that, although nothing indicates that requirement/expectation in current codebase except one `configure.ac:AC_PROG_CC_C99` line.

[...]

PR #825, opened 2020-10-09
Reshuffle Travis CI matrix of allowed_failure jobs

jimklimov:
Follows up for PR #824 answering to issue #823 to bring green (nowarn) jobs into the realm of configs no longer allowed to fail, and fix a minor issue with `solis.c` driver to have at least no-warn builds of NUT in gnu89 mode too.

This also documents in `developers.txt` that we support C99 at this time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -383,7 +383,7 @@ matrix:
 # Note: "env" lines below must exactly describe a matrix option defined above
   allow_failures:
   - env: NUT_MATRIX_TAG="cDefault-gcc-default-nowarn" BUILD_TYPE=default-all-errors CFLAGS="-std=gnu99" CXXFLAGS="-std=gnu++99"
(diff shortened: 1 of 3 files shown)


jimklimov:
FYI to NUT Team members: I added "Branch protection" to "master" in GitHub settings, so PRs would generally require to succeed in CI and be based against latest master. Technically this is to make the "Update Branch" (This branch is out-of-date with the base branch) button appear so that sufficiently privileged reviewers can push updated Travis code into the original branches of PRs being evaluated for merging.

Team admins should be still allowed to merge PRs that are not "Updated" or green-lit by CI, allowing to quickly merge a series of pull requests that are assumed good and not conflicting each other, without waiting for CI rebuilds after each merge. 

A rule of thumb for quick merges with admin privileges is that the top list of Travis job configurations not allowed to fail should be successful (including the spellcheck job for documentation changes).

PR #826, opened 2020-10-09
docs/developers.txt : document coding style examples for tabs vs spaces

jimklimov:
I suppose this would cause a fair bit of bikeshedding as any style doc does, so put it as a separate PR :) 
But AFAIK, this just spells out the existing precedent so should not stall for too long.


aquette (approved):
Lgtm

PR #829, opened 2020-10-10
More flexible Travis CI matrix test options

jimklimov:
Follow-up for #823 : adding more options for building with different compilers, C standards and target platforms (adding MacOS X and Windows potentially, or different CPU architecture builds).

Separated the numerous build cases into "yaml includable" blocks that developers can combine into a job matrix for running their branches - e.g. when someone gets to work on fixing cases that are far in the red now, such as MacOS X or Windows builds that are costly to set up for the Travis infra team and so are a limited resource we should not waste by default in the master branch - not until those cases get fixed by contributors eventually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,11 @@ os:
 
 sudo: false
 
-# Tests for cppunit require C++11 which requires gcc-4.8 or newer
-# This is available in either "trusty" or newer distros, or in "docker" envs
-#dist: trusty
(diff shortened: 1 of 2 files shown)


jimklimov:
PR #831 would aim to block by default the test cases that are not expected to pass, to conserve Travis CI resources by default. It would be posted separately in order to keep an easily accessible build history point with all test cases enabled, to help conjure up solution to failing cases.

PR #830, opened 2020-10-12
docs/acknowledgements.txt: revise the last paragraph for listing corporate contributors

jimklimov:
Another potential bike-shedding moment, so posting the small change as a separate PR.

### DIFF
--- a/docs/acknowledgements.txt
+++ b/docs/acknowledgements.txt
@@ -167,7 +167,12 @@ and a UPStation GXT2 with the Web/SNMP card for development of the liebert
 driver and expansion of the existing snmp-ups driver.
 Liebert has since been acquired by link:http://www.emerson.com[Emerson].
 
-NOTE: If a company or individual isn't listed here, then we probably don't have
-enough information about the situation. Developers are requested to report vendor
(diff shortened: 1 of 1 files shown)


jimklimov:
One point to consider going forward - how much we prefer the mailing list ("send us some mail") vs. raising issues and PRs on GitHub? I'd rather drop the suggestion to "send mail" so it is less likely we'd miss the message.

aquette (approved):

aquette:
Also something that was on my todo:
Going GH is better imo nowadays

clepple:
I think GitHub is certainly more efficient for code review of larger changes, but consider the user who is not a developer: below a certain size threshold, it is not worth their time to read the TOS and sign up for GitHub (it would be a little different if this were a self-hosted GitLab instance maybe, but I’m not necessarily suggesting that would be better for us). I’ve filed issues for things brought up on the mailing list in the past, and have no problem continuing to do so (time permitting). After all, if I need to research a NUT issue, I’m going to go to my email archive first, not search GitHub.

clepple (approved):

jimklimov:
Ok, keeping the dual mail-github wording here then :)

Issue #835, opened 2020-10-21
2.7.4.1 e24b9a84 fails to connect to see UPS


cschuber:
FreeBSD 13-CURRENT
ups.model: UPS RF1025      
ups.productid: 0501

Config that worked correctly under 2.7.4 and under 2.7.4.1 e24b9a84 but now causes an error under 79a2f758:

listening on 10.1.2.254 port 3493
listening on 10.1.1.254 port 3493
listening on 127.0.0.1 port 3493
Can't connect to UPS [cps] (usbhid-ups-cps): No such file or directory

Same configuration:

[cps]   
        driver = usbhid-ups
        port = auto
        # pollfreq = 5
        pollfreq = 30
        desc = "cwfw"

The ups matches this in the nut supplied devd.conf:

#  Dynex DX-800U?, CP1200AVR/BC1200D, CP825AVR-G, CP1000AVRLCD, CP1000PFCLCD, CP
1500C, CP550HG, etc.  - usbhid-ups
notify 100 {
        match "system"          "USB";
        match "subsystem"       "DEVICE";
        match "type"            "ATTACH";
        match "vendor"          "0x0764";
        match "product"         "0x0501";
        action "chgrp uucp /dev/$cdev; chmod g+rw /dev/$cdev";
};

cschuber:
usbhid-ups is dying on a bus error at dstate.c:583.

Program received signal SIGBUS, Bus error.
0x000000000021faa2 in dstate_poll_fds (timeout=..., extrafd=-1) at dstate.c:583
583		FD_SET(sockfd, &rfds);
(gdb) bt
#0  0x000000000021faa2 in dstate_poll_fds (timeout=..., extrafd=-1)
    at dstate.c:583
#1  0x000000000021efbb in main (argc=0, argv=0x7fffffffea48) at main.c:749
(gdb) p sockfd
$1 = -1
(gdb) p rfds  
$2 = {__fds_bits = {0 <repeats 16 times>}}
(gdb) p &rfds
$3 = (fd_set *) 0x7fffffffe5b8
(gdb)

clepple:
@aquette I'm thinking there might be a typo in 73c8644d5

`dump_data` seems to be initialized to 0, but this checks for `< 0` before calling `dstate_init()`:

[...]

PR #836, opened 2020-10-22
Fix dstate init call WRT data dump

aquette:
Closes: https://github.com/networkupstools/nut/issues/835


jimklimov (approved):
Given the discussion in issue, LGTM. I'm not where I can see the codebase well, so... are these all the use-cases of that flag?

aquette:
@jimklimov not sure what you meant by your "all the use cases" statement.
I tested this time more extensively the data dump and standard modes, and it's fine.
As for improvements, we may want to warn when using -a (classic daemon mode) with -d (datadump), since this last will exit fast. But one may still want to just test a configuration, prior to invoking systemd.

PR #838, opened 2020-10-26
snmp-ups: Eaton ePDU outlet management improvements

aquette:
This includes:
* improved visibility on whole-unit and per-outlet switchability visibility
* fix on outlet identifier

### DIFF
--- a/docs/nut-names.txt
+++ b/docs/nut-names.txt
@@ -488,6 +488,8 @@ of the user manual.
 |===============================================================================
 | Name                           | Description                 | Example value
 | outlet.count                   | Total number of outlets     | 12
+| outlet.switchable              | General outlet switch ability
+                                   of the unit (yes/no)        | yes
 | outlet.n.id                    | Outlet system identifier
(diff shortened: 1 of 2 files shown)


jimklimov (approved):
As you are holding the pulse of the MIB and HW development, or those guys are nearby I suppose, I'd trust the changes.

One matter for @clepple and @zykh to speak up on, as in similar cases before, is to agree on introduced new "nut-name" `outlet.switchable`.

Would it be fair to merge in, say, a week - if no vetoes appear?

clepple (commented):
I don't use NUT's PDU support enough to have a strong opinion on the changes. The new variable seems okay, I guess, though it duplicates information in the individual `outlet.n.switchable` variables (and NUT doesn't provide much between getting one or all of the variables).

In the future, it would probably be better to have `ST_FLAG_BOOLEAN` or similar to cover yes/no variables, but we have a few of these in the codebase already - so I don't want to hold this PR up over that.

Issue #833, opened 2020-10-16
Verify that changing sys/termios.h to termios.h does not change behavior on *BSD


clepple:
FreeBSD:

    /usr/include/sys/termios.h:3:2: warning: "this file includes <sys/termios.h> which is deprecated, use <termios.h> instead" [-W#warnings]

This shouldn't be a big deal to fix, but I want to make sure that the auto-detection works as expected, and the new header doesn't cause unintended behavior.

jimklimov:
I suppose it can be a configure-time check (e.g. clean compile without warnings), so in C sources we include one or another pathname based on the ifdef'ed flag macro.

cschuber:
I can probably look at that.

clepple:
@cschuber that would be awesome, thanks!

To be honest, I hadn't looked at the code for this in a while, and it turns out we already have `#if defined(HAVE_SYS_TERMIOS_H)`. (So my earlier comment about auto-detection is not well informed.)

The litmus test is probably to simplify [serial.h](https://github.com/networkupstools/nut/blob/master/drivers/serial.h#L8-L12) to just `#include <termios.h>` and make sure that a serial driver still works on BSD. (I am optimistic - seems like we needed `speed_t` which [POSIX.1](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/termios.h.html) says should be in `termios.h`).

@jimklimov minimal warnings is a reasonable goal, but I don't think we want to put the cart before the horse and say that builds need zero warnings (a la `-Werror`) to be successful. If `sys/termios.h` is deprecated, and its removal does not cause side effects, let's take that opportunity to simplify.

cschuber:
I'll try to carve some time to look at it and base a patch on 2.7.4 (not master, as master usbuhid-ups bus faults on me since my update to head of master yesterday).

cschuber:
The easy fix would be to check for FreeBSD in serial.h (either include sys/param.h and check for __FreeBSD_version or __FreeBSD__) but that only puts a band-aid over the problem. Ideally a check for termios.h should be made and if it's found do not check for sys/termios.h in configure.ac.

I'm not willing to create a separate patch for the port because once sys/termios.h is removed it will revert to termios.h. IMO, as a FreeBSD developer and maintainer of the FreeBSD nut ports, this is not a serious issue. The OP can temporarily remove sys/termios.h to verify that this is the case.

cschuber:
[...]

PR #842, opened 2020-11-04
Revise termios.h and detection of ability to build serial drivers

jimklimov:
Follows up from #833 discussion.

CC @cschuber and @clepple


jimklimov:
Per http://buildbot.networkupstools.org/public/nut/builders/FreeBSD-x86/builds/937 and http://buildbot.networkupstools.org/public/nut/builders/OpenBSD-x64/builds/793 this state of codebase no longer has problems with sys/termios.h and builds on systems with either variant, quietly.

This also improves detection, if only basic, whether the current host CAN build serial drivers, to help make the choice during `./configure --with-serial=auto` runs.

PR #844, opened 2020-11-08
Fightwarn groundwork

jimklimov:
This PR follows up from #823 and is part of a larger effort to detect and fix warnings issued by newer compilers and exposing issues in NUT codebase; it just lays the groundwork for subsequent PRs that would propose fixes for different groups of problems (logic, maintainability, overflow, ...) detected.

It also enables optional Travis CI builds for the larger test matrix for branches whose names contain  "fightwarn" (or PRs targeted at branches so named) so that usual builds of NUT and its pull requests would not waste CI resources predictably-in-vain.

Over time as some test cases do become "green" and are moved to not being optional successes, this optional buildability would impact less and less cases, hopefully.

One practical point worth mentioning to the team is the addition of `developers.txt` section below on using `goto` as an explicit way to implement `switch` case fall-through; after several attempts, seems this is the only portable way to achieve that (and also quite self-documenting). Subsequent PRs would post the fixes for existing fall-through statements; for most of then the intention was quite apparent, but a few I'm not sure about - these would be posted separately to not block merging the main bunch.

This PR I intend to merge with little need for discussion, maybe within a week. Still, review and approval would be warmly appreciated :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -838,9 +838,9 @@ _matrix_master:
 ### Enabled for branches with names containing "fightwarn" to perform
 ### builds that intend to actively fix the issues which preclude the
 ### following items from becoming green:
-#  - *_matrix_allowfail_linux
-#  - *_matrix_allowfail_osx
-#  - *_matrix_windows
+  - *_matrix_allowfail_linux
+  - *_matrix_allowfail_osx
+  - *_matrix_windows
 
 _matrix_fixbugs:
   include: &_matrix_fixbugs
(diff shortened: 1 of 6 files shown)


jimklimov:
To fetch the entirety of "classified" PRs mentioned above along with this one, grab https://github.com/networkupstools/nut/tree/fightwarn (current state is in commit https://github.com/networkupstools/nut/commit/a31977eb0b5002cb1f0d08e849c55b67dbebcbf7).

Note that at this time there is one known issue without a solution proposed, raised in #850.

PR #846, opened 2020-11-08
Fightwarn fallthrough questionable - upscode2

jimklimov:
* drivers/upscode2.c: comment the fallthrough in upsc_simple() switch, so far consider it explicitly desired as it was before


clepple (commented):
No idea, but this driver has been around for a while, so I suspect someone would have noticed if the behavior was not correct.

jimklimov:

PR #847, opened 2020-11-08
Fightwarn fallthrough questionable - mge-xml

jimklimov:
* drivers/mge-xml.c: avoid implicit switch fall-through (and comment whether that wondering is the right thing to do though)


clepple (commented):
No idea, sorry.

jimklimov:

Issue #734, opened 2019-10-09
snmp-ups doesnt work for APC UPS: testOID provided and doesn't match MIB


harsh-w:
Running Network UPS Tools 2.7.4 on Ubuntu 18.04 and trying to work with an APC SMT2200 with an AP9631

```
$ sudo upsd -V
Network UPS Tools upsd 2.7.4
```

For some reason the upsd driver cannot seem to match any OID, either the ietf or the apcc ones.

The contents of the ups.conf file
```
[APC_UPS]
                   driver = snmp-ups
                   port = 192.168.42.63
                   snmp_version = v3
                   secLevel = authPriv
                   secName = shutdown_user
                   authPassword = *********
                   authProtocol = SHA
                   privPassword = *********
                   privProtocol = AES
                   desc = "APC UPS"
```
Running `$ sudo /lib/nut/snmp-ups -DD -a APC_UPS` gives 

```
Network UPS Tools - Generic SNMP UPS driver 0.97 (2.7.4)
   0.000000	debug level is '2'
   0.000772	SNMP UPS driver: entering upsdrv_initups()
   0.000790	SNMP UPS driver: entering nut_snmp_init(snmp-ups)
   0.005638	Setting SNMP retries to 5
   0.005658	Setting SNMP timeout to 1 second(s)
   0.031828	SNMP UPS driver: entering load_mib2nut(auto)
   0.031840	trying the new match_sysoid() method
   0.049989	Can't get sysOID value
   0.050026	load_mib2nut: trying classic method with 'apcc' mib
   0.050034	Testing ups.model using OID .1.3.6.1.4.1.318.1.1.1.1.1.1.0
   0.050087	load_mib2nut: testOID provided and doesn't match MIB 'apcc'!
   0.050111	load_mib2nut: trying classic method with 'mge' mib
   0.050238	Testing ups.model using OID .1.3.6.1.4.1.705.1.1.1.0
   0.050306	load_mib2nut: testOID provided and doesn't match MIB 'mge'!
   0.050393	load_mib2nut: trying classic method with 'netvision' mib
   0.050422	Testing ups.model using OID .1.3.6.1.4.1.4555.1.1.1.1.1.1.0
   0.050515	load_mib2nut: testOID provided and doesn't match MIB 'netvision'!
   0.050540	load_mib2nut: trying classic method with 'pw' mib
   0.050576	Testing ups.model using OID 1.3.6.1.4.1.534.1.1.2.0
[...]

PR #850, opened 2020-11-08
drivers/snmp-ups.c: usmAESPrivProtocol is not an array to properly use a sizeof()

jimklimov:
````
snmp-ups.c: In function 'nut_snmp_init':
snmp-ups.c:599:66: error: division 'sizeof (oid * {aka long unsigned int *}) / sizeof (oid {aka long unsigned int})' does not compute the number of array elements [-Werror=sizeof-pointer-div]
  599 |    g_snmp_sess.securityPrivProtoLen =  sizeof(usmAESPrivProtocol)/sizeof(oid);
      |                                                                  ^
In file included from /usr/include/net-snmp/snmpv3_api.h:27,
                 from /usr/include/net-snmp/net-snmp-includes.h:78,
                 from snmp-ups.h:79,
                 from snmp-ups.c:39:
/usr/include/net-snmp/library/transform_oids.h:50:26: note: first 'sizeof' operand was declared here
   50 | NETSNMP_IMPORT oid      *usmAES128PrivProtocol; /* backwards compat */
      |                          ^~~~~~~~~~~~~~~~~~~~~
make[1]: *** [Makefile:1953: snmp_ups-snmp-ups.o] Error 1
````

In my recent fight against warnings in NUT codebase, augmented by #844 and following from #823, this issue remains the only one I don't have an idea about solving. Help would be very much welcome to make a lot of additional Travis CI test-cases green.


jimklimov:
Some investigation showed that the token we use is defined in current net-snmp headers as a legacy pointer type, `oid *`, while many other names in the same header are arrays typed like `oid[10]`. There are also named macros to represent the amount of entries in those arrays, which is what we would use as a priority if available.

Bi11:
It is more than a warning. Because incorrect length of `usmAES128PrivProtocol`, `generate_Ku` fails and AES doesn't work. 
The similar issue is affecting `nut-scanner` too. 

See https://github.com/networkupstools/nut/issues/734#issuecomment-741982795

jimklimov:
Hell @Bi11 and sorry for the delay. I've read through this and linked comments, and my understanding is that with the current state of the master branch (or at least with these fixes applied to your build) AES works for you, while in the (aged) 2.7.4 release it did not?

Can you clarify please if there remains something to fix in `nut-scanner` along the same lines, or is it also already fixed? In the past weeks I juggled almost a thousand commits to fix warnings here and there, and lost track a bit :) So verifications like this are very welcome, whatever the verdict ;)

Bi11:
I've just tried the master branch, the "sizeof" issue in 2.7.4 has been fixed with your commits, both in `drivers/snmp-ups` and `nut-scanner`. I'm really appreciated for your improvements made to this project, @jimklimov.

However, `nut-scanner` is still not working with my UPS, and there are some other trivial issues. I'll create a new Issue or PR once I make it work with my UPS.

jimklimov:
Thank you very much for confirmations! I know about 1200 warnings remain
per clang-9, though most are relatively trivial double-vs-float and
int-unsigned-size_t which should not be fatal.

A few left me scratching my head, as attempts to quickly fix grew into a
big can of worms to pull. So I wanted to fix others first and then better
see impact of remaining efforts, like when warnings appear in previously
green code :)

If you want to join this crusade, note the "fightwarn" branch often
contains a few fixes ahead of those committed to master.

PR #852, opened 2020-11-08
Fightwarn - find use for "extra" argument in instcmd()s all around the place

jimklimov:

Got a few nagging concerns about the chosen solution (to print the "extra" value to log if we failed to support the command):
* what if `extra` is NULL? (Would logging routine not break - easy to test but was not done yet; can be worked around with lots of `extra ? extra : "<null>"` replacements)
* is there a chance to leak something secret into the log?

### DIFF
--- a/drivers/al175.c
+++ b/drivers/al175.c
@@ -1247,7 +1247,7 @@ static int instcmd(const char *cmdname, const char *extra)
 		return (!err ? STAT_INSTCMD_HANDLED : STAT_INSTCMD_FAILED);
 	}
 
-	upslogx(LOG_NOTICE, "instcmd: unknown command [%s]", cmdname);
+	upslogx(LOG_NOTICE, "instcmd: unknown command [%s] [%s]", cmdname, extra);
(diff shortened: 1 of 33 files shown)


jimklimov:
Made a small test case to be sure:

````
// upsdebug.c
#include "common.h"

int main(void) {
    const char *s1 = "!NULL";
    const char *s2 = NULL;
    upsdebugx(0, "D: '%s' vs '%s'", s1, s2);
    return 0;
}
````

````
# Makefile
$ grep deb Makefile.am 
bin_PROGRAMS += upsdebug
upsdebug_SOURCES = upsdebug.c
upsdebug_LDADD = ../common/libcommon.la
````

````
$ make upsdebug && ./upsdebug ; echo $?
D: '!NULL' vs '<NULL>'
0
````

So no crash, no error-codes, a `<NULL>` string is substituted - so unspecified `extra` arguments should not break `instcmd()` all around the place.

jimklimov:
Put the test above into `tests/nutlogtest.c` for good measure, and found the C++ test needs a little polishing.

PR #853, opened 2020-11-08
Fightwarn - floating-point abs()

jimklimov:
A number of maths comparisons for negligibly small values were originally done using `abs()` which is (per warnings from many systems) intended for integers. This PR changes those cases to use `fabs()` from `math.h`.


clepple (approved):
I would put the config check for `abs` on the previous line, since it is in `stdlib.h` not `math.h`, but should work either way.

PR #863, opened 2020-11-09
FTY remerge from opensource/master as of 20201109

jimklimov:
Resync to 42ity from NUT master


jimklimov:
Wrong target repo, sorry

PR #866, opened 2020-11-10
libshut.c riello_ser.c nut-scanner bits: banish non-standard u_char and u_short

jimklimov:
Probably standard C99 "uint8_t" and 16(?) bit short would be more proper,
especially for serial-line protocols. But the original code used types
derived from hardware implementation sized "char" and "short" - so they
stayed this way here.

Stems from efforts on #823 and #844 to get more clean and green system setups in CI. Due to the comment above, review would be welcome.


aquette (approved):
Seems indeed good, with a 5 seconds review. Testable for non reg on any Eaton unit with a serial port and having USB HID (ie not older power ware with XCP)

clepple (approved):
LGTM

jimklimov:
Thanks @aquette for the suggestion, now I intend to test this on an OmniOS box that has Eaton 9PX 5000i UPS attached by USB cable as well as seen over SNMP :) Not sure how much the USB drivers would help with bug-hunt in serial port code, but at least similar variable type change in SNMP code is more auditable.

Not sure what to look out for, probably diff's of outputs from `upsc` with one or other same-named driver build.

aquette:
testing through USB won't give anything, since libshut is the equivalent marshaling layer of libusb.
Then it's the common HID core above.
AFAIR, 9PX also have a serial port (RS 232 / DB9). Keep in mind that P2P ports are mutually exclusive.
So when testing RS232, remove the USB cord from the UPS ;)
hope it helps, I've no more ups@home, so can't help further

jimklimov:
In this case, my physical setup that is stranded in the now-inaccessible office won't be of much help.

I looked at systems headers I could get at, everywhere that has an `u_char` defined it is an `unsigned char` (same for `short`s), so this PR does not change the outcome for those.

I suppose a further pedantic change to turn such arch-dependent cases into `uint8_t` and `uint16_t`respectively, would need more testing with real hardware.

PR #879, opened 2020-11-17
Fightwarn - macos llvm - noreturn

jimklimov:
Follows up from #823 and #844 to address warnings from clang10 on MacOS builds: pre-declare attributes for routines that `exit()` so the compiler can optimize for no return path handling.


jimklimov:
Error reports from clang were cut short by their reporting limits, some more "appeared" in view after the cleanup. PR would grow :)

PR #880, opened 2020-11-17
nutclient.cpp:  match a copy constructor with a copy assignment…

jimklimov:
…operator (Device, Variable, Command classes)

https://lgtm.com/rules/2165180572/


jimklimov:
At least in `gnu++17` mode with clang-9, this file shows a lot more different warnings in other aspects of code. The one issue originally posted seems solved (at least syntax-wise, not really tested yet...)

PR #881, opened 2020-11-17
scripts/augeas/gen-nutupsconf-aug.py: drop "global" modifier outside …

jimklimov:
…a local function context


jimklimov:
Seems this change is not portable; e.g. Travis fails to regenerate the lenses. Python 2 vs 3 maybe?

aquette:
indeed, leftover from planned but unfinished development.
while on the augeas side, IIRC, I started to add the augeas verification (auto, when avail) but stopped due to a specific failure I don't remember... If you resurrect this, it may pop...

PR #882, opened 2020-11-17
Fightwarn - lgtm.com - bogus comparisons

jimklimov:

We have a number of cases where either an uint type is tested for (non-)negativity, or a signed int was recently assigned from a function that returns unsigned types (the latter I am reluctant to "fix" as different OSes might return different types and ranges), or was assigned from an element of an array of uint's.

### DIFF
--- a/drivers/bcmxcp.c
+++ b/drivers/bcmxcp.c
@@ -1207,7 +1207,7 @@ void init_limit(void)
 
 	/* Horn Status: */
 	value = answer[BCMXCP_EXT_LIMITS_BLOCK_HORN_STATUS];
-	if (value >= 0 && value <= 2) {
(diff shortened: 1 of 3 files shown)


drivers/bcmxcp.c
```diff
@@ -1207,7 +1207,7 @@ void init_limit(void)
 
 	/* Horn Status: */
 	value = answer[BCMXCP_EXT_LIMITS_BLOCK_HORN_STATUS];
-	if (value >= 0 && value <= 2) {
+	if (value <= 2) {
```
jimklimov, line 1210: Note: currently `answer[]` is an array of unsigned values, so compilers warn about check for `... >= 0` as redundant.

PR #883, opened 2020-11-17
Fightwarn - macos llvm - ssize_t for read()/write()/send()

jimklimov:
Routines for `read()` and `write()` or `send()` are defined with `ssize_t` wherever I found to look.

Builds on MacOS complained about bitness range mismatch of `int` and `ssize_t` so the chain of calls was modernized here.

Follows up from #823 and #844 efforts.


jimklimov:
At least partially tested by connections (driver, upsd, upsc) to snmp-ups and usbhid-ups devices.

PR #885, opened 2020-11-18
Fightwarn - macos llvm - nonliteral format strings

jimklimov:
Follows up from #823 and #844 to address warnings from clang10 on MacOS builds: NUT extensively uses string variables from tables or calculated at runtime to pass into various likes of `snprintf()` and NUT's own routines with fmt+varargs syntax.

This PR highlights such cases of intentional use of non-literal format strings with pragmas to quiesce GCC and CLANG warnings for this.

The main possible point of PR review is to double-check if these cases (now wrapped in pragmas, sometimes spanning large blocks of code with many `snprintf`s inside) are all intentional and unavoidable. Otherwise, there is little change from previous state of codebase (one avoidable case avoided), so this PR is not likely to break anything :)

### DIFF
--- a/clients/upssched.c
+++ b/clients/upssched.c
@@ -339,7 +339,15 @@ static int send_to_one(conn_t *conn, const char *fmt, ...)
 	char	buf[US_SOCK_BUF_LEN];
 
 	va_start(ap, fmt);
+#if defined (__GNUC__) || defined (__clang__)
+#pragma GCC diagnostic push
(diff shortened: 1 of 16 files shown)


jimklimov:
Error reports from clang were cut short by their reporting limits, some more "appeared" in view after the cleanup. PR would grow :)

jimklimov:
Several more cases to go, but will go in another PR later, to have gains now and to keep changesets small.

PR #887, opened 2020-11-18
Fightwarn - lgtm com - hiddenargs usbhid

jimklimov:

Changes proposed in this PR concern me a bit more than in others: "should we save the value of function arguments into the global variables for this driver instance?" (and would it play well with several instances running), so additional review and sanity-checking would be very welcome.


jimklimov:
At least partially tested by connections (driver, upsd, upsc) to snmp-ups and usbhid-ups (in branch for libusb-1.x support) devices.

PR #889, opened 2020-11-18
Fightwarn - lgtm com - hiddenargs netxml

jimklimov:
drivers/netxml-ups.c: send_http_request(): do not shadow global varnames "session" and "uri" with func arguments

Changes proposed in this PR concern me a bit more than in others: "I wonder if we should we save the values of function arguments into the global variables for this driver instance?" (and would it play well with several instances running), so additional review and sanity-checking would be very welcome.


jimklimov:
At least partially tested by connections (driver, upsd, upsc) to netxml-ups devices.

jimklimov:
May be at least partially tested by connections (driver, upsd, upsc) to netxml-ups devices - Eaton 9PX. I am not sure if that codepath does involve mge-xml sources, though.

PR #890, opened 2020-11-18
Fightwarn - lgtm com - hiddenargs mgexml

jimklimov:
drivers/mge-xml.c: numerous routines: do not shadow global varname "val" (for XML element temp store?) with same-named func argument

Changes proposed in this PR concern me more than in others: what role does the global `val[]` play - is it a scratch buffer, or does it carry pieces of data between routines e.g. during XML parsing?

I wonder if we should we save the values of function arguments into the global variables for this driver instance? (and would it play well with several instances running), so additional review and sanity-checking would be very welcome.

### DIFF
--- a/drivers/mge-xml.c
+++ b/drivers/mge-xml.c
@@ -131,9 +131,9 @@ typedef struct {
 						   NULL if no further processing is required) */
 } xml_info_t;
 
-static const char *online_info(const char *val)
+static const char *online_info(const char *arg_val)
 {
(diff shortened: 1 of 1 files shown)


aquette (approved):
LGTM

jimklimov:
May be at least partially tested by connections (driver, upsd, upsc) to netxml-ups devices - Eaton 9PX. I am not sure if that codepath does involve mge-xml sources, though.

PR #891, opened 2020-11-18
Fightwarn - lgtm com - hiddenargs nut-scanner serial

jimklimov:
tools/nut-scanner/scan_eaton_serial.c: shut_synchronise(): do not shadow global varnames "upsfd" (from scan_eaton_serial.c and main.c) with func argument "upsfd"

Changes proposed in this PR concern me a bit more than in others: "should we save the value of function arguments into the global variables for this driver instance?" - was there some use for global `upsfd` that might be expecting to use the same fd in different places, so additional review and sanity-checking would be very welcome.


jimklimov:
Merged as part of other PRs that build upon this change.

PR #892, opened 2020-11-18
Fightwarn - lgtm com - hiddenargs tripplite-usb

jimklimov:
drivers/tripplite_usb.c: subdriver_match_func(): do not shadow global varname "hd" with func argument "hd"

Changes proposed in this PR concern me a bit more than in others: "I wonder if we should we save the values of function arguments into the global variables for this driver instance?" (and then would it play well with several instances running), so additional review and sanity-checking would be very welcome.

### DIFF
--- a/drivers/tripplite_usb.c
+++ b/drivers/tripplite_usb.c
@@ -162,11 +162,16 @@ static usb_device_id_t tripplite_usb_device_table[] = {
 	{ -1, -1, NULL }
 };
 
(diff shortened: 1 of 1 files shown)


clepple:
> ...and then would it play well with several instances running

I don't think local/global changes would matter with several driver instances, since each driver is a separate process. (`tripplite_usb` has difficulty telling non-HID UPSes apart anyway, unless they are different models, so this is hard to test).

PR #895, opened 2020-11-18
scripts/augeas/gen-nutupsconf-aug.py: drop unused "global" vars

jimklimov:
scripts/augeas/gen-nutupsconf-aug.py: drop unused variables
    
    Originally PR meant to drop "global" modifier outside a local
    function context, but found that nothing defines or uses these
    three names. Must be some leftover from planned but unfinished
    development.
    
    Re-ran ./autogen.sh and it generates the same file for the build.

And replaces PR #881


aquette:
as per my comments on #881

PR #902, opened 2020-11-25
fightwarn - nutclient.{h,cpp} : fix an horde of warnings

jimklimov:
After a practical fix for #880, at least 50 errors were emitted by clang-9 for the remaining C++ code in this file. Most though not all taste of syntactic sugar, but whatever - if it helps the code be clean and nice.

Follows up from #823 / #844 efforts.

### DIFF
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -60,6 +60,23 @@ static inline void *xrealloc(void *ptr, size_t size){return realloc(ptr, size);}
 static inline char *xstrdup(const char *string){return strdup(string);}
 #endif /* HAVE_NUTCOMMON */
 
+/* To stay in line with modern C++, we use nullptr (not numeric NULL
+ * or shim __null on some systems) which was defined after C++98.
+ * The NUT C++ interface is intended for C++11 and newer, so we
+ * quiesce these warnigns if possible.
+ * An idea might be to detect if we do build with old C++ standard versions
+ * and define a nullptr like https://stackoverflow.com/a/44517878/4715872
+ * but again - currently we do not intend to support that officially.
+ */
+#ifdef HAVE_PRAGMAS_FOR_GCC_DIAGNOSTIC_IGNORED_CXX98_COMPAT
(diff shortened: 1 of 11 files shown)


clients/nutclient.cpp
```diff
@@ -730,7 +730,7 @@ std::map<std::string,std::map<std::string,std::vector<std::string> > > TcpClient
 			{
 				std::vector<std::string>& vals = *it2;
 				std::string var = vals[0];
-				vals.erase(vals.begin()),
+				vals.erase(vals.begin());
```
jimklimov, line 741: @aquette @boricj : please tell me I did fix a typo and not some intentional C++ trick? ;)

boricj: I'd say harmless typo.

bkuhls:
This commit breaks cross-builds:
```
checking for practical support to pritnf("%s", NULL)... configure: error: in `nut-fbe7806e83f2838da21a0c7d56649777780881c1':
configure: error: cannot run test program while cross compiling
```
A hackish fix is to add ax_cv__printf_string_null=yes to the configure environment.

jimklimov:
Interesting, thanks for the report. Can you clarify the cross-setup you
have (platforms involved, compiler versions, other-platform packages maybe,
configure command line) -- I wonder if I can reproduce it so that CI builds
could care for that case and the main codebase would work with it? I think
we already do somewhat similar to cross-builds in some of Travis
branches... but maybe not enough :)

bkuhls:
> Can you clarify the cross-setup you have

The issue occured while bumping the buildroot package from version 2.7.4
https://github.com/buildroot/buildroot/blob/master/package/nut/nut.mk
to its current git HEAD: https://github.com/bkuhls/buildroot/commit/8d135b7bb81d792e941c4bd478c6090f9f4a9d8d

To reproduce the problem use buildroot to build nut with my commit but without https://github.com/bkuhls/buildroot/blob/master_patches/package/nut/nut.mk#L33

PR #905, opened 2020-11-25
Fightwarn - the remainder of unreachable code warnings, and random small fixes

jimklimov:
Follows up from #823 / #844 efforts to remove several classes of warnings from being seen in NUT build reports.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ addons:
     - libmodbus-dev
     - libnss3-dev
     - libssl-dev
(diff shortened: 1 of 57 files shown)


jimklimov:
At least partially tested for nut-scanner fixes (in the overall fightwarn effort branch applied) for discovery of nutclient, SNMP and NetXML devices - these worked. An USB driver for two ports was occupied and two port-busy reports appeared. (With offending drivers stopped, a re-run of nut-scanner discovered these devices).

PR #910, opened 2020-11-26
fightwarn - avoid variable-length arrays, their support became optional after C99

jimklimov:
Follows up from #823 / #844 efforts, this PR avoids initializing variable-length arrays in the code. In most cases it was changed to dynamic `xmalloc/free-before-return` handling. A couple of subsequent fixes for `rhino.c` also got here, which depend on these changes.

In the `solis.c` driver it was just replaced by the macro which (in header) initializes the unmodifiable (right?) typed variable:
````
#define PACKET_SIZE 25
static const int packet_size = PACKET_SIZE;
````


clepple (approved):
Changes seem reasonable, but as with a lot of the other changes, we'll want to mention in the release notes (or better yet, on the lists beforehand) that testing is needed.

PR #912, opened 2020-11-26
Use HID_ID to match devices instead of USB specific sysfs attributes

hadess:
This should make it possible to use `hid-recorder` and `hid-replay` to emulate UPSes.

See https://gitlab.freedesktop.org/libevdev/hid-tools/-/issues/25

@bentiss, does this look correct to you?

(I haven't tested this beyond checking that the generated files look correct)

### DIFF
--- a/tools/nut-usbinfo.pl
+++ b/tools/nut-usbinfo.pl
@@ -108,7 +108,7 @@ sub gen_usb_files
 	# Udev file header
 	open my $outUdev, ">$outputUdev" || die "error $outputUdev : $!";
 	print $outUdev '# This file is generated and installed by the Network UPS Tools package.'."\n\n";
-	print $outUdev 'ACTION!="add|change", GOTO="nut-usbups_rules_end"'."\n";
+	print $outUdev 'ACTION=="remove", GOTO="nut-usbups_rules_end"'."\n";
 	print $outUdev 'SUBSYSTEM=="usb_device", GOTO="nut-usbups_rules_real"'."\n";
 	print $outUdev 'SUBSYSTEM=="usb", GOTO="nut-usbups_rules_real"'."\n";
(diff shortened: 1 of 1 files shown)


hadess:
(This definitely doesn't work, I'll do some more debugging in a bit, at least the matching isn't right...)

hadess:
And now it should work once https://gitlab.freedesktop.org/libevdev/hid-tools/-/issues/25 gets fixed. The hiddev device isn't getting instantiated right now.

hadess:
I've split off the fixes I made in https://github.com/networkupstools/nut/pull/913

This MR should have been about using `HID_ID` for matching so `hid-replay` could be used to mock UPSes, but NUT (and upower) use hiddev, which is only available if the transport is USB, not for uhid/hid.

Ideally, the NUT code would be migrated to use hidraw, so the codepath would be the same for USB UPSes, and "replayed" devices. That's not the case as of now though, and I don't have the bandwidth to work on that right now.

jimklimov:
Thanks for giving this a try! If you get the bandwidth for this later, possibly after we choose and merge one of the libusb-1.x integration branches, seems like it would be a useful to a higher-level dummy-ups.

clepple:
> This MR should have been about using HID_ID for matching so hid-replay could be used to mock UPSes, but NUT (and upower) use hiddev, which is only available if the transport is USB, not for uhid/hid.

For future reference: I haven't looked at upower, but NUT (primarily the `usbhid-ups` driver) is not actually using hiddev - it is using `libusb` after detaching the kernel drivers (e.g. hiddev). This was an early design decision, since the original Linux hiddev API did not allow NUT to distinguish between two "Voltage" Usage IDs, say if one was nested under "Battery" and the other under the collection for line power. A user-space HID parsing library was written that used libusb, with the advantage that it still works on non-Linux systems. To mock an UPS, it doesn't seem too farfetched to replay USB pcaps into libusb.

PR #913, opened 2020-11-26
tools/nut-usbinfo.pl: Fixes

hadess:
Misc fixes


jimklimov (approved):

jimklimov:
LGTM for most of the changes, just gotta trust your research and judgement about the udev rules :) Thanks!

I will leave this PR open for a few days to give a chance for others to comment if something would concern them.

hadess:
> LGTM for most of the changes, just gotta trust your research and judgement about the udev rules :) Thanks!

You might want to test them, even summarily, as I don't have the hardware to test it.

clepple:
@hadess I probably need to reread the LWN article and the bugs linked from it, but assuming that the NUT rules are changing permissions on nodes created from ADD events, what is the harm in skipping that when a BIND event comes in?

It sounds like users of 4.14+ kernels should be seeing regressions of some sort, but I'm not sure what that looks like for NUT.

I have no problem testing the two later commits at the same time, but scheduling time to test on non-production hardware might be tough (I had a few batteries die on non-critical UPSes around the house, and I haven't replaced them yet), so I want to understand what I'm looking for before rewiring things.

hadess:
> @hadess I probably need to reread the LWN article and the bugs linked from it, but assuming that the NUT rules are changing permissions on nodes created from ADD events, what is the harm in skipping that when a BIND event comes in?

The properties you set when processing an event aren't kept across subsequent events, so your device would end up without any properties. That's why the original rules also handled change events.

hadess:
> It sounds like users of 4.14+ kernels should be seeing regressions of some sort, but I'm not sure what that looks like for NUT.

Newer kernels and older systemd would see no properties set on the NUT handled devices. This is worked-around in newer versions of systemd.

clepple:
@hadess did a quick sanity check, and everything seems to work fine with Ubuntu 20.04, plus the following NUT, systemd, and kernel versions. Thanks again for the fix and patient explanations.

commit 38b5f96f3c6dd55782ba9eaf5a505a1c2dca72a8

```
$ systemd --version
systemd 245 (245.4-4ubuntu3.3)

$ uname -a
Linux voyager 5.4.0-58-generic #64-Ubuntu SMP Wed Dec 9 08:16:25 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

PR #927, opened 2020-11-30
fightwarn - avoid "bad function cast" by casting explicitly the return value of the function

jimklimov:
Follows up from #823 effort.

This PR has potential to somehow change interpretations of numbers in NUT calculations, so careful review is encouraged. Also I hope I did not mix up parenthesis grouping and negations, but it is late night here so everything is possible (caught some of those before posting)...

Would merge after someone LGTMs this :)


clepple (approved):
LGTM

PR #928, opened 2020-11-30
Fightwarn - Introduce nut_float.h to compare floating-point numbers in a safer manner

jimklimov:
Follows up from #823 effort.

This PR has potential to somehow change interpretations of numbers in NUT calculations, so careful review is encouraged. Also I hope I did not mix up parenthesis grouping and negations, but it is late night here so everything is possible (caught some of those before posting)...

Would merge after someone LGTMs this :)

### DIFF
--- a/include/nut_float.h
+++ b/include/nut_float.h
@@ -40,9 +40,15 @@
  *     double fabs(double x);
  *     float fabsf(float x);
  *     long double fabsl(long double x);
+ * To be on the safer side, we compare the difference to be smaller or equal to
+ * the Epsilon for the respective numeric type, in order to be equivalent to a
+ * zero for our needs. Various libs define it as "the next representable number
+ * after 1.0 which is not equal to 1.0" for the discrete maths involved; no talk
+ * about exactly comparing it to zero or whether it is the smallest representable
+ * non-zero value...
  */
-#define f_equal(x, y) ( fabsf((float)(x) - (float)(y)) < FLT_EPSILON )
-#define d_equal(x, y) ( fabs((double)(x) - (double)(y)) < DBL_EPSILON )
-#define ld_equal(x, y) ( fabsl((long double)(x) - (long double)(y)) < LDBL_EPSILON )
+#define f_equal(x, y) ( fabsf((float)(x) - (float)(y)) <= FLT_EPSILON )
+#define d_equal(x, y) ( fabs((double)(x) - (double)(y)) <= DBL_EPSILON )
+#define ld_equal(x, y) ( fabsl((long double)(x) - (long double)(y)) <= LDBL_EPSILON )
 
 #endif	/* NUT_FLOAT_H_SEEN */
--- a/include/nut_float.h
+++ b/include/nut_float.h
@@ -1,0 +1,48 @@
+/*
+ * nut_float.h - Network UPS Tools include files for floating point routines
+ * and ways to compare if non-integers are close enough to assume equal
+ *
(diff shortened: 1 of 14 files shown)


include/nut_float.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * nut_float.h - Network UPS Tools include files for floating point routines
+ * and ways to compare if non-integers are close enough to assume equal
+ *
+ * Copyright (C) 2011	Arjen de Korte <adkorte-guest@alioth.debian.org>
```
clepple, line 5: Is this from one of Arjen's other files? If not, an updated copyright year and author might be good.

include/nut_float.h
```diff
@@ -0,0 +1,48 @@
...
+/* Modern compilers frown upon direct comparisons of floating point numbers
+ * since their imprecise internal representations can misfire, and really we
+ * care if they are indiscernably close. To aviod warnings like -Wfloat-equal
+ * we prefer to compare with methods defined below. Note: despite the "floating"
+ * name, fabs() seems to be defined over doubles or wider types, e.g.:
+ *     double fabs(double x);
+ *     float fabsf(float x);
+ *     long double fabsl(long double x);
+ */
+#define f_equal(x, y) ( fabsf((float)(x) - (float)(y)) < FLT_EPSILON )
```
clepple, line 44: Should this be `< 2*FLT_EPSILON` or `<= FLT_EPSILON`? Otherwise, if epsilon is the closest number to zero, the only way the delta can be less than epsilon is if it is exactly 0 (which is the case the compiler is complaining about).

clepple (changes requested):
I agree that something like this header should be used for these kinds of comparisons. Not sure about the implementation, though.

I can't say this is a rigorous analysis, but the `*_EPSILON` comparisons look incorrect to me (see inline comment). Happy to defer to test results or someone else's analysis.

include/nut_float.h
jimklimov: Good point, did not commit the change. This file is new :)

include/nut_float.h
jimklimov: Interesting consideration. I don't think any discussions on float comparisons I saw in the research leading up to this solution went into this detail, although there were quite a few examples with strict "less". To err on side of caution, might go with "small difference still means zero for our needs" :)

Per various layman definitions in headers etc., these epsilons are defined like:
* "The macro yields the smallest X of type float such that 1.0 + X != 1.0." - https://www.qnx.com/developers/docs/6.4.1/dinkum_en/cpp/float.html
* "This is the difference between 1 and the smallest floating point number of type float that is greater than 1. It’s supposed to be no greater than 1E-5." (1E-9 for double and long double) - https://www.gnu.org/software/libc/manual/html_node/Floating-Point-Parameters.html
* "difference between 1.0 and the next representable value for float, double and long double respectively" - https://en.cppreference.com/w/c/types/limits

So it is probably not exactly the minimum representable value in the bits of the discrete floating point numeric type, I don't see a direct correlation for comparisons to zero. Got me wondering if the math should be like `fabs( x - y ) + 1.0 > 1.0 + FLT_EPSILON` or some such :)

clepple: > Got me wondering if the math should be...

I think adding 1.0 to a very small number makes the comparison problem worse.

The `<= *_EPSILON` comparison should keep future compilers quiet (and in most of these cases, NUT was well-served by the original round-to-int semantics anyway).

clepple (approved):
Changes look good (though as before, we will want people to test for unintentional errors).

PR #933, opened 2020-12-15
Fix incorrect usmAESPrivProtocol import in nut-scanner

Bi11:
In libnetsnmp, usmAES128PrivProtocol is a pointer, while usmAESPrivProtocol is an array.
This change is the same as in c7cf507454bcf6f40b24b58e61c130c7772c5bf2.

### DIFF
--- a/tools/nut-scanner/scan_snmp.c
+++ b/tools/nut-scanner/scan_snmp.c
@@ -62,7 +62,7 @@
 #include "nutscan-snmp.h"
 
 /* Address API change */
-#ifndef usmAESPrivProtocol
+#if ( ! NUT_HAVE_LIBNETSNMP_usmAESPrivProtocol ) && ( ! defined usmAESPrivProtocol )
 #define USMAESPRIVPROTOCOL "usmAES128PrivProtocol"
 #else
 #define USMAESPRIVPROTOCOL "usmAESPrivProtocol"


Bi11:
`usmAES128PrivProtocol` and `usmAESPrivProtocol` are NOT exchangeable in binary level.

```
Symbol table '.dynsym' contains 1302 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
   ...
   833: 00000000000b81e0     8 OBJECT  GLOBAL DEFAULT   24 usmAES128PrivProtocol
   ...
   178: 00000000000b7ce0    80 OBJECT  GLOBAL DEFAULT   24 usmAESPrivProtocol
```

jimklimov:
I'm not at code at the moment; should other changes from that commit (lengths etc) apply here?

Bi11:
The length is correct ( it was fixed in 54a61655059a3f19b2b1de2d772aba049f32a4b2 ). 
But current code won't compile with old version of libnetsnmp which has`usmAES128PrivProtocol` defined but no `usmAESPrivProtocol`.
But it is another task to fix compile issue with old version of libnetsnmp.

jimklimov:
Thanks for the research and the fix.

If the older dependencies are relevant in practice, a (configure/define gated?) fix is also welcome :)

PR #936, opened 2020-12-16
snmp-ups: Restore legacy Eaton ePDU switchability info

jimklimov:
Use a hack to also have switchability for both the unit and its
outlets on legacy Eaton G2 ePDU

Partially cherry-picked from 42ity fork, and bumped to their faster moving
MIB mapping revision number. See https://github.com/42ity/nut/pull/118

As for the behavior of the fix:

* on G2, the first outlet.switchable doesn't exist, so try the 2nd,
  which is the ugly hack: try on the first outlet to see if switchable

* on G3, the first is good, so flag unique does the rest

Tested on MA G2 and G3.

### DIFF
--- a/drivers/eaton-pdu-marlin-mib.c
+++ b/drivers/eaton-pdu-marlin-mib.c
@@ -34,7 +34,7 @@
 /* Eaton PDU-MIB - Marlin MIB
  * ************************** */
 
-#define EATON_MARLIN_MIB_VERSION	"0.49"
+#define EATON_MARLIN_MIB_VERSION	"0.56"
 #define EATON_MARLIN_SYSOID			".1.3.6.1.4.1.534.6.6.7"
 #define EATON_MARLIN_OID_MODEL_NAME	".1.3.6.1.4.1.534.6.6.7.1.2.1.2.0"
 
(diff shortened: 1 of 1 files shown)


drivers/eaton-pdu-marlin-mib.c
```diff
@@ -709,7 +722,11 @@ static snmp_info_t eaton_marlin_mib[] = {
 	/* outletControlSwitchable */
 	{ "outlet.%i.switchable", ST_FLAG_STRING, SU_INFOSIZE,
 		".1.3.6.1.4.1.534.6.6.7.6.6.1.9.%i.%i",
-		"no", SU_OUTLET | SU_TYPE_DAISY_1, &marlin_outlet_switchability_info[0] },
+		"no", SU_OUTLET | SU_FLAG_UNIQUE | SU_TYPE_DAISY_1, &marlin_outlet_switchability_info[0] },
+	/* FIXME: handle non switchable units (only measurements), which do not expose this OID */
```
aquette, line 726: note that this FIXME could have been removed, since the absence of the pointed OID (on non-switchable units) will result in the non publication of the data

aquette (approved):

Issue #940, opened 2020-12-25
Problem: arcane knowledge needed for specifying subdrivers


jimklimov:
Trying to see a remote system's UPS with `usbhid-ups` and `blazer_usb` drivers, I got a hint from the help message that subdrivers can be specified. Indeed, some are mentioned in the manpage. I suppose this can take a while to unravel for a new NUT user, or one that has not played with USB in a while.

TODO:
* Subdrivers supported by a current build of a driver (some are ifdef-dependent) are listed in the corresponding array, and so their string names can be listed in `help()` messages.
* For `usbhid-ups` in particular, if no subdriver matches the connected devices, suggest that the user can specify the `-x explorer` flag to collect some data about the devices that did look like unsupported UPSes, to pass to the NUT devs mailing list and/or post a Github issue.
* Bump debuggging around driver matching so that users have more understanding what was tried and why it failed (e.g. "Tried your explicit VendorID vs. value from UPS and it did not match" to help catch typos in config file)

jimklimov:
PR #941 addresses most of the goals. In the end, it did not get to suggesting the `explore` option (or other USB drivers) if usbhid-ups fails to find stuff. There are also fewer drivers with user-selectable subdrivers than I first anticipated, those (two) are now covered to suggest the strings in help text. Selection of USB subdrivers (especially reasons for mismatch decision) is now traced where I could check in practice.

jimklimov:
Keeping open for now, in case someone finds not-traced subdriver selection codepaths, or addresses the helpful suggestions where a driver matched nothing, and proposes PRs to fix those cases.

PR #941, opened 2020-12-25
Help troubleshoot USB driver setups

jimklimov:
Following up with issue #940, this PR adds some debug tracing for device-vs.-driver matching in codepaths I've encountered in practice. A cursory look across the codebase exposed similar logic used in some other drivers, but it is not something I can test quickly now, nor decide if similar change is welcome there too. Anyway, later PRs can be modeled after this one ;)

Also just two drivers were recognized to support the "subdriver" flag, `nutdrv_qx` (only for USB mode) and `blazer_usb` so only their help routines were updated to list the USB subdrivers. Example output:
````
$ ./drivers/nutdrv_qx -h
...

Acceptable values for 'subdriver' via -x or ups.conf in this driver: cypress, phoenix, ippon, krauler, fabula, fuji, sgs

Read The Fine Manual ('man 8 nutdrv_qx')
````

There is also `riello_usb` whose code structure seems very related, and it registers one `cypress` subdriver into the equivalent array, but does not register the `subdriver` keyword support (although can handle its value, still).

Equivalent logic in `snmp-ups` library of MIBs is handled already, by `snmp-ups -s test -port=test -x mibs=--list` to display a table of details about recognized MIBs.

The `usbhid-ups` driver in the end only matches devices it knows to the subdrivers it picks, but there does not seem to currently be a way for user to influence that choice.

### DIFF
--- a/drivers/blazer_usb.c
+++ b/drivers/blazer_usb.c
@@ -508,9 +508,33 @@ int blazer_command(const char *cmd, char *buf, size_t buflen)
 #endif	/* TESTING */
 }
 
+#ifndef TESTING
+const struct subdriver_t {
+	const char	*name;
+	int		(*command)(const char *cmd, char *buf, size_t buflen);
+} subdriver[] = {
+	{ "cypress", &cypress_command },
+	{ "phoenix", &phoenix_command },
+	{ "ippon", &ippon_command },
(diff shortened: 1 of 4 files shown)


clepple (approved):
I don't have any devices that can test this code, but it sounds like a good idea to me.

Issue #869, opened 2020-11-13
Consider the options migrating from Travis CI .ORG


jimklimov:
According to https://travis-ci.community/t/builds-hang-in-queued-state/10250/24 and https://travis-ci.community/t/org-to-com-migration-deadline/10260 the Travis CI team had poorly communicated that the free .ORG service will cease end of this year, and projects should move to .COM. Also there is https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions and https://docs.travis-ci.com/user/billing-overview/, and the timeline and discussion of options at https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

* From the latter, the point on "I don't have enough time in my day to send off emails every few [days|weeks|months] requesting extra build credits so I can continue maintaining my open source projects." sounds all too familiar and painful...

In fact, like many others in those threads, I only found out by looking for reasons of the recent lags while the status says all systems are green. Further in the first linked thread, there are some concerns about how much and how easily FOSS projects can build in the new setup, and it is even unclear whether their free allotment is a single-time deal or a monthly quota. If a quota is overflowed, it may be possible to ask for a bump which would likely mean regular bureaucracy and lag in PR tests to live through a month, and/or possibly a more active need for sponsorship from users and companies to pay for a plan, month to month.

Alternate options include going to GitHub Actions, or spinning up free VMs in the cloud providers (some provide a way to have free low-profile services; not sure if running a builder occupying a CPU core 24/7 counts as fair use for that sort of service), or buying some VMs or physical builders, hosted at providers or at home for the community members - but again can become a matter of sponsorship in the end. On the up-side, dedicated workers can be used with varied OSes not provided by common cloud CI.

Considering recent work to build in NUT as diverse environments as we could grab (#823, #844) the availability of multiple options in Travis was certainly helpful to bring code up into shape against standards and OSes. Thanks to them for providing as much as they did in the free .ORG service, in any case.

But the future default matrix of tests, whether on Travis CI .COM or on some other solution, likely would have to be reduced to conserve their resources and to speed our PR CI turnaround:
[...]

PR #946, opened 2020-12-26
Fightwarn - standardize warnings we care about via configure script arguments

jimklimov:
The options chosen into pre-sets are what we use for Travis CI tests since #823, and revised further for what I used locally for some builds.

This PR adds support for:
* `--enable-Werror(=yes/no)` - make warnings fatal
* `--enable-warnings(=arg)` - enable certain warning presets:
** `gcc-hard`, `clang-hard`, `gcc-minimal`, `clang-minimal`, `all` - actual definitions that are compiler-dependent (the latter just adds `-Wall` which may be relatively portable)
** `hard` or `minimal` - if current compiler is detected as CLANG or GCC, apply corresponding setting from above (or `all` otherwise)
** `gcc` or `clang` - apply the set of options (regardless of detected compiler) with default "difficulty" hard-coded in configure script, to tweak as our codebase becomes cleaner
** `yes`/`auto` - if current compiler is detected as CLANG or GCC, apply corresponding setting with default "difficulty" from above (or `all` otherwise)
** intentionally undocumented `gcc-legacy` pre-set (applied from `legacy` argument if compiler is GCC or detected claims to be compatible), this is the default value of the `--enable-warnings` option to behave same as before this PR: apply a few non-fatal warnings if compiling with GCC

As a side effect, this simplifies the Travis CI test case specifications and allows to define e.g. Jenkins test matrix or local workstation builds more easily and hands-off. This simplification may play a role in addressing issue #869 at some point.

Hopefully this mechanism is extensible enough if we would need to add more compilers and/or "difficulty levels" in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -913,9 +913,6 @@ _matrix_osx_gnustd_nowarn:
       - $HOME/Library/Caches/Homebrew
       - $HOME/.ccache
 
-# -Wno-reserved-id-macro -- configure script tends to define _GNU_SOURCE_, __EXTENSIONS__ etc.
-# -Wno-unused-macros -- system headers define a lot of stuff we do not use, gotta be fatal right?
-# -Wno-padded -- NSPR and NSS headers get to issue lots of that
 _matrix_osx_gnustd_warn:
   include: &_matrix_osx_gnustd_warn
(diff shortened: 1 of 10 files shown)


jimklimov:
[WIP] Travis does not anymore require the few `*-warn-*` test cases that it did. Back to drawing board...

jimklimov:
Now documented and spell-checked. Planning to merge soon.

PR #947, opened 2020-12-28
Introduce a dynamic Jenkinsfile pipeline script to act similar to Travis CI script

jimklimov:
This is a rough-draft setup that works in my local laptop-based testing across several VMs. On one hand this includes a fair bit of hardcoding to automate in the future, but on another already can serve to illustrate what we can achieve in this direction for addressing issue #869 if need arises (follows up from #946 in that regard, and from #823 overall)...

The resulting run in Jenkins new BlueOcean interface looks like this:

nut-jenkinsfile-869-screenshot


jimklimov:
It seems the Jenkins BlueOcean interface (or the version I have) has a bug about not visibly marking failed steps wrapped into `catchError()` clauses (so that the whole matrix gets tested anyway). Marking with `warnError()` clauses works better:

nut-jenkinsfile-869-screenshot-warnError

jimklimov:
This effort taught me quite a bit about limitations of the standard Matrix in Jenkins Pipelines, and led to development of our own hopefully-generic dynamatrix and a pipeline for NUT based on it. Parts of this PR went there.

PR #954, opened 2021-01-03
Modbus driver for Huawei UPS2000/2000A (1kVA-3kVA)

biergaizi:
I recently purchased a Huawei UPS2000 for my server, but the vendor only provided a Windows management program. The program is heavy, it's written in Java that comes with a whole web server and a MySQL database, and it goes without saying that it could not be used in other operating systems - and they want to sell people an SNMP adapter card for that, which would make the UPS to be as expensive as other competing vendors. No thanks...

Fortunately, documentation is publicly available. I reasoned that it would be a waste of effort if it remains unpublished for my private use - implementing it as a NUT driver and upstreaming the driver could benefit others, thus I'm now submitting a Pull Request.

This is only an initial draft of Huawei UPS2000 (1kVA-3kVA) driver on the RS-232 interface. This driver is still a work-in-progress, so far only voltage, frequency, current, power, and temperature monitoring is implemented (I think submitting it right now is helpful to get some useful feedback), I plan to also implement UPS status flag and battery monitoring in the following day, and also to do some coding style fixes. 

Notably, I had to work around a significant technical problem. The UPS uses a Modbus-based serial protocol, and right now in NUT, Modbus is already supported via `libmodbus`. However, `libmodbus` has a serious technical limitation - it only supports register reads and writes. In this driver, it needs to read device identification information to ensures that it has been hooked up to the correct UPS - which is unsupported by libmodbus. Worse, although libmodmus supports sending custom commands via `modbus_send_raw_request()`, but `modbus_receive_confirmation()` assumes the response message is formatted in a certain way with a length in the header - which is not true for our purpose - it simply stops reading in the middle. To add insult to injury, `libmodbus` does not provide any low-level public API for us to receive the raw message from the serial device. Thus, implementing device identification is literally impossible in `libmodbus`. 

To solve the problem, I devised the following workaround:

1. Open the serial port via `ser_open()`
2. Open the same serial port via libmodbus.
3. Construct a Modbus device identification message manually via `ser_send_buf()` and `ser_get_buf()`, and manually parse the response. Also, the CRC-16 verification code has to be copied and pasted into the driver.
4. After the device identification sequence has been completed, sending and receiving normal Modbus registers reads via libmodbus as usual. Because `ser_*()` functions are never used anymore at this point, it's safe to do so. 

Other than that, the rest of the implementation is trivial. See the code comments in the patch for more technical details.

Your code review is appreciated, thanks for your time.

    Signed-off-by: Yifeng Li <tomli@tomli.me>


drivers/huawei-ups2000.c
```diff
@@ -0,0 +1,616 @@
...
+		if (strlen(ups2000_desc[i]) == 0) {
+			fatalx(EXIT_FAILURE, "desc field %d is missing!", i);
+		}
+	}
+}
+
+
+void upsdrv_initinfo(void)
+{
+	bool in_list;
```
biergaizi, line 387: Yes, this variable should be initialized. I need to fix it later... It's strange that even with `-Wall -Wextra`, compilers cannot detect a possible use of uninitialized variable... To be fair, I do aware that it's a hard computation problem.

biergaizi: Fixed.

aquette (approved):
Lgtm, apart maybe from the go-to. But no big deal

biergaizi:
@aquette I just eliminated `goto` in the latest commit - it was purely coincidental, the `goto` was simply no longer needed after some logic reorganization for `status_set()` implementation.

biergaizi:
I just implemented status and alarm reporting. The driver is almost complete.

drivers/huawei-ups2000.c
jimklimov: I can't see the context anymore, but if the originally-uninitialized variable was not read in any codepath before getting assigned, it is deemed okay.

With the "fightwarn" effort I found that most of the time there are no warnings left from clang-8 and older and from gcc-5..gcc-10 (don't have newer locally), so the hordes of warnings to fix come from clang-9, and occasionally from some other set-ups. And then also there are gcc-4.x (and older?) still relevant on systems NUT runs on in practice, whose warnings also matter - but they lately only popped up with introduced uses of GCC pragmas whose syntax support evolved between GCC releases.

biergaizi: @jimklimov The context is: it *was* used in the code path. There was a `if (in_list)` check, and it was possible for an uninitialized `in_list` to make the decision, oops, a bug. But strangely, the compiler never issued any warning, not GCC 9 and not clang 10. Thus my comment about correcting issuing warnings for uninitialized variable is a hard computation problem...

biergaizi:
I must have triggered a bug in the the LGTM.com static analyzer. It says

     uint16_t val;
    ...
    if (val != 0 || val != 1)
    /* Alert: Comparison is always true because val <= 0. */

This alert is obviously nonsense...

biergaizi:
95% complete, nearly everything is implemented, only need some minor tweaks.

I also successfully recruited an alpha tester entirely by accident (out of curiosity I searched my UPS firmware revision, and unexpectedly came across a NUT tutorial in a forum written by an author who happens to own the UPS2000-G, is it the most unlikely way to find testers...). He's busy at work but should be able to report back within a month. When I get the report, the man page can be updated to include a full list of supported model.

jimklimov:
So, how it it looking to the author? :) Is this still "WIP" or feature-complete - so can the PR be reviewed and merged? Thanks :)

jimklimov:
Gentle bump? :)

yurym:
Huawei UPS2000 series are clones of FSP upses.

biergaizi:
Apologize for my absence. Back in January, I originally found two potential testers on the web, however, one couldn't get the driver working for some reasons, and another doesn't have the time to do the testing. Meanwhile I was also get distracted by other projects, so kept this pull request aside for a while.

Today I was fortunate enough to find another tester. The result was a success. In the following days, I'll rebase my driver, do a minor update to the driver and documentation, and complete the final revision of this pull request.

biergaizi:
@yurym Interesting. Does it mean that the same driver codebase can be used on those UPS units as well?

biergaizi:
Also, if there's no further technical problems left, I'd like to see the RFC earlylog patch (#965) getting merged as well. Otherwise this driver will be affected by the missing log problems in #965.

jimklimov:
Ok, #965 is merged now :)

biergaizi:
It's ready for a final review. The driver may still have some sharp edges, but this should be good enough for a `DRV_EXPERIMENTAL`. -;)

biergaizi:
Please do not merge the PR yet. I just realized the device driver for the MaxLinear RX21V1410 USB chip has been upstreamed since Linux 5.13, it means the driver should also support USB by now. I need to do some retests, and if it really works, update the code comments and documentations. On the other hand, no code change is necessary, it's strictly a kernel update, code reviews are still free to proceed.

biergaizi:
USB support is confirmed on Linux 5.12. Code comments and documentation has been updated. It's the final version of this pull request, ready for review.

biergaizi:
Bump. This PR is merge-ready since last week.

jimklimov (approved):
Sorry it took a while to rectify the NUT CI side as well to get to the PRs in queue, great thanks for your contribution and patience!

Issue #953, opened 2021-01-03
Cannot shutdown APC SU-1000 with AP9617


dwmw2:
When I attempt to shut down I get an error:
```
   0.121544	Initiating UPS shutdown
   0.121547	[D1] upsdrv_shutdown...
   0.121552	[D2] entering su_setOID(instcmd, shutdown.return, (null))
   0.121559	[D3] su_find_info: "shutdown.return" found
   0.121565	[D1] entering nut_snmp_set(.1.3.6.1.4.1.318.1.1.1.6.1.1.0, t, 200)
   0.151626	[su1000] nut_snmp_set: can't set .1.3.6.1.4.1.318.1.1.1.6.1.1.0: Error in packet: wrongType (The set datatype does not match the data type the agent expects)
   0.151632	[D1] su_setOID: cannot execute command 'shutdown.return'
```

Using v2.7.4 from Fedora 33 the error is different:
```
   0.086717	Initiating UPS shutdown
   0.086720	upsdrv_shutdown...
   0.086726	Unknown template type: shutdown.return
   0.086732	entering su_instcmd(shutdown.return, (null))
   0.086740	su_find_info: "shutdown.return" found
   0.086745	entering nut_snmp_set(.1.3.6.1.4.1.318.1.1.1.6.1.1.0, i, 2)
   0.127311	[su1000] nut_snmp_set: can't set .1.3.6.1.4.1.318.1.1.1.6.1.1.0: Error in packet: (genError) A general failure occured
   0.127321	su_instcmd: cannot set value for shutdown.return
```

```
 $ snmpget  -u nut 192.168.122.104 .1.3.6.1.4.1.318.1.1.1.6.1.1.0
SNMPv2-SMI::enterprises.318.1.1.1.6.1.1.0 = INTEGER: 1

 $ snmpset  -u nut 192.168.122.104 .1.3.6.1.4.1.318.1.1.1.6.1.1.0 i 2
Error in packet.
Reason: (genError) A general failure occured
Failed object: SNMPv2-SMI::enterprises.318.1.1.1.6.1.1.0

 $ snmpset  -u nut 192.168.122.104 .1.3.6.1.4.1.318.1.1.1.6.1.1.0 t 200
Error in packet.
Reason: wrongType (The set datatype does not match the data type the agent expects)
Failed object: SNMPv2-SMI::enterprises.318.1.1.1.6.1.1.0
```
Using SNMPv1 doesn't make much difference.

```
[...]

PR #956, opened 2021-01-08
snmp-ups: fix instant commands using numeric parameters

jimklimov:
Instant commands using numeric parameters, like on/off/cycle
commands, were failing due to a bad value submitted type (time
instead of integer)

Cherry-picked from https://github.com/42ity/nut/pull/120


clepple:
@aquette @jimklimov do you think this would fix #953?

PR #961, opened 2021-01-20
SNMP MIB subdrivers: rename lookup tables to unique identifiers

jimklimov:
Ported from 42ity fork, this PR makes the SNMP MIB helper lookup table names unique across the codebase.

Originally this was required by DMF capability (to load tables by names from XML descriptor files), but now having the same fix in upstream also helps cross-pollination of the fork and upstream repos (more relevant diffs) as they both actively evolve.

Uniqueness verifiable with:
````
$ grep info_lkp_t drivers/*-mib.c | sed 's,^.*info_lkp_t.* \(.*\) =.*$,\1,' | sort | uniq -c | sort -n
````

Note there are some commented-away entries (seen by leading asterisk in the line when just grepping)

### DIFF
--- a/drivers/apc-ats-mib.c
+++ b/drivers/apc-ats-mib.c
@@ -29,26 +29,26 @@
 #define APC_ATS_SYSOID       ".1.3.6.1.4.1.318.1.3.11"
 #define APC_ATS_OID_MODEL_NAME ".1.3.6.1.4.1.318.1.1.8.1.5.0"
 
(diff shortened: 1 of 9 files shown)


clepple (approved):
*shrug* they're static, so hopefully this is a no-op as far as code generation. As usual, no SNMP hardware to test with here.

jimklimov:
Yes, for existing static `snmp-ups` drivers this should be just a cosmetic change in the source and none exposed in the binary.

Issue #962, opened 2021-01-23
 Voltage value from upsc is 0 on AVR750U (0x3024).


tjmc:
From upsc
```                                                                                                                                                                                                                                          battery.charge: 100                                                                                                                                                                                                                          battery.charge.warning: 30                                                                                                                                                                                                                   battery.runtime: 1480
battery.type: PbAc
battery.voltage: 0.0
battery.voltage.nominal: 12.0                                                                                                                                                                                                                device.mfr: Tripp Lite                                                                                                                                                                                                                       device.model: AVR750U                                                                                                                                                                                                                        device.serial: 3028BV4OM882401230                                                                                                                                                                                                            device.type: ups                                                                                                                                                                                                                             driver.name: usbhid-ups                                                                                                                                                                                                                      driver.parameter.pollfreq: 30                                                                                                                                                                                                                driver.parameter.pollinterval: 2                                                                                                                                                                                                             driver.parameter.port: /dev/bus/usb/002/009                                                                                                                                                                                                  driver.parameter.productid: 3024                                                                                                                                                                                                             driver.parameter.synchronous: no
driver.version: 2.7.4-2112-gb263e476
driver.version.data: TrippLite HID 0.82
driver.version.internal: 0.43                                                                                                                                                                                                                input.frequency: 6020.0                                                                                                                                                                                                                      input.transfer.high: 145.0
input.transfer.low: 89.0
input.voltage: 0.0
input.voltage.nominal: 120                                                                                                                                                                                                                   output.current: 120.0                                                                                                                                                                                                                        output.frequency: 6020.0
output.frequency.nominal: 60
output.voltage: 0.0
[...]

PR #963, opened 2021-01-23
Add Tripplite AVR750U (0x3024). Use smart1500lcdt_scale 

tjmc:
Fix for https://github.com/networkupstools/nut/issues/962

### DIFF
--- a/drivers/tripplite-hid.c
+++ b/drivers/tripplite-hid.c
@@ -80,6 +80,7 @@ static usb_device_id_t tripplite_usb_device_table[] = {
 	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x1003), battery_scale_0dot1 },
 	/* e.g. TrippLite AVR750U */
 	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x1007), battery_scale_0dot1 },
+	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x3024), smart1500lcdt_scale },
 	/* e.g. TrippLite ECO550UPS */
 	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x1008), battery_scale_0dot1 },
 	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x1009), battery_scale_0dot1 },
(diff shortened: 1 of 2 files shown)


drivers/tripplite-hid.c
```diff
@@ -80,6 +80,7 @@ static usb_device_id_t tripplite_usb_device_table[] = {
 	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x1003), battery_scale_0dot1 },
 	/* e.g. TrippLite AVR750U */
 	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x1007), battery_scale_0dot1 },
+	{ USB_DEVICE(TRIPPLITE_VENDORID, 0x3024), smart1500lcdt_scale },
```
jimklimov, line 83: Thanks! Looking at precedent in this table, the ID's are alphabetically sorted which helps maintaining it without conflicts. Can you please move the line lower, probably adding a comment with the device model again (maybe adjusted for difference from existing AVR750U entry... "newer model"? revision/years known?)

jimklimov (commented):
Looks useful; a little bit to fix for easier code maintenance.

And I can't test with a device, so have to hope the change is correct for the hardware in question :)

tjmc:
> 
> 
> Looks useful; a little bit to fix for easier code maintenance.
> 
> And I can't test with a device, so have to hope the change is correct for the hardware in question :)

Thanks for looking at the request. 

from https://github.com/networkupstools/nut/issues/962

```/lib/nut/usbhid-ups -D -a office -u root```
```
   0.044669     [D1] Path: UPS.PowerSummary.Input.Voltage, Type: Feature, ReportID: 0x31, Offset: 0, Size: 16, Value: 0.001192
   0.065565     [D1] Path: UPS.BatterySystem.Battery.Voltage, Type: Feature, ReportID: 0x20, Offset: 0, Size: 16, Value: 0.000137
   0.079124     [D1] Path: UPS.PowerConverter.Input.Voltage, Type: Feature, ReportID: 0x18, Offset: 0, Size: 16, Value: 0.001192
```

I hope I understood the response to make sure it's correct for the newer AVR750U

Issue #821, opened 2020-10-07
Looking at C sources it seems that using '-D' precludes backgrounding, but the nuance is not mentioned in usage()


jimklimov:
* For drivers:
In help text at https://github.com/networkupstools/nut/blob/1db6f1cef880f37f0c8ed4db521991f7175fac81/drivers/main.c#L109 it does not convey that backgrounding will be disabled. At https://github.com/networkupstools/nut/blob/1db6f1cef880f37f0c8ed4db521991f7175fac81/drivers/main.c#L615-L616 it prepares for backgrounding, and at https://github.com/networkupstools/nut/blob/1db6f1cef880f37f0c8ed4db521991f7175fac81/drivers/main.c#L718-L721 it actually does it only if `(nut_debug_level == 0)` among other conditions.
* For server:
Similarly for the server part, making the choice at https://github.com/networkupstools/nut/blob/d5e29025fed7aac5f6fe3400bf17164cd4a6150a/server/upsd.c#L1282 and printing the usage at https://github.com/networkupstools/nut/blob/d5e29025fed7aac5f6fe3400bf17164cd4a6150a/server/upsd.c#L1080
* For upsmon:
Decision is made at https://github.com/networkupstools/nut/blob/f6167703fcdba4c351436e7dcb098045f5ac35ba/clients/upsmon.c#L1973 and usage printed at https://github.com/networkupstools/nut/blob/f6167703fcdba4c351436e7dcb098045f5ac35ba/clients/upsmon.c#L1652

TODO:

- [ ] Update usage() in main.c
- [ ] Update help text in upsd.c
- [ ] Update help text in upsmon.c
- [x] Revise if READMEs and manpages say that

jimklimov:
For manpages:
* Documented for drivers in general: https://github.com/networkupstools/nut/blob/a87849c3a05d25f9d58b023946f11b9a69a8aba6/docs/man/nutupsdrv.txt#L72
* Sort of mentioned for upsmon: https://github.com/networkupstools/nut/blob/a87849c3a05d25f9d58b023946f11b9a69a8aba6/docs/man/upsmon.txt#L48
* [ ] Not documented for upsd: https://github.com/networkupstools/nut/blob/a87849c3a05d25f9d58b023946f11b9a69a8aba6/docs/man/upsd.txt#L47

jimklimov:
[...]

PR #965, opened 2021-01-23
[RFC] Write early startup logs from upsdrv_initinfo/updateinfo() to syslog

biergaizi:
When a NUT driver is started (via upsdrvctl start), it runs the following sequence:

1. Run as a foreground process.
2. Do some basic initialization, log to stderr.
3. If not debugging:
   - fork()
   - Run as a daemon, stderr off, syslog on.

The problem, however, is that `upsdrv_initinfo()` and `updateinfo()` are first called before forking, as a result, the log messages is only sent to stderr, not syslog. On a traditional init + syslog server setup, those startup messages are lost. However, those lost messages can be urgent - for example, it can be a warning about a failed UPS inverter. If the server is freshly started, the admin may miss them until much later, or keep staring at the mysterious alarm word without finding anything relevant in syslog. Similarly, other types of warnings, such as a rounded or clipped `ondelay` value, should also be visible in syslog.

When the driver is running in non-debug fork mode, this patch checks the debugging flags and turns on the syslog earlier, allowing the logs during the initial `upsdrv_initinfo()` and `upsdrv_updateinfo()` to be recorded in syslog before forking.

Another potential solution - checking `isatty()` and turning on syslog if the output is not - has also been suggested by Rob Crittenden. But...

1. It cannot reliably tell whether the output is visible. A driver started by init still has access to the system console before it forks, even if nobody is here to read it.
2. Running the driver in debug mode and redirecting stdout/stderr to a file is common during troubleshooting. If we switch between stderr/syslog using isatty(), the debug log is unexpectedly sent to syslog instead, so we still need to check the debug flags first.
3. Another option is not to touch stderr, keep it as-is, and only to decide whether to use syslog. But we don't fork in debug mode, and a non-daemon should not spam the syslog. So we still need to check the debug flags. 

So in the end, I think what matters is the debug flag, not the output. 

Mailing list discussion:

https://alioth-lists.debian.net/pipermail/nut-upsdev/2021-January/007508.html

> Signed-off-by: Yifeng Li <tomli@tomli.me>


biergaizi:
Found some related open issues: #821 #349 #683

It seems the backgrounding/foregrounding policy in the current project is subject to change.

jimklimov:
Ok, the proposed change seems a bit orthogonal to issues on the related subject linked above (in the sense that fixes there, like separation of "debug means foreground" into separately tunable flags, are still doable with this feature added).

PR #968, opened 2021-01-26
add snr subdriver to blazer ups driver

iryabovnag:
This request add support for the SNR-UPS-LID-600 by adding a subdriver snr to blazer_usb driver

### DIFF
--- a/drivers/blazer_usb.c
+++ b/drivers/blazer_usb.c
@@ -324,6 +324,79 @@ static int krauler_command(const char *cmd, char *buf, size_t buflen)
 }
 
 
+static int snr_command(const char *cmd, char *buf, size_t buflen)
+{
+	/*
+	 * Still not implemented:
+	 * 0x6	T<n>	(don't know how to pass the parameter)
+	 * 0x68 and 0x69 both cause shutdown after an undefined interval
+	*/
+	const struct {
+		const char	*str;		/* Megatec command */
+		const int	index;	/* SNR string index for this command */
+		const char	prefix;	/* character to replace the first byte in reply */
+	}  command[] = {
+		{ "Q1\r", 0x03, '(' },
+		{ "F\r",  0x0d, '#' },
+		{ "I\r",  0x0c, '#' },
+		{ "T\r",  0x04, '\r' },
+		{ "TL\r", 0x05, '\r' },
+		{ "Q\r",  0x07, '\r' },
+		{ "C\r",  0x0b, '\r' },
+		{ "CT\r", 0x0b, '\r' },
+		{ NULL, 0, '\0' }
+	};
+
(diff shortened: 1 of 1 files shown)


aKazancev96 (approved):

drivers/blazer_usb.c
```diff
@@ -360,9 +433,19 @@ static void *phoenix_subdriver(USBDevice_t *device)
...
+
+	subdriver_command = &snr_command;
+	return NULL;
+}
+
+
 static usb_device_id_t blazer_usb_id[] = {
 	{ USB_DEVICE(0x05b8, 0x0000), &cypress_subdriver },	/* Agiler UPS */
 	{ USB_DEVICE(0x0001, 0x0000), &krauler_subdriver },	/* Krauler UP-M500VA */
+	{ USB_DEVICE(0x0001, 0x0000), &snr_subdriver },		/* SNR UPS LID600 */
```
jimklimov, line 448: I am concerned about use of same values as in the previous line to identify the USB device. Alas, some vendors do take shortcuts and/or use same controllers with different firmware, or as seems to be case here - hijack(?)/use a sort of unregistered value.

This may need an end-user to reference explicitly the `snr` subdriver in their `ups.conf` settings.

Which reminds me, docs were not updated by this PR e.g. https://github.com/networkupstools/nut/blob/master/docs/man/blazer-common.txt#L154 and probably nut.dict will sant to accept the new valid word.

Also, if this work is done for a device vendor, you're encouraged to update https://github.com/networkupstools/nut/blob/master/docs/acknowledgements.txt

jimklimov (changes requested):

drivers/blazer_usb.c
jimklimov: Also, looking at the closed #967 - do not worry about posting the PR'ed change all as one commit, there may be many small meaningful changes to make a feature and PR (and there's rebasing and squashing for small commits like typo fixes) ;)

jimklimov:
Hello again, and sorry for lagging so much on this small contribution. I was recently advised by the previous maintainer that blazer family of drivers were an older generation (should get removed eventually), and newer code for compatible devices should better land into nutdrv_qx* driver sources. Your change seems quite compact - can you please see if applying this USB dialect subdriver to nutdrv_qx.c directly would work for your UPS? If so, that would be a much better PR in terms of future maintenance. Thanks!

iryabovnag:
Hi @jimklimov. Thanks for your feedback. 
Yes, I saw nutdrv_qx* and will move my PR to it. 
I am going to improve the calculation of the battery.runtime. It calculates only by the battery.charge and precision is very poor. I think need some updates with battery.voltage considering.

PR #974, opened 2021-02-11
Add USB device number matching

efuss:
I needed a `device` argument for USB matching in order to independently match two identical UPSen attached to the same bus.

The changes to usb-common.[hc] and libusb.c are straightforward (sans the `DEBUG_EXACT_MATCH_DEVICE`, which I modelled after `DEBUG_EXACT_MATCH_BUS`, whatever that is used for).

The changes to the various USB drivers are mechanical. Maybe some `#define` for the regex array size (or even seven more `#defines` for the magical index values) would be useful.

I documented the `device` field only where the existing `bus` field was documented (which is not the case everywhere it's actually used).

### DIFF
--- a/drivers/usb-common.c
+++ b/drivers/usb-common.c
@@ -117,6 +117,13 @@ static int match_function_exact(USBDevice_t *hd, void *privdata)
 		return 0;
 	}
 #endif
+#ifdef DEBUG_EXACT_MATCH_DEVICE
+	if (strcmp_null(hd->Device, data->Device) != 0) {
+		upsdebugx(2, "%s: failed match of %s: %s != %s",
+		    __func__, "Device", hd->Device, data->Device);
+		return 0;
+	}
+#endif
 	return 1;
 }
 
@@ -152,6 +159,9 @@ int USBNewExactMatcher(USBDeviceMatcher_t **matcher, USBDevice_t *hd)
 #ifdef DEBUG_EXACT_MATCH_BUS
 	data->Bus = hd->Bus ? strdup(hd->Bus) : NULL;
 #endif
+#ifdef DEBUG_EXACT_MATCH_DEVICE
+	data->Device = hd->Device ? strdup(hd->Device) : NULL;
+#endif
 	*matcher = m;
 
 	return 0;
@@ -173,6 +183,9 @@ void USBFreeExactMatcher(USBDeviceMatcher_t *matcher)
 	free(data->Serial);
 #ifdef DEBUG_EXACT_MATCH_BUS
 	free(data->Bus);
+#endif
+#ifdef DEBUG_EXACT_MATCH_DEVICE
+	free(data->Device);
 #endif
 	free(data);
 	free(matcher);
@@ -285,7 +298,7 @@ static int match_regex_hex(regex_t *preg, int n)
 
 /* private data type: hold a set of compiled regular expressions. */
 typedef struct regex_matcher_data_s {
-	regex_t	*regex[6];
+	regex_t	*regex[7];
 } regex_matcher_data_t;
 
 /* private callback function for regex matches */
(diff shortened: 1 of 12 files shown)


drivers/usb-common.c
```diff
@@ -361,18 +374,23 @@ static int match_function_regex(USBDevice_t *hd, void *privdata)
 		    __func__, "Bus", hd->Bus);
 		return r;
 	}
+
+	r = match_regex(data->regex[6], hd->Device);
+	if (r != 1) {
+		return r;
```
jimklimov, line 380: To be consistent with other code around, can you please add the `upsdebugx()` call for the failure, probably similar to one you added at https://github.com/networkupstools/nut/pull/974/files#diff-6aea68febe9f0fc9ad47c0d821a43073c65461e02ca9f0648ba78981482f98b8R122 ?

Otherwise, looks good to me.

efuss:
> To be consistent with other code around
Sure. I thought I'd done that, but probably I missed that part when adapting my patch (against the stable 2.7.4 version) to Git master.

cd33fc706ff0766ac2a3cc813f7fbff7d1229d9d.

jimklimov (approved):
LGTM, so going to merge soon unless some other maintainers voice objections?

One minor nit, inherited from earlier blocks in the readme's, is the Linux-specific path names and `lsusb`, but this can be addressed later to disclaim those examples: `e.g. lsusb(8) on Linux` or some such wording.

clepple:
> LGTM, so going to merge soon unless some other maintainers voice objections?
> 
> One minor nit, inherited from earlier blocks in the readme's, is the Linux-specific path names and `lsusb`, but this can be addressed later to disclaim those examples: `e.g. lsusb(8) on Linux` or some such wording.

`lsusb` is one of the first things I install from the ports tree on *BSD :-) so I don't think we need to call it out as Linux-specific. For the other instances of the device, we could maybe reference driver debug messages.

clepple (commented):
I'm on the fence about this. The mechanics of the patch are solid, but device numbers are not likely to be stable. I'd maybe consider this an interim fix until someone has time to look at the physical port location stuff that is specific to libusb-1.0 (and document the options as such).

I am also concerned about how this will affect mergeability of #300.

efuss:
> device numbers are not likely to be stable
But it's all we've got, no?

For my Masterguard UPSen, I check the "slave address" (ups.id) to make sure I'm not talking to the wrong one.

Unfortunately, the underlying logic in NUT is that once the USB matching has succeeded, the driver/protocol layer can only claim or refuse to claim, but refusing to claim will not allow another driver instance to claim the same device.

So combining device and slave_address parameter at least gives me a chance that it will attach correctly and fail if device numbers changed.

clepple:
> > device numbers are not likely to be stable
> But it's all we've got, no?

this is the physical port location stuff I was thinking of: https://stackoverflow.com/questions/14279796/linux-libusb-get-usb-device-path

> For my Masterguard UPSen, I check the "slave address" (ups.id) to make sure I'm not talking to the wrong one. Unfortunately, the underlying logic in NUT is that once the USB matching has succeeded, the driver/protocol layer can only claim or refuse to claim, but refusing to claim will not allow another driver instance to claim the same device. So combining device and slave_address parameter at least gives me a chance that it will attach correctly and fail if device numbers changed.

I just think that if we are merging this into the main branch for a release, we should set expectations that some manual intervention will be needed when using bus and device number matching. Fixing the claim logic or testing the libusb-1.0 branch would be even better, IMHO.

efuss:
> we should set expectations that some manual intervention will be needed when using bus and device number matching.
What about the disclaimers I just added in 92672be3?

clepple:
Maybe "Note that device numbers are not likely to be stable across reboots or device re-plugging."

efuss:
> Maybe "Note that device numbers are not likely to be stable across reboots or device re-plugging."
I was just in the process of re-phrasing it, but then re-consindered it.
I don't want to start nit-picking, but if you re-boot without fiddling with the cabling in betweeen, one would expect the attachments being done the same way as last time, no?

clepple:
> I was just in the process of re-phrasing it, but then re-consindered it. I don't want to start nit-picking, but if you re-boot without fiddling with the cabling in betweeen, one would expect the attachments being done the same way as last time, no?

Ideally, yes. However, let's say you plug in, unplug, plug in, configure with device number, reboot - the device number will likely be lower after the reboot. A newer kernel might also reorder the devices.

jimklimov:
Am I correct assuming we are still discussing about, or waiting for, a commit with better caveats-emptor warning to the documentation for this feature?

efuss:
> Am I correct assuming we are still discussing about, or waiting for, a commit with better caveats-emptor warning to the documentation for this feature?
No idea. I regard my warning as sufficient, but I'm not the project and not in a position to make a decision.

eski-dev:
Forgive me as I am not well versed in github etiquette, not do I have anything particularly useful to add to the conversation, however this is a feature that would be key to my current roadblock implementing two identical Tripp Lite SMART1000RM1U UPSes. Is there an estimated timeline for having this change formally added to the code?

jimklimov:
Sorry for stalling this a bit, hope to merge soon now. Thanks for the contribution!

Just in case: was this tested for setups without a specified `device` configuration option, that this would not preclude identifying a single attached device, etc.? By code looks like it would behave same as unspecified any other such parameter, but got me wondering how we really handle empty/null values there...

efuss:
> Just in case: was this tested for setups without a specified `device` configuration option
I'm afraid not. I always have two identical UPSen connected to a server.

PR #975, opened 2021-02-11
Add Masterguard A/E driver based on Q* framework

efuss:
Add a new driver for Masterguard A and E series based on the Q* framework (_nutdrv_qx_).

Based on vendor documentation of the command set and USB protocol I was both asked not to disclose.

Supports “new” A700/1000/2000/3000(-19) and E40/60/100(-19) (latter totally untested due to lack of hardware).

This is still sort of work-in-progress due to a bunch of open questions on NUT integration:

### General/Q* related:

- `instcmd()` gets an `extradata` parameter which the Q* framework passes to the handler function (in `value`). Where does that data come from? Is there any documentation on it?
Looking at example code, **test.battery** seems to use that data, while **shutdown.{reboot,stayoff}** use **ups.delay.{shutdown,start}** instead. Is that simply for historical reasons?

- I need to issue a command to the UPS in order to learn the supported values for **output.voltage.nominal**. In the Q* framework, is there a way to simply ignore the result (after it has been digested in a preprocess function to set the corresponding `info_rw_t` enum, of course)? I currently use a custom var **output_voltages** for it.

- On certain commands, the UPS doesn't report an answer. Not an empty one, just nothing. Is there a saner way to handle that than the USB subdriver knowing these commands and skipping the result-reading part of the function for them?

- In case a variable can be modified by an instcmd (e.g. **ups.beeper.status** by **beeper.enable**), how do I express that in the `qx2nut` table?

- In case modifying one variable (**output.voltage.nominal**) is going to modify another one (**output.voltage.current**), how am I supposed to express that?

- Given the Q* framework, how can I report battery.charger.status? It depends on several command's output.

### Translating my UPS's values to NUT:

- The UPS reports the battery's running time (not in the **battery.runtime** sense, the time since last replacement). How do I report that (I'm using ups.uptime now, which doesn't really seem correct)?

- The UPS has a (fixed-size) fault log (and a command to retrieve it, of course). I'm now using custom variables for it? Better ideas?

- The UPS reports the nominal hold time at both 50% and 100% load. Is there a way to report this in NUT variables?

### Variable/Command intended semantics:

- Am I supposed to report **ups.model/.mfr** when I report **device.***?

- What's the difference between the **load.off.delay** and **shutdown.stayoff** commands?

- What different from **test.battery** is **test.failure.start** supposed to do? Deploy the battery and then turn off the load?

- What's **calibrate.start** different from **test.battery.start.deep**?

- What's the difference between **ambient.temperature.high** and **ambient.temperature.high.critical**?

- Am I supposed to report both **ups.temperature** and **ambient.0.temperature**?

- The documentation of the **ups.shutdown** variable says “Enable or disable UPS shutdown ability (poweroff)” and the _mge-hid_ driver is the only one to use it. Is this a safeguard against accidentally issuing the shutdown commands?


efuss:
Some CI script seems to choke on
 ```
{ USB_DEVICE(0x06da, 0x0002),   "Phoenixtec Power","USB Cable (V2.00)", &phoenixtec_subdriver },/* Masterguard A Series */
```

because of the parenthesis:

```
Regenerating the USB helper files...
Unmatched ( in regex; marked by <-- HERE in m/(#define|#DEFINE)\s+"USB Cable ( <-- HERE V2.00\s+(\dx(\d|\w)+)/ at ./nut-usbinfo.pl line 279.
```

**Comment by efuss** [comment]
Fixed the simple-minded matching in `tools/nut-usbinfo.pl`.

**Comment by jimklimov** [comment]
What is `vis.h` and where does it come from? Is that a not-committed part of the PR development? Or a third-party dependency that becomes required to build nutdrv_qx driver?

File: drivers/nutdrv_qx.c [hunk 1]
```diff
@@ -1209,6 +1216,68 @@ static int	fuji_command(const char *cmd, char *buf, size_t buflen)
...
+	if (e != NULL && ++e < buf + buflen) {
+		*e = '\0';
+		if (nut_debug_level >= 4) {
+			stravis(&vis, buf, VIS_NL | VIS_CSTYLE);
+			upsdebugx(4, "< %s", vis);
+			free(vis);
+		}
+		return e - buf;
+	} else {
+		upsdebugx(3, "read: buflen %zd too small", buflen);
```
jimklimov, line 1275: The `buflen` is unsigned `size_t`, so format value is `%zu`. Kudos for `%z` part though ;)

drivers/nutdrv_qx.c
```diff
@@ -1209,6 +1216,68 @@ static int	fuji_command(const char *cmd, char *buf, size_t buflen)
...
+			*buf = '\0';
+			return ret;
+		}
+		if ((e = memchr(p, '\r', ret)) != NULL) break;
+	}
+	if (e != NULL && ++e < buf + buflen) {
+		*e = '\0';
+		if (nut_debug_level >= 4) {
+			stravis(&vis, buf, VIS_NL | VIS_CSTYLE);
+			upsdebugx(4, "< %s", vis);
```
jimklimov, line 1270: Already commented about the use of `vis.h` and `stravis()` - I think these can be dropped for many benefits. Also (and especially then) the check for `if (nut_debug_level >= 4) {...}` also becomes redundant since `upsdebugx(4...)` already checks for the requested level.

Same for the case ~50 lines above.

drivers/nutdrv_qx.c
```diff
@@ -40,6 +42,8 @@
 #include "attribute.h"
```
jimklimov, line 42: Note to self: bump DRIVER_VERSION when merging is imminent

jimklimov (changes requested):

efuss:
> What is `vis.h` and where does it come from?
Sorry, that's a BSD extension.

> Is that a not-committed part of the PR development?
I don't get that.

> Or a third-party dependency that becomes required to build nutdrv_qx driver?
It's used in level 4 debugging code.
As strvis() etc. are non-POSIX (sorry, forgot to check), that debugging code may simply be dropped.

efuss:
> The `buflen` is unsigned `size_t`, so format value is `%zu`.
Yes, sorry.

> the check for `if (nut_debug_level >= 4) {...}` also becomes redundant
Yes, sure. I just wanted to save the strvis() in case nothing is printed.

jimklimov:
>> Is that a not-committed part of the PR development?
>I don't get that.

My original guess, before I found the BSD header, was that you developed
some "vis" helper codebase but did not commit it. That misunderstanding is
now resolved :)

efuss:
> The `buflen` is unsigned `size_t`, so format value is `%zu`.
082e40f6ed4783dc992ee61671053149946dfafb

efuss:
> What is `vis.h` and where does it come from?
dca9a523c21686e1249e07c818dc6b6a235d3b67

jimklimov:
````
[2021-04-21 15:11:40] [build-stderr] nutdrv_qx.c: In function ‘phoenixtec_command’:
[2021-04-21 15:11:40] [build-stderr] nutdrv_qx.c:1258:5: warning: ‘e’ may be used uninitialized in this function [-Wmaybe-uninitialized]
[2021-04-21 15:11:40] [build-stderr]  1258 |  if (e != NULL && ++e < buf + buflen) {
[2021-04-21 15:11:40] [build-stderr]       |     ^
````

````
[2021-04-21 15:11:43] [build-stderr] nutdrv_qx_masterguard.c: In function ‘masterguard_claim’:
[2021-04-21 15:11:43] [build-stderr] nutdrv_qx_masterguard.c:991:21: error: ‘ptrdiff_t’ undeclared (first use in this function)
[2021-04-21 15:11:43] [build-stderr]   991 |     if (p - cmd >= (ptrdiff_t)sizeof cmd - 1) break;
[2021-04-21 15:11:43] [build-stderr]       |                     ^~~~~~~~~
[2021-04-21 15:11:43] [build-stderr] nutdrv_qx_masterguard.c:26:1: note: ‘ptrdiff_t’ is defined in header ‘<stddef.h>’; did you forget to ‘#include <stddef.h>’?
````

File: drivers/nutdrv_qx_masterguard.c [hunk 4]
```diff
@@ -0,0 +1,1049 @@
...
+	/* several values are deduced from the T field */
+	{ "series",			0,		NULL,			"WH\r",	"",	113,	'(',	"",	16,	16,	"%s",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	masterguard_series },
+	{ "device.model",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	16,	45,	"%s",	QX_FLAG_STATIC,				NULL,	NULL,	masterguard_model },
+	{ "ups.power.nominal",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	16,	45,	"%s",	QX_FLAG_STATIC,				NULL,	NULL,	masterguard_power },
+/* not used, use GS instead because the value is settable
+	{ "battery.packs",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	47,	47,	"%.0f",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+*/
+	{ "input.voltage.nominal",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	49,	51,	"%.0f",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+	{ "input.frequency.nominal",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	53,	57,	"%.2f",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+	{ "number_of_battery_cells",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	59,	61,	"%.0f",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	masterguard_numcells },
```
jimklimov, line 661: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+	 *    01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012
+	 *    0         1         2         3         4         5         6         7         8         9         0         1
+	 *    (00 10.06 03.09 A  700  + 0 Bat Pack-19        0 230 50.00 012 02.30 006.00 012.00 018 10 40 160 276 47.00 53.00
+	 */
+	/* type				flags		rw			command	answer	len	leading	value	from	to	dfl	qxflags					precmd	preans	preproc */
+	{ "ups.id",			ST_FLAG_RW,	masterguard_r_slaveaddr,"WH\r",	"",	113,	'(',	"",	1,	2,	"%.0f",	QX_FLAG_SEMI_STATIC | QX_FLAG_RANGE,	NULL,	NULL,	masterguard_slaveaddr },
+	{ "ups.firmware",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	4,	8,	"%s",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+	{ "ups.firmware.aux",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	10,	14,	"%s",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+	/* several values are deduced from the T field */
+	{ "series",			0,		NULL,			"WH\r",	"",	113,	'(',	"",	16,	16,	"%s",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	masterguard_series },
```
jimklimov, line 653: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+	{ "device.model",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	16,	45,	"%s",	QX_FLAG_STATIC,				NULL,	NULL,	masterguard_model },
+	{ "ups.power.nominal",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	16,	45,	"%s",	QX_FLAG_STATIC,				NULL,	NULL,	masterguard_power },
+/* not used, use GS instead because the value is settable
+	{ "battery.packs",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	47,	47,	"%.0f",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+*/
+	{ "input.voltage.nominal",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	49,	51,	"%.0f",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+	{ "input.frequency.nominal",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	53,	57,	"%.2f",	QX_FLAG_STATIC,				NULL,	NULL,	NULL },
+	{ "number_of_battery_cells",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	59,	61,	"%.0f",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	masterguard_numcells },
+	{ "nominal_cell_voltage",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	63,	67,	"%.2f",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	NULL },
```
jimklimov, line 662: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+	{ "battery.voltage.nominal",	0,		NULL,			"WH\r",	"",	113,	'(',	"",	63,	67,	"%.2f",	QX_FLAG_STATIC,				NULL,	NULL,	masterguard_battvolt},
+	{ "runtime_half",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	69,	74,	"%.0f",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	masterguard_mmm_ss },
```
jimklimov, line 664: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+	{ "runtime_full",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	76,	81,	"%.0f",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	masterguard_mmm_ss },
```
jimklimov, line 665: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+	{ "recharge_time",		0,		NULL,			"WH\r",	"",	113,	'(',	"",	83,	85,	"%.0f",	QX_FLAG_STATIC | QX_FLAG_NONUT,		NULL,	NULL,	masterguard_hhh },
```
jimklimov, line 666: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+
+	/*
+	 * > [GSN,XX\r]
+	 * < [(XX,SSN=SSSSSSSSSSSnnnnn\r]
+	 *    0123456789012345678901234
+	 *    0         1         2
+	 *    (00,SSN=6A1212      2782
+	 */
+	/* type			flags	rw	command		answer	len	leading	value	from	to	dfl	qxflags				precmd				preans	preproc */
+	{ "device.part",	0,	NULL,	"GSN,XX\r",	"",	25,	'(',	"",	8,	18,	"%s",	QX_FLAG_STATIC | QX_FLAG_TRIM,	masterguard_add_slaveaddr,	NULL,	NULL },
```
jimklimov, line 759: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+
+	/*
+	 * > [MSO\r]
+	 * < [(220 230 240\r]
+	 *    0123456789012
+	 *    0         1
+	 *    (220 230 240
+	 */
+	/* type			flags	rw	command		answer	len	leading	value	from	to	dfl	qxflags				precmd	preans	preproc */
+	{ "output_voltages",	0,	NULL,	"MSO\r",	"",	5,	'(',	"",	1,	0,	"%s",	QX_FLAG_STATIC | QX_FLAG_NONUT,	NULL,	NULL,	masterguard_output_voltages },
```
jimklimov, line 782: @aquette @clepple : Any suggestions for standard field naming?

drivers/nutdrv_qx_masterguard.c
```diff
@@ -0,0 +1,1049 @@
...
+	 *    0         1         2         3         4         5         6         7         8         9         0
+	 *    (00,7 0043 0000:16:48:06 0 0000 0000:00:00:00 0 0000 0000:00:00:00 0 0000 0000:00:00:00 0 0000 0000:00:00:00
+	 *    (01,9 0010 1780:14:57:19 7 0046 0000:21:14:41 0 0000 0000:00:00:00 0 0000 0000:00:00:00 0 0000 0000:00:00:00
+	 */
+	/* type		flags	rw	command		answer	len	leading	value	from	to	dfl	qxflags		precmd				preans	preproc */
+	{ "fault_1",	0,	NULL,	"FLT,XX\r",	"",	108,	'(',	"",	4,	23,	"%s",	QX_FLAG_NONUT,	masterguard_add_slaveaddr,	NULL,	masterguard_fault },
+	{ "fault_2",	0,	NULL,	"FLT,XX\r",	"",	108,	'(',	"",	25,	44,	"%s",	QX_FLAG_NONUT,	masterguard_add_slaveaddr,	NULL,	masterguard_fault },
+	{ "fault_3",	0,	NULL,	"FLT,XX\r",	"",	108,	'(',	"",	46,	65,	"%s",	QX_FLAG_NONUT,	masterguard_add_slaveaddr,	NULL,	masterguard_fault },
+	{ "fault_4",	0,	NULL,	"FLT,XX\r",	"",	108,	'(',	"",	67,	86,	"%s",	QX_FLAG_NONUT,	masterguard_add_slaveaddr,	NULL,	masterguard_fault },
+	{ "fault_5",	0,	NULL,	"FLT,XX\r",	"",	108,	'(',	"",	88,	107,	"%s",	QX_FLAG_NONUT,	masterguard_add_slaveaddr,	NULL,	masterguard_fault },
```
jimklimov, line 809: @aquette @clepple : Any suggestions for standard field naming?

jimklimov (commented):
Tentatively merged. I am not sure about some strings in the mapping table, but most of these are listed in `masterguard_makevartable()`.

I suppose we need some way of tagging experimental drivers which add workable support if merged, but may have "uncommitted API" for the interactions which need standardization. Several driver PRs are up in the air due to that...

drivers/nutdrv_qx_masterguard.c
clepple: `device.part` is listed in the 2.7.4 names table, or are you referring to something else?

drivers/nutdrv_qx_masterguard.c
clepple: If this is a set of selectable output voltages, it seems like it would be best to list them as enum options in an R/W variable, but I realize that might be difficult with the qx framework (haven't looked at that much). I have no preference, other than matching an existing name prefix like `output.voltage.*`

drivers/nutdrv_qx_masterguard.c
clepple: Maybe `ups.fault.N` could be a more detailed version of `ups.alarm` (which is just space-separated keywords)?

drivers/nutdrv_qx_masterguard.c
efuss: `battery.numcells`?

drivers/nutdrv_qx_masterguard.c
efuss: `device.series`?

drivers/nutdrv_qx_masterguard.c
efuss: `battery.voltage.percell`?

drivers/nutdrv_qx_masterguard.c
efuss: `battery.runtime.nominal.halfload`?

drivers/nutdrv_qx_masterguard.c
efuss: `battery.runtime.nominal.fulload`?

drivers/nutdrv_qx_masterguard.c
efuss: `battery.charge.time`?

drivers/nutdrv_qx_masterguard.c
efuss: As I wrote in my original message:

I need to issue a command to the UPS in order to learn the supported values for **output.voltage.nominal**. In the Q* framework, is there a way to simply ignore the result (after it has been digested in a preprocess function to set the corresponding `info_rw_t` enum, of course)? I currently use a custom var **output_voltages** for it.

efuss:
There's also a large bunch of questions still open from my original message. Some overlap with the naming questions by @jimklimov, but most don't. I feel slightly uneasy with my guesswork and would appreciate a senior developer commenting on the subject.

jimklimov:
UPDATE: This PR was among the reasons we as community decided in #1046 to add an `experimental.*` namespace to un-block submitted drivers with names that did not follow existing standard patterns.

I merged this one before the decision was made (just missed the notes, I guess), so one useful step before releasing NUT-2.7.5 would be to go over its changes to identify and rename those data points, adding the prefix.

PR #979, opened 2021-02-26
Add support for Ablerex Dk+

ups3628:
Manufacturer: Ablerex
Model:  Dk+ series

This request add support for Ablerex Dk+ series.


jimklimov:
Hello, sorry for the delay and thank you very much for the contribution. I've cleaned up the coding style a little (to have consistent whitespaces), and added the documentation paragraphs to mention this new feature.

Is there anything else the documentation should mention?

jimklimov:
Also, if any user of another blazer_usb subdriver can build and test this codebase to make sure nothing broke for existing support, would be great. From looking at the changes, does not seem such breakage is probable however :)

jimklimov:
For reference, there are more details in the mailing list discussion at https://alioth-lists.debian.net/pipermail/nut-upsdev/2021-March/007517.html

* The "DK+" is a customer model name for ODM.
* The below table is a description of the new driver as we provide.
  Include
  a. Add NUT Function List
  b. Add NUT's function for Each UPS model List
  blazer_usb-ablerex-20210310

jimklimov:
Hello @ups3628 (and @zykh as our QX expert): during a discussion in team, I was reminded that blazer and a number of other drivers for Qx protocols were headed to obsoletion, with their abilities consolidated in a better (more modularly) architected nutdrv_qx driver with its many subdrivers tracked in independent sources `drivers/nutdrv_qx_*.{c,h}`. The old drivers remained mostly for backwards-compatibility of already deployed configurations and to balance regressions possible in the new implementations.

Unfortunately, while this "migration" and obsoletion was mentioned in some documentation here and there, it was not made apparent in the impacted driver sources - I'll post better visible comments right there in heading boilerplates.

So while this was not your fault placing the contribution into the blazer_usb driver, I must ask you to relocate it into a new nutdrv_qx_ablerex subdriver, so that we as a project can maintain it in the long term (e.g. when the obsoleted driver sources may eventually get removed from main codebase) and not lose the support added now for your devices. Also, without the device access we are not in position to rearrange the code and guarantee that it still works correctly.

jimklimov:
Closing in favor of #1135; information provided here is however valuable for future reference.

Issue #960, opened 2021-01-19
alarm_commit() silently truncates a 1024-char message to 256 characters


biergaizi:
Currently, an alarm message is limited to 1024 characters (`LARGEBUF`), a developer may check the source and see `alarm_buf[LARGEBUF]` and naively assuming the alarm buffer is adequate.

    static char     status_buf[ST_MAX_VALUE_LEN], alarm_buf[LARGEBUF];

However, when `alarm_commit()` is called, it attempts to create a status variable, but a status variable is only limited to 256 characters (`ST_MAX_VALUE_LEN`), far below the 1024-character `alarm_buf`. If the UPS alarm string is too long, it's silently truncated without any warning.

    /* write the status_buf into the info array */
    void alarm_commit(void)
    {
            if (strlen(alarm_buf) > 0) { 
                    dstate_setinfo("ups.alarm", "%s", alarm_buf);
                    alarm_active = 1; 
            } else {
                    dstate_delinfo("ups.alarm");
                    alarm_active = 0; 
            }
    }

This is unexpected from a developer's point of view.

1. The alarm message is truncated when it's committed as a variable even if the alarm buffer itself is large enough. And when `alarm_commit()` truncates an alarm, it doesn't return any error code.

2. Nowhere in the documentation has a warning about the length of the alarm string. In fact, the documentation treats a `alarm_set()` call as something different from a `dstate_setinfo()`, without mentioning the fact that `alarm_set()`'s underlying implementation is a  `dstate_setinfo()` and the same restrictions apply.

## Proposed Solution(s)

1. Raise the length limit of a variable to 1024, the same size of `alarm_buf`, or
2. Lower the limit of the `alarm_buf` to 256, the same size of `status_buf`, or
3. Don't silently truncates an alarm message, at least returns an error code, or
4. Keep the existing code, but improve the documentation `new-drivers.txt` to warn the developers about these limitations.

jimklimov:
I agree this is an unexpected trap. For the least surprise - being generally not sure what consumers, including external libs and binaries, exist for a dstate status variable, a safe bet currently seems to be the option number 2 and get these buffer sizes in sync, and maybe add debug messages at least for overflows in `alarm_set()`.

It seems that @aquette was the last to touch all lines with `alarm_buf` so I hope he's available to comment here.

[...]

PR #986, opened 2021-03-12
drivers/dstate.c: limit alarm_buf string to fit into a dstate info…

jimklimov:
…status, and log errors and overflows from alarm_set() [issue #960]

### DIFF
--- a/drivers/dstate.c
+++ b/drivers/dstate.c
@@ -36,7 +36,7 @@
 
 	static int	sockfd = -1, stale = 1, alarm_active = 0, ignorelb = 0;
 	static char	*sockfn = NULL;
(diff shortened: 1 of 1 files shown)


aquette:
I would be in favor of logging the discarded or truncated alarm... We may miss something important!

jimklimov:
Note that originally those `buf` were just discarded quietly - now we began to know this happens :)
Added a commit to also log the contents of the `buf`, if it is a sanely short zero-terminated string (reviving the original `LARGEBUF` limit for that), otherwise the logged message is truncated to that.

aquette (approved):
Lgtm, thx @jimklimov

Issue #840, opened 2020-10-28
Moving away from master/slave terminology


MelGrubb:
Let me start by saying that I do not want to start a political war here, and will not be rendering any judgement of the choice of technology terms which predate my time on this planet.

Having said that, there is an industry-wide trend to move away from the terms "master" and "slave". This recently became an issue for me since I cover configuring NUT in a video course I did. As part of a company-wide move toward more inclusive language, the transcript of my course was flagged for re-evaluation and update. I have explained that since I didn't write NUT, there's not much I can do about it apart from injecting an acknowledgement of the troublesome language into the course. It brings up a problem we all may face in a future though, where tools and technology choices may be evaluated and possibly rejected on criteria other than their technical merit.

As a software developer, and part of the IT world, I realize that the terms are an industry standard, and while they may be considered in poor taste in retrospect, they were not chosen specifically to hurt or injure anyone. Having said that, the entire IT industry has started moving in the direction of shedding these terms and replacing them with main/secondary, leader/follower, or server/client where appropriate. Even uses of the term "master" without an associated "slave" such as with git branches or "master recordings", are being frowned on and replaced.

To maintain backwards compatibility, the existing terms will need to be supported, of course, but can additional terms such as primary/secondary or leader/follower be added as synonyms in the configuration system so that people may choose these alternate terms and achieve the same effect?

clepple:
@MelGrubb Thanks for opening this issue. This has been on my mind for a while, but I think you articulated it better than the draft I was working on.

Part of my dilemma of not publicly raising the issue is related to the gap between the last NUT release and now. We're still getting bug reports from people running 2.7.2 (release six years ago), but we're getting complaints that the latest release (2.7.4) is four years old. I don't want to hold up the next release (which I personally do not have enough time to lead) while we check for unintended side effects of what is ostensibly a simple addition of synonyms. We've already had an embarrassing "just one simple fix" bug (that we know of) while working on the release.

That said, I personally think we do need to pick an alternate set of terms, and even suggest those new terms to users when they upgrade, in order to not propagate master/slave. If NUT were more a part of the "move fast and break things" cult, releasing multiple versions in a year, I think we would be obliged to move more quickly on this. But NUT doesn't move at that pace (and neither do some of its users, judging by the bug reports). In that light, and considering the large number of NUT users and developers for whom English isn't their first language, I hope our apparent inaction until now is not judged too harshly.

[...]

PR #992, opened 2021-03-13
Retire some terminology

jimklimov:
This PR aims to address the bulk of the issue raised in #840.

At the moment of this posting, community discussion is underway to select the replacement terminology, so at the moment this PR serves more as a map of source locations to fix than as a final solution (unless we happen to settle on the words used here).
UPDATE: Terminology was chosen: "primary" and "secondary" modes for upsmon (and eventually possibly NUT protocol aliases)

Even then, there are some more points to consider that were not covered yet:
* documentation sources refer to images, these probably contain the terms being obsoleted and should also be updated when new ones are elected;
* currently there was no change to protocol and to ABI exposed in headers;
* for the protocol in particular, it may be possible to create a backwards compatible solution (so older upsmon's can talk to newer upsd, and older upsd can talk to newer upsmon) which would avoid a required protocol version bump and redeployment of all NUT installations to new code (can be complicated for package users);
* as a bonus, an `upsmon` built from this branch may already support both old and new terms in its configuration to test how that goes (neither built nor tested yet however). It should so far talk the old protocol on the network;
* as highlighted in "docs/history.txt: reflect that slave/master terms were deprecated" committed separately, there is (was?) a precedent using the chosen word for a different purpose: "Manager mode was added to allow changing the value of read/write variables in certain UPS models" but I did not find any related mention of this word in current codebase;
* a few documentation changes that were too cumbersome to backport into "master/slave" terminology have landed here. Most of the documentation and other changes not directly related to the obsoletion of these words for upsmon context are in PRs #989 #990 #991, to keep the change size of this one under control; minor merge conflicts of this PR with those are anticipated and will be addressed when they collide in the PR merges. UPDATE: conflicts handled :)

Note that currently this PR aims to minimally impact the active codebase, only adding the support for "primary" as alias to "master" and "secondary" as alias to "slave" in processing of upsmon configuration, to match contemporary documentation promoting the new words.

Changes to NUT protocol handling and/or public function names may come (and points to modify in code were commented among this PR's changes), but this would be handled in some other PR(s) commanding more thorough testing for unbroken communications across different releases of NUT servers and clients.

### DIFF
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -802,7 +802,7 @@ void TcpClient::deviceLogin(const std::string& dev)
 	detectError(sendQuery("LOGIN " + dev));
 }
 
-/* FIXME: Protocol update needed to handle master/manager alias
+/* FIXME: Protocol update needed to handle master/primary alias
  * and probably an API bump also, to rename/alias the routine.
  */
 void TcpClient::deviceMaster(const std::string& dev)
@@ -1304,7 +1304,7 @@ void Device::login()
 	getClient()->deviceLogin(getName());
 }
 
-/* FIXME: Protocol update needed to handle master/manager alias
+/* FIXME: Protocol update needed to handle master/primary alias
  * and probably an API bump also, to rename/alias the routine.
  */
 void Device::master()
@@ -1711,7 +1711,7 @@ int nutclient_get_device_num_logins(NUTCLIENT_t client, const char* dev)
 	return -1;
 }
 
-/* FIXME: Protocol update needed to handle master/manager alias
+/* FIXME: Protocol update needed to handle master/primary alias
  * and probably an API bump also, to rename/alias the routine.
  */
 void nutclient_device_master(NUTCLIENT_t client, const char* dev)
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -802,6 +802,9 @@ void TcpClient::deviceLogin(const std::string& dev)
 	detectError(sendQuery("LOGIN " + dev));
 }
 
+/* FIXME: Protocol update needed to handle master/manager alias
+ * and probably an API bump also, to rename/alias the routine.
+ */
 void TcpClient::deviceMaster(const std::string& dev)
 {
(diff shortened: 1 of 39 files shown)


clients/nutclient.cpp
```diff
@@ -1705,6 +1711,9 @@ int nutclient_get_device_num_logins(NUTCLIENT_t client, const char* dev)
 	return -1;
 }
 
+/* FIXME: Protocol update needed to handle master/manager alias
```
aquette, line 1714: Agreed. Some pragma warning for compile time will be good too

jimklimov: It seems that compiler support for the feature varies: a pragma may be a hammer too blunt, impacting all users of the file (packagers, home users - non-devs just building NUT and with little say over actual use or not of the deprecated routines).

Recent compiler versions support (sometimes even in the same manner) ways of providing attributes to functions - so our and third-party consumer code that uses them would see the warning and be in position to update:

* clang and gcc allow for `__attribute__((deprecated))` decorations, with optional message:
````
void explode(void) __attribute__((deprecated("extremely unsafe, use 'combust' instead!!!")));
````
** https://clang.llvm.org/docs/LanguageExtensions.html#messages-on-deprecated-and-unavailable-attributes
** https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Function-Attributes.html

* clang and MSVS allow for `__declspec(deprecated)`, not sure yet if MSVS also handles `__attribite__` in any version
````
__declspec(deprecated("Please use F2")) void F1();
````

* C++14 (for nutclient.cpp/.hpp in particular) introduces the `deprecated` keyword into C++ language

Thanks for posters to https://stackoverflow.com/questions/26429243/difference-between-pragmadeprecated-and-declspecdeprecated for providing many hints for the above summary.

So following up with this would be an interesting topic, probably with a fair bit of M4 to check which keyword best applies to the current compiler (and language revision) and peppering code with that. Or even "worse" -- at least in some cases we are talking about public headers for third-party consumers, so the complete solution should better be embedded in them.

PR #994, opened 2021-03-20
New driver: microsol-apc

ygorre:
Derived from solis driver supporting newer APC Back-UPS BR hardware.
Support added for APC Back-UPS BZ1500, BZ2200BI and BZ2200I.
Tested on BZ2200BI.
Created common file that may be used on solis as well.


rpvelloso:
I'll be testing this! 👍👍

drivers/microsol-apc.c
```diff
@@ -0,0 +1,184 @@
...
+#include "main.h"
+#include "serial.h"
+#include "nut_float.h"
+#include "timehead.h"
+
+#include "microsol-common.h"
+#include "microsol-apc.h"
+
+#define DRIVER_NAME	"APC Back-UPS BZ series UPS driver"
+#define DRIVER_VERSION	"0.68"
```
jimklimov, line 36: Please check if the DRIVER_NAME is right? It seems to mismatch the description in the file heading a bit. Also, for a new driver makes sense to start with new numbering, unless it is very closely related to this version (minus one?) of another driver used as a starting point?

jimklimov:
@rpvelloso : how did the testing go? :)

@ygorre: thanks for the contribution! Is this a vendor-backed update (to reflect in docs/acknowledgements.txt)?
Also, can I ask you to please add the manpage for the new driver, probably also starting as a copy of solis with respective modifications?

rpvelloso:
Had not time to test it yet.

drivers/microsol-apc.c
ygorre: Just pushed a new commit fixing the DRIVER_NAME. As for the version, does the project have any guidelines for the numbering? If not, I would use version 1.0.

jimklimov: Good question, I don't remember any such guidance. Looking at existing sources, most drivers have `0.XY` numbers with a few claiming non-zero major number and a few having one digit after decimal. I don't think there were particular reasons for any number, other than increasing the value as features/logic were changed or added. Probably the zero reflects a "young" driver that should graduate into higher numbers as many devices are proven to be supported and bugless, but I think this was never automated or pursued manually.

@clepple @aquette : any insights on DRIVER_VERSION assignments?

aquette: Indeed X.YZ. probably around the new-driver doc, or simply inherited from what Russel did setup. Would need official formalization and possible enforcement mister new Leader ;)

PR #998, opened 2021-03-25
nut-scanner: improve SNMP extractor script

aquette:
Also remove EOL comments


jimklimov:
Just in case of style "violations", should this also look for double slashes to cut comments?

PR #1003, opened 2021-03-30
drivers/dstate.c: alarm_set(): stretch slightly absurd ways to avoid …

jimklimov:
…overflow and fix signed/unsigned mismatch (follow up for #986)


jimklimov:
Note: LGTM alerts are valid for a particular width of int (build bitness). The assertion, like elsewhere in code, protects from overflows in other bitnesses NUT can be built for, and pragmas making up most of the change-set should keep "production" compilers hush about this comparison line.

PR #1005, opened 2021-04-06
Fix a small error in powercom driver output voltage processing

dev-aaront-org:
This fixes a typo in the powercom driver that can cause the output voltage to be calculated incorrectly for some KIN models.


jimklimov:
Good point, great catch, thanks!

PR #1008, opened 2021-04-09
nutdrv_qx: add support for SNR-UPS-LID-XXXX

askazancev:
Hello everyone. 
In this PR I suggest adding support for UPSes series SNR-UPS-LID-XXXX to nutdrv_qx as a USB device.
An end-user should reference explicitly the `snr` subdriver in `ups.conf` settings because of using the same values of VendorID/ProductID as fabula_subdriver, fuji_subdriver, and krauler_subdriver.


jimklimov (changes requested):
Looks quite solid to me, thanks! Got one comment for the code to fix, but otherwise looks ok to merge if nobody vetoes this in the next few days :)

Hope @zykh can take a look at this regarding Qx driver/protocol evolution in general.

Just in case, these devices do not need a separate subdriver file for some more complex logic or lookup tables as in e.g. https://github.com/networkupstools/nut/pull/638/files#diff-a94c7d9bf3942c23f12a818a5f0d6f75c8cfb259488ea3c1b93991c0e09b7ca6 ?

drivers/nutdrv_qx.c
```diff
@@ -1008,6 +1008,100 @@ static int	fuji_command(const char *cmd, char *buf, size_t buflen)
...
+
+		if (strcmp(cmd, command[i].str)) {
+			continue;
+		}
+
+		for (retry = 0; retry < 10; retry++) {
+
+			int	ret;
+
+			ret = usb_get_string(udev, command[i].index, langid_fix, buf, 102);
```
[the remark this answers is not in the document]
askazancev: Thank you for taking a look at it!

Do you mean to add something like that before the first cycle ([line 1029](https://github.com/networkupstools/nut/pull/1008/files/b1b39d9b642a16696a831eec1699858439c7a866#diff-cb890ac7b82cb7a4f861284bcf39cc4f168312b61cb9455eb2755e345c0aa627R1029))?
```
	if (buflen < 102) {
		upsdebugx(4, "size of buf less than 102 byte!");
		return 0;
	}
```
I think it will protect against overflows if the buffer is too small.

jimklimov (approved):
Thanks for the added check, looks safer now :)

So, waiting a few days if someone else from the community has any concerns....

jimklimov:
One more point came up: the limitation of 102 bytes seemed reminiscent of something - the "hunnox" subdriver introduced with #638 seems to have spell it too at https://github.com/networkupstools/nut/pull/638/files/3a5a758dbbfbb8bdf22a45cdde06fe10d4e0249b#diff-cb890ac7b82cb7a4f861284bcf39cc4f168312b61cb9455eb2755e345c0aa627R922 although the list of commands on top of the (originally fabula) method differs from one introduced for "snr" here.

jimklimov:
Note: nutdrv_qx.c DRIVER_VERSION not bumped here since it was updated a minute ago with merge of PR #638.

docs/man/nutdrv_qx.txt
```diff
@@ -324,7 +324,7 @@ The argument is a regular expression that must match the bus name where the UPS
 
 *subdriver =* 'string'::
 Select a serial-over-USB subdriver to use.
-You have a choice between *cypress*, *fabula*, *fuji*, *hunnox*, *ippon*, *krauler*, *phoenix* and *sgs*.
+You have a choice between *cypress*, *fabula*, *fuji*, *hunnox*, *ippon*, *krauler*, *phoenix*, *sgs* and *sgs*.
```
askazancev, line 327: Double *sgs*

jimklimov:
Thanks, sorry, copy-paste error :(

docs/man/nutdrv_qx.txt
jimklimov: Fixed in master branch after merge, thanks for bringing the attention to this and sorry for the typo.

Issue #1023, opened 2021-04-22
hidparser.c: GetValue() can return incorrect result when compiled for 64-bit environment


nbriggs:
I've discovered, with the attached test case (rename the .txt to .c), using
[testgetvalue.txt](https://github.com/networkupstools/nut/files/6355450/testgetvalue.txt), code copied from hidparser.c, that extracting a value from a report:
```
  const unsigned char Buf[5] = {0x8d, 0xFF, 0xFF, 0xFF, 0xFF};
  HIDData_t Data = {.LogMax=2147483647, .LogMin = -1, .Size = 32};
  long Value;
```
with `GetValue(&Buf, &Data, &Value)`, independent of whether the processor is big-endian or little-endian, produces Value = -1 when compiled in an ILP32 environment and Value = 0 when compiled in an LP64 environment.

I'll track down the problem and propose a fix.

nbriggs:
In GetValue --
```
        /* figure out how many bits are significant */
        range = pData->LogMax - pData->LogMin + 1;
        if (range <= 0) {
                /* makes no sense, give up */
                *pValue = value;
                return;
        }
```
If you're in an ILP32 environment, range (long) will be < 0 if there was overflow in the calculation, which there is in this case of Min=-1, Max=2147483647, and so it will never do any checking and just return the value.
I think there are other things wrong with the logic, but this explains why in ILP32 environments it returns the expected value, but in LP64 it presses on and subsequently gets it wrong.

nbriggs:
I've got a revised implementation for `GetValue`, but won't be able to submit it until next week.
The main difference is that it uses the absolute magnitude of the `LogMin` and `LogMax` values, rather than the range, to determine how much of the input gets masked off.  Are there test cases capturing input from some of the less well conforming UPSs?

jimklimov:
Sorry for the delay. Unfortunately, currently there are not many tests present, and those that are - are mostly for code quality/syntax. There was a long-standing plan to mock some device activities (so dummy-ups appeared) to test the upsd server and its clients, but probably nothing in place for testing against real hardware - including workarounds for known-broken hardware (and the chain of OS drivers, libs, etc. leading to that) especially as there is no persistent pool of HW for the project... just some team members or users and contributors reporting results with what they have on hand.

blecher-at:
While working on #616 I noticed the following issue caused by this change.
Without the changes to drivers/hidparser.c, I see 0x1709 converted to 232,7 and 0x14 to 20, which is correct

[...]

PR #1024, opened 2021-04-23
Extract values in HIDParse correctly when compiled in 64-bit (LP64) mode

nbriggs:
Reworks the conversion of bytes in little-endian order to host order to
operate independently of the sizeof(long) and the byte-order of the host system.
The extracted value is stored as uint32_t as the code already performs casts
to signed/unsigned of varying sizes as appropriate to the context of use.

Closes issue #1015

### DIFF
--- a/drivers/hidparser.c
+++ b/drivers/hidparser.c
@@ -150,14 +150,14 @@ static long FormatValue(uint32_t Value, uint8_t Size)
  * -------------------------------------------------------------------------- */
 static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 {
-	int	Found = -1;
+	int	Found = -1, i;
 
 	while ((Found < 0) && (pParser->Pos < pParser->ReportDescSize)) {
 		/* Get new pParser->Item if current pParser->Count is empty */
 		if (pParser->Count == 0) {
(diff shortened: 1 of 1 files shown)


nbriggs:
See withdrawn/closed PR #1016 for additional discussion.

jimklimov:
Thanks, so in this incarnation, does the byte-stream-to-number conversion work reasonably for all host types and word lengths involved?

nbriggs:
Yes, the new conversion of a little-endian number in the byte stream is independent of word size (32/64 bit) and endian-ness.  It doesn't address the bug I raised in issue #1023 which will need some more thought.

jimklimov:
Just in case - is it also producing numbers as expected for shorter integers (8, 16 bits), in particular with no garbage in uninitialized memory of bytes not touched? Seems all should be OK, but better double-checked than sorry :)

nbriggs:
Yes, there is no possibility of there being uninitialized bytes in the result as the entire result (`pParser->Value`) is zeroed before anything is accumulated into it, and the only values subsequently accumulated are the 0, 1, 2, or 4 bytes as specified by the length of the item.

This code is used to parse the report *descriptors* but not the values that the UPS reports back as its state changes (that's `GetValue`)

PR #1027, opened 2021-04-26
nutdrv_qx.c: add correction of runtime estimation

askazancev:
Hi everyone! Here a little improvement in runtime estimation when the battery has some degradation. The faster batt.volt.act dropping the lower will be batt.runt.est. 
We have done series of experiments on different batteries and loads and suppose this will be useful for more accurate battery runtime estimation.

To use the improved runtime estimation in ups.conf must be defined the following [extra arguments](https://networkupstools.org/docs/man/nutdrv_qx.html#_extra_arguments)
-         default.battery.voltage.high
-         default.battery.voltage.low
-         runtimecal
-         chargetime

[Math model of battery.runtime estimation.xlsx](https://github.com/networkupstools/nut/files/6376449/Math.model.of.battery.runtime.estimation.xlsx)

[Experimental data](https://docs.google.com/spreadsheets/d/14zKytk-7WgEY4UfrITtRvgGzlIfs8BgdXhhjHlQjILI/edit#gid=18449429)

MAINTAINER UPDATE: Attached a copy 2021-10-16
* [Experimental data NUT + LID-600 (PR 1027).ods](https://github.com/networkupstools/nut/files/7357494/Experimental.data.NUT.%2B.LID-600.PR.1027.ods) (Note: Excel users may need to change graph axis from "date" to "text")
* [Experimental data NUT + LID-600 (PR 1027).pdf](https://github.com/networkupstools/nut/files/7357497/Experimental.data.NUT.%2B.LID-600.PR.1027.pdf)

I would greatly appreciate it if you kindly give me some feedback.


askazancev:
@jimklimov hello. Do you have any thoughts about that ?

jimklimov:
Hello, sorry for the delay reviewing this change. Looking at more context per https://github.com/askazancev/nut/blob/nutdrv_qx_improve_runtimecalc/drivers/nutdrv_qx.c#L2991

If I understand correctly, the driver already attempted various estimations of `batt.runt.est` at different moments in life cycle (initial, walk) with guesses based on readings from the UPS and for the case near your proposed change - adjusting it incrementally based on expected-linear charging time if it is online, or based on load if discharging. 

With your addition, if the battery voltage data are available, that estimation is replaced by `voltage_battery_charge * batt.runt.nom` (where `voltage_battery_charge` is the fraction of current charge vs full charge) but this replacement only happens if actual `voltage_battery_charge` is smaller than expected by previous (load-based) estimation, so adapting to a battery too old and otherwise non-linear, right?

Seems reasonable :)

jimklimov:
For the PR itself, I suppose it makes sense to update the driver documentation to mention that we estimate remaining run-time so and so (and I suppose that this UPS family/protocol does not report the value natively), and that this estimation is enabled by setting the options you listed above, perhaps with an example configuration section.

clepple (commented):
I think it would be good to have a few `upsdebugx` printouts to troubleshoot the battery voltage estimation (and as Jim mentioned earlier, to indicate that `battery.voltage` isn't the value provided by the UPS).

jimklimov:
Correction: battery voltage I think *is* provided by the UPS; just the remaining run time estimation is not.

jimklimov:
Hello @askazancev : there were numerous subdriver changes recently merged for nutdrv_qx, as well as general NUT support of libusb-1.0 API (in addition to earlier support of just libusb-0.1 API). Would it be possible for you to re-test this PR (master synced into it just now) that the predictions still work as expected after all the codebase changes?

PR #1043, opened 2021-06-12
server: Add option to disable weak SSL protocols/ciphers.

dougnazar:
Adds a boolean option DISABLE_WEAK_SSL. Defaults to off to maintain compatibility with existing installs.

On OpenSSL that supports at least TLSv1.2, disables everything below. Otherwise falls back to next best protocol level. Also enforces server cipher preference order.

On NSS which supports `SSL_VersionRange*()` functions (3.14+) it disables everything below TLSv1.2 if supported, else uses TLSv1.1. It also disables several weaker ciphers (RC4 & 3DES). On older systems I just copied the existing code but disabled SSLv3 instead. Hopefully I didn't break backwards compatibility but have no way to test anything that old.

Tested with [testssl](https://testssl.sh/) which gives it now an A+ grade vs B for OpenSSL & C for NSS.


jimklimov:
CC @NUT-RogerPrice for SSL-related proposal

server/conf.c
```diff
@@ -219,6 +225,17 @@ static int parse_upsd_conf_args(size_t numargs, char **arg)
 #endif /* WITH_CLIENT_CERTIFICATE_VALIDATION */
```
jimklimov, line 228: @aquette : looking at this older code above, should client certs validation only apply to NSS but not OpenSSL?

server/netssl.c
```diff
@@ -36,22 +36,22 @@
...
- *   /usr/include/mps/key.h:9:9: note: '#pragma message: key.h is deprecated. Please include keyhi.h instead.'
- *   /usr/include/mps/keyt.h:9:9: note: '#pragma message: keyt.h is deprecated. Please include keythi.h instead.'
- * If this becomes a warning or error in the future, it can be addressed
- * with a trick like done elsewhere for best pick of (sys/)types.h support
- * for the specific build target platform.
- */
+#if defined(NSS_VMAJOR) && NSS_VMAJOR > 3 || (NSS_VMAJOR == 3 && NSS_VMINOR >= 39)
+	#include <keyhi.h>
+	#include <keythi.h>
+#else
```
jimklimov, line 42: @dougnazar : Kudos for investigating and rectifying this bit :)

jimklimov (approved):
Looks quite solid to me, thanks for picking up this subject. I would wait some days for comments from other community members, before merging.

One further idea could be to report (once per startup) if the flag is disabled after parsing the configuration, leaving this NUT instance less secure than it could be, as a way to nudge existing deployments to improve their setup before this protection eventually becomes enabled by default.

dougnazar:
Okay, good. You fixed the spellcheck issue.

I've been fighting with it this morning. It really doesn't want to run if you use separate build directories.

jimklimov:
With "it doesn't want to run", did you mean the spell checker? It is quirky with the web of makefiles indeed. Most of its history, most of NUT could only be build via root directory. This is hopefully improving over time - many subdirs with their makefiles can be rebuilt in-place during development - but maybe some less frequented locations are still "neglected" :(

Otherwise, as most other proper autotoools-based projects, NUT supports out-of-tree complete builds (as e.g. the `distcheck` ability verifies regularly in CI buidls).

dougnazar:
Yes, the spellchecker. I saw the travis failure and went to go follow the instructions but I was using separate build dirs (openssl/nss) and it failed. Thinking it was a missing $(srcdir) or something went to take a look and then found that the other subdirs use the docs/Makefile to run the spellchecker. Makes it a bit harder to fix. I'll try to make some time to look at it.

I added a warning in `init_ssl()` as suggested. I'll wait a few days to see if anything else needs work before pushing.

dougnazar:
@jimklimov, anything I can do to help push this forward?

jimklimov:
Thanks, not at the moment - struggling to get new CI working properly now that Travis is gone... sorry about the delays.

server/conf.c
jimklimov: Posted as issue #1096 to follow up later

jimklimov (approved):

Issue #985, opened 2021-03-04
Support of generic HID-compliant UPS with Interface greater than 0


abratchik:
Current implementation of HID-compliant UPS support (usbhid_ups) assumes only Interface 0. This may or may not be valid for composite devices, which may combine the UPS function with some other interface. 
Besides, [the USB HID Power Device specification](https://www.usb.org/sites/default/files/pdcv11.pdf) does not limit the end point to Interface 0 only.  
Is it possible to make the interface number configurable or auto-detecting the correct interface? That would help to make networkupstools more universal.

jimklimov:
With the PR merged, I think this issue may be closed. Thanks for the info and the effort!

DavidLin1577:
mark

PR #1044, opened 2021-06-12
HID-compliant UPS support with interface gt than 0 + Arduino UPS support

abratchik:
Fixes issue #985 

This pull request enables support of USB HID-compliant UPS, which are composite devices. Example - Powervar UPS, which is a composite device utilizing interface greater than 0.

General idea of proposed changes - additional fields added to the usb_communication_subdriver_t structure as follows:

- **hid_rep_index** - interface number. Default is 0.
- **hid_desc_index** - descriptor number. Default is 0
- **hid_ep_in** - input interrupt endpoint number. Default is 1
- **hid_ep_out** - output interrup endpoint number. Default is 1.

These parameters can later be overridden in a **usbhid-ups** subdriver in order to set the correct values and remove the necessity to hardcode VID/PID in the **libusb.c** module. The most convenient place to override these settings in the subdriver is the **claim** method as it is called at the time of subdriver matching before any HID requests:

```
/* this function allows the subdriver to "claim" a device: return 1 if
 * the device is supported by this subdriver, else 0. */
static int arduino_claim(HIDDevice_t *hd)
{
	int status = is_usb_device_supported(arduino_usb_device_table, hd);

	switch (status)
	{
	case POSSIBLY_SUPPORTED:
		/* by default, reject, unless the productid option is given */
		if (getval("productid")) {
                        usb->hid_ep_in=4;          /* <== setting output interrupt end point to 4 */
                        usb->hid_ep_out=5;        /* <== setting output interrupt end point to 5 */
                        usb->hid_rep_index = 2;  /* <== setting interface to 2 */
			return 1;
		}
		possibly_supported("Arduino", hd);
		return 0;

	case SUPPORTED:
                usb->hid_ep_in=4;
                usb->hid_ep_out=5;
                usb->hid_rep_index = 2;
		return 1;

	case NOT_SUPPORTED:
	default:
		return 0;
	}
}
```

Additional **usbhid-ups** subdriver added to support [Arduino-based HID-compliant UPS project](https://create.arduino.cc/projecthub/abratchik/hid-compliant-ups-with-arduino-75198c?ref=user&ref_id=1670256&offset=0). The subdriver uses the interface 2 and interrupt endpoints 4 and 5 thus utilizing the flexibility introduced by this PR. The idea of introducing this driver is to enable creation and prototyping of smart UPS based on Arduino, which is inexpensive versatile platform very much suitable for custom fully open-source UPS development. 

Powerwar UPS HID subdriver is also updated -  update of hid_rep_index based on hardcoded VID/PID removed from libusb.c. The hid_rep_index for this device is now set in the subdriver itself similarly to the code snippet above.


jimklimov:
Thanks for this PR. It seems that defining more IDs in the sources got the upower rules file regenerated, but this change was not committed as part of the PR and so some of CI scenarios complained: see https://travis-ci.org/github/networkupstools/nut/jobs/774366109#L1994 for the expected changes.

.gitignore
```diff
@@ -12,6 +12,8 @@
 /_install_pkgprotodir/
 Makefile
 Makefile.in
+CMakeCache.txt
+CMakeFiles/
```
jimklimov, line 16: Did you (locally) have any CMake recipes to build NUT? In the future, such contribution might be welcome too, although currently the autoconf-based scriptware does the job and is quite large to make a 1:1 equivalent too complex to achieve.

Or was NUT source built for you as part of some larger cmake-driven project which left noise in your workspace?

drivers/arduino-hid.c
```diff
@@ -0,0 +1,145 @@
...
+
+/* USB IDs device table */
+static usb_device_id_t arduino_usb_device_table[] = {
+	/* Arduino Leonardo, Leonardo ETH and Pro Micro*/
+        { USB_DEVICE(ARDUINO_VENDORID, 0x0036), NULL },
+	{ USB_DEVICE(ARDUINO_VENDORID, 0x8036), NULL },
+        { USB_DEVICE(ARDUINO_VENDORID2, 0x0036), NULL },
+	{ USB_DEVICE(ARDUINO_VENDORID2, 0x8036), NULL },
+        { USB_DEVICE(ARDUINO_VENDORID2, 0x0040), NULL },
+        { USB_DEVICE(ARDUINO_VENDORID2, 0x8040), NULL },
```
jimklimov, line 47: Cosmetic: mismatch of tabs and spaces to indent (also in other parts of the PRed code)

drivers/libusb.c
```diff
@@ -246,13 +239,8 @@ static int libusb_open(usb_dev_handle **udevp, USBDevice_t *curDevice, USBDevice
...
 			upsdebugx(2, "- Device release number: %04x", curDevice->bcdDevice);
 
-			/* Powervar UPS is a composite device. Index must be 1 to select the right interface */
-			if ((curDevice->VendorID == 0x4234) && (curDevice->ProductID == 0x0002)) {
-				hid_rep_index = 1;
-				upsdebugx(2, "Set rep index to 1");
-			}
 			if ((curDevice->VendorID == 0x463) && (curDevice->bcdDevice == 0x0202)) {
-				hid_desc_index = 1;
+				usb_subdriver.hid_desc_index = 1;
```
jimklimov, line 243: @aquette : generally the change in this PR makes sense to me. I wonder if the comment earlier in the original file (where it defined `hid_desc_index` in relation to `bcdDevice`) should be moved here to explain why we assign the index. Currently it has moved to libusb.h...

@abratchik: Even better might be to move this assignment to an eaton-specific subdriver same as this PR did for Powervar UPS (and introduced Arduino support), so the shared `libusb.c` code has less weird cases handled directly.

drivers/libusb.c
```diff
@@ -552,7 +540,7 @@ static int libusb_get_interrupt(usb_dev_handle *udev, unsigned char *buf, int bu
 	}
 
 	/* FIXME: hardcoded interrupt EP => need to get EP descr for IF descr */
-	ret = usb_interrupt_read(udev, 0x81, (char *)buf, bufsize, timeout);
+	ret = usb_interrupt_read(udev, USB_ENDPOINT_IN + usb_subdriver.hid_ep_in, (char *)buf, bufsize, timeout);
```
jimklimov, line 543: In the PR changes, I did not see any definition of `USB_ENDPOINT_IN` as `0x81` (or near that, to shift by a small `hid_ep_in` value). Just worrying if this change is correct.

drivers/libusb.c
```diff
@@ -582,5 +570,9 @@ usb_communication_subdriver_t usb_subdriver = {
 	libusb_get_report,
 	libusb_set_report,
 	libusb_get_string,
-	libusb_get_interrupt
+	libusb_get_interrupt,
+        0,
+        0,
+        1,
+        1
```
jimklimov, line 577: Cosmetic: indentation

Also given the "human-readable" names of values above, it seems useful to define meaningfully named macros for default values, to explain these zeroes and ones here and in other subdrivers that would use them.

scripts/upower/95-upower-hid.rules
```diff
@@ -150,6 +152,14 @@ ATTRS{idVendor}=="10af", ATTRS{idProduct}=="0001", ENV{UPOWER_BATTERY_TYPE}="ups
 ATTRS{idVendor}=="10af", ATTRS{idProduct}=="0004", ENV{UPOWER_BATTERY_TYPE}="ups"
 ATTRS{idVendor}=="10af", ATTRS{idProduct}=="0008", ENV{UPOWER_BATTERY_TYPE}="ups"
 
+# Arduino
+ATTRS{idVendor}=="2341", ATTRS{idProduct}=="0036", ENV{UPOWER_BATTERY_TYPE}="ups"
+ATTRS{idVendor}=="2341", ATTRS{idProduct}=="8036", ENV{UPOWER_BATTERY_TYPE}="ups"
+
+# Arduino
+ATTRS{idVendor}=="2A03", ATTRS{idProduct}=="0036", ENV{UPOWER_BATTERY_TYPE}="ups"
+ATTRS{idVendor}=="2A03", ATTRS{idProduct}=="8036", ENV{UPOWER_BATTERY_TYPE}="ups"
```
jimklimov, line 161: It seems that defining more IDs in the sources got the upower rules file regenerated, but not all of this change was committed as part of the PR and so some of CI scenarios complained: see travis-ci.org/github/networkupstools/nut/jobs/774366109#L1994 for the expected changes.

IIRC, you need to re-run `autogen.sh` locally to have the file updated in the workspace.

jimklimov (changes requested):
Thanks for this PR! Looks reasonable, but got a few nits to address (including consistent indentation). 

Also would like some review from other community members, to make sure I did not miss an elephant in the room.

scripts/upower/95-upower-hid.rules
jimklimov: Posted that fix myself :)

FYI: It is generally recommended to use a separate branch (not same name as in an upstream project) for new features and sources of PRs, to reduce surprises in later codebase re-syncs :)

.gitignore
abratchik: I built it locally to experiment with Arduino board. Sorry, I'm still new to CICD. Please kindly advise what I need to change, I will do necessary fixes and resubmit.

drivers/libusb.c
abratchik: @jimklimov I fully agree but this is a bit more work. I can cleanse this hardcode although it was not in plans :). My main intent was to support Arduino boards but they are composite devices by definition, so they use interface 0 for sketch flashing purposes.

.gitignore
jimklimov: With this bit, I was rather curious why you had to git-ignore cmake related files at all, since NUT does not dabble in that build system currently. Might as well stay ignored for the scratch files, in case someone adds real cmake recipes later on.

drivers/libusb.c
clepple: Not a strong preference, but I think just having a comment to the right of the 1's and 0's to indicate e.g. `hid_ep_in` would be clearer than a macro/enum. TBH we should be parsing the EP numbers from a descriptor anyway.

clepple (changes requested):
Overall a useful improvement to the code, thanks!

I think the short commit message should match the PR title, or even better, the commit should be split into the two separate commits (one for interface > 0, and one for Arduino support). The "request changes" feedback is for the comments in the rest of my review, not splitting the commit, but smaller and more specific commits make it easier if someone is using `git bisect` or manually reverting commits during regression testing.

drivers/libusb.c
```diff
@@ -47,9 +47,6 @@ upsdrv_info_t comm_upsdrv_info = {
 
 #define MAX_REPORT_SIZE         0x1800
 
-/* Used for Powervar UPS to make sure we use the right interface in the Composite device */
-static int hid_rep_index = 0;
-
 static void libusb_close(usb_dev_handle *udev);
 
 /*! Add USB-related driver variables with addvar().
```
jimklimov, line 52: Follow-up idea/question: would it make sense to add a driver configuration tunable here, for people to try a non-zero interface number? I suppose over time there can be more HID devices exposing an UPS PDC in a way that current NUT of the day does not know how to find...

Also, is it possible to detect and report if a device has more than one (the zero-numbered) interface, especially if no UPS support is found on #0, as a hint for log readers to try looking further?

jimklimov: e.g. looking at the user report log from https://alioth-lists.debian.net/pipermail/nut-upsuser/2021-May/012410.html / https://alioth-lists.debian.net/pipermail/nut-upsuser/attachments/20210518/30968f29/attachment.htm with errors like:

````
           0.056845     Trying to match device
           0.057251     Device matches
           0.057334     failed to claim USB device: could not claim interface 0: No such file or directory
           0.057475     failed to detach kernel driver from USB device: could not detach kernel driver from interface 0: No route to host
           0.057554     failed to claim USB device: could not claim interface 0: No such file or directory
           0.057600     failed to detach kernel driver from USB device: could not detach kernel driver from interface 0: No route to host
           0.057768     failed to claim USB device: could not claim interface 0: No such file or directory
           0.057825     failed to detach kernel driver from USB device: could not detach kernel driver from interface 0: No route to host
           0.057870     failed to claim USB device: could not claim interface 0: No such file or directory
           0.058006     failed to detach kernel driver from USB device: could not detach kernel driver from interface 0: No route to host
           0.058050     Can't claim USB device [0764:0601]: could not detach kernel driver from interface 0: No route to host
````

I can't prove or disprove quickly if there is some new device revision with "interface 0" used for something while another interface got introduced for UPS activity. Or if that is just some system setup error (permissions, libusb version, ...) on that user's side...

abratchik: I thought about this. Placing interface and end points  in ups.conf would be better of course. Another alternative - Win and OSX hid ups drivers find the right ones without any settings. In theory, one can cycle through the interfaces and determine the one with UPS by reading the descriptor.

jimklimov: That "lockpicking" would be a nice option too (though something with a toggle to enable/disable, just in case). I do not know if there is a way to "know" as a libusb consumer that an interface actually exists on a device, and if the range is contiguous (non-replying interface means we stop iterating). And if the errors like in the log above are related to this and can be used as hints for the iterator :)

.gitignore
abratchik: @jimklimov I think I understood your question finally, sorry it took time :) I believe I tried to ccmake when attempting to build nut from source and this created garbage files, which I then tried to exclude through gitignore. I will revert this change in the next commit.

drivers/arduino-hid.c
abratchik: @jimklimov will fix in the next commit. I think this is because the file was generated first and then fixed through NetBeans IDE, which converts tabs to spaces automatically.

drivers/libusb.c
abratchik: @jimklimov I checked the changes required for removing hardcode for Eaton- it is a bit more complex than I thought. Since I don't have Eaton device for testing, I suggest we can leave it as-is, without touching Eaton. Powervar UPS changes were relatively straightforward and harmless so I decided to propose refactoring. Regarding Eaton - one can do it later in a  in a separate PR if required.

drivers/libusb.c
abratchik: @jimklimov agreed, I suggest we can do in the next PR in order not to pack too many changes in this one.

drivers/libusb.c
abratchik: @clepple @jimklimov thank you for review, fixed in the latest commit, added macros and comments

abratchik:
> Overall a useful improvement to the code, thanks!
> 
> I think the short commit message should match the PR title, or even better, the commit should be split into the two separate commits (one for interface > 0, and one for Arduino support). The "request changes" feedback is for the comments in the rest of my review, not splitting the commit, but smaller and more specific commits make it easier if someone is using `git bisect` or manually reverting commits during regression testing.

@clepple thank you for review - I agree this should be 2 commits. On the other hand, the change was only to add Arduino support and open the door for custom UPS development. Enhancement to libusb is a side change, just to avoid hardcoded VID/PID.

jimklimov:
Well, technically... they do sort of pursue different goals and might be cherry-picked separately by some NUT fork or release to bring different benefits. Non-zero interface support is a welcome feature on its own (especially if made configurable), and Arduino driver is "just" one of several consumers for such feature.

clepple (approved):
LGTM, thanks!

abratchik:
@jimklimov I have decomposed the commits into two as requested, please kindly check

drivers/arduino-hid.c
abratchik: fixed

jimklimov (approved):
Sorry it took a while to rectify the NUT CI side as well to get to the PRs in queue, great thanks for your contribution and patience!

jimklimov:
Had a few "cosmetic" nits, an important one being the copyrights in new files. Since I can't edit the source for this PR, I proposed the changes in https://github.com/abratchik/nut/pull/1 - so intend to delay merging this one for a few days until that one is accepted, or would add those commits to NUT later after merging this PR anyway :)

abratchik:
> 
> 
> Had a few "cosmetic" nits, an important one being the copyrights in new files. Since I can't edit the source for this PR, I proposed the changes in [abratchik#1](https://github.com/abratchik/nut/pull/1) - so intend to delay merging this one for a few days until that one is accepted, or would add those commits to NUT later after merging this PR anyway :)

hi @jimklimov I accepted your changes, thanks a lot!

PR #1047, opened 2021-06-14
Fix spellcheck to also run from separate build dirs

dougnazar:


jimklimov:
Thanks, I hope to give it a spin soon before merging :)

Just in case, your idea for the fix is for `make spellcheck` to go into corresponding builddirs with their makefiles, but to look at files from `SPELLCHECK_SRC` from `srcdir` (since we do not build the txt files but provide them as original "sources")?

dougnazar:
That's correct. It seemed the simplest fix for the issue and since `builddir == srcdir` it still works with building in the source directory.

PR #1060, opened 2021-07-05
RFC: add new variable battery.packs.external

aquette:


clepple (commented):
I'd recommend using the adjective-noun pattern seen in "Number of bad battery packs", so "Number of internal battery packs" and "Number of external battery packs".

Also, I remember some UPSes would return a zero-padded string, but I don't think we should continue doing that. Could also have some values other than 1.

aquette:
hi Charles,
thanks for your comments, I've just applied your recommended improvements

PR #1061, opened 2021-07-07
genericups: Add support for FTTx battery backup.

dogtopus:
battery

(Picture from CyberPower website)

A lot of those "UPSes" use a 4-wire open-drain-like contact closure signaling (dubbed "telemetry interface") to report their status (see https://dl4jz3rbrsfum.cloudfront.net/documents/CyberPower_UM_CBN50U48A-1.pdf for example). They are sorta compatible with APC 940-0020B/C cables in terms of pin assignments, with extra signals for replace battery and battery missing and lack of shutdown signal.
* [Copy of CyberPower_UM_CBN50U48A-1.pdf](https://github.com/networkupstools/nut/files/7206067/CyberPower_UM_CBN50U48A-1.pdf)

nut-fttx

This patch adds a sample device type (23) for this kind of device. It also adds support for the extra signal types (RB for replace battery and BYPASS for disconnected battery) used by these devices, as well as an option to unassign/disable an unused signal in the config file by setting it to NULL (I can also change it to `none` to be more consistent with the docs. It's your choice).

Note that there is no standard cable for this kind of device, so type 23 assumes ~~RS232 control signaling~~ active low TTL signaling (typical for USB to serial converter chips such as CH340x and FT232 series, as well as RS232 transcievers)  with pull-up and RTS as the voltage source for input pin pullups.


jimklimov (approved):

docs/man/genericups.txt
```diff
@@ -219,6 +229,11 @@ UPS TYPES
 
      [CP=RTS] [OL=CTS] [LB=-DCD] [SD=DTR]
 
+23 = Generic FTTx (Fiber to the x) battery backup
+     with 4-wire telemetry interface
+
+     [CP=RTS] [OL=CTS] [LB=-DCD] [RB=-RNG] [BYPASS=-DSR] [SD=none]
```
jimklimov, line 235: @dogtopus : In the patch description you mentioned a "NULL" vs. "none" - is this `SD=none` one of such cases? I did not find an earlier particular handling for "none" string in `genericups.c` so got no strong opinion... just wondering if that `strstr` call checking for "NULL" string should also check for "or `none`" and that's it?

dogtopus: @jimklimov Yes, that's exactly what I meant. Documentations have none as the placeholder for not applicable/disabled signals but I used NULL instead (mostly because I implemented the feature before I actually read that part of the doc).

jimklimov: Followed up with https://github.com/dogtopus/nut/pull/1

Wondering now if an `SD=none` as documented with the original PR would have been in fact handled when parsing the `upstype` from `ups.conf` as documented? Just in case, did you check that ability to customize?

jimklimov:
Thanks for bearing with me, and for the contribution and explanations of course :)

PR #1062, opened 2021-07-07
Extend battery information collection

aquette:
Add battery.status, battery.capacity.nominal and Li-ion battery cell min/max voltage / temperature

### DIFF
--- a/docs/nut-names.txt
+++ b/docs/nut-names.txt
@@ -401,6 +401,8 @@ battery: Any battery details
 | battery.capacity.nominal     | Nominal battery capacity (Ah)       | 8.0
 | battery.current              | Battery current (A)                 | 1.19
 | battery.current.total        | Total battery current (A)           | 1.19
+| battery.status               | Health status of the battery
+                                 (opaque string)                     | ok
 | battery.temperature          | Battery temperature (degrees C)     | 050.7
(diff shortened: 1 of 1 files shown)


clepple (changes requested):
This probably needs clarification as to what the min/max represent: observations (reset by an instant command?) or allowed operating voltage/temperature range.

Would also recommend a review by @NUT-RogerPrice.

aquette:
thanks again for your comments @clepple 
On the Eaton unit I used, both min/max temperature and voltage are measurements, so observed values, not settings.
I also don't see a command to reset.

NUT-RogerPrice:
From the point of view of your humble I-D editor, this is the same situation as #1063.  May I suggest handling it as if the I-D was already an RFC, by announcing the proposed addition in mailing list nut-upsuser to get the "rough consensus"   (apparently that means 95% agreement).   Roger

aquette:
small update: these were announced to the mailing list as an RFC for a week, without feedback so far:
https://alioth-lists.debian.net/pipermail/nut-upsuser/2021-August/012485.html

Is a "no answer / objection" considered as a rough consensus? If so, is there some time limit / deadline attached, to consider the merge of such proposal?

PR #1063, opened 2021-07-07
RFC: clarify and complete battery dates

aquette:
The existing battery.date means the installation or last
replacement date, as it seems to be implemented by various
drivers.

Add an additional battery.date.maintenance for the next
replacement or maintenance date

### DIFF
--- a/docs/nut-names.txt
+++ b/docs/nut-names.txt
@@ -403,7 +403,10 @@ battery: Any battery details
 | battery.runtime.restart      | Minimum battery runtime for UPS
                                  restart after power-off (seconds)   | 120
 | battery.alarm.threshold      | Battery alarm threshold             | 0 (immediate)
-| battery.date                 | Battery change date (opaque string) | 11/14/00
+| battery.date                 | Battery installation or last change
+                                 date (opaque string)                | 11/14/20
+| battery.date.maintenance     | Battery next change or maintenance
+                                 date (opaque string)                | 11/13/24
 | battery.mfr.date             | Battery manufacturing date
                                  (opaque string)                     | 2005/04/02
 | battery.packs                | Number of battery packs             | 001


clepple (commented):
No strong opinions here, but you should probably coordinate with @NUT-RogerPrice

aquette:
Thanks @clepple 
@NUT-RogerPrice  any comment?

NUT-RogerPrice:
The addition of a new variable is a modification to the command/response protocol so should also be reflected in the file doc/nut-names.txt which (if the I-D becomes an RFC) will become a Recording Document for the RFC and will have the authority of an RFC.  I argued in the I-D that there was no need to burden IANA with the registration of variable names since this can be handled efficiently in the NUT project mailing list.  This proposal has passed the ISE review, so I suggest that we act as if the I-D was already an RFC and give publicity to the proposed addition in the NUT mailing list nut-upsuser.  If there are no objections then 
 1) We have the "rough consensus" that the IETF call for,
 2) The change is publicly documented.  (At least more publicly than this GitHub issue discussion)

Roger

aquette:
@NUT-RogerPrice  thanks for your feedback. I'll mail on nut-upsuser these various additions.
Since I'm now away from NUT for quite some time, is there any preferred format for this (title, content)?

NUT-RogerPrice:
_Addition of new variables to the future RFC via Recording Document /doc/nut-names.txt._  Since there are few of us, I don't think much formality is needed.  If the requester can convince us in his/her own words and get "rough consensus". then that's enough.  Roger

aquette:
small update (replay of #1062): these were announced to the mailing list as an RFC for a week, without feedback so far:
https://alioth-lists.debian.net/pipermail/nut-upsuser/2021-August/012485.html

Is a "no answer / objection" considered as a rough consensus? If so, is there some time limit attached?

Thanks again @NUT-RogerPrice

PR #1076, opened 2021-08-20
Propose to express dates using ISO 8601 when possible

aquette:


jimklimov:
It is a bit unclear what "or otherwise and Combined date and time representations" means. An example of ISO 8601 date representation (one of many offered by the standard) "or otherwise"? Which combined date and time would we take - e.g. `YYYYMMDDTHHMMSSZ` (literal T separator and Z for "zulu" UTC timezone)? Or with dashes and colons? Or...?

Otherwise, as a generally uniform and standardized approach - LGTM :)

jimklimov:
PS: Hope to escape from the beach sometime and finish the master-branch Jenkinsfile to pass common PRs better and faster.

aquette:
thanks for your feedback @jimklimov 
I'm gonna clarify by amending the PR

jimklimov (approved):

jimklimov:
Regarding RFC 3339 vs. ISO 8601, I am not sure we should press for one over another, and at least to my experience (including manpages and arg names of commands like `date`, the ISO spec is just better known). I believe the examples in text are a subset of either spec, and we do not go to full spec like Nth day of the year anyway.

Discussion below raises a few points of note:
* Negative TZ offsets seem to not be allowed in ISO, at least not `-00:00`?
* Year-month-day without dashes is not okay for RFC
* Context and agreement of the parties (e.g. date parser libs) is what matters most/a lot

https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats#:~:text=Pretty%20much%2C%20yes%20%2D%20RFC%203339,has%20some%20small%2C%20subtle%20differences.

PR #1078, opened 2021-08-23
snmp-ups: Add a generic function to convert date

aquette:
This provides conversion from US date (mm/dd/yyyy) to ISO
Calendar date (yyyy-mm-dd).
Also adapt the function callback mechanism to make it more
generic, and only char* to char* reformating, beside from the
int to char* value lookup.


aquette:
implement proposal of #1076

jimklimov (approved):
Can't help thinking if the usdate to isodate routine fits better in "str" or "common" helper sources, to avoid duplication or discrepancy when/if other code eventually adds this feature to go with #1076

Otherwise LGTM

aquette:
so, going ahead and merging. we can still refine later for relocating this code, if worth

PR #1089, opened 2021-09-11
ci_build.sh: work around handling cppunit and libgd/cgi faults on NUT CI farm agents

jimklimov:
For the first shot, this disables their use in build scenarios known to fail for some systemic reasons.

Next shot, if possible, would be to fix these situation (maybe in NUT, maybe in distro):
* FreeBSD with GCC fails to link cppunit based tests; passes okay with CLANG
* OpenIndiana fails with cppunit (segfaults with GCC)
* OpenIndiana fails libgd => CGI detection in configure script when using gcc-4.9 and gcc-4.4.4-il with either bitness; okay in other GCC and CLANG versions

For cppunit, possibly some shared library should be linked; differences of GCC vs CLANG imply some ABI mismatch though? In any case, a more proper solution would be for configure script to detect if it can build and run a trivial test, not for CI systems to set up exclusions.

For libgd and gcc-4.x, more log digging is needed.

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -92,8 +92,8 @@ if [ -z "${CANBUILD_LIBGD_CGI-}" ]; then
     [[ "$CI_OS_NAME" = "windows" ]] && CANBUILD_LIBGD_CGI=no
 
     # NUT CI farm with Jenkins can build it; Travis could not
-    #[[ "$CI_OS_NAME" = "freebsd" ]] && CANBUILD_LIBGD_CGI=no
-    [[ "$TRAVIS_OS_NAME" = "freebsd" ]] && CANBUILD_LIBGD_CGI=no
+    [[ "$CI_OS_NAME" = "freebsd" ]] && CANBUILD_LIBGD_CGI=yes \
+    || [[ "$TRAVIS_OS_NAME" = "freebsd" ]] && CANBUILD_LIBGD_CGI=no
 fi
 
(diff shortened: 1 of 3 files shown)


jimklimov:
UPDATE: For (stock) libgd vs. older compilers on OpenIndiana, there will be no real fix in CI.

People who would require to build on older systems (produced before the switch to gcc-7+ as default compiler) might just succeed; maybe if CI adds an older-version environment it would "just work" too.

However in the (current) 2021.04 release, it complains of insufficiently capable gcc-4.x toolchain for using the binary library:
````
ld: fatal: file /usr/gcc/4.9/lib/gcc/i386-pc-solaris2.11/4.9.4/../../../libstdc++.so.6: version 'CXXABI_1.3.9' does not exist:
        required by file /usr/lib/libgd.so
ld: fatal: file processing errors. No output written to conftest
````

jimklimov:
CI notifier needs some more work... Posted https://github.com/networkupstools/jenkins-dynamatrix/issues/3 for UNSTABLE stages (meaning github "neutral" not "failed" for us)

The dangling notifications for which we see no verdict while the build is finished are also annoying.

Anyhow, the PR's goal is achieved, per https://ci.networkupstools.org/blue/organizations/jenkins/nut%2Fnut/detail/PR-1089-head/4/pipeline/ there are no more red-state failures in scenarios that "must succeed".

PR #1091, opened 2021-09-12
Jenkinsfile-dynamatrix: reduce complexity by dynamatrixAxesCommonEnvCartesian

jimklimov:
Like competing #1090, this PR also includes the changes from #1089.

Considering future maintenance of the build recipe, this PR is preferred over #1089 which follows the structure of older experiments. Especially if it would yield similar or better results savings-wise.

### DIFF
--- a/Jenkinsfile-dynamatrix
+++ b/Jenkinsfile-dynamatrix
@@ -72,7 +72,12 @@ import org.nut.dynamatrix.*;
     dynacfgBase['commonLabelExpr'] = 'nut-builder'
     dynacfgBase['dynamatrixAxesLabels'] = //[~/^OS_.+/]
         ['OS_FAMILY', 'OS_DISTRO', '${COMPILER}VER', 'ARCH${ARCH_BITS}']
-    dynacfgBase['dynamatrixAxesCommonEnv'] = [ ['LANG=C', 'TZ=UTC'] ]
+    // This default will be used as is or merged with other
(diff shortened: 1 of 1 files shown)


jimklimov:
Seems something is wrong with either the dynamatrixAxesCommonEnvCartesian implementation, or with merging of config objects, or with brackets... Latest iterations manage to yield an absurd setup like:
````
WITHAGENT: (ARCH_BITS=32&&ARCH32=i386&&COMPILER=CLANG&&CLANGVER=9&&OS_DISTRO=openindiana&&OS_FAMILY=illumos)
 && (nut-builder) && BITS=32&&CSTDVARIANT=gnu&&CSTDVERSION_c=11&&CSTDVERSION_cxx=11 
 &&  BUILD_TYPE=default-all-errors
 && BUILD_WARNFATAL=no && BUILD_WARNFATAL=yes
 && BUILD_WARNOPT=auto && BUILD_WARNOPT=hard && BUILD_WARNOPT=minimal
 && LANG=C && LC_ALL=C && TZ=UTC)
````

So it seems that at some point (due to config array merge?) the BUILD_WARNFATAL and BUILD_WARNOPT values defined in *different* config bundles get into same level of the config array instead of being matched and iterated properly.

For now, sibling PR #1090 wins; it produces fewer combos than original code and their contents are reasonable sets of envvars to assign.

PR #1099, opened 2021-09-21
Use ci_build.sh for CI spellchecks, and check some more docs files

jimklimov:
Let the first spellcheck attempt be quiet in the log; only redo the work verbosely if that fails.

NOTE: This was originally part of a PR to track which files were spell-checked, so that retry would only double-check and report the failed documents -- improving the UX for developers who would see all the errors concisely. The first shot at that solution worked well, but only in GNU make, so has to be tracked as a separate effort.


jimklimov:
Ready for merging: spellcheck recipe part passed; there are some warnings from recently added PRs that are pursued elsewhere.

PR #1055, opened 2021-07-01
Create test harness for GetValue() report value extraction (usbhid-ups)

nbriggs:
Adds source and updates Makefile.am to build tests/getvaluetest,
a test harness for the report value extraction function GetValue()
in hidparser.c.

getvaluetest has some built-in test cases, which are easily extensible,
but also accepts a single test specification on the command line to
allow for easy experimentation.

getvaluetest -h for usage

This code is independent and could be merged before PR #1040 so that it's easy to check (on a 64-bit system) for the before/after behavior of `GetValue()`.

PR #1138, opened 2021-10-20
Research fallout from usbhid-ups fix for LP64 bit maths

jimklimov:
Follows up from https://github.com/networkupstools/nut/issues/1023#issuecomment-947108993 and PRs #1040 and #1055

This PR increases logging detail for report posted in the issue (apparently impacted development on #616), and brings back code of implementation obsoleted by PR #1040 so it is easier to compare the two in real time.

Hopefully this branch will get tested for different endiannesses by the QEMU side of NUT CI farm.

CC @nbriggs @aquette @blecher-at


blecher-at:
min max were always 0, -1 (which can't be correct) on the wrongly reported values.
see the full report [issue_1138_full.txt](https://github.com/networkupstools/nut/files/7381977/issue_1138_full.txt)

nbriggs:
First:  With respect to the Unit code, 15782177, (aka 0xF0D121) -- that is indeed the unit code for Volts.

Second:  I fed the report descriptor to [USB Report Descriptor Parser](https://eleccelerator.com/usbdescreqparser/) -- an independent implementation of the HIDParser -- and this is the result:
[report-decoded.txt](https://github.com/networkupstools/nut/files/7384950/report-decoded.txt)

To pick one of the problematic items:
```
   0.297229     [D1] Path: UPS.Input.Voltage, Type: Feature, ReportID: 0x0f, Offset: 0, Size: 16, Unit: 15782177, UnitExp: 6, LogMin: 0, LogMax: -1, PhyMin(notused): 0, PhyMax(notused): 0, Value: -0.1
```
we see that:
```
0x85, 0x0F,        //     Report ID (15)
0x75, 0x10,        //     Report Size (16)
0x09, 0x30,        //     Usage (0x30)
0x67, 0x21, 0xD1, 0xF0, 0x00,  //     Unit (System: SI Linear, Length: Radians, Mass: Gram)
0x55, 0x06,        //     Unit Exponent (6)
0x26, 0xFF, 0xFF,  //     Logical Maximum (-1)
0x16, 0x00, 0x00,  //     Logical Minimum (0)
```
which indeed says that the acceptable range is 0..-1.   If they meant to say 0..65535, they should have described the LogicalMaximum in a 4-byte field (as they did in Report ID 9) instead of a 2-byte field:

```
   0.292855     [D1] Path: UPS.PowerSummary.ConfigVoltage, Type: Feature, ReportID: 0x09, Offset: 0, Size: 16, Unit: 15782177, UnitExp: 6, LogMin: 0, LogMax: 65535, PhyMin(notused): 0, PhyMax(notused): 0, Value: 12
```
which had a report descriptor of
```
0x85, 0x09,        //     Report ID (9)
0x75, 0x10,        //     Report Size (16)
0x27, 0xFF, 0xFF, 0x00, 0x00,  //     Logical Maximum (65534)
```
(One could be forgiven for wondering why this report decoder thinks that FFFF0000 (little-endian 32-bit value) is 65534 not 65535!! -- I'm not going to read his Javascript to find out)

I do notice that since I last visited the specs there has been an update to the [USB Usage Tables for HID Power Devices](https://www.usb.org/sites/default/files/pdcv11.pdf) (May 29, 2020) -- unfortunately I think the extent of the update is to newly disclaim responsibility for using the information it contains.  :-(

So... I think there are a few things to be done

1. Tell the UPS manufacturer that they got the report descriptor values wrong and ought to fix it in some newer revision of their firmware (fat chance of that happening, I would guess!)
2. Put a hack in the code that parses the Report Descriptor (not the code that I fixed to extract the values) so that if it is parsing a LogicalMaximum value, and that value is negative and less than the LogicalMinimum (or perhaps only if it's all-bits-set for the size it's using), then it should reinterpret it as an _unsigned_ value and log it as buggy data from the device.   Since this happens only when the initial connection is made to the device, not each time it reports the actual value (e.g., for UPS.Input.Voltage), you don't need to do any special handling to reduce the chatter.   I think It might be reasonable to put this at Warning level.
3. Have a look at *all* the report descriptors from any devices you can get your hands on and see if they make sense.

Comments?

jimklimov:
Thanks @nbriggs for the fine research, I guess the "hack in the code" makes sense for the time being; at least it would make devices usable again even if with lesser peace of mind that the values are trustworthy protocol-wise.

Unfortunately I do not have USB-connected UPSes currently, so might not be able to help with data-point collections in short term.

jimklimov:
NOTE: Rebased over recent master with its bunch of recipe fixes; should make it easier to `./autogen.sh && ./configure && (cd tests && make all)` (or `cd drivers` etc.) without necessarily rebuilding all of NUT, as one would iterate on "proper-hack" fixes.

jimklimov:
Thinking out loud: are you aware of any software emulators of UPS protocols - scripts, VMs, etc. that would talk on a Unix or network socket to act like a serial/snmp/... UPS? Maybe even something that would even "be" an USB device (maybe as a kernel driver emulating one, or something that can be "attached" to a VM vUSB port)?

I suppose that would be very helpful for NUT driver debugging and unit-testing, especially around different CPU endianness and word size definitions (while interacting with the same bit stream on the cable).

jimklimov:
Closing the PR, since this is probably not something we would actually merge, and to avoid rebuilds whenever master changes.

The branch remains available for testing if would be needed (and invoking CI automation to test in QEMU machines with many architectures).

PR #1147, opened 2021-10-27
drivers/snmp-ups.c: fix build warnings (with daisy-chain code et al)

jimklimov:
Follows up from #823 to fight warnings in NUT codebase

CC @aquette about:
* Not sure if the template offsets may be validly negative?
* Also, compiler did not mind but should other `SU_*` macros be forced as `unsigned long`?
* ...And maybe the `snmp_info_t->info_flags` field should also be increased to `unsigned long`... or rather a fixed-bitness type for both sorts of flags?

### DIFF
--- a/docs/daisychain.txt
+++ b/docs/daisychain.txt
@@ -190,7 +190,7 @@ Two functions are available to handle alarms on daisychain devices in your
 driver:
 
 * device_alarm_init(): clear the current alarm buffer
-* device_alarm_commit(const int device_number): commit the current alarm
+* device_alarm_commit(const size_t device_number): commit the current alarm
 buffer to "device.<device_number>.ups.alarm", and increase the count of
 alarms. If the current alarms buffer is empty, the count of alarm is
 decreased, and the variable "device.<device_number>.ups.alarm" is removed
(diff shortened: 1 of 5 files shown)


jimklimov:
Going over off-line comments (thanks @aquette):
> Numbering of templated device outlets => the auto guessing algo (using either walking or smthg.count) should be guilty
> look for (iirc) process_template or instantiate_smthing

>> RE: in particular, was there some valid use for negative values (could be used to walk back and forth by adding them - no idea if that happened or was intended)

> no
> it's a bug
> index starts from 0 or 1, set this offset and then it's the count of device/outlets/groups that matters that's used to iterate
> guessed and persisted for each such dev/outlets/groups at first iteration (set to -1 before), then reused - so again, definitely a bug in this init

jimklimov:
Abandoning this side quest for now. Original problem with build warnings dissipated due to other fixes in the codebase. Some findings about snmp-ups driver were addressed in other PRs. And breakage due to change of indexes to unsigned numbers was ultimately not much investigated (that change in this PR did complicate maths and range-checks however, for questionable gain).

PR #1151, opened 2021-10-30
Fix the web of Makefile cross-references across subdirs

jimklimov:
* We had issues (links welcome) with builds of parts of codebase from sub-directories, rather than builds arranged normally from the root directory. Large part of the mess was the absence of dependencies on files (e.g. `*.la` archives) located in other sub-directories, so those were not automatically built when we wanted to e.g. `configure && cd tests && make`. Now such situations should resolve correctly.

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -749,7 +749,7 @@ bindings)
     echo ""
     ./autogen.sh
     #./configure
-    ./configure --with-cgi=auto --with-serial=auto --with-dev=auto
+    ./configure --with-cgi=auto --with-serial=auto --with-dev=auto --with-doc=skip
     $MAKE all && $MAKE check
     ;;
 *)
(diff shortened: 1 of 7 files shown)


jimklimov:
Note for posterity: at least one build on a Debian 8 derived environment ended up creating directories literally named `$(top_srcdir)` and could not resolve some `*.Plo` files for the 4 sources impacted here, while compiling `nut-scanner`.

Commenting those 4 resulting `include` lines away from generated `tools/nut-scanner/Makefile` helped complete that build.

This was not seen anywhere else so I'd say it was either a strange system or some bug in older autotools.

But if this bites someone else, posted issues or better yet PRs are welcome. Returning to explicit `../` paths might be an option, but "really fixing" the issue for those and modern toolkits would be better.

Issue #505, opened 2017-11-28
Building PDF files in parallel with a2x often fails


clepple:
Ran into this while testing pull #418. A `make -j10` with PDF output enabled will often fail due to parallel a2x processes using the same names for intermediate files.

Proposed solution: uniquely named temporary directories.

StefanBruens:
In kicad this was "solved" by using a dependency between the various output formats. Thus different manuals are built in parallel, but e.g. user-manual.{txt,pdf} were built sequentially.

bigon:
Adding `.NOTPARALLEL:` at the end of the Makefile in the `docs/` directory workaround the problem

It will prevent that makefile to be run in parallel and prevent the build to fail

jimklimov:
Should be fixed with #1154 (and possibly other contemporary efforts)

PR #1152, opened 2021-10-30
Try to fix parallel docs builds with collision avoidance per a2x-shared files

jimklimov:
Despite all the best efforts regarding dependency definitions, parallel builds and/or distchecks of NUT tend to fail. One toxic chain of events was linked to documentation -- the `a2x` program tends to write and otherwise manipulate same-named files in the current directory, so build threads conflict with each other. A number of different workarounds was applied to reduce the chance of such conflicts (at a cost to build speed in some cases), although it seems problematic to rule them out completely: seems there is no *portable* feature for varied `make` implementations to force running some targets sequentially.


jimklimov:
UPDATE: Seems related to earlier efforts and discussions at #505, #418. One of those suggests just adding a hint to `make` with a `.NOTPARALLEL:` pseudo-target at the end of the Makefile. Not sure this is portable, but given how other prospective ways (e.g. subdirs) failed or are too complex, that hack is at least worth trying. At worst, GNU make would be safe for parallel builds, and those who do not accept this hint - won't.

jimklimov:
It seems current a2x and asciidoc can both be coerced into using output directories and so (checking) to not-conflict. Will explore that and maybe post an alternative PR if it seems a fruitful approach.

jimklimov:
Closed due to functional and better solution in #1154

PR #1154, opened 2021-11-01
Try to fix parallel docs builds with subdirs

jimklimov:
Picks up from #1152 to allow actual parallel builds (and distchecks) with each piece of documentation rolling in a dedicated uniquely named directory while it is being generated. This allows to avoid the conflict of requesting same-named targets from different rules, as well as of a2x using same-named helper files to build different docs, ruling it out on the OS/filesystem level. And to gain parallelism instead of safely evading it with earlier trickery attempts.

Subject also touched in #505 and #418 discussions.


jimklimov:
TODO: If the different `make` implementations in CI farm (GNU, BSD, Sun) all do not choke on this change of recipe for at least sequential builds, next step is to activate parallel rules in `Jenkinsfile-dynamatrix` to gain the benefit on CI farm. Or maybe, temporarily, to *add* such scenarios as "permitted to fail", to see if this change causes any intermittent/racey non-100% faults.

jimklimov:
For a rough benchmark, on a laptop this branch passed `make clean -k; make distcheck -j15` in 2m4s (there are far less than 15 cores, so this was more to check for collisions), and a sequential `make clean -k; make distcheck` in 4m18s. Both succeeded. (C/C++ overhead was negligible thanks to ccache; so this was mostly configure, make and asciidoc/a2x CPU time).

PR #1158, opened 2021-11-03
Limit nut-scanner thread count

jimklimov:
If extremely many threads are spawned, e.g. scanning a very large IP address range, the `nut-scanner` program can crash due to resource exhaustion or hitting some constraints in pthreads implementation (various effects were seen on different generations of OSes) or third-party libraries used in particular scanners.

This PR adds a (CLI-configurable as `--thread N`) limit to amount of threads that would be spawned by different parallelized scanners (snmp, nut, netxml, serial), and if the limit gets hit - it would `pthread_tryjoin_np()` to wait for some of the threads to complete their work, and only then proceed to spawn another.

Note that while there is locking to increase or decrease the overall thread counter, checks are not mutex'ed and I anticipate that small accounting errors are possible (e.g. launching up to +1 "extra" thread for each scan type).

Given that scanning usually means sending a request and waiting if any reply comes within a timeout, the processing load of each item scan is negligible while considerable wall-clock time is spent. Lack of parallelism does add up: in tests, an SNMP scan of a /24 subnet took ~5 seconds without hitting a limit, 20 seconds with a limit of 64 threads, and ~1200 seconds with a limit of 1 thread.

The default limit in this PR is arbitrarily set to 1024 threads overall, to accomodate fast scans of a typical /24 subnet with several protocols at once.
UPDATE1: The default can get reduced if allowed file descriptor `ulimit` is smaller.
UPDATE2: Also individual protocol scanners can be subjected to their own limits; whichever is smaller (and gets hit first) wins.

PS: This is a separate issue from broken netmask processing, addressed in PR #1157

### DIFF
--- a/tools/nut-scanner/scan_snmp.c
+++ b/tools/nut-scanner/scan_snmp.c
@@ -841,34 +841,36 @@ nutscan_device_t * nutscan_scan_snmp(const char * start_ip, const char * stop_ip
 	}
 
 #ifdef HAVE_PTHREAD
-	upsdebugx(2, "%s: all planned scans launched, waiting for threads to complete", __func__);
-	for (i = 0; i < thread_count; i++) {
-		int ret;
+	if (thread_array != NULL) {
+		upsdebugx(2, "%s: all planned scans launched, waiting for threads to complete", __func__);
+		for (i = 0; i < thread_count; i++) {
+			int ret;
 
-		if (!thread_array[i].active) continue;
+			if (!thread_array[i].active) continue;
 
-		ret = pthread_join(thread_array[i].thread, NULL);
-		if (ret != 0) {
-			upsdebugx(0, "WARNING: %s: Clean-up: pthread_join() returned code %i",
-				__func__, ret);
-		}
-		thread_array[i].active = FALSE;
+			ret = pthread_join(thread_array[i].thread, NULL);
+			if (ret != 0) {
+				upsdebugx(0, "WARNING: %s: Clean-up: pthread_join() returned code %i",
+					__func__, ret);
+			}
+			thread_array[i].active = FALSE;
 # ifdef HAVE_PTHREAD_TRYJOIN
-		pthread_mutex_lock(&threadcount_mutex);
-		if (curr_threads > 0) {
-			curr_threads --;
-			upsdebugx(5, "%s: Clean-up: Joined a finished thread #%i",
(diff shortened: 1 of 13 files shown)


aquette:
side note:
years ago, I uncovered a limitation in Net SNMP shared lib: when spawning more than 1024 handles, you crash it.
Worth to track this

tools/nut-scanner/scan_snmp.c
```diff
@@ -748,32 +747,131 @@ nutscan_device_t * nutscan_scan_snmp(const char * start_ip, const char * stop_ip
...
+		 * account them in thread_array[] as something to not wait
+		 * for below in pthread_join()...
+		 */
+
+		/* TOTHINK: Should there be a threadcount_mutex when
+		 * we just read the value in if() and while() below?
+		 * At worst we would overflow the limit a bit due to
+		 * other protocol scanners...
+		 */
+		if (curr_threads >= max_threads) {
```
aquette, line 763: this can probably be commonalized, part of future improvements, to avoid code redundancy

aquette (approved):
LGTM, as discussed

jimklimov:
Thanks, though trying to confirm that (in a Debian 11 derived enviromnent) by scanning 8 * /24 subnets and a huge limit of jobs, I did not get a crash and got many responses. Need to check with older systems...

It did however not work well from the first try: needed to bump `ulimit -n 131072` in the shell (default 1024) at least for net-snmp to be able to search for its global and per-host config files, probably for net sockets, etc. It did complain otherwise.

Next improvement here would be to not hardcode the default thread count limit, but to derive it from detected current file descriptor ulimit (probably not the other way around - forcing the ulimit bump to match, since this should not generally run as root to be able to bump).

jimklimov:
After recent commits, I can no longer crash `nut-scanner -M` for NetXML scans alone (even for larger subnets); however when scanning with several protocols (e.g. netxml + snmp), I get `buffer overflow detected` and core-dumps again.

tools/nut-scanner/scan_snmp.c
jimklimov: Had the same feeling, around making third or fourth copy :\

jimklimov:
TODO later: replicate 42ity approach with semaphores (where available) instead of simple counters (possibly not too thread-safe)?

PR #1159, opened 2021-11-05
Fix Makefile cleanup

jimklimov:
A git clone, build, check, possible `make distcheck` and finally `make distclean` should leave the Git workspace in pristine condition - no generated files remaining, no tracked files removed :)

Note: in this case, `distclean` removes just the files that can be recreated by running the `configure` script, and `realclean` wipes also files created by running `./autogen.sh` including the generated `configure` script and files that autotools copies from its own local installation. It was tested that both do not preclude the generated "dist tarball" to include such files, and be buildable with `make distcheck`.

The dist tarball contains same list of files as before (this took a bit of back and forth).

Due to peculiarities of some `make` implementations, and our inter-directory dependencies much improved by #1151, this simple goal proved tricky to achieve; in particular alas `$<` is not always usable.


jimklimov:
TODO/TRY: for later commits to make do with distchecks, instead of testing that linked source file is missing (dist seems to package it, and as a file not symlink), but better use NODIST keywords to not deliver the unneeded file at all.

Issue #1168, opened 2021-11-08
CI: Revise builds with older autoconf and related tool kits


jimklimov:
After #1151 and #1159 improvements to Makefile structure, some older systems (including some of those on Buildbot - across Linux, BSD and MacOS) no longer build the master branch, claiming issues with the out-of-dir sources, e.g. http://buildbot.networkupstools.org/public/nut/builders/Debian-x64-clang-stretch/builds/567/steps/compile/logs/stdio:
````
make[4]: Entering directory '/data/buildbot/slave/Debian-x64-clang-stretch/build/tools/nut-scanner'
Makefile:682: .deps/libnutscan_la-bcmxcp_ser.Plo: No such file or directory
Makefile:683: .deps/libnutscan_la-serial.Plo: No such file or directory
make[4]: *** No rule to make target '.deps/libnutscan_la-serial.Plo'.  Stop.
make[4]: Leaving directory '/data/buildbot/slave/Debian-x64-clang-stretch/build/tools/nut-scanner'
Makefile:599: recipe for target 'distdir' failed
````

On main NUT CI farm, and on BuildBot worker with "buster", this does not happen.

Earlier there were issues with literal `$(top_builddir)` directory created during build, confusing the expected hierarchy. Here with "dist" being the problem, the "dist-hook" is worth investigating too.

PR #1170, opened 2021-11-09
CI: Revise builds with older autoconf and related tool kits

jimklimov:
This should address issue #1168 (some fallout after PRs #1151 and #1159) - at least this passed on a test system with Ubuntu 14.04 that failed to build before the PRed change.


jimklimov:
Now this PR also rearranged the `ci_build.sh` script to not generally start (loops) with a distcheck, destined to spend a minute re-configuring to update a Makefile to delete it, and *then* `configure` again for the build.

jimklimov:
Also discovered a bit of heads-up for people like me, who develop or debug shell scripts in the product or ecosystem nearby: I recently found that the `expr` shell utility/builtin is geared towards boolean evaluations, so any expressions that return a numeric zero mean a "false" exit-code.

For example, `expr 1 - 1` would fail while `expr 2 - 1` would not.

Since many of our infra and maybe prod scripts do portable maths with `expr` (so they don't rely on particular shell's abilities like bash's `N=$(($N-1))` or availability of third-party tools like `bc`), without realizing such edge case before, this is a "strange failure" mode to be on lookout for when something "crazy" happens.

Safety fencing with `|| true` or checking in case of error that the returned string is `"0"` (and not some other error) may be due.

PR #1180, opened 2021-11-13
Redefine bit-flag map values for snmp-ups.h

jimklimov:
This PR follows up from findings of PRs #1147 and #1177, and aims to just redefine the bit-flags used to identify nuances of data points in MIB2NUT mappings.

Given that many value ranges had to move around and expand to avoid overlapping/conflicts (several macros had zero values, a few had different numbers shifted by different offsets to same result, etc.), some defined values were relocated considerably to group them better.


jimklimov:
Tested with a build of the driver against the same daisy-chain of 3 ePDUs, that it collects same sort of info (modulo minor value differences for voltages etc.)

jimklimov:
Variant in https://github.com/42ity/nut/pull/156 was tested against various Eaton UPS, ePDU (+daisy-chained ePDUs) and ATS models and did not expose any regressions compared to earlier builds of that fork.

PR #1182, opened 2021-11-13
Add CyberPower RMCARD205 which is already supported

dogsbody:
Support for CyberPower RMCARD205 was added in #686 and #793.

This pull request simply adds the card to the Hardware compatibility list

Thank you to everyone involved in making this awesome application


jimklimov:
Thanks for confirmation and for the PR :)

Issue #151, opened 2014-09-10
User manual: change wording of ownership to match suggestion


clepple:
http://article.gmane.org/gmane.comp.monitoring.nut.user/8680 bullet 3: -> "owned by group" or similar

PR #1186, opened 2021-11-16
 FTY remerge from upstream NUT as of 2021-11-16

jimklimov:
Follows up from #151 to include some upstream changes, including those of direct interest to 42ity:
* https://github.com/networkupstools/nut/pull/1177 (begin to rectify the use of bit-flags in snmp-ups, actual fix pending with https://github.com/networkupstools/nut/pull/1180 later)
* https://github.com/networkupstools/nut/pull/1173 (improved Eaton sensor readings for modern devices)


jimklimov:
Wrong repo :\

PR #1188, opened 2021-11-16
Semi-cosmetic fixes to snmp-ups SNMP version processing

jimklimov:
One is a potential bug (copy-paste typo allowed `strncmp` to look beyond fixed string literal potentially, although just into its ending nul char).

Another is "cosmetic" update to the usage message.

### DIFF
--- a/drivers/snmp-ups-helpers.c
+++ b/drivers/snmp-ups-helpers.c
@@ -24,12 +24,14 @@
  *
  */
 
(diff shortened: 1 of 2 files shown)


jimklimov:
Also rearranged include files so `config.h` with our system-impacting settings is used first (notably, before system headers that might fall back to defining something else).

PR #1196, opened 2021-11-22
docs/FAQ.txt: update for use-cases of dummy-ups in relay mode

jimklimov:
subj, community comments welcome

### DIFF
--- a/docs/FAQ.txt
+++ b/docs/FAQ.txt
@@ -618,6 +618,36 @@ This means all packages have been built by a third party.  If you
 have an issue that's related to packaging, you will need to seek
 help with whoever built it for you.
 
+== My UPS is directly connected to an appliance with a limited version of NUT, how can I monitor the UPS from arbitrary clients?
+
+You can set up a separate general-purpose system as the NUT server for
+your "arbitrary clients", using `dummy-ups` in "relay mode" as the driver.
+This instance of `dummy-ups` would in turn be the NUT client allowed to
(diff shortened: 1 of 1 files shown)


docs/FAQ.txt
```diff
@@ -618,6 +618,36 @@ This means all packages have been built by a third party.  If you
...
+other systems as clients of these NUT servers.
+
+Developers who are working on NUT, its drivers, or further projects and
+appliances based on NUT, and who need to monitor their UPS from multiple
+systems using the complete NUT stack on each system (e.g. during testing),
+can benefit from dedicating a separate general-purpose system as the NUT
+server using the real (networked) driver for the UPS, while using the
+`dummy-ups` in "relay mode" as the driver connected to this dedicated
+server on each tested system.
+
```
tessus, line 650: Thanks for the update. I was specifically searching for the terms "proxy" and "forward".

The text certainly helps and I might find my answers in the FAQ by reading every single entry. Is there a chance we could add an entry like:

```suggestion
== How can I setup NUT as a proxy (setup a server to forward/relay client data)?

This can easily be achieved by using the `dummy-ups` driver. The `port` field acts
as the reference to the "other" UPS on another NUT server.

Example:

    [proxy]
        driver = dummy-ups
        port = upsname@ip-or-hostname[:port]
        desc = "UPS proxy for UPS upsname on server ip-or-hostname"
```

jimklimov: Thanks, accepted :)

Issue #1066, opened 2021-07-13
Support of Huawei UPS2000-*-1/2/3KRT*


whc2001:
## Background
Product of UPS2000 series are online (double conversion) UPSes manufacturered by Huawei. They come with two different form factor: UPS2000-A (tower factor) and UPS2000-G (rack-mount factor). They are all equipped with a standard DB9 RS232 serial communication port, and a USB port that emulates virtual serial port. The software protocol is based on **Modbus RTU** and [the protocol document can be downloaded on their official webisite](https://support.huawei.com/enterprise/zh/doc/EDOC1000110695) ([~~all in Chinese~~](https://github.com/networkupstools/nut/issues/1066#issuecomment-975646685)). I have a UPS2000-G-3KRTS on hand. I will attempt to translate the most important part of the protocol document, and try to implement the driver based on the existing Modbus driver [phoenixcontact_modbus.c](https://github.com/networkupstools/nut/blob/master/drivers/phoenixcontact_modbus.c). 

## Note
Devices of this series also have an "intelligent slot" for their extension cards. That includes the SNMP network card, and RS485 card which also runs on Modbus RTU protocol but comes with changeable slave addresses and more functionalities. However currently I don't have any intelligent cards on hand so I am focusing on the on-device serial port.

whc2001:
## Querying Registers

### Note:

- All values are stored as U16
- Gain: the value read out must divide this value
- Type: the data type AFTER applying the gain
  - e.g. Voltage reading from the register is 2262
  - Gain is 10
  - Type is Real
  - -> Voltage = 226.2
- All register address may add UPS ID multiplied by 10000. However when using on-device serial port, the UPS ID is **1** fixed. **Querying only with register address without the UPS ID also works when there is only one device.**
  - e.g. Address of Input Voltage register is 1000
  - UPS ID is 1
  - -> Register ID used to send request is 11000 (0x2AF8), or 1000 (0x03E8)
- All reserved bits in bit flags shall be 0 unless noticed otherwise.

|Name|Gain|Type|Unit / Range|Address|Length (in U16)|R/W|
|-|-|-|-|-|-|-|
|Input Voltage|10|Float|V|1000|1|R|
|Input Frequency|10|Float|Hz|1003|1|R|
|Bypass Voltage|10|Float|V|1004|1|R|
|Bypass Frequency|10|Float|Hz|1007|1|R|
|Output Voltage|10|Float|V|1008|1|R|
[...]

PR #1198, opened 2021-11-22
Huawei UPS2000: Update Man Page, Change Logic for "load.on" command

biergaizi:
This commit introduces three minor changes after #954 merge.

* Thanks to the user report by @whc2001 in #1066, we finally have a beta tester with a UPS2000-G model with positive result. This commit adds UPS2000-G-3KRTS to the list of tested model in the man page, with slightly updated instructions for submitting new user reports. Two minor typos in the man page are also fixed.
* Currently the driver prevents the user from executing `load.on` when the UPS is already on, and reports a `STAT_INSTCMD_FAILED` with error logs. While executing "load.on" when the UPS is already in "normal-mode on" has no effect, it's still a legal command and should not be rejected arbitrarily, which can potentially confuse downstream applications and users (Update: But we still need to do this if the UPS is in bypass mode, because the UPS would otherwise enter normal mode - "load.on" is not supposed to do that).
* The previous driver also contains two misleading typos in the error logs: (1) it incorrectly suggests `bypass.off` instead of `bypass.stop`, and (2) reports `failed: reason unknown.` when the reason is simply an already-on UPS.

    This commit allows `load.on` to execute when it's already on, and changes the log messages from errors to warnings. Typos in the log messages are also fixed.

### DIFF
--- a/docs/man/huawei-ups2000.txt
+++ b/docs/man/huawei-ups2000.txt
@@ -30,16 +30,20 @@ UPS with the following characteristics.
 newer kernels, read the section *Cabling* carefully).
 
 The UPS2000 series has two variants: UPS2000-A with a tower chassis,
-and UPS2000-G with a rack-mount chassis. Only the UPS2000-A variant
-has been tested, but UPS2000-G units are probably also supported.
+and UPS2000-G with a rack-mount chassis. Both should be equally supported,
+but more testers are needed.
 
-It has been tested on the following models.
+Currently, it has been tested on the following models.
 
 * UPS2000-A-1KTTS (firmware: V2R1C1SPC40, P1.0-D1.0)
 * UPS2000-A-2KTTS (firmware: V2R1C1SPC50, P1.0-D1.0)
+* UPS2000-G-3KRTS (firmware: V2R1C1SPC40, P1.0-D1.0)
 
-We encourage users of other models to report your successful or
-unsuccessful results to the bug tracker or the mailing list.
+If your model is not in the list, we encourage you to report successful
+or unsuccessful results to the bug tracker or the mailing list.
+Make sure to include the full model number of your UPS manually
+in your report, because the firmware only reports "UPS2000-A"
+for all models, including the G series.
 
 huawei-ups2000 uses the libmodbus project, for Modbus implementation.
 
--- a/docs/man/huawei-ups2000.txt
+++ b/docs/man/huawei-ups2000.txt
@@ -300,7 +300,7 @@ Serial port becomes unresponsive
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Some malformed commands are known to lock up the serial port (including
-USB, which a USB to serial device) Upon receiving them, UPS2000 stops
+USB, which is a USB-to-serial device). Upon receiving them, UPS2000 stops
 all serial communications. The result is a completely unresponsive UPS,
 regardless of what you do - restarting NUT, rebooting the computer -
 cannot restore connectivity, as if someone has unplugged the RS-232 cable.
@@ -328,7 +328,7 @@ requires a device-specific driver, which is only available on Linux
 
 On an unsupported system, the USB device can still be recognized as a
 USB ACM device, but in reality, communication is impossible. It can
-only be fixed by implementing a driver for your system nothing can
+only be fixed by implementing a driver for your system, nothing can
 be done within NUT.
 
 Finally, in the unlike scenario that you are using NUT on Microsoft
--- a/docs/nut.dict
+++ b/docs/nut.dict
@@ -492,6 +492,7 @@ KNutClient
 KNutSetting
 KOLFF
 KRT
+KRTS
 KTTS
 Kain
 Kaminski
--- a/drivers/huawei-ups2000.c
+++ b/drivers/huawei-ups2000.c
@@ -1509,24 +1509,42 @@ static int ups2000_instcmd_load_on(const uint16_t reg)
 	status_init();
 	r = ups2000_update_status();
 	if (r != 0) {
+		/*
+		 * When the UPS status is updated, the code must set either OL, OB, OL ECO,
+		 * BYPASS, or OFF. These five options are mutually exclusive. If the register
+		 * value is invalid and set none of these flags, failure code 1 is returned.
+		 */
 		dstate_datastale();
 		return STAT_INSTCMD_FAILED;
 	}
 	status_commit();
 
-	/* is it off? */
 	status = dstate_getinfo("ups.status");
-	if (!strstr(status, "OFF")) {
+	if (strstr(status, "OFF")) {
+		/* no warning needed, continue at ups2000_write_register() below */
+	}
+	else if (strstr(status, "OL") || strstr(status, "OB")) {
 		/*
-		 * You cannot turn it on if it's not off. It's probably in
-		 * bypass mode. If so, use "bypass.stop".
+		 * "Turning it on" has no effect if it's already on. Log a warning
+		 * while still accepting and executing the command.
 		 */
-		if (strstr(status, "BYPASS"))
-			upslogx(LOG_ERR, "load.on failed: UPS is already on and in bypass mode. "
-				"To enter normal mode, use bypass.off");
-		else
-			upslogx(LOG_ERR, "load.on failed: reason unknown.");
-
+		upslogx(LOG_WARNING, "load.on: UPS is already on.");
+		upslogx(LOG_WARNING, "load.on: still executing command anyway.");
+	}
+	else if (strstr(status, "BYPASS")) {
+		/*
+		 * If it's in bypass mode, reject this command. The UPS would otherwise
+		 * enter normal mode, but "load.on" is not supposed to affect the
+		 * normal/bypass status. Also log an error and suggest "bypass.stop".
+		 */
+		upslogx(LOG_ERR, "load.on error: UPS is already on, and is in bypass mode. "
+				 "To enter normal mode, use bypass.stop");
+		return STAT_INSTCMD_FAILED;
+	}
+	else {
+		/* unreachable, see comments for r != 0 at the beginning */
+		upslogx(LOG_ERR, "load.on error: invalid ups.status (%s) detected. "
+				 "Please file a bug report!", status);
 		return STAT_INSTCMD_FAILED;
 	}
 


jimklimov:
For the sake of code evolution, I'd explicitly check if the status is "on" and not assume that not-off not-bypass means "on". And then the remaining reason could be an error and an unknown.

jimklimov (approved):
Looks good, with one suggestion to make it more robust.

jimklimov:
Since my comments above can be addressed "immediately" or later, I'll shelf this PR for a day or two and merge (if there is no feedback before that, to change the plan)

biergaizi:
> I'd explicitly check if the status is "on" and not assume that not-off not-bypass means "on". And then the remaining reason could be (or not be) an error case and an unknown situation.

According to NUT's coding standard on status flags ([docs/new-drivers.txt, line 218](https://github.com/networkupstools/nut/blob/master/docs/new-drivers.txt#L218)), the allowed status flags are, `OL`, `OL`, `LB`, `HB`, `RB`, `CHAG`, `DISCHRG`, `BYPASS`, `CAL`, `OFF`, `OVER`, `TRIM`, `BOOST`, or `FSD`, and their combinations.this is exactly what my driver does. The lookup table for setting the power status is defined in [`ups2000_status_reg[]`, line 676](https://github.com/networkupstools/nut/blob/fa86fc87da478aa0b84a3e719b7b22853d3c4fc0/drivers/huawei-ups2000.c#L676)

Since there is no single `ON` flag to check, I originally coded it to check the absence of the `OFF` flag. Here, I read your suggestion as adding three separate string tests for `OL`, `OB`, and `BYPASS`. Is it what you meant?

> And then the remaining reason could be (or not be) an error case and an unknown situation. 

The lack of an `OFF` flag conclusively decides that the UPS is on. When the UPS status is updated, the code must set either `OL`, `OB`, `OL ECO`, `BYPASS`, or `OFF`. 


Otherwise the code logs an error and report a data stale immediately. 


I have no strong opinion on whether the driver should try to proceed issuing "load.on" to the device - if really the worst case is "no effect", then go for it :)

I looked around and am not convinced at the moment whether dstate ups.status is always a single string token, or if it can have several flags listed (OL CHRG and so on), so decided to not barge into your PR code with an unilateral fix :)


Likewise, please check (maybe later with another PR if needed) if the UPS can have such a composite status for "OFF" and "BYPASS" cases.

biergaizi:
> I'd explicitly check if the status is "on" and not assume that not-off not-bypass means "on". And then the remaining reason could be (or not be) an error case and an unknown situation.

According to NUT's coding standard on status flags ([docs/new-drivers.txt, line 218](https://github.com/networkupstools/nut/blob/master/docs/new-drivers.txt#L218)), the allowed status flags are, `OL`, `OB`, `LB`, `HB`, `RB`, `CHAG`, `DISCHRG`, `BYPASS`, `CAL`, `OFF`, `OVER`, `TRIM`, `BOOST`, or `FSD`, and their combinations. 

This is exactly what my driver does. I originally coded it to check for the absence of the `OFF` flag, because there is no one single `ON` flag to check. Here, I read your suggestion as adding three separate string tests for `OL`, `OB`, and `BYPASS`. Does it sound reasonable to you? If so, I'll push a new revision to include this change.

> And then the remaining reason could be (or not be) an error case and an unknown situation. 

In the original code, the lack of an `OFF` flag conclusively decides that the UPS is on. When the UPS status is updated, the code must set either  `OL`, `OB`, `OL ECO`, `BYPASS`, or `OFF`, because it is one of the five possible mutually exclusive values returned by the "power supply mode" register (*Table 2-1 Collection signal list*, number 1024, datasheet page 32)

> 0: no power supplied
> 1: bypass mode
> 2: mains mode
> 3: battery mode
> 5: mains ECO

If the value in the register is invalid, it triggers both `dstate_datastale();` and `STAT_INSTCMD_FAILED` with a fatal error logged. Therefore, the logic to check the lack of the `OFF` flag is conclusive and complete. 

On the other hand, I do agree that an extra `else` for an unreachable condition is still a good programming practice, and checking for the positives instead of the negatives makes the code more readable.

> I have no strong opinion on whether the driver should try to proceed issuing "load.on" to the device - if really the worst case is "no effect", then go for it :)

Thanks for the reminder. As I was writing this, I do realize that the UPS has 5 possible states. I'll do an exhaustive test to ensure it's really the case in all conditions. 

> I looked around and am not convinced at the moment whether dstate ups.status is always a single string token, or if it can have several flags listed (OL CHRG and so on), so decided to not barge into your PR code with an unilateral fix :)

Register handling in this driver is declarative and data-driven. Let me elaborate on the logic of the code and how the status flags are currently set in the driver. 

First, the lookup table for setting the dstate (both `ups.status`, and `battery.charger.status`) is defined in [`ups2000_status_reg[]`, line 676](https://github.com/networkupstools/nut/blob/fa86fc87da478aa0b84a3e719b7b22853d3c4fc0/drivers/huawei-ups2000.c#L676).

 	{ 1024, 1 /* bool must_set_flag */ , {
            /* status,  val, bit */
		{ "OFF",      0, -1, NULL, NULL },
		{ "BYPASS",   1, -1, NULL, NULL },
		{ "OL",       2, -1, NULL, NULL },
		{ "OB",       3, -1, NULL, NULL },
		{ "OL ECO",   5, -1, NULL, NULL },
		{ NULL,      -1, -1, NULL, NULL },
	}},
    /* definition for additional registers follows */

Function `ups2000_update_status()` loops over the register definitions in `ups2000_status_reg[]` one by one, reads a register and set the corresponding flags via `status_set()` or `dstate_setinfo()` accordingly. The first entry, register 1024, set either the `OFF`, `BYPASS`, `OL`, `OB`, or `OL ECO` flag according to the value of the register, which can be `0`, `1`, `2`, `3` or `5`.

There's a risk that the register may contain an invalid value and set none of the status, to prevent this, there's Boolean condition `must_set_flag` in the register definition. When `ups2000_update_status()` loops over this register but does not set any flag, the function returns failure code `1` to the caller.

After the first register 1024 is processed, more UPS status flags governed by additional registers are subsequently append to the status flag in the next loop iteration over the lookup table. 

> Likewise, please check (maybe later with another PR if needed) if the UPS can have such a composite status for "OFF" and "BYPASS" cases.

While other combinations, like `OL CHRG` are possible, a combination of different power conditions, like `OFF` and `BYPASS` is not, because the register can only take 1 out of 5 mutually exclusive values.

jimklimov:
Ok, great thanks for the clarifications and additional pass over this :)

Then, what would you currently propose: should I merge this PR "as is", or do you plan some further tests/development (the "else" case?) to be sure, in the coming few days?

biergaizi:
Please don't merge it for now. I'll revise the patch & do a force-push later. Thanks.

biergaizi:
Patch has been updated. 

Here's the relevant portion of the commit message: The logic for the off test is reworked to improve readability. The original code checks whether the UPS is on by testing the negative "not OFF" condition, which is not easy to understand - the reader must have intimate knowledge on the code to know its correctness. Namely, When the UPS power status is updated, 1 of the 5 mutually exclusive flags, OL, OB, OL ECO, BYPASS, or OFF is always set, thus the absence of the OFF flag always reliably indicates an ON status. This commits converts the single negative "not OFF" test to three positive tests for "OL", "OB" and "OFF", with additional comments added on reachable and unreachable conditions. Thus, previous uncertainty is eliminated.

jimklimov (approved):
Thanks a lot! Do you consider this improvement completed, can it be merged? :)

biergaizi:
This patch needs some additional work. I retested it and found the UPS actually automatically enters normal mode upon receiving the "load.on" command in bypass mode. Possibly, this was my original and forgotten motivation for rejecting this command. Oops.

This is an edge case and I'm not sure what's the best way to handle this. Should we reject this command or accept this behavior?

1. If we define "load.on" as "ask the UPS to deliver power to the load", we should accept it. But it seems to make `bypass.stop` redundant, and it's not something we want.

2. If we define "load.on" as providing power to the load, regardless of its source...
    * When the UPS is in bypass mode, we should reject this command.
    * However, this has additional implications - when the UPS is in bypass mode but external power is not available, it should enter normal mode.

3. If we define "load.on" strictly as a normal mode power switch, we should reject it whenever the UPS is in bypass mode.

After some thoughts, I think Option 3 is the preferred option. It'll be a one-line change to add an extra `return STAT_INSTCMD_FAILED;`. Also, I think NUT needs to standardize this behavior in the documentation. Other drivers may have the same problem.

biergaizi:
This problem also needs additional clarifications on other UPS systems if they have separate output power controls. For example, if a hypothetical UPS has `load.on` and `load.off` that only control the output power relay without any association to its current power status, online or bypass, then it would be a non-issue. Defining "`load.on` strictly as a normal mode power switch" would be a bad definition.

A more appropriately definition can be: load.on: Turn on the load immediately, without entering or exiting from bypass mode.

jimklimov:
Thanks, those are very good questions. I've CC'ed some experienced team members whose review about the architecture of NUT and goals of some toggles may be more relevant, so I'd learn something too :)

jimklimov:
Offline comment from @aquette : "between 2 and 3, with a preference for 3"

biergaizi:
I updated the commit. Now it rejects `load.on` in bypass mode. This change has been tested on my system.

jimklimov (approved):

whc2001:
Just tried almost everything on my device, seems like everything is working!

- `upsc` Display Info
  
  - [x] Info
  
  - [x] Stale & Recover from Stale

- Flags Test
  
  - [x] OL
  
  - [x] OB
  
  - [ ] ECO (To test later, need to consult the user manual for enabling it)
  
  - [x] BYPASS
  
  - [x] OFF
  
  - [x] CAL
  
  - [x] DISCHRG

- Buzzer
  
  - [x] On
  
  - [x] Off
  
  - [x] Toggle

- Bypass
  
  - [x] Start
    
    - [x] Shall ignore `load.on` when in bypass
      - In `2.7.4-3515-gad84d9d2`, `upscmd` still returns `OK` when executing `load.on` during `BYPASS`, but in syslog there actually is `load.on error: UPS is already on, and is in bypass mode. To enter normal mode, use bypass.stop`.
  
  - [x] Stop

- Load
  
  - [x] On
  
  - [x] Off

- Test
  
  - [x] Quick
  
  - [x] Deep
  
  - [x] Stop

- Shutdown
  
  - [x] Stay Off
  
  - [x] Return
  
  - [x] Reboot
  
  - [x] Graceful Reboot

- Parameters
  
  - [x] OnDelay=60
  
  - [x] OnDelay=180
  
  - [x] OffDelay=RebootDelay=6
  
  - [x] OffDelay=RebootDelay=30
  
  - [x] RebootDelay=60

jimklimov:
Possibly the question of #1391 is relevant in this driver too, seeing the comment above for https://github.com/networkupstools/nut/blob/fa86fc87da478aa0b84a3e719b7b22853d3c4fc0/drivers/huawei-ups2000.c#L676

PR #1200, opened 2021-11-23
docs/FAQ.txt: extend Q&A on "NUT proxying" with mention of "clone" driver

jimklimov:
Follows up in a way to #1196 so CC @tessus

### DIFF
--- a/docs/FAQ.txt
+++ b/docs/FAQ.txt
@@ -684,6 +684,13 @@ Example with `clone` driver:
 		load.status = outlet.1.status
 		[...]
 
+This allows to group load attached to a separately manageable
+outlet (or group of outlets) on larger UPS and ePDUs, in order
+to power those devices on/off together. This may be also useful
+to delegate management of feeds to devices for purposes like
+hosting or supporting hardware for smaller teams sharing a rack
+in a larger company.
+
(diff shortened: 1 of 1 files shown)


tessus:
Wow, ok. Another thing I did not know about. (It seems there are quite a lot of things I do not know when it comes to this project. ;-) )

However, I am not sure I entirely understand the purpose of using such a clone. It's basically just an alias, so why not use the original instead? The relay is clear to me, because it forwards/relays data.

But the clone makes no sense to me. Or at least from this explanation I don't understand why I would have to use such a clone. It is still a 1:1 relation so the only thing it hides is the name.

jimklimov:
I believe historical rationale was to group load on a separately manageable
outlet or group of outlets on larger UPS and ePDUs, to power on/off
together. Maybe also to delegate for hosting/teams/....

tessus:
@jimklimov out of curiosity... when are the html pages refreshed? I am just asking because other projects just use a git hub action after a PR has been merged.

jimklimov:
What started as a short answer here, became a mailing-list post instead :)
https://alioth-lists.debian.net/pipermail/nut-upsuser/2021-December/012597.html

PR #1241, opened 2021-12-23
Fightwarn libusb 1.0+0.1

jimklimov:
This PR updates the libusb-1.0+0.1 branch with "fightwarn" considerations (following up from effort in #823 to get the NUT codebase CI-tested and free from more and more classes of warnings, so we can enable non-regression testing for them with future iterations), allowing to proceed with issue #300 to get libusb 1.0 API supported by NUT officially. In fact, this PR would complete the current definition of "medium" level warnings, allowing to set them as default requirement.

One important change here, that was questionable to myself but got "rationalized", is the step back from fixed C-standard fixed-size numeric value types like `size_t`, to `typedef ... usb_ctrl_...` with types defined by respective backend API for libusb-1.0, libusb-0.1, libhid, libshut and some drivers. For statically linked drivers this should not matter in fact, and this change does not preclude (and even helps) later efforts to re-define those typedefs identically for all backends, and add more range-checks and casts to make the complete codebase build with that.

This PR also reduces duplication of code (libusb 0.1 vs 1.0 definitions were repeated in many drivers and libs) offloading that to usb-common.h

At least, such is the theory - testing against real hardware would be very helpful; if you do - please comment the good, the bad and the ugly outcomes in #300.

Sibling PR #1240 addresses similar refresh for the original libusb-1.0 branch, but this one (libusb-1.0+0.1) is planned to become part of NUT master codebase when the dust settles.

During the co-evolution of these branches, it was confirmed (and later assured by git-merges) that libusb-1.0+0.1 includes and builds upon libusb-1.0. The original difference was the ability to `configure` a build of NUT with one or another libusb API. As part of this PR, a newer m4 codebase to do so was applied to both branches, so now they should only differ in other bugfixes added in the libusb-1.0+0.1 branch.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -403,8 +403,13 @@ dnl the usual builds can pass by default on systems without asciidoc.
 NUT_ARG_WITH([docs], [build and install documentation (alias to --with-doc)], [man=auto])
 NUT_ARG_WITH([doc], [build and install documentation (see docs/configure.txt for many variants of the option)], [${nut_with_docs}])
 
-NUT_ARG_ENABLE([warnings], [enable warning presets that were picked as useful in maintainership and CI practice (variants include gcc-minimal, gcc-medium, gcc-hard, clang-minimal, clang-medium, clang-hard, all; auto-choosers: hard, medium, minimal, yes=auto='gcc or clang or all at hardcoded default difficulty')], [legacy])
-NUT_ARG_ENABLE([Werror], [fail the build if compiler emits any warnings (treat them as errors)], [no])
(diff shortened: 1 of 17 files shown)


jimklimov:
UPDATE: This PR got frankensteined into a mess of changes hard to revise due to a history of merges to keep in sync with master; so I sat to cherry-pick and will now be posting another to increment libusb-1.0 branch with a general fightwarn effort running-up to the "questionable" change noted above, then will rebase this PR on top of that with fewer changes to revise specific to this iteration.

jimklimov:
PR rebased over the less significant change set, now merged into target branch.

PR #1274, opened 2022-02-06
upssched "read confirmation failed, trying again" fix

dtsecon:
- fix start timer failure with log message "read confirmation failed, trying again"
- parent read from child using select() instead of alarm()

### DIFF
--- a/clients/upssched.c
+++ b/clients/upssched.c
@@ -497,14 +497,20 @@ static int sock_read(conn_t *conn)
 			if ((ret == -1) && (errno == EAGAIN))
 				return 0;
 
-			if (ret == 0)		/* nothing to parse yet */
-			continue;
+			/* O_NDELAY with zero bytes means nothing to read but
+			   since read() follows a succesful select() with
+			   ready file descriptor, ret shouldn't be 0. */
+			if (ret == 0)
+				continue;
 
 			/* some other problem */
 			return -1;	/* error */
 		}
 
 		ret = pconf_char(&conn->ctx, ch);
+
+		if (ret == 0) 	/* nothing to parse yet */
+			continue;
 
 		if (ret == -1) {
 			upslogx(LOG_NOTICE, "Parse error on sock: %s",
--- a/clients/upssched.c
+++ b/clients/upssched.c
@@ -497,14 +497,14 @@ static int sock_read(conn_t *conn)
 			if ((ret == -1) && (errno == EAGAIN))
 				return 0;
 
+			if (ret == 0)		/* nothing to parse yet */
+			continue;
+
 			/* some other problem */
 			return -1;	/* error */
 		}
 
 		ret = pconf_char(&conn->ctx, ch);
-
-		if (ret == 0)		/* nothing to parse yet */
-			continue;
 
 		if (ret == -1) {
 			upslogx(LOG_NOTICE, "Parse error on sock: %s",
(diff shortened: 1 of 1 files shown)


clients/upssched.c
```diff
@@ -757,30 +747,39 @@ static void sendcmd(const char *cmd, const char *arg1, const char *arg2)
 
 		/* we're connected now */
 
-		ret = write(pipefd, enc, enclen);
+		ret = write(pipefd, enc, sizeof(enc));
 
 		/* if we can't send the whole thing, loop back and try again */
-		if ((ret < 1) || (ret != (ssize_t) enclen)) {
+		if ((ret < 1) || (ret != sizeof(enc))) {
```
jimklimov, line 753: What was wrong with `enclen` caching the `sizeof(enc)` (IIRC)?

Also, lost casting for comparing `sizeof` to the known-positive `ret` here, so open to signed/unsigned mismatch warnings which now should be fatal (`ssize_t ret` vs. `size_t sizeof`).

Can't remember looking from phone, if `enclen` was also range-checked to fit in `SSIZE_MAX`...

clients/upssched.c
```diff
@@ -757,30 +747,39 @@ static void sendcmd(const char *cmd, const char *arg1, const char *arg2)
...
+		do {
+    			/* set timeout every time before call select() */
+    			tv.tv_sec = 1;
+    			tv.tv_usec = 0;
+
+    			FD_ZERO(&fdread);
+    			FD_SET(pipefd, &fdread);
+
+    			ret_s = select(pipefd + 1, &fdread, NULL, NULL, &tv);
+				switch(ret_s) {
```
jimklimov, line 769: indentation looks funny here

jimklimov:
good catches, thanks! got a few comments though, posted in review...

clients/upssched.c
dtsecon: - `enclen` cashes the length of string pointed by `buf` (`enclen = strlen(buf)`). There is a check also of `enclen` against `SSIZE_MAX` which is pointless taking into account that `buf` / `enc` has a fixed length of `SMALLBUF` = 512 / (512 + 8). The result was that `enc` wasn't written but the size of string pointed by `buf` 
- sizeof(enc) has been casted prior to comparison

clients/upssched.c
dtsecon: code alignment has been fixed as well

clients/upssched.c
jimklimov: Oh, that must have been a copy-paste typo, buflen elsewhere, enclen in some places.

As for SMALLBUF, pedantically, given that it is a coding-time "variable" (something people can change), and how int sizes vary (hopefully no 8/16 bit builds of NUT are relevant, but...) static analyzers wanted that resulting sizeof(stuff) checked for range.

dtsecon:
@jimklimov any idea why mr Jenkins cannot build the PR?

clients/upssched.c
```diff
@@ -757,30 +747,39 @@ static void sendcmd(const char *cmd, const char *arg1, const char *arg2)
 
 		/* we're connected now */
 
-		ret = write(pipefd, enc, enclen);
+		ret = write(pipefd, enc, sizeof(enc));
```
jimklimov, line 750: Got to a computer now for another look at the code: the contents of `buf` and `enc` are prepared by a series of `snprintf`'s, so they contain some string that is no longer than the buffer size - but may be smaller. So writing the "string, null-char, garbage" padding to the buffer size in this line's change may be pointless at best, unsafe somehow at worst.

So I'm in favor of fixing this bit back to `enclen`/`buflen` equaling the respective `strlen()` result, and will post a commit shortly.

clients/upssched.c
jimklimov: Note to self: The enclen vs buflen mess-up fixed in a later commit

jimklimov:
> any idea why mr Jenkins cannot build the PR?
Checking. After Fosshost datacenter migration to another sponsor, things are either slow or very laggy - I suppose their hypervisor may be swapping or something. So sometimes it hits some connection timeouts etc. and build agents get lost :\

Anyhow, the recent new commit here should trigger a relevant rebuild.

jimklimov:
Well... today it lags a lot worse than usual :( Agents were unable to checkout - connect to github - within the timeout...

jimklimov:
Cool, so as of https://ci.networkupstools.org/job/nut/job/nut/job/PR-1274/7/ Mr. Jenkins did not complain on any of the build scenarios it ran!

@dtsecon : is the current codebase state (with my patches added) okay with you to merge as your PR? :) Thanks!

Issue #1271, opened 2022-01-26
configure --with-drivers=X does not require its dependencies


jimklimov:
As raised in #1017:

It seems that `configure` script does not really know or care about the ever-changing population of drivers, and is geared to enable/disable whole categories (e.g. with/without modbus) so did not require to have libmodbus as a dependency for that particular enabled driver - just opted to use one if available.

> We've just bought a UPS2000-A-1KTTS. I'm not very familiar with building C
> code, i had some issues. (fresh master 88e8a83)
> Linux mgmt01 4.4.0-210-generic #242
> <https://github.com/networkupstools/nut/issues/242>-Ubuntu SMP Fri Apr 16
> 09:57:56 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
> Ubuntu 16.04 (yeah, old, currently under replacement)
> config: ./configure --with-user=ups --with-group=nut --with-usb
> --with-drivers=huawei-ups2000
> i'm not sure if it's a bug, but configure won't complain about missing
> modbus lib, but the make will fail. apt-get install libmodbus-dev is needed.

_Originally posted by @jimklimov in https://github.com/networkupstools/nut/issues/1017#issuecomment-1022228476_

Issue #1017, opened 2021-04-16
feature request : Add support for Huawei Modbus UPSs


sumsethan:
here comes the Huawei UPS2000 series UPS Modbus  Protocol  Manual
https://support.huawei.com/view/PdfRead/EDOC1000110696/SUPE_DOC/k001/document.pdf

It has a USB port and a RS232 port
USB port seems like behind a translation chip " ID 04e2:1410 Exar Corp. XR21V1410 USB-UART IC"(on lsusb command) though RS232

Any ideas guys?

sumsethan:
connecting usb port recognized as a  /dev/XRUSB0 device

jimklimov:
I believe currently there is one modbus-aware driver, `phoenixcontact_modbus`, with some efforts discussed to introduce a second. While the media connection is serial (or emulated to that), though I think RS-485 is more frequent than RS-232 for that (not that they would be easy to tell apart by looking), NUT may be lacking a driver that talks an actual protocol for that.

Actually, one of the discussed efforts is indeed a Huawei UPS2000 driver in #954 which was not merged yet due to a burst of activity and then silence from the contributor. But I suppose it would help if you can build that branch from biergaizi:ups2000 and confirm whether it works for your device.

whc2001:
Any progress on that?

biergaizi:
> @sumsethan: connecting usb port recognized as a /dev/ttyXRUSB0 device

This is new information for me. What system are you using? 

The RX21V1410 chip in UPS2000 doesn't work on the official upstream Linux kernel, it can be recognized as a CDC device, but not functional (presumably because the chip is in a special configuration). On your system, if you see that the device shows as a nonstandard name like `ttyXRUSB0`, it means you're using an unofficial, out-of-tree driver unsupported by the upstream Linux kernel, so it must be a local patch from your Linux distro.

I didn't even realize there is actually a distro that included out-of-tree driver for the chip. In all my documentation I always claim that USB is entirely unsupported, but apparently with the unofficial driver installed, USB should actually work.

Again, what is your distro?

biergaizi:
Following the same line of investigation, apparently the driver for MaxLinear RX21V1410 has just been mainlined to the upstream in January and April, and an official driver (with the standard serial port names, not the nonstandard `ttyXRUSB0`) is already included in Linux 5.13, 5.14, and 5.15-RC. This is an important development.

Perhaps it was a good decision not to merge my patch back in January. Time to retest the driver and update the documentation.

[...]

PR #1277, opened 2022-02-08
configure.ac: improve "--with-drivers=name[,name...]" handling [issue #1271]

jimklimov:
As  raised in #1017 and detailed in #1271, until now we could request specific drivers to build, but that also required to still specify dependencies involved.

With this PR, requiring a driver would require the corresponding dependencies - same as if `--with-modbus` etc. was specified, unless some other flag for that dependency is requested explicitly.

Also this fixes handling of `--with-linux-i2c=yes` and `--with-macosx-ups=yes` to fail on unsuitable platforms (unable to fulfill explicit requirement), but to not default them to "yes" blindly everywhere (via `--with-all` or `--with-drivers=all`), just on those platforms ("auto" on others to try building anyway but not failing if can't configure such build).

Closes: #1271

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -477,7 +477,21 @@ AC_ARG_WITH(drivers,
 		if test -z "${with_powerman}"; then with_powerman="yes"; fi
 		if test -z "${with_modbus}"; then with_modbus="yes"; fi
 		if test -z "${with_ipmi}"; then with_ipmi="yes"; fi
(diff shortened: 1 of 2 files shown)


jimklimov:
Note to self: had CI-side issues building docs, not related to changes in this PR but maybe worth looking into later: all 3 failed scenarios were about packager-guide.chunked (HTML) on FreeBSD. Maybe just lagged worse than recently-usual at that time range...

Issue #1288, opened 2022-02-14
Allow to persistently run certain drivers as certain user accounts


jimklimov:
Currently a root can request that various NUT daemons, including drivers, are started as some non-default user name (not the configured `RUN_AS_USER` built-in value) with a common `-u USER` option.

Drivers also support a `user = "USER"` option among the global settings in `ups.conf`, but not a tailored setting that only a certain device might require (e.g. keep running everything as `nut` or `nobody`, except that pesky USB auto-connection that wants `root` or too much magic otherwise).

Quest: add support for driver-level "user" configuration in `ups.conf` and `drivers/main.c` (and docs).

PR #1287, opened 2022-02-14
drivers/upsdrvctl: add "-d" option to pass current debug level from upsdrvctl to drivers

jimklimov:
Closes: #1036

(Note: that issue discusses other possible improvements around this subject, which are synergetic with this one and posted as separate issues and PRs)

PR #1293, opened 2022-02-15
drivers/main.c: allow to specify a user for each driver via ups.conf

jimklimov:
Closes: #1288

Enhances earlier ability to start ALL drivers as a non-built-in account, by setting `user = "name"` in `ups.conf` global section, adding a similar ability to override this further at each driver's level.

Tested to accept and report the flag (along with debug-level passthrough from `upsdrvctl`, done by #1287), e.g.:
````
# ./drivers/upsdrvctl -DDDd start nutdev1
Network UPS Tools - UPS driver controller 2.7.4-4660-gc31748b
   0.000000     [D1] Starting UPS: nutdev1
   0.000859     [D2] 1 remaining attempts
   0.001456     [D2] exec:  /usr/local/ups/bin/usbhid-ups -DDD -a nutdev1
Network UPS Tools - Generic HID driver 0.45 (2.7.4-4660-gc31748b)
USB communication driver (libusb 1.0) 0.42
   0.000000     [D3] main_arg: var='driver' val='usbhid-ups'
   0.000673     [D3] main_arg: var='port' val='auto'
...
   0.005066     [D3] main_arg: var='user' val='root'
   0.005440     [D1] Overriding previously specified user 'nobody' with 'root' specified for driver section
...
````

Anticipated as most-useful when counter-acting non-persistent device filesystem node permissions, so local-media drivers (USB, maybe serial).

Additionally, this fixes a possible issue with earlier global-level `user` setting, where the filesystem socket could be created by the driver as not accessible to unprivileged `upsd` process. Now this should behave better by default, and can be twisted explicitly with new `group` setting in global and/or driver level configuration.


jimklimov:
Testing with a `user = ...` specified at both global level and in the driver section:
````
   0.000000     [D1] Overriding previously specified user 'nobody' with 'default' specified in global section
...
   0.005517     [D1] Overriding previously specified user 'default' with 'root' specified for driver section
...
...
````

And for command-line option having precedence (though apparently the order of processing differs from my expectations when writing the message texts):
````
# ./drivers/upsdrvctl -DDDd -u www-data start nutdev1
...
   0.000000     [D1] Overriding previously specified user 'nobody' with 'default' specified in global section
...
   0.005441     [D1] Overriding previously specified user 'default' with 'root' specified for driver section
   0.005598     [D1] Built-in default user for drivers 'root' was ignored due to 'www-data' specified on command line
...
````

Issue #799, opened 2020-06-17
Tripp Lite AVRX550U using wrong protocol. Issuing restart command to UPS after shutdown?


objecttothis:
- OS name and version: FreeBSD 11.3-Release
- exact NUT version, 2.7.4
- NUT installation method: Prepackaged with FreeNAS
- exact device name and related information (manufacturing date, web pointers, ...), Tripp Lite AVRX550U
- complete problem description:
When the UPS loses AC power and receives a shutdown command from the server after the server shuts down the UPS powers down but then after two seconds powers back on even though it is still under battery power.  When I contacted Tripp Lite about the issue, they pointed out that the AVRX550U uses the 2012 protocol, but the NUT compatibility list has it using the 2010 protocol.  If Tripp Lite is correct then the protocol in AVRX550U needs to be corrected to 2012, not 2010.

Second, it's unclear if NUT is issuing the command for the UPS to power back on after powering off, but it should not be issuing that command.  I've attached the protocol documentation that Tripp Lite sent me.  You probably already have it, but just in case it's helpful, here it is.

[Communication Protocol 2012 for customer.pdf](https://github.com/networkupstools/nut/files/4794808/Communication.Protocol.2012.for.customer.pdf)

Here is the upsmon.conf file:
````
MONITOR ups@localhost:3493 1 upsmon [REDACTED] MASTER
NOTIFYCMD "/usr/local/sbin/upssched"
NOTIFYFLAG ONBATT SYSLOG+EXEC
NOTIFYFLAG LOWBATT SYSLOG+EXEC
NOTIFYFLAG ONLINE SYSLOG+EXEC
NOTIFYFLAG COMMBAD SYSLOG+EXEC
NOTIFYFLAG COMMOK SYSLOG+EXEC
NOTIFYFLAG REPLBATT SYSLOG+EXEC
NOTIFYFLAG NOCOMM SYSLOG+EXEC
NOTIFYFLAG REPLBATT SYSLOG+EXEC
NOTIFYFLAG NOCOMM SYSLOG+EXEC
NOTIFYFLAG FSD SYSLOG+EXEC
NOTIFYFLAG SHUTDOWN SYSLOG+EXEC
SHUTDOWNCMD "/sbin/shutdown -p now"
POWERDOWNFLAG /etc/killpower
HOSTSYNC 15
````

ups.conf
````
[ups]
        driver = usbhid-ups
        port = /dev/uhid0
        desc =
````

upssched.conf
````
[...]

PR #1303, opened 2022-02-17
scripts/systemd/nut-driver@.service.in: make sure drivers always try to start

jimklimov:
Work around default throttling where systemd gives up eventually.

Closes: #799 

See also https://github.com/systemd/systemd/issues/2416


jimklimov:
Fast merge since CI does not test units anyway.

PR #1310, opened 2022-02-21
NUT-Monitor: Port to Python3+PyQt5

luke-jr:
Everyone is removing support for Python 2.x, and I find NUT-Monitor a nice interface, so I ported it to Python 3.

GTK 2 wasn't available, and I hate GTK anyway, so I also ported it to PyQt5.

There's a few minor UI changes, but I think it's good enough for now.


jimklimov:
Hello, thanks for doing this, looks like quite a large effort to modernize this part :)

I just wonder about NUT aiming to support also deployments on older systems where wholesale upgrades are not always easy or feasible. That means, NUT might work well there (and older NUT-Monitor if there's GUI), but there would be no Python 3.x (or 3.10+ as seems implied here).

Earlier reshuffle made this client and its python library dual-version aware, hopefully (with the `*.py.in` renames and build-time substitution of path to interpreter; tested to behave the same for its defined test suite). I am afraid that this change as proposed "hard-codes" the python3 effectively, so that builds with python2 would just not work, but did not try it yet so just don't know.

Maybe it makes sense to instead provide the modified application in a separate directory as a separate project (and packages eventually, e.g. nut-monitor-ui2 vs nut-monitor-ui3 or similar - with clearly defined different dependencies for py2/3 vs py3 and qt vs gtk)? This would allow end-users of modern NUT on those older systems to still have a working tool.

luke-jr:
I did the port with Python 3.8, but I tested also in 3.9, and don't know any reason it wouldn't work with older/newer versions (I did intentionally avoid introducing newer Python feature usage). It might even work with 2.7 with some minor changes, but I no longer have the ability to easily test that - if someone cares about maintaining 2.7 support, perhaps they can put the effort into that? :/

>Earlier reshuffle made this client and its python library dual-version aware,

Definitely not the current client in NUT. PyGTK2 doesn't even exist for Python 3.

jimklimov:
Well, after giving this much thought - including that NUT should keep on working wherever it did in earlier releases (and those systems users are often not qualified programmers), I am inclined to have both cakes and eat them :)

I think it can be achieved by including both versions somehow - perhaps renaming clients clumsily but explicitly, with the older client as `NUT-Monitor-py2gtk2` and a copy of your codebase as `NUT-Monitor-py3qt5`, and provide the `NUT-Monitor` name as some (python? shell? packaged symlink?) dispatcher to call one or another based on actually available interpreter (eventually also UI library) version?..

jimklimov:
> Definitely not the current client in NUT. PyGTK2 doesn't even exist for Python 3.

Finally got a way to check, and can confirm that :) Indeed, UI client from master branch does not work with `python3.9` as interpreter.

jimklimov:
Ended up posting PR #1354 (could not update yours cleanly, especially as I rebased over current master branch) which should cover both user bases. Can you please check if that works reasonably for you?

Also it seems I could not enable localization for the Qt5 client, in the way I could (exporting LANG) for the older Gtk one. Should new localization files be generated, or some methods called in the updated variant of the script?

luke-jr:
I don't know anything about porting localisations from Gtk. Just wanted something that actually works on a modern system. >_<

jimklimov:
Fair enough. At least, each starts and renders :)

PR #1318, opened 2022-02-27
APC HID UPS: Add ability to set battery battery.mfr.date

WereCatf:
Some APC UPSes allow for setting battery.mfr.date, so it can be
easily referenced later, even if there were no stickers or papers
left to refer to.

This patch has been tested to work with APC Back-UPS ES 550G.

Signed-off-by: Nita Vesa <werecatf@outlook.com>

### DIFF
--- a/drivers/apc-hid.c
+++ b/drivers/apc-hid.c
@@ -128,8 +128,24 @@ static const char *apc_date_conversion_fun(double value)
 	return buf;
 }
 
+static double apc_date_conversion_reverse(const char *date_string)
+{
+	int year, month, day;
+	long date;
+
+	sscanf(date_string, "%04d/%02d/%02d", &year, &month, &day);
+	if(year >= 2070 || month > 12 || day > 31)
+		return 0;
+	year %= 100;
+	date = ((year / 10 & 0x0F) << 4) + (year % 10);
+	date += ((month / 10 & 0x0F) << 20) + ((month % 10) << 16);
+	date += ((day / 10 & 0x0F) << 12) + ((day % 10) << 8);
+
+	return (double) date;
+}
+
 static info_lkp_t apc_date_conversion[] = {
-	{ 0, NULL, apc_date_conversion_fun, NULL }
+	{ 0, NULL, apc_date_conversion_fun, apc_date_conversion_reverse }
 };
 
 /* This was determined empirically from observing a BackUPS LS 500 */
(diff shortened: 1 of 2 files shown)


jimklimov:
Looks good! 

I'll be without a computer for a while, so sanity-checking: should that be a `double` type? Not a `long` like in `snmp-ups` for example?

WereCatf:
The function prototype has been defined as returning a double, so it 
wouldn't really change anything, even if the function did the intermediary 
steps with a long, would it? I'm not against making any changes, though, if 
you insist: just tell me what you want.

The code does work with my two APC devices just fine, though.

jimklimov:
Wanter to say "Okay then" (at least until/unless this is a bug for someone)
- "just sanity-checking so far".

But then I wondered if 32-bit, or ARM/MIPS/... CPU builds would take kindly
to bit shifts in a double. It does actually feel safer to manipulate bits
in a `long` (or better e.g. `uint32_t`) temp var, and cast to `double` for
just the return.

Moving computers, can't test myself :\

WereCatf:
Is there a direct link to a logfile or something somewhere, so I could see what the CI-build failed at? The website never loads for me, no matter what I do. It just stalls and so I can't check what's going on.

jimklimov:
Unfortunately the sponsor-provided VMs behind NUT CI are quite weak or overwhelmed since the new-year migration, so it does take a while for Jenkins to render or to think a next thought :(

The website does open, at least for the legacy interface (cut the URL at build number) but indeed it can take a minute too. I'll take a look at making legacy UI the default there; the new fancy UI autorefresh kicks in before it renders the thing, it seems.

The build page should have links to specific logs of failed scenarios, primarily seek the "*.check.gz" files in the list; the general "console" log is too cluttered to make sense of usually.

WereCatf:
If I'm reading that right, it seems the build failed because...the build-process got disconnected?

```
Build agent lost in flight

A build agent got disconnected (network issue, OOM on their host, other crash). Job should be restarted to see the "real" results of that build matrix combination.
[Indication 1](https://ci.networkupstools.org/job/nut/job/nut/job/PR-1318/lastUnsuccessfulBuild/consoleFull#-15882823778fad603d-05df-417f-bee1-2f432162d8a8)
```

jimklimov:
"There is no cloud, just somebody else's " (leftover) "computers..."
sigh :(

jimklimov (approved):

PR #1320, opened 2022-02-28
Cast to unsigned type when interpreting HID descriptor length bytes (libusb 0.1)

nbriggs:
The libusb 0.1 interface definition declares a (signed) char type for
control messages.  The HID descriptor length contained within a control
message is intended to be interpreted as a pair of unsigned bytes so
we must cast to uint8_t when doing the arithmetic rather than trip over
the sign bit.

Closes #1261, closes #1312.

PR #1325, opened 2022-03-07
LibUSB related code - revise for bit-maths with signed char

jimklimov:
As recently found, the chosen approach with conversion to "typedef'ed" support of libusb/libhid/libshut/... codebases and respective APIs, including the buffer-passing that ultimately evaluated to arrays of signed or unsigned chars, had backfired a bit in cases where code did some maths (and bit maths) with individual bytes of those buffers.

As @nbriggs suggested, I went over the codebase looking for files that contain `usb_ctrl_char` string and further for operations in them that matched regex `(>>|<<|[^&~|^][&^|~][= (0-9A-Za-z])` and in fact there were not many places that looked doubtful, PR #1320 caught almost all of them which seemed significant :)

Building and testing in practice with various architectures (32/64-bit, x86, SPARC, ARM...) would be very useful I guess, to be more sure about quality of pending NUT release.

### DIFF
--- a/drivers/blazer_usb.c
+++ b/drivers/blazer_usb.c
@@ -684,7 +684,7 @@ void upsdrv_initups(void)
 		 * This should allow automatic application of the workaround */
 		ret = usb_get_string(udev, 0, 0, (usb_ctrl_charbuf)tbuf, sizeof(tbuf));
 		if (ret >= 4) {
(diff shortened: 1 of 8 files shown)


jimklimov:
I thought about that, for a bit, but it seemed too close to GetValue tests in spirit (byte merges to make a number) and target (usb)...

PR #1327, opened 2022-03-10
SNMP subdriver generation script: fixes and improvements

aquette:

### DIFF
--- a/scripts/subdriver/gen-snmp-subdriver.sh
+++ b/scripts/subdriver/gen-snmp-subdriver.sh
@@ -90,30 +90,30 @@ MODE=0
 NAME=gen-snmp-subdriver
 TMPDIR="${TEMPDIR:-/tmp}"
 SYSOID_NUMBER=".1.3.6.1.2.1.1.2.0"
-DEBUG=`mktemp "$TMPDIR/$NAME-DEBUG.XXXXXX"`
-DFL_NUMWALKFILE=`mktemp "$TMPDIR/$NAME-NUMWALK.XXXXXX"`
-DFL_STRWALKFILE=`mktemp "$TMPDIR/$NAME-STRWALK.XXXXXX"`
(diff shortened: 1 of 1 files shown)


scripts/subdriver/gen-snmp-subdriver.sh
```diff
@@ -270,14 +272,8 @@ generate_C() {
...
-	cat >> "$CFILE" <<-EOF
-
-		/* end of structure. */
-		{ NULL, 0, 0, NULL, NULL, 0, NULL }
-	};
-
-	mib2nut_info_t	${LDRIVER} = { "${LDRIVER}", ${UDRIVER}_MIB_VERSION, NULL, NULL, ${LDRIVER}_mib, ${UDRIVER}_DEVICE_SYSOID };
-	EOF
+	printf "\n\t/* end of structure. */\n\t{ NULL, 0, 0, NULL, NULL, 0, NULL }\n};\n\n" >> "$CFILE"
+	printf "mib2nut_info_t	${LDRIVER} = { \"${LDRIVER}\", ${UDRIVER}_MIB_VERSION, NULL, NULL, ${LDRIVER}_mib, ${UDRIVER}_SYSOID };\n" >> "$CFILE"
```
jimklimov, line 276: Out of curiosity: why a cumbersome `printf` and lots of file appendings instead of the single-operation append that was here before? To ensure tab chars - I think the `cat` can handle that too (especially with verbatim mode, `cat << 'EOF'` style)?

aquette: Indeed, formatting purpose...

jimklimov:
@aquette : Can you please give a shot to the current PR state, if it addresses your issue and generates valid walks from real devices?

Also found and fixed an issue with HOSTNAME (did not accept dots, so entering an IP was troublesome).

jimklimov (approved):

aquette:
feedback: good. Only the mib2nut_info_t entry that got tabbed

Issue #1263, opened 2022-01-14
v2.7.5/v2.8.0 release?


j-broo:
Please excuse me for opening an issue with this question, but is there any indication on when we can expect v2.7.5 to be released? How close is the current build to a release candidate?

Maybe make this issue pinned as I'm sure others have this question too and I don't see any info.

teoberi:
A sign [here](https://github.com/networkupstools/nut/milestones)?

jimklimov:
Good point, thanks for reminding of that mechanism... gotta remember now how to assign issues/PRs to milestones.

In practical terms, I hope to get the NUT CI farm (recently migrated to another Fosshost's another slower sponsor) to churn through some of the remaining open PRs for device supports and HCL bumps, possibly fix some smaller issues (that may wait for another release name, but would be cool to fix earlier), and go through some HW testing to catch some low-hanging-fruit issues.

With libusb-1.0 merged, and literally thousands of build warnings eradicated, as well as many screenfulls of added and updated drivers, etc. coming from the community, I hope we are quite close to releasing a solidly featured version. Pragmatic caution tells me to expect a stream of issue reports (due to the sheer amount of changes, something gotta give in!) and a subsequent release afterwards, though. Hopefully shortly afterwards, if due to that.

jimklimov:
Also thanks for the suggestion to pin. Referencing earlier similar questions for cross-linking:
* #819 
* #699

blaa:
Since 2.7.4 was released in 2016, and master is currently 4698 commits away from this tag, I believe it would make more sense to tag it as 2.8. I'm unsure if the project uses semantic versioning or not, but that's certainly not just a "patch" release, and 2.7.5 would be widely understood like a small change over 2.7.4.

Would be cool to have a release that would make it into the next stable Debian release. Bullseye was released 2021-08, so there's some time for that, but downstream packaging might take a lot of time as well.

jimklimov:
Well, it is supposed to be semantic. Speaking of it, the goals were indeed for a 2.7.5, but as I found in recent sprints - over time incompatibilities arose, such as new key words for ups.conf etc. that make config files acceptable for current master using its newest features not usable with 2.7.4. In that sense it may be indeed a 2.8 material, thanks for the reminder.

jimklimov:
[...]

PR #1336, opened 2022-03-23
Announce NUT_NETVERSION=1.3

jimklimov:
Update the `configure.ac` to report the new version in `netmisc.c`, and update docs that referred to "2.7.5" as the new release to speak about "2.8.0" which sings to similar tune as #1263.

Also documents `PRIMARY` command as alias to `MASTER` (per #840) in protocol 1.3, although the practical implementation is expected to be compatible with protocol 1.2 client/server (see #1328).

CC @NUT-RogerPrice

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -373,7 +373,7 @@ Release notes for NUT 2.8.0 - what's new since 2.7.4:
    few changes also happened in header files installed for builds configured
    `--with-dev` and so may impact `upsclient` and `nutclient` (C++) consumers.
    At the very least, binaries for those consumers should be rebuilt to remain
-   stable with NUT 2.7.5 and not mismatch int-type sizes and other arguments.
+   stable with NUT 2.8.0 and not mismatch int-type sizes and other arguments.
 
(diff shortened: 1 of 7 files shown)


aquette (approved):

clepple (commented):
You know my thoughts on numeric protocol versions.

The other parts of the PR look fine.

jimklimov:
I might have missed those thoughts in older context, so would try a guess:
should they be integer numbers? :)

I don't have a strong opinion here since NUT code itself does not make use
of the version (string?) AFAIK. But numbers would compare more easily :)

clepple:
If two network protocol upgrades are being discussed, and both branches in Git propose an incremented protocol version, one will have to be updated a second time to avoid a conflict if both are not merged at the same time. It seems like very little gain for a lot of maintenance burden. (And that's if `NETVER` doesn't get renamed...)

Other protocols have text-based "feature flags" where they indicate a set of features that are supported.

In all of the previous NUT protocol iterations, a client could simply try a new command, and check for an appropriate "ERR" response.

PR #1342, opened 2022-03-29
tools/nut-usbinfo.pl: Use hwdb for UPower rules

benzea:
A udev hwdb is faster to process, so switch to use that. Doing so means that the attributes are set further up in the chain, so we still need a rule to copy them over.

Alternatively, we could consume it differently in UPower (i.e. fetch the "usb" subsystem parent first), either works, but this was a bit more straight forward and it is hard to test otherwise.


tools/nut-usbinfo.pl
```diff
@@ -125,17 +125,8 @@ sub gen_usb_files
 	print $outputUPower '# This file was automatically generated by NUT:'."\n";
 	print $outputUPower '# https://github.com/networkupstools/nut/'."\n#\n";
 	print $outputUPower '# To keep up to date, monitor upstream NUT'."\n";
-	print $outputUPower '# https://github.com/networkupstools/nut/commits/master/scripts/upower/95-upower-hid.rules'."\n";
+	print $outputUPower '# https://github.com/networkupstools/nut/commits/master/scripts/upower/95-upower-hid.hwdb'."\n";
 	print $outputUPower "# or checkout the NUT repository and call 'tools/nut-usbinfo.pl'\n\n";
```
jimklimov, line 129: It may also help to leave a persistent link to update the file "in-place":
```suggestion
	print $outputUPower "# fetch https://raw.githubusercontent.com/networkupstools/nut/master/scripts/upower/95-upower-hid.hwdb\n";
```
...or maybe not, since support of devices depends on the binary NUT code deployed on the system, and this list seems generated from C sources (not `data/drivers.list` like in some other cases).

scripts/upower/95-upower-hid.hwdb
```diff
@@ -0,0 +1,179 @@
...
+usb:v2e51p*
+ UPOWER_VENDOR=Ever
+
+usb:v2e66p*
+ UPOWER_VENDOR=Salicru
+
+usb:v4234p*
+ UPOWER_VENDOR=Powervar
+
+usb:v03f0p0001*
```
jimklimov, line 80: Is there a particular reason to not include a comment (or `UPOWER_VENDOR` or something) to keep info about UPS makers/sellers here?

The original rules file had blocks with a top comment ("Tripplite" etc.) for groups of IDs, and here is just a wall of text with less detail. Would be great if it can be made functionally equivalent to the old approach.

jimklimov:
Generally looks useful, comments posted.

Is there any known limitation such as "hwdb starts with OS releases as recent as..." so we'd need the older style file generated for older releases that are still of practical interest (boxes used as minimally changed appliances)?

On the other hand, any rough estimate how much faster is the hwdb compared to text rules (like "1 sec vs 5 sec to parse")?

tools/nut-usbinfo.pl
benzea: We could also move the generation into the UPower repository. That is, if grabbing `data/drivers.list` from the NUT repository is enough to generate it there.

I suppose the reason for it being this way is that the data is parsed from the C source code in the first place.

jimklimov: Looking at the files more, I think the `drivers.list` is not suitable here, it is more about documenting (often post-factum) that certain drivers understand and manage certain devices - how-ever those get branded on the market, etc. 

In particular, it is not about only USB, and does not convey VID/PID info systemically (for some devices these IDs are listed among comments or names, but not exhaustively nor in a parsable manner).

scripts/upower/95-upower-hid.hwdb
jimklimov: Could you please add that then? :)

benzea: OK, how about this? It limits setting the vendor on actual UPS devices, which is sufficient. I only did that because it seems quite readable this way.

tools/nut-usbinfo.pl
benzea: > ... support of devices depends on the binary NUT code deployed on the system ...

Not sure what you mean here. NUT does not need to be installed for UPower to show warnings and trigger an emergency shutdown eventually.

scripts/upower/95-upower-hid.hwdb
```diff
@@ -0,0 +1,200 @@
...
+
+# Dell
+usb:v047cpffff*
+ UPOWER_BATTERY_TYPE=ups
+ UPOWER_VENDOR=Dell
+
+# ST Microelectronics
+usb:v0483pa113*
+ UPOWER_BATTERY_TYPE=ups
+ UPOWER_VENDOR=ST Microelectronics
```
jimklimov, line 42: Just in case: should strings with spaces be quoted?

scripts/upower/95-upower-hid.hwdb
```diff
@@ -0,0 +1,200 @@
...
+usb:v03f0p1fe0*
+usb:v03f0p1fe1*
+usb:v03f0p1fe2*
+usb:v03f0p1fe3*
+usb:v03f0p1fe5*
+usb:v03f0p1fe6*
+usb:v03f0p1fe7*
+usb:v03f0p1fe8*
+ UPOWER_BATTERY_TYPE=ups
+ UPOWER_VENDOR=Hewlett Packard
```
jimklimov, line 26: I am not sure about the hwdb format; does it apply the `UPOWER_*` details to just one `usb:*` line above it, or to all lines above? If just one, then I think it does not suffice - the other IDs are also UPSes of different models (and possibly protocol revisions) from the same vendor.

scripts/upower/95-upower-hid.hwdb
```diff
@@ -0,0 +1,200 @@
...
+usb:v03f0p0001*
+usb:v03f0p1f06*
+usb:v03f0p1f08*
+usb:v03f0p1f09*
+usb:v03f0p1f0a*
+usb:v03f0p1fe0*
+usb:v03f0p1fe1*
+usb:v03f0p1fe2*
+usb:v03f0p1fe3*
+usb:v03f0p1fe5*
```
jimklimov, line 21: For known product IDs does the line need to end with asterisk (are there further fields in the `usb:...` pattern) or can it be a faster exact match?

scripts/upower/95-upower-hid.hwdb
jimklimov: Looks better, posted more questions to clarify specifics :)

tools/nut-usbinfo.pl
jimklimov: Guess I gotta read up more on what UPower does and how... I suppose it can deal with standard USB HID devices for at least basic commands/readings/alerts served on standard paths; however the `USB_DEVICE` macro also identifies the other USB connected UPSes, like those with effectively (protocol-wise) a built-in serial-to-USB converter and usually one of Megatec Qx protocol variations.

jimklimov:
@benzea : Note for possible future fixes and force-pushes - I added a couple of commits above ;)

tools/nut-usbinfo.pl
benzea: It implements standard USB HID for reading basic status information. As I understand it, the code in the perl script is filtering the information to only include USB HID devices.

scripts/upower/95-upower-hid.hwdb
benzea: Yes, and that is the usual pattern that is also used inside of systemd for example. I guess the generated binary hwdb file is optimised for prefix searches. The real string we match against is visible from `udevadm info -e` as the MODALIAS, e.g. `MODALIAS=usb:v0463pFFFFd0100dc00dsc00dp00ic03isc00ip00in00`.

jimklimov: Hm, looking at that example - should "our" strings be upper-cased hex? Or does it match insensitively?

benzea: Wow, yep, I really got that one wrong. The MODALIAS for the usb: entries in the hwdb are explicitly uppercase (while the idVendor sysfs attributes are lowercase).

tools/nut-usbinfo.pl
jimklimov: Oh, right, found it at last:
````
                        # UPower device entry (only for USB/HID devices!)
                        if ($vendor{$vendorId}{$productId}{"driver"} eq "usbhid-ups") ...
````

**Comment by benzea** [comment]
I don't have any hardware either. I just checked the rules were working by testing them against another USB HID device that I have (a yubikey in this case).

**Review by jimklimov — COMMENTED** [review]
@benzea : Cheers, got some second thoughts which I hope you can clarify here :)

One question is posted about the code separately. Another is more generic: the original issue discussion wording, and the https://www.freedesktop.org/software/systemd/man/hwdb.html man page, seem to imply that "hwdb" is an "udev" implementation detail, instead of `udev.rules`...

Why did this PR end up changing `UPower` configuration, which seems similar and is also from FreeDesktop project (as a sort-of-descendant of HAL and DeviceKit-power), but is not udev, right?.. I am not convinced now that *this* is the file we want edited about systemd udev/hwdb implementation details. Is it even compatible? 

Alternately: are the same HWDB files reusable by both? After all, we still do generate a full-scale `"$TOP_BUILDDIR/scripts/udev/nut-usbups.rules.in";` - see "$TOP_BUILDDIR/scripts/udev/nut-usbups.rules.in"; and https://github.com/networkupstools/nut/blob/master/tools/nut-usbinfo.pl#L196-L200  

* https://upower.freedesktop.org/
* https://www.freedesktop.org/software/systemd/man/hwdb.html
* https://www.freedesktop.org/software/systemd/man/udev.html

**Comment by benzea** [comment]
> One question is posted about the code separately. Another is more generic: the original issue discussion wording, and the https://www.freedesktop.org/software/systemd/man/hwdb.html man page, seem to imply that "hwdb" is an "udev" implementation detail, instead of udev.rules...

I am a bit confused. hwdb is just a way to set variables on devices that udev knows about. What we do here, is use an existing matching for USB devices that is run anyway, to inject our own variables. One can also trigger such a match against the hwdb from a udev rule.

What is absolutely true, is that having UPOWER_* variables here is odd. Ideally, we would use some more generic names for variables and use those. At which point, we might be able to ask for the hwdb to be included in upstream systemd.

File: scripts/upower/95-upower-hid.rules [hunk 6]
```diff
@@ -1,192 +1,2 @@
...
-SUBSYSTEM=="usbmisc", GOTO="up_hid_chkdev"
-# only support USB, else ignore
-SUBSYSTEM!="usb", GOTO="up_hid_end"
-
-# if usbraw device, ignore
-LABEL="up_hid_chkdev"
-KERNEL!="hiddev*", GOTO="up_hid_end"
-
-# if an interface, ignore
-ENV{DEVTYPE}=="usb_interface", GOTO="up_hid_end"
```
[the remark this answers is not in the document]
benzea: from udev(7)
```
       SUBSYSTEM
           Match the subsystem of the event device.

       SUBSYSTEMS
           Search the devpath upwards for a matching device subsystem name.     
```
 * The hwdb tags USB devices (not the usbmisc device!)
 * We then match a "usbmisc" device that has a "usb" parent (which is probably redundant), so yes, it is honoured
 * then, with KERNEL=="hiddev*", exactly the same match as before, just inverted for obvious reasons
 * then, with IMPORT{}, we copy those tags down to the hiddev

It is not 100% equivalent. As, it for example copies the ID_PRODUCT/ID_VENDOR entries down to the `usbmisc` device (not sure why I did that, that might actually be redundant and `IMPORT{UPOWER_*}` should be sufficient).

I think `usb_interface` is just an optimization. i.e. don't bother with executing hundreds of rules for *every* endpoint. So yes, the exemption is missing but it will not matter as it is only one simple match/import rule. But, we can also add it back without any negative side effect. After all, no one needs UPOWER_* variables on every endpoint.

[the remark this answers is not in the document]
benzea: Another difference is obviously that the USB device itself also gets the UPOWER_* variables, which before it had not.

jimklimov:
Thanks for the prompt reply, even if it rather reinforced my recent concerns:

> having UPOWER_* variables here is odd

...um, because it is an UPower configuration file. We shot well, but picked the wrong target, it seems.

scripts/upower/95-upower-hid.rules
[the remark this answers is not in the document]
jimklimov: My reading was that this original matching was a sort of "if/switch" statement:
* if `SUBSYSTEM==usbmisc`, go to mid-way label `up_hid_chkdev`
* if `SUBSYSTEM==usb`, go to mid-way label `up_hid_chkdev` (by default, because otherwise we bail)

So if `SUBSYSTEM` is any one of these two - try next steps (consider hiddev and interfaces); otherwise don't bother.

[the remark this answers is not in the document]
jimklimov: As for interfaces and settings on "every endpoint", not being a domain expert I am in fact not sure what we should want here technically. In the past couple of years more and more devices appear which serve UPS HID protocols on a non-zero USB interface number, through libusb. If these translate on Linux to different filesystem nodes with their separate access permissions, and if this is the filter which separates base device (and/or interface #0 used for ages) from the non-zero interfaces, then perhaps we do want to skip the filter. It might however break some other software, I guess (whose generic/storage/scanner/whatever endpoints would get owned by `nut` and inaccessible to their specific software).

I just don't know for sure here, so voicing what looks like reasonable-guess concerns.

jimklimov:
CC @mbiebl

jimklimov:
Looking at code, where a copy of this database landed in v0.99.18 (by you, so I guess you know whatcha doing after all ;p)...

https://gitlab.freedesktop.org/search?group_id=1217&project_id=139&repository_ref=v0.99.18&scope=blobs&search=%22UPOWER_ literally does go for
````
type = g_udev_device_get_property (native, "UPOWER_BATTERY_TYPE");
````
and 
````
	/* prefer UPOWER names */
	vendor = g_udev_device_get_property (native, "UPOWER_VENDOR");
	if (vendor == NULL)
		vendor = g_udev_device_get_property (native, "ID_VENDOR");
...
````

in https://gitlab.freedesktop.org/upower/upower/-/blob/v0.99.18/src/linux/up-device-hid.c

However this code may be tailored to what you have in the database. "WUP" uses a somewhat different name set, other files may be provided (or not) by distros or users - but are not in codebase here, and I have yet to find a 1-2-3 instruction on setting it up which would sort of imply a schema :D

jimklimov:
On a side note, does "linux" in source name imply that others (*bsd) do not use this rule set at all?

benzea:
I have the feeling that there must be a slight misunderstanding on how it all works (and why like this).

So, udev (part of systemd), is the infrastructure to discover devices on the system. udev uses rules to e.g. set properties, bind drivers, etc. The hwdb is just an optimisation at the end of the day. There is a set of default rules maintained by systemd upstream[^1] and everyone is allowed to add their own rules as suits them.

Now, we have UPower and NUT, which both need the same information (what HID device is actually a UPS). We choose to set properties in udev as that is a really good solution to the discovery problem.

-----

> * revert this (if upower after all does not play well with HWDB format)?

I do not understand this point. Using the HWDB is a **much superior** solution here.

> * change/replace the other rules file generation?

Sure, happy with that. We just need some hwdb that is getting applied. Doesn't matter what it looks like exactly.

> * do we need two rules files simultaneously? …

* udev is pretty much guaranteed to be available (non-systemd distros also have it; OpenWRT might be a niche were it is different)
* Duplicating a small set of udev rules is completely fine, the performance impact is minimal.
* I think that two rule files make sense from a distribution point of view (they might even be identical). The reason is that a direct depdencency of UPower on NUT is just not fun for anyone.

So, if both UPower and nut ship the same hwdb (but different rules), that would be a good solution! `systemd-hwdb` would merge them and the resulting binary hwdb will be the same. Still really efficient for udev to process and avoids all of the distribution issues.

[^1]: Systemd upstream again pulls rules/hwdb information from other sources. Upstreams for this information are e.g. ChromeOS and libfprint, IIRC.
[^2]: In the libfprint case the rules enable USB power management. And that is desirable in particular when you are not using the fingerprint reader.

benzea:
> However this code may be tailored to what you have in the database. "WUP" uses a somewhat different name set, other files may be provided (or not) by distros or users - but are not in codebase here, and I have yet to find a 1-2-3 instruction on setting it up which would sort of imply a schema :D

Just invent a new schema that everyone can use in the future. I would say what UPower does is quite solid, but maybe you disagree ;-)

> On a side note, does "linux" in source name imply that others (*bsd) do not use this rule set at all?

I suppose, *bsd does not have udev. So obviously, no. But, on linux you *really should* use udev.

benzea:
> * something new for me is about systemd merging different rules …

Not quite. Logically it is executing everything in sequence (including the hwdb). The different directories are just different sources, all files are treated equally and are sorted by name (hence they always have a 2 digit numeric prefix).

I was saying the merging about the hwdb. The rule here is also "last matching entry wins". hwdb is an optimised binary format for doing the matching in the hwdb file and systemd-hwdb should optimise it if the input had duplicate entries.

> * Conversely though, if two rules/hwdb files have same (or worse - slightly different, e.g. different GROUP) content and different base names - won't that double the run-time overheads, possibly open a door for race-conditions, etc?

As said above, the result is simple: last one wins. There is no race, because everything waits for the udev daemon to be done processing the rules.

> * "non-systemd distros" also having udev still implied Linux, not other OSes (many of which have their own equivalents, like devd... does upower cooperate with that?)

UPower has some code, but it is in a bad state and only handles a small subset of what the Linux codebase can do.

PR #1357, opened 2022-04-08
drivers/mge-hid.c: mge_claim(): test that VendorID==PHOENIXTEC 0x06da…

jimklimov:
… has also Vendor or Product containing "AEG"

Follows up from investigation for #564 and #560 that we have a conflict for USB Vendor ID 0x06da in the upshid-ups subdrivers. This PR aims to separate the detection in a way that the mge-hid.c would only claim devices whose Vendor or Product strings contain "AEG", and so allow liebert-hid.c to still support the rest of devices with this chip.

This is an opportunistically proposed "logically meaningful" change as I have no such devices to really test against. Looking at existing code, I think that liebert-hid.c got shadowed after commit a08a2292afc559d6dfb9eede014a941d8f6b511d (in 2016) and all devices with this VID were grabbed by mge-hid.c after NUT 2.7.4.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -180,6 +180,10 @@ refer to this change set (too long in the making) as NUT 2.7.5.
      of firmware for) CyberPower CPS*EPFCLCD [issue #439, PR #1245]
    * added `onlinedischarge` option for UPSes that report `OL+DISCHRG`
      when wall power is lost [PR #811]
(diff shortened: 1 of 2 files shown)


clepple (approved):
LGTM, but no AEG hardware to test, and unclear when I will be able to rebuild on the box with the MGE UPS.

PR #1361, opened 2022-04-10
Report deprecation messages when starting older Qx drivers

jimklimov:
Seeing quite a few `blazer_ser` and `blazer_usb` mentions in the issue tracker (for varied NUT version TBH), I thought that it would be helpful to point users who copy-paste configurations from older tickets and blogs to try `nutdrv_qx` instead.

I looked for drivers whose source mentions `"Q1"` or `"Q3"` and came up with the ones in this PR:
* bestups
* blazer
* masterguard

There are also some whose protocol I am not sure about, but per "fightwarn" effort they needed similar fixes (at least among themselves) :)
* riello
* bcmxcp
* richcomm

### DIFF
--- a/UPGRADING
+++ b/UPGRADING
@@ -50,10 +50,11 @@ Changes from 2.7.4 to 2.8.0
 
 - oldmge-shut has been removed, and replaced by mge-shut.
 
-- New drivers for devices with "Qx" family of protocols should be developed
-  as sub-drivers in the `nutdrv_qx` framework for USB and Serial connected
-  devices, not as updates/clones of older e.g. `blazer` family and `bestups`.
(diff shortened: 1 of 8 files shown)


jimklimov:
Thinking of it, maybe the man pages should reflect this too.

For reviewers: is the wording generally okay to copy-paste further?

clepple (changes requested):
`tripplite.c` is the serial version of `tripplite_usb.c`, but it is not a Qx protocol, and is most likely not covered by `nutdrv_qx` (unless there was a refactoring that I was not aware of).

others LGTM.

jimklimov:
Thanks, updating it back :)

PR #1383, opened 2022-04-20
Hook NIT into (dist)check targets

jimklimov:
Recent PRs added the "usable PoC" of NIT (NUT Integration Testing suite), which is a portable-shell script to set up the sandbox with a running `upsd` and some dummy-ups instances, and tests them with a few `upsc` queries as well as PyNUT self-test suite (so that code gets regularly covered also).

In some ways it is similar to Ubuntu QART `test-nut.py` (see issue #3) which tests a packaged installation; however NIT should be able to test a recent build iteration right in the source/build tree, as well as a separate (packaged) recent version of NUT with compatible CLI, without impacting system configuration, processes and communication ports/sockets.

This PR refines the script a bit and integrates it with `make check` and `make distcheck` recipes for NUT.
It also adds and calls a simple C++ client `cppnit` (optional, where we can build it) to test `libnutclient`, and updated `dummy-ups` to address issue #1385 uncovered during this work.

Further plans include:
* devising a way to modularize this script, calling externally tracked test suites against prepared sandbox (and giving them tools to recycle configs and daemons as needed) instead of keeping a huge single script for everything;
* integrate with TAP protocol instead of arbitrary run-and-log approach: https://www.gnu.org/software/automake/manual/html_node/Using-the-TAP-test-protocol.html and https://www.gnu.org/software/automake/manual/html_node/Scripts_002dbased-Testsuites.html
* reduce noise by default (e.g. stderr from daemons, especially where we expect them to fail with invalid configs, etc.)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -118,7 +118,7 @@ all-docs check-docs \
 man all-man man-man check-man man-html all-html:
 	cd $(builddir)/docs && $(MAKE) $@
 
-check-NIT:
+check-NIT check-NIT-devel:
 	cd $(builddir)/tests/NIT && $(MAKE) $@
 
 # This target adds syntax-checking for committed shell script files,
--- a/NEWS
+++ b/NEWS
@@ -303,8 +303,8 @@ refer to this change set (too long in the making) as NUT 2.7.5.
    in upscmd
 
  - dummy-ups can now specify `mode` as a driver argument, and separates the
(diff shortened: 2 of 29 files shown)


jimklimov:
This work led to some improvements in the codebase overall, as well as the addition of these tests (and making sure that pre-release codebase behaves well), but the test script as it is now remains noisy and somewhat time-consuming.

I will make it an optional part of the build routine with `configure --enable-check-nit` which would add this target to usual `(dist)check` list.

Issue #1369, opened 2022-04-14
Enhance USB device and subdriver matching in usbhid-ups to be on par with nutdrv_qx


jimklimov:
Currently nutdrv_qx sources and features seem more advanced than those in usbhid-ups in regard to device matching abilities to pick a sub-driver out of the box. While the former has an extensive table with not only vendor/product IDs but also optionally string matching (so devices with same interface chips but different protocols can be differentiated), the latter assumes that each vid/pid combo can mean only one subdriver (modulo MGE HID for AEG or Liebert/Phoenixtec per #1357 hack), and just processes the list of subdrivers it knows until there is a hit.

This issue proposes to approach the situation holistically, and see if the solution used in nutdrv_qx can be ported into usbhid-ups?

Also, a user-driven `subdriver` option would be useful to try handling unknown devices.

PR #1384, opened 2022-04-20
nut-scanner does not detect UPS with unknown ProductID

vincentto13:
nut-scanner is not able to detect properly UPS if product id
is defined as 0xffff in usb_device_table. For instance entry:

{ 0x2e51, 0xffff, "usbhid-ups" }

Does not recognize properly my UPS, which appears as

Bus 003 Device 002: ID 2e51:0000 EVER ECO PRO AVR CDS

with lsusb command, but is fully supported.
I don't know if setting 0xffff in usb_device table was intentional,
for this purpose, but this commit treats ProductID from
usb_device_table, as bitmask while traversing over usb devices.

The following approach resulted in, the following output running
nut-scanner with my Ever UPS attached:

./nut-scanner -U
Scanning USB bus.
[nutdev1]
    driver = "usbhid-ups"
    port = "auto"
    vendorid = "2E51"
    productid = "0000"
    product = "ECO PRO AVR CDS"
    serial = "XXXXXXXXXXXXXXXXXXXXXXX"
    vendor = "EVER"
    bus = "003"

### DIFF
--- a/tools/nut-scanner/scan_usb.c
+++ b/tools/nut-scanner/scan_usb.c
@@ -189,7 +189,7 @@ static char* is_usb_device_supported(usb_device_id_t *usb_device_id_list,
 
 	for (usbdev = usb_device_id_list; usbdev->driver_name != NULL; usbdev++) {
 		if ((usbdev->vendorID == dev_VendorID)
-		 && (usbdev->productID == dev_ProductID)
+		 && ((usbdev->productID == dev_ProductID) || ((usbdev->productID & dev_ProductID) == dev_ProductID))
 		) {
 			return usbdev->driver_name;
 		}


jimklimov:
Hi, a ProductID (and VendorID for that matter) is not a mask but a number of entity assigned by some authority (products listed by their vendor, vendors registered with USB IF). Except that many boxes and their makers do a poor job at this or blatantly lie, using bogus numbers.

Proper diligently used Product IDs let us differentiate an MS keyboard from an MS mouse from an MS headphone or a hypothetical MS UPS - and separate the UPSes of same vendor using different protocols over USB. And they let us match a driver (and dialect subdriver) suitable for this or that device.

So here a better solution is to add the (parsed) comment+macro for new vendors, and USB_DEVICE entry for the new vid:pid pair to the driver which works best (also bump driver.list.in for HCL and the DDL if possible). And hope this really is the only protocol using such pair...

jimklimov (changes requested):
Mask is not the way.

vincentto13:
Cool I pretty agree with proposed solution. I wanted to make it smart, and bitmask seemed to be a generic solution, but you're right that, that if Vendor 0x2e51 would release keyboard with different ProductID, this commit would find keyboard as UPS ;). I was thinking about inheriting from driver common, as the partial_support. But this would be like major refactor, which I'd like to avoid, as I don't know the NUT code.


I'll go with your proposed solution, but can you please explain what is HCL, and DDL in this context ? Btw thank you for commits that made Ever UPS supported by NUT :)

jimklimov:
Regarding HCL, you found it :) Currently it is more of a source for documentation (and generated web site) that allows to semi-structurally collect user reports (and some comments) of what drivers worked for which devices. It is also a reference for developers e.g. to identify unrelated products (brand, protocol) having same basic identification (often not re-branded USB chip info).

The DDL is tracked in a nearby repo https://github.com/networkupstools/nut-ddl and represented at the website; section https://networkupstools.org/ddl/#_file_naming_convention details how to contribute to it. This is a detailed report of which devices were known to work with which drivers in which NUT releases, what data points they reported, what commands they supported, etc.

The line you posted seems to be in `tools/nut-scanner/nutscan-usb.h` which is a generated file, based on collection from roughly `drivers/*-hid.c` during build, so the SCM-trackable solution is to update in this case the original `drivers/ever-hid.c` file. Since this is about a new product ID for existing vendor, see if adding a line to `ever_usb_device_table[]` there and rebuilding NUT would let your deployment discover your UPS model "out of the box".

As for refactoring existing committed code, sure there are a lot of "improvement opportunities" so feel welcome to dig more into NUT codebase to make it better :) Keep in mind that nut-scanner in particular may be complicated, since on one hand it uses `libltdl` to optionally use third-party shared objects - e.g. its binary is not pre-compiled to require libusb, libsnmp and so on, but picks them up at run-time if available or skips if not. On another hand, most of its logic is delivered as a separately usable `libnutscan` for which the `nut-scanner` front-end program is just one of possible consumers. So sharing more code between it and the bulk of NUT may require some trickery with source code layout and/or "helper libs" in the makefiles to compile same sources into different object files; there already is precedent for multiple-use of serial-port code, etc.

vincentto13:
Thanks for detailed explanation. I've created another Pull Request #1387, and once the change will be merged, I'll create a second one against https://github.com/networkupstools/nut-ddl. This one can be discarded I guess.

jimklimov:
As a heads-up, recent NUT releases have added a `subdriver` option for `usbhid-ups` driver, so you can force it to try a mapping of your choice against given (explicitly) `vendorid`, `productid` and perhaps other fields that let us identify the device uniquely, see #1369 and #2058

PR #1406, opened 2022-04-27
WIP: Support for Voltronic Axpert P30 protocol inverters

minfrin:
Adds support for the Voltronic Axpert inverters based on the P30 protocol.

Updated the nutdrv_qx driver to support the optional sending of commands
protected by CRC.

Initial support for query commands. Based on work done in the Voltronic
Sunny driver.

Still TODO:

- Implement commands to write to the inverter.
- Remove commented code originating from sunny driver.
- Further testing.
- Update docs (manpages, acknowledgements...)


drivers/nutdrv_qx.c
```diff
@@ -928,7 +935,7 @@ static int	krauler_command(const char *cmd, char *buf, size_t buflen)
 
 		int	retry;
 
-		if (strcmp(cmd, command[i].str)) {
+		if (strncmp(cmd, command[i].str, cmdlen)) {
```
jimklimov, line 938: Sanity-checking this change here and in many other routines: should it be `strncmp` to only compare the first N characters of the `cmd`? By the new argument, seems so, but still...

Avoiding an unintended `strncmp("shutdown", "shut up!", 4) == true` sort of mishap ;)

drivers/nutdrv_qx.c
```diff
@@ -3915,8 +3920,8 @@ static int	qx_process_answer(item_t *item, const size_t len)
 
 	/* Short reply */
 	if (item->answer_len && len < item->answer_len) {
-		upsdebugx(2, "%s: short reply (%s)",
-			__func__, item->info_type);
+		upsdebugx(2, "%s: short reply (%s) %d<%d",
```
jimklimov, line 3923: for `size_t len` formatting symbol is `%zu` not `%d`

drivers/nutdrv_qx.c
```diff
@@ -3966,12 +3972,12 @@ int	qx_process(item_t *item, const char *command)
 
 	/* Prepare the command to be used */
 	memset(cmd, 0, cmdsz);
-	snprintf(cmd, cmdsz, "%s", command ? command : item->command);
+	snprintf(cmd, cmdsz, "%s%n", command ? command : item->command, &cmd_len);
```
jimklimov, line 3975: Cute trick with `%n`, but why not just check the return value of snprintf (error if negative, number of chars without the ending null-byte otherwise)?

jimklimov (changes requested):
At least for changes in existing codebase, there are mismatches that gotta fail compilations ;)

Try to iterate with `BUILD_TYPE=fightwarn-clang ./ci_build.sh` and then just `make`/rinse/repeat to iterate small fixes until no warnings remain, then polish with `BUILD_TYPE=fightwarn-gcc ./ci_build.sh`

minfrin:
This is a Work-In-Progress based on a very old branch containing code for a sister driver, on the case getting this to compile clean for master, and then 2.7.x.

drivers/nutdrv_qx.c
minfrin: Overflow, mainly.

s(n)printf from what I can see returns the theoretical length assuming an infinite buffer, rather than the actual length constrained by the buffer.

Ideally the snprintfs should be memcpys, as the buffer could be either a zero terminated string, or a non zero terminated buffer.

The non-zero terminated buffer part is because of the voltronic's CRC implementation, which is two binary possibly-zero digits before the \r.

This is why we need all the changes to what was until now just a simple string of non-zero characters.

drivers/nutdrv_qx.c
```diff
@@ -154,7 +156,7 @@ static struct {
 
 /* == Support functions == */
 static int	subdriver_matcher(void);
-static ssize_t	qx_command(const char *cmd, char *buf, size_t buflen);
+static int	qx_command(const char *cmd, size_t cmdlen, char *buf, size_t buflen);
```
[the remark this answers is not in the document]
minfrin: Very old branch, tested against 2.7.4. I'm trying to keep the commit attributions of the CRC implementation code, not sure if git is going to figght with me on that.

I've updated the patch to master, going to commit it after some food :)

minfrin:
Replaced by https://github.com/networkupstools/nut/pull/1407

Issue #1419, opened 2022-05-10
Set up (CircleCI for) testing of NUT builds on MacOS


jimklimov:
Following up from #823 and #869 more specifically, after the creation of new NUT CI farm we have a big selection of open-source systems running as agents for the NUT CI farm, but so far no MacOS machines were offered as builders. The venerable Buildbot setup contributed by a team member is also supposed to go away this summer, and for the past ~100 builds of the master branch the Mac worker was not available. And then we got a #1415 in the recent release.

This ticket serves as reminder that MacOS should be tested, and that CircleCI seems to offer some 6000 minutes per month for free:
* https://circleci.com/execution-environments/macos/
* https://circleci.com/pricing/
* https://circleci.com/docs/2.0/hello-world-macos/
* https://circleci.com/docs/2.0/sample-config/
* https://circleci.com/docs/2.0/configuration-reference/#docker-machine-macos-windows-executor
* https://circleci.com/developer/orbs/orb/circleci/macos

jimklimov:
PoC of the CircleCI setup evolved to a parameterized job making it easy to define several scenarios to build on MacOS (gcc/clang with different revisions of gnu/strict-C language).

One notable downside at the moment is that the preparation of VM (install of dependencies) takes about 1.5 minutes per scenario, and there seems no easy neat way to run several scenarios using the same machine, other than explicitly running one opaque job with many steps. So wasting ~7 minutes per build (4 scenarios) from our 6000 minutes per month free allocation.

Issue #1415, opened 2022-05-07
nut 2.8.0 build failure on MacOS


chenrui333:
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>build error</summary>

```
dummy-ups.c:238:24: error: no member named 'st_mtim' in 'struct stat'
                                        if (datafile_stat.st_mtim.tv_sec != fs.st_mtim.tv_sec) {
                                            ~~~~~~~~~~~~~ ^
dummy-ups.c:238:45: error: no member named 'st_mtim' in 'struct stat'
                                        if (datafile_stat.st_mtim.tv_sec != fs.st_mtim.tv_sec) {
                                                                            ~~ ^
2 errors generated.
```

</details>

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/2287073413
relates to Homebrew/homebrew-core#101035

jimklimov:
That would be MacOS? Which version? Hope I can find some clues in public
headers. Maintaining compat with obscure proprietary systems is hard
guesswork :\

Aldernately, can you locate and experiment how tge file modification time
field is called in `struct stat`, if present?

jimklimov:
Thanks for the nudge, I managed to get a MacOS build environment (on CircleCI) willing to build NUT, so iterated some more attempts to see and fix (or quickly hide) some more platform-specific issues which issued warnings and CI is set to treat those as fatal. Will post a PR soon-ish when that job finally passes :)

chenrui333:
Thanks @jimklimov！

PR #1421, opened 2022-05-10
CircleCI project setup and some fixes for MacOS

jimklimov:
https://app.circleci.com/pipelines/github/networkupstools/nut?filter=all

Closes: #1415 
Closes: #1419 

I'd say this is a "usable PoC quality", in that builds apparently pass, but implementation (detection instead of magic-passing of env vars for build settings) can be refined so it just works out of the box... later, and better done by someone with an actual Mac :)

CC @clepple : there were some fixes for macos driver, can you give those a second look please? and/or check on a live system, to make sure it still behaves well? :)


clepple (approved):
Tested the driver (commit d8144f7) in standalone mode on 10.15 and it seems to work fine.

jimklimov:
Great to hear, thanks!

PR #1422, opened 2022-05-12
Adding insufficient memory libusb error handling to reconnect UPS ins…

tdy91:
…tead of doing nothing.

For example, this error appears after 1 to 8 hours on an Eaton MGE 1500 UPS (0463:ffff) and
causes many "Can't retrieve Report 0a: Resource temporarily unavailable" errors, followed by a
passage of UPS in ups.tatus "OL OFF" (Online Off) which immediately cuts off computer power supply !!!
Trying to reconnect solves this prohibitive problem


jimklimov:
Thanks! Looks reasonable, will wait for CI to confirm these symbols exist on all OSes it covers.

PR #1431, opened 2022-05-19
docs/config-notes.txt: Fix link to the blazer_usb manual page

federicomenaquintero:
There's no man page for blazer[8]; I think blazer_usb is the intended
one, since all the links in that set are to the USB drivers.

(I was reading https://networkupstools.org/docs/user-manual.chunked/ar01s06.html and that link was broken.)


jimklimov:
Thanks, seems like a right thing to do and a remnant from older times.

Issue #1427, opened 2022-05-14
2.8.0 does not construct ups.status string for Cyberpower UPSes due to "NULL" string in cyberpower-mib


awused:
* FreeBSD 13.0-release
* NUT 2.8.0 (installed via ports)
* UPS: Cyberpower PR1500LCD + RMCARD205
  * Previously used the same network card with another UPS without issue.
  * Was on firmware version 1.1.2, tried updating to 1.3.4. No apparent change. Not going to update to 1.3.5 as I still have a Linux client on 2.7.4.

With 2.7.4 I had no connectivity problems, but with 2.8.0 as soon as I start nut_upsmon it begins broadcasting `Communications with UPS cyberpower@localhost lost` and `UPS cyberpower@localhost is unavailable`. It never seems to establish a connection. All other tools seem to work just fine over snmp v1. My ups.conf is extremely basic, only specifying driver, community, and port. I did try specifying `synchronous = no` but it made no difference.

<details>
<summary>upsc output (2.8.0)</summary>

```
battery.charge: 100
battery.current: 3.20
battery.runtime: 3960
battery.runtime.elapsed: 0
battery.temperature: 34
battery.voltage: 27.60
battery.voltage.nominal: 24
device.contact: Administrator
device.description: UPS SNMP Card
device.location: Server Room
device.mfr: CYBERPOWER
device.model: PR1500LCD
device.serial: PYTLZ2000045
device.type: ups
driver.name: snmp-ups
driver.parameter.pollinterval: 2
driver.parameter.port: 192.168.42.127
driver.parameter.synchronous: auto
driver.version: 2.8.0
driver.version.data: cyberpower MIB 0.51
driver.version.internal: 1.21
input.frequency: 60
input.voltage: 118.90
output.current: 3.20
output.frequency: 60
output.voltage: 118.90
ups.delay.reboot: 0
ups.delay.shutdown: 180
ups.delay.start: 0
ups.firmware: CR01203J9O1
ups.load: 26
ups.mfr: CYBERPOWER
ups.model: PR1500LCD
ups.serial: PYTLZ2000045
ups.status:
```
</details>

Notably ups.status is blank, and querying the IDs from https://github.com/networkupstools/nut/blob/master/drivers/cyberpower-mib.c for `ups.status` I get these values, which should show up as `OL` if I understand the code correctly. That is what I see on 2.7.4.
```
[...]

PR #1381, opened 2022-04-18
drivers/cyberpower-mib.c: update cyberpower_power_status[]…

jimklimov:
… with new values [#1377] [#590]

To reviewers: is "OL TRIM" etc. correct to couple tokens, or should one be used like e.g. "BYPASS" in some drivers?

Issue #1377, opened 2022-04-16
Update CPS MIB for new power states


jimklimov:
Latest RMCARD FW (1.3.5 - https://www.cyberpowersystems.com/product/software/firmware/rmcard205305-firmware-v135/ ) changed the ups.status OID when on ECO mode. It looks like prior firmware were setting status to onLine (2) when in ECO mode.
Changelog for 1.3.5 ( https://dl4jz3rbrsfum.cloudfront.net/documents/CyberPower_RN_RMCARD205_305-Firmware-v1.3.5.pdf )

- Corrects the OID value of the UPS Output Status when UPS is in ECO Mode.

Looking into MIB 2.9 - there is a specific code of 8 for onECO mode.
```
upsBaseOutputStatus OBJECT-TYPE
   SYNTAX INTEGER  {
      unknown(1),
      onLine(2),
      onBattery(3),
      onBoost(4),
      onSleep(5),
      off(6),
      rebooting(7),
      onECO(8),
      onBypass(9),
      onBuck(10),
      onOverload(11)
   }
```

However the current CPS SNMP driver does not have status option for ups status code 8:

```
static info_lkp_t cyberpower_power_status[] = {
	{ 2, "OL", NULL, NULL },
	{ 3, "OB", NULL, NULL },
	{ 4, "OL BOOST", NULL, NULL },
	{ 5, "OFF", NULL, NULL },
	{ 7, "OL", NULL, NULL },
	{ 1, "NULL", NULL, NULL },
	{ 6, "OFF", NULL, NULL },
	{ 0, NULL, NULL, NULL }
} ;
```

Thus when ups is on ECO mode using RMCARD FW 1.3.5 the ups.status is blank/unknown

_Originally posted by @dcglabs in https://github.com/networkupstools/nut/issues/590#issuecomment-1099450552_

jimklimov:
PR merged, hope it works :)

Issue #590, opened 2018-08-13
Support CyberPower RMCARD205


C4Wiz:
can you add support for this card please?

username115:
I would like to second this request

dogsbody:
+1

This card supports SNMP v1 & v3

The MIB file is [CyberPower_MIB_v2.9.MIB.TXT](https://github.com/networkupstools/nut/files/3993029/CyberPower_MIB_v2.9.MIB.TXT) which was downloaded (and renamed for uploading) from https://www.cyberpower.com/uk/en/product/sku/rmcard205#downloads

aquette:
Please look ar #416 to prépare the needed data. I'll do my best in january
ti look ar this.

Le dim. 22 déc. 2019 à 22:33, Dan Benton <notifications@github.com> a
écrit :

tnagels:
@aquette I'd love to help with this as I am also in need of the update. I don't understand what you mean by referring to [#416](https://github.com/networkupstools/nut/issues/416), there is a ton of information there. What exactly do you expect from me/us?

aquette:
hi @tnagels 
look for the instructions around gen-snmp-subdriver.sh to provide me a base.
I'll then turns this into a branch like in https://github.com/networkupstools/nut/issues/416#issuecomment-554307433, so that you can test and we can progress

Update: note that we already have https://github.com/networkupstools/nut/blob/master/drivers/cyberpower-mib.c
I'd be first interested in seeing the results from this.
Also note that NUT 2.7.4 (last official release) only have version 0.1 of this driver, while the GIT version has 0.2, so it would be better to use GIT (in which case, use the above referenced comment to get and compile)

tnagels:
@aquette I will try to check this later this week on a separate system. The setup I am using now does not allow me to do this (it is in a managed container I have little control over).

Meanwhile, I'll also check the driver you referenced by just comparing the MIB entries with what I get when doing an snmp walk. If they match up there is a decent chance the driver will work i guess.

Thomas

aquette:
@tnagels ack. it's possible that this current driver needs some update and completion though. Time (and your tests) will tell.

freecableguy:
Would also appreciate seeing RMCARD205 support added.

cybrnook:
@aquette I have one of these cards in my UPS at home, so have free access to it anytime. However, trying to read above, and your reference to the link for #416 , I don't even know where to start to help you and the community. Just a normal Joe, so would need really step by step to do this.

[...]

PR #1432, opened 2022-05-23
Update cyberpower-mib.c C table ordering

jimklimov:
Closes: #1427

Introduced by  #1381 for addressing #1377 (raised in #590 discussion).

### DIFF
--- a/drivers/cyberpower-mib.c
+++ b/drivers/cyberpower-mib.c
@@ -24,7 +24,7 @@
 
 #include "cyberpower-mib.h"
 
-#define CYBERPOWER_MIB_VERSION		"0.51"
+#define CYBERPOWER_MIB_VERSION		"0.52"
 #define CYBERPOWER_OID_MODEL_NAME	".1.3.6.1.4.1.3808.1.1.1.1.1.1.0"
 
 /* CPS-MIB::ups */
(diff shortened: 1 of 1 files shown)


drivers/cyberpower-mib.c
```diff
@@ -44,6 +43,10 @@ static info_lkp_t cyberpower_power_status[] = {
 	{ 9, "OL BYPASS", NULL, NULL },	/* onBypass */
 	{ 10, "OL TRIM", NULL, NULL },	/* onBuck */
 	{ 11, "OL OVER", NULL, NULL },	/* onOverload */
+
+	/* Note: a "NULL" string must be last due to snmp-ups.c parser logic */
+	{ 1, "NULL", NULL, NULL },	/* unknown */
```
awused, line 48: Should this just be "" like it is below in cyberpower_battery_status? If the value returned being checked is 1 it will stop iteration without adding "NULL" to the status line, though in practice it's the same result whether it's "" or "NULL".

jimklimov: Sorry, can't really say and can't investigate deep at the moment.

In original git commit from a decade ago that introduced this table, it was already like this. Later https://github.com/networkupstools/nut/commit/2438ff5e7ae2dcc505934add1b1bfc2cb0ab51f8 changed some `"NULL"` strings to known states, leaving this one in the middle of remaining table.

PR #1448, opened 2022-06-18
Huawei UPS2000: support newer UPS2000G/CH341 variants.

biergaizi:
We received a report of the existence of a new hardware variant from the bug tracker [1].

This UPS unit has the model number UPS2000-G-3KRTS just like a model we've previously tested, and it has the same V2R1C1SPC50 firmware, which was tested by us as well (although not on UPS2000-G-3KRTS but on UPS2000-A-2KTTS).

However, there are two obvious hardware differences. First, this unit reports itself as "UPS2000G" via the protocol instead, which is a new value not found in the datasheet or previously encountered by us. Previously, even a "UPS2000G" model reports itself as the "UPS2000A". Another difference is the use of WCH CH341 serial-to-USB converter chip instead of the MaxLinear/Exar RX21V1410 chip.

    Bus 002 Device 005: ID 1a86:5523 QinHeng Electronics CH341 in serial mode, usb to serial port converter

The production label on the UPS says 2021-12-04. So it's likely a newer hardware revision or variant. Was it redesigned to improve USB serial compatibility? Or my personal guess, due to chip shortage? Who knows.

Nevertheless, once the "UPS2000G" model string is added to the supported list, it was reported that the driver correctly functions without doing anything else. Thus, from a software perspective there's no difference.

The driver version has been bumped to v0.03. Its man page has also been updated with additional instructions for CH341 users.

Thanks GitHub user @Qinka for reporting it.

[1] https://github.com/networkupstools/nut/issues/1066#issuecomment-1158417669

### DIFF
--- a/docs/man/huawei-ups2000.txt
+++ b/docs/man/huawei-ups2000.txt
@@ -26,33 +26,42 @@ UPS with the following characteristics.
 
 1. Output power: 1 kVA to 3 kVA (higher power models are unsupported).
 
-2. Connection: USB or RS-232 (USB is only supported on Linux 5.12 and
-newer kernels, read the section *Cabling* carefully).
+2. Connection: USB or RS-232 (for most UPS models, USB is only supported
+on Linux 5.12 and newer kernels, but there are exceptions, read the section
+*Cabling* carefully).
 
 The UPS2000 series has two variants: UPS2000-A with a tower chassis,
 and UPS2000-G with a rack-mount chassis. Both should be equally supported,
 but more testers are needed.
-
 Currently, it has been tested on the following models.
 
-* UPS2000-A-1KTTS (firmware: V2R1C1SPC40, P1.0-D1.0)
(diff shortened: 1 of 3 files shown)


biergaizi:
USB information confirmed, ready to merge. Please take a look.

jimklimov:
Looks great, thanks to both. Will let CI chew this a bit, but don't expect fatal surprises.

biergaizi:
CI has failed but in a non-obvious way. I believe it's not the code that failed, instead, the build VM somehow crashed? Please take a look, and restart the CI if necessary.

jimklimov:
Yes, seems like intermittent CI issues (network flicker).

biergaizi:
Hi. Is it okay to merge this commit to the upstream? I don't think the CI/CD is ever gonna work, and there is literally only one-line of actual code change...

biergaizi:
CI/CD passed just after I said it doesn't have a chance. Magic!

jimklimov:
Sorry, was mostly-offline for a month.

PR #1216, opened 2021-12-06
Fightwarn: huawei-ups2000

jimklimov:
CC @biergaizi (so split from #1215 and earlier PRs with other similar fixes): Fixing CI warnings for the new huawei-ups2000 driver from PR #1198 and #954 for issue #1017 

````
$ (cd drivers/ && make -k huawei-ups2000.o)
depbase=`echo huawei-ups2000.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
/usr/lib/ccache/clang-9 -DHAVE_CONFIG_H -I. -I../include   -I/export/home/jim/nut-DMF/tmp/include -I../include  -I/usr/include/neon   -I/usr/include/modbus -I/export/home/jim/nut-DMF/tmp/include -std=gnu17 -m64 -Wno-unknown-warning-option -D_REENTRANT -ferror-limit=0 -Wall -Wextra -Weverything -Wno-unused-macros -Wno-reserved-id-macro -Wno-padded -Wno-documentation -Wno-cast-qual -pedantic -Wno-float-conversion -Wno-double-promotion -Wno-implicit-float-conversion -Werror -MT huawei-ups2000.o -MD -MP -MF $depbase.Tpo -c -o huawei-ups2000.o huawei-ups2000.c &&\
mv -f $depbase.Tpo $depbase.Po
huawei-ups2000.c:128:6: error: no previous extern declaration for non-static variable 'ups2000_desc' [-Werror,-Wmissing-variable-declarations]
char ups2000_desc[UPS2000_DESC_MAX_FIELDS][UPS2000_DESC_MAX_LEN] = { { 0 } };
     ^
huawei-ups2000.c:128:1: note: declare 'static' if the variable is not intended to be used outside of this translation unit
char ups2000_desc[UPS2000_DESC_MAX_FIELDS][UPS2000_DESC_MAX_LEN] = { { 0 } };
^
huawei-ups2000.c:131:11: error: no previous extern declaration for non-static variable 'modbus_ctx' [-Werror,-Wmissing-variable-declarations]
modbus_t *modbus_ctx = NULL;
          ^
huawei-ups2000.c:131:1: note: declare 'static' if the variable is not intended to be used outside of this translation unit
modbus_t *modbus_ctx = NULL;
^
huawei-ups2000.c:329:24: error: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
[...]

PR #1465, opened 2022-07-11
huawei-ups2000:  fightwarn hypercorrectionism cleanup

biergaizi:
In Fightwarm fix #1216, several bugs have been fixed, but it also introduced some changes that I perceive as unnecessary hypercorrectionism. For example, it used three type casts when it's only necessary to use just one type cast (as far as I can see, it's the case, but the CI/CD should let us know if the compiler disagrees). This is a minor cleanup Pull Request to remove these changes.

### DIFF
--- a/drivers/huawei-ups2000.c
+++ b/drivers/huawei-ups2000.c
@@ -2106,5 +2106,5 @@ static uint16_t crc16(uint8_t * buffer, uint16_t buffer_length)
 		crc_lo = table_crc_lo[i];
 	}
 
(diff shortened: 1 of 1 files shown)


biergaizi:
CI/CD has failed, but it's actually caused by an unrelated spellcheck regression in the current branch, and it has nothing to do with the actually code. Please merge the fix in #1466 then we can restart the CI/CD again.

jimklimov:
CI faults in last iteration were due to CI build farm slowness it seems (some daemons did not start/connect within timeout); a previous iteration (merge from master, not rebased as done later) succeeded fully. Merging, thanks :)

Issue #1455, opened 2022-06-25
upscode2 fails with Compaq T2400h, works at high debug level


marekm72:
The driver needs a minimum of three -D options to recognize responses from the UPS correctly.
At lower debug levels, just says "Missing UPCL after UPCL" - actually receives "UPCL\n" which fails to compare with "UPCL".
UPS responses on the serial port are the same, it's just that debug option changes something in the way it is received (with newline as part of the command).  After each "Supports command: UPxx" debug message there is an empty line.

nut-2.8.0-r0 package from Alpine Linux
Network UPS Tools - UPScode II UPS driver 0.90 (20220328-3481-gccbf3eb6c6)
Warning: This is an experimental driver.
Some features may not function correctly.

marekm72:
Looking at the code, I see the bug in upscode2.c line 1053:

		upsdebugx(3, "upscrecv: %zd bytes:\t'%s'", res-1, str_rtrim(buf, ENDCHAR));

which should be replaced with:

		str_rtrim(buf, ENDCHAR);
		upsdebugx(3, "upscrecv: %zd bytes:\t'%s'", res-1, buf);		

upsdebugx is a macro, its other args are only evaluated if debug level >= first arg.
str_rtrim has side effects (modifies the string in place and returns pointer to it).
The trailing newline should be removed regardless of debug level, of course.
It may be good to review the code (other drivers too) for other similar bugs.

jimklimov:
Nice catch! FWIW, it was changed from function to macro for a bit of efficiency in the default case (no debug - no overhead to call methods).

jimklimov:
Thanks, preparing a PR. FWIW, I gave up a bit on using a source formatter to simplify the searches, and went for this one-liner to trawl the code with my bare eyes (and some `less` search):

````
git grep -A10 -E 'upsdebug(x|_hex|_ascii|_with_errno)' | (H=false; while IFS= read LINE ; do case "$LINE" in *upsdebug*) H=true ;; esac; $H && echo "$LINE" ; case "$LINE" in *");") H=false; echo "---" ;; esac ; done )| uniq | less
````

Just in case similar needs to be done again...

Issue #182, opened 2015-02-03
Add support for external Data Mapping File


aquette:
NUT drivers (SNMP, XML/PDC, USB/HID, ...) hold hard-coded textual data, that maps NUT variables to the protocols-native ones. For example, each interesting SNMP OID (data) as an entry (line) in the snmp-ups driver that tells it how to retrieve and process this data:
https://github.com/networkupstools/nut/blob/master/drivers/eaton-mib.c#L253
https://github.com/networkupstools/nut/blob/master/drivers/apc-mib.c#L153
https://github.com/networkupstools/nut/blob/master/drivers/mge-xml.c#L903

These data should be moved out the drivers source code, into some text files, so that these data can be created and edited without the need to recompile NUT.

A specification is under writing to describe the Data Mapping File (DMF / .dmf)

aquette:
@clepple , @balooloo , @zykh : specs pushed to the wiki:
https://github.com/networkupstools/nut/wiki/Data-Mapping-File
Calling for review and approval

jimklimov:
Arno asked me to look for a small JSON parser library, and "jsmn" seems like a good fit for this use-case. It is small, so might be embedded into the NUT codebase rather than expected as an OS package (or as many projects do, pick the specific or newest variant between OS package and private copy during the "./configure" phase), and it is MIT-licensed. And extremely fast for both small and large documents, compared to others.

To start with, I was suggested to look at the detailed Russian-language blog from Lev Walkin (aka lionet) at http://lionet.livejournal.com/118853.html on C JSON parsers. Note that it is from September 2012, so the projects reviewed probably have moved on and improved since then.

In the first pass the author goes over a dozen parsers and projects suggested on http://json.org website. Apparently he looked carefully into the code and related files of each project, since he makes an opinion about the coding quality and test coverage (basic with valid data, invalid data, fuzz – none of these) and picks a few for the second pass – the practical comparison.

Parsers that remained in queue for his testing are: yajl, json-c, json-parser, jsonsl, jsmn, jansson, cson. All or most of these had some critical comments and drawbacks, with some the blog author hesitated whether to test at all. Those that got into finals are MIT, BSD, ISC or public-domain licensed (none of the (L)GPL ones made it through pass 1).

Ultimately, his test was a benchmark posted at http://github.com/vlm/vlm-json-test which:
1.	Parses an 8MB file 100 times, to calculate MB/s performance for large documents (bigger = better)
[...]

PR #1495, opened 2022-07-14
Avoid meaningful work in upsdebug*() macro call [#1455]

jimklimov:
Thanks to @marekm72 for noticing in #1455 that `str_rtrim()` modifies the original string, but macroization of `upsdebugx()` et al in #685 caused this method to not get called unless debug level is high - so less-than-expected data goes down the datapath.

Not sure if it practically impacted `upsd`, but for the `upscode2` driver this did cause a regression in NUT v2.8.0 release compared to earlier releases.

I went over the codebase to look at other arguments to existing debug calls, with the following grep, and did not notice other similar "offenses":

````
:; git grep -A10 -E 'upsdebug(x|_hex|_ascii|_with_errno)' | (\
    H=false; while IFS= read LINE ; do
        case "$LINE" in *upsdebug*) H=true ;; esac;
        $H && echo "$LINE" ;
        case "$LINE" in *");") H=false; echo "---" ;; esac ;
    done )| uniq | less
````

It is not a 100% perfect selection, catches a bit more than expected, but is good enough for the code trawling.

Possibly, similar inspection makes sense for code added/modified with the Windows #5, DMF #182/#469, and other long-standing feature branches.

### DIFF
--- a/drivers/upscode2.c
+++ b/drivers/upscode2.c
@@ -1050,7 +1050,9 @@ static ssize_t upscrecv(char *buf)
 	} else if (res == 0) {
 		upsdebugx(3, "upscrecv: Timeout");
 	} else {
-		upsdebugx(3, "upscrecv: %zd bytes:\t'%s'", res-1, str_rtrim(buf, ENDCHAR));
+		/* Note: s should end up same as buf */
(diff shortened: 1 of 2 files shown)


clepple (commented):
Makes sense to me.

PR #1502, opened 2022-07-17
News update for NUT v2.8.1

jimklimov:
Initially the committed state of the `NEWS` file contained a lot of planned work for 2.8.1 as catching up with ideas not finished (not urgent) before cutting the NUT v2.8.0 release a few months ago.

As we accumulated (and fixed in master branch) some reports of regressions in that release, it becomes more urgent to cut a bug-fix release soon so the better-functioning code can get into distributions etc. Many of those are still catching up with updating the recipes to provide something after 2.7.4 anyway.

Milestone definitions in Github were updated, although work not completed in practice by whenever we decide that 2.8.1 is ready to roll would move in planning to 2.8.2 just at that point in time (moving of lines in `NEWS` in this PR now may be not final).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,7 +5,7 @@ ChangeLog file (generated for release archives), or to the Git version
 control history for "live" codebase.
 
 ---------------------------------------------------------------------------
-PLANNED: Release notes for NUT 2.8.2 - what's new since 2.8.1:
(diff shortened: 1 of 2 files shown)


jimklimov:
Review requests here are more for FYI overview of recent changes and known fixed breakage.

Issue #1467, opened 2022-07-11
Man page generation fails when building only partial list of drivers


Kaapeli64:
I was writing an ebuild for Gentoo and ran into an issue whenever only some of the drivers are built, eg with:

`./configure --with-drivers=usbhid-ups`

The build will always fail:
```
asciidoc: ERROR: upsd.txt: line 185: [blockdef-filter] missing closing delimiter
a2x: ERROR: "xmllint" --nonet --noout --valid "/tmp/nut.bad/docs/man/tmp/man8.nutupsdrv.8.1443001/nutupsdrv.xml" returned non-zero exit status 4

asciidoc: ERROR: upsd.txt: line 185: [blockdef-filter] missing closing delimiter
mv: cannot stat './tmp/man8.nutupsdrv.8.1443001/nutupsdrv.8': No such file or directory
make[2]: *** [Makefile:1646: nutupsdrv.8] Error 1
make[2]: *** Waiting for unfinished jobs....
mv: cannot stat './tmp/man8.upsd.8.1443040/upsd.8': No such file or directory
make[2]: *** [Makefile:1646: upsd.8] Error 1
make[2]: *** Waiting for unfinished jobs....
mv: cannot stat './tmp/man8.upsd.8.1443085/upsd.8': No such file or directory
make[2]: *** [Makefile:1646: upsd.8] Error 1
a2x: ERROR: "xmllint" --nonet --noout --valid "/tmp/nut.bad/docs/man/tmp/man8.nutupsdrv.8.1443021/nutupsdrv.xml" returned non-zero exit status 4

mv: cannot stat './tmp/man8.nutupsdrv.8.1443021/nutupsdrv.8': No such file or directory
```
The reason is that linkman-driver-names.txt is empty.

I made a fix patch into Gentoo bug tracker https://836776.bugs.gentoo.org/attachment.cgi?id=791087 which looks sane to me. This is the first time I'm playing with autotools, so I'm not quite sure if I understood what I was doing. The results with this fix look right, the man pages contain only those drivers that were built and no more.

Regardless of the fix, the issue is real. I could try make a pull request too if you think this is done right.

Kaapeli64:
[...]

PR #1505, opened 2022-07-17
Fix man-page building when just a few drivers are built by name

jimklimov:
Regression for custom-builders in NUT 2.8.0 release

Closes: #1467

Thanks: @Kaapeli64 for the finding and suggesting a mostly-correct fix


jimklimov:
CI hiccups are addressed elsewhere

Issue #1405, opened 2022-04-27
Revise NIT to reliably grab a NUT_PORT for the test run and make it part of CI routine


jimklimov:
#1380 and #1383 introduced NIT (NUT Integration Testing) suite for running some dummy drivers, data server, and a few clients to test them, and the bindings from supported language.

When it was enabled by default in NUT CI farm runs, which involved running some builds on slower non-native agents and agents with several executors, some tests failed with reports of inability to grab the NUT_PORT. While nothing in logs seemed to point to a conflict (maybe it just was too slow for embedded timeouts), flakiness of the test precluded making it a default activity in CI runs (that, and also the noisiness addressed with another issue #1404).

This needs more investigation and adjustment, in order to enable NIT for at least some CI scenarios to make sure our API and protocols and bindings do not regress as we iterate.

jimklimov:
PR #1534 which added ways to check if the port is busy (at least for CI farm platforms) seems to alleviate the practical problem of false-positive failed CI tests. Such support may need to be addressed in the future if more platforms are added and their probing ways differ.

PR #1508, opened 2022-07-18
Fix/troubleshoot NIT on slow systems

jimklimov:
This PR addresses part of issue #1405 to help troubleshoot (and in one situation try to fix) flakiness of the new NUT integration tests.

Empirically, when the NUT CI farm has build storms (e.g. a PR or few have landed, and some got merged, etc.) or the VMs involved are otherwise slow (Fosshost hypervisors overloaded by other projects?), the `make check-NIT` part of the builds tends to randomly fail one or a few times across the build matrix, requiring log investigation by NUT team to make sure the problem was not with contributed codebase.

This is something hard to reproduce otherwise, so the assumption is that the root cause gotta be about timeouts, with ordinarily sub-second operations such as driver start-up, connection to upsd, and responding to queries suddenly taking order(s) of magnitude more wall-clock time.

### DIFF
--- a/tests/NIT/nit.sh
+++ b/tests/NIT/nit.sh
@@ -345,6 +345,7 @@ isPidAlive() {
 }
 
 FAILED=0
+FAILED_FUNCS=""
 PASSED=0
 
 testcase_upsd_no_configs_at_all() {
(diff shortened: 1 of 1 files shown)


jimklimov:
The fix is not fully complete: several cases were logged where `testcase_upsd_allow_no_device()` tried to "Test that UPSD responds to UPSC" (listing devices when there were none) and `upsc` failed with `Error: Server disconnected`.

While the test script might retry this and similar requests to rule out lagginess (or in real-life situation, network hiccups etc.), it should not be indefinitely relaxed to the situation as the problem generally may be a NUT regression too.

PR #1515, opened 2022-07-23
huawei-ups2000.c: rename NULL macros, close #1474.

### LINKED ISSUE #1474: huawei-ups2000.c:1434: warning: "REG_NONE" redefined (WIN32 builds)
With native MSYS2 Windows builds there is a warning:

````
huawei-ups2000.c:1434: warning: "REG_NONE" redefined
 1434 | #define REG_NONE -1, -1
      |
In file included from C:/msys64/mingw64/include/minwindef.h:163,
                 from C:/msys64/mingw64/include/windef.h:9,
                 from C:/msys64/mingw64/include/windows.h:69,
                 from C:/msys64/mingw64/include/winsock2.h:23,
                 from C:/msys64/mingw64/include/modbus/modbus-tcp.h:16,
                 from C:/msys64/mingw64/include/modbus/modbus.h:292,
                 from huawei-ups2000.c:47:
C:/msys64/mingw64/include/winnt.h:8926: note: this is the location of the previous definition
 8926 | #define REG_NONE (0)
      |
````

that should be fixable by a rename in NUT sources.

CC @biergaizi

biergaizi:
Originally, in the struct array `ups2000_cmd[]`, a `REG_NONE` macro was used as a short-hand for a nonexistent register value. Unfortunately it was in conflict with a Windows NT macro of the same name (previously not found because it was only tested on Unix systems). Thus, this commit renames it to `REG_NULL`. To avoid visual confusions with the function pointer field (which also uses C's NULL for its nonexistence), this commit also replaces it with a new macro `FUNC_NULL`, which is a simple alias to C's `NULL`.

It's not exactly the best practice, a `UPS2000_` prefix should've been used in principle. However, using it in this case creates exceedingly long lines and makes the struct array definition more difficult to read, which defeats the very purpose of defining these macros to begin with. Considering that it's only an one-off use, it's unlikely to cause further problems.

This PR closes #1474.

### DIFF
--- a/drivers/huawei-ups2000.c
+++ b/drivers/huawei-ups2000.c
@@ -1422,7 +1422,8 @@ static int ups2000_delay_set(const char *var, const char *string)
  * used to handle commands that needs additional processing.
  * If "reg1" is not necessary or unsuitable, "-1" is used.
  */
-#define REG_NONE -1, -1
+#define REG_NULL  -1, -1
+#define FUNC_NULL NULL
 
 static struct ups2000_cmd_t {
 	const char *cmd;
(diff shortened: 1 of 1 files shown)


jimklimov:
Thanks for taking care of this on short notice :)

PR #1524, opened 2022-07-28
Master branch update to minimize differences vs. Windows branch (and apply some fixes from it)

jimklimov:
This PR updates the `master` branch with fixes (some to code, some cosmetic, mostly "non-functional" either way) applied on the NUT for Windows branch, so that the remaining differences of the two branches are (with a few exceptions) only about code compiled `#ifdef WIN32`.

This aims to help with a concern raised in discussion of issue #5 - to get a grip on what the code added for Windows might break in currently supported multitude of platforms, e.g. what it changes (or not) in POSIX-conformant codebase.

This effort also allowed to cursorily review those changes and update the issues listed in https://github.com/orgs/networkupstools/projects/2/views/1 project, so someone sometime would be able to focus on tidying up this platform support starting from the current Windows branch state.

Barring typos, it is expected that CI would confirm lack of breaking changes from this PR. But human review would be appreciated too.


jimklimov:
Offloaded changesets worthy of separate mention and attention into PRs of their own.

PR #1530, opened 2022-07-29
Modbus drivers: fix to use `uint16_t` and `unsigned int`

jimklimov:
Fix code using less-than-standard type names.

CC @dtsecon - FYI


jimklimov:
Originally published in PR #1524

PR #1536, opened 2022-08-01
Fix case for EATON manufacturer

aquette:
This must be upper case to avoid duplicate discoveries

Closes: IPMPROG-4489

(cherry picked from commit a10f5967154e6c6d2dfa53975de6f874b9966804)


aquette:
It's a pure static mapping for older NMC. The other side (XML/PDC) reports with upper case, like others

PR #1541, opened 2022-08-07
fix link when viewed on Github

akwan:
This fixes a trivial Github link rendering issue on the README from http://www.networkupstools.org/stable-hcl.html[online] to http://www.networkupstools.org/stable-hcl.html

### DIFF
--- a/README
+++ b/README
@@ -171,7 +171,7 @@ For example, it is available on Debian systems as:
 
 	/usr/share/nut/driver.list
 
-This table is also available link:http://www.networkupstools.org/stable-hcl.html[online].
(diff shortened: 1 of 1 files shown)


jimklimov:
Thanks for the effort, but that's actually part of `asciidoc` markup for links, so there would be an HTML a-href tag saying "online" in visible text and leading to the specified URL.

Probably GitHub renders the extension-less `README` as `markdown` by default; and this reminds me that there was an abandoned effort to try and fix it (perhaps by renaming files to specific extensions and making sure `make` recipes are all updated). => #669 #226

PR #1582, opened 2022-08-18
snmp-ups: fix alarm processing

aquette:
only walk the number of present alarms, instead of INT_MAX,
to avoid potential blocking

### DIFF
--- a/drivers/snmp-ups.c
+++ b/drivers/snmp-ups.c
@@ -4,7 +4,7 @@
  *
  *  Copyright (C)
  *	2002 - 2014	Arnaud Quette <arnaud.quette@free.fr>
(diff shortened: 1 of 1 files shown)


eric3873 (approved):

jimklimov:
Hi, nice catch. Just to make sure I understand it right: previously the driver would walk alarms subtree until it does INT_MAX requests or SNMP server says it has no more OIDs there - "naively" hoping for the latter to work (this did not block every time in the previous 7 years I suppose)? And now it would do at most `value` requests in any case?

aquette:
exactly @jimklimov

jimklimov:
A bit of pedantic bug here: `nut_snmp_walk()` takes an `int` argument, and `value` is `long` so it can overflow. Should blurp an error in the unlikely case `if (value > INT_MAX)`

PR #1604, opened 2022-08-21
upslog: Add support for multiple UPSs

cschuber:
upslog is a utility that logs UPS status at regular intervals, specified
by the -i option. Unfortunately upslog supports only on UPS. For sites
that need to monitor multiple UPSs the options are to cobble an rc script
for each or doctor up the nut_upslog.in script to support cloning of the
script. Unfortunately an rc script capable of being cloned would become
the source of more PRs and would require significanly more tehcnical
documentation that by itself might become confusing for the average
system administrator.

Therefore a new -m option is added to support multiple UPSs using the
same invocation of upslog. The patch parses a -m option and forks
almost immediately following the getopt(3) invokation to monitor each
individual UPS using a separate upslog process. This is not ideal but
better than scripting. Lightweight threads might be a better solution.

## General points

  comments or text documents.

  case.

  or its equivalent.

  concept.

  the changed codebase.

### DIFF
--- a/clients/upslog.c
+++ b/clients/upslog.c
@@ -32,6 +32,10 @@
  */
 
 #include "common.h"
+#include <signal.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
 #include "nut_platform.h"
 #include "upsclient.h"
 
@@ -41,16 +45,31 @@
 #include "upslog.h"
 
 	static	int	reopen_flag = 0, exit_flag = 0;
+	static	int	wait_status;
 	static	uint16_t	port;
 	static	char	*upsname, *hostname;
 	static	UPSCONN_t	ups;
 
 	static	FILE	*logfile;
-	static	const	char *logfn, *monhost;
+	static	char *logfn, *monhost;
 	static	sigset_t	nut_upslog_sigmask;
 	static	char	logbuffer[LARGEBUF], *logformat;
 
 	static	flist_t	*fhead = NULL;
+	struct 	monhost_child {
+		char	*monhost;
+		char	*logfn;
+		char	*pidfilebase;
+		pid_t	pid;
+		struct	monhost_child	*next;
+	};
+	static	struct	monhost_child *monhost_child_anchor = NULL;
+	static	struct	monhost_child *monhost_child_current = NULL;
+	static	struct	monhost_child *monhost_child_prev = NULL;
+	static	struct	sigaction upslog_sigaction;
+	static	int	trapped_signals[] = { SIGHUP, SIGINT, SIGTERM, SIGCHLD };
+	static	pid_t	daemon_pid;
+
 
 #define DEFAULT_LOGFORMAT "%TIME @Y@m@d @H@M@S% %VAR battery.charge% " \
 		"%VAR input.voltage% %VAR ups.load% [%VAR ups.status%] " \
@@ -131,6 +150,8 @@ static void help(const char *prog)
 	printf("  -p <pidbase>  - Base name for PID file (defaults to \"%s\")\n", prog);
 	printf("  -s <ups>	- Monitor UPS <ups> - <upsname>@<host>[:<port>]\n");
 	printf("        	- Example: -s myups@server\n");
+	printf("  -m <tuple>	- Monitor UPS <ups,logfile,pidfile>,\n");
+	printf("		- Example: -m myups@server,/var/log/myups.log,/var/run/myups.pid\n");
 	printf("  -u <user>	- Switch to <user> if started as root\n");
 
 	printf("\n");
@@ -393,9 +414,25 @@ static void run_flist(void)
 	 * -u <username>
 	 */
 
+static void term_handler(int signo)
+{
+	if (signo != SIGCHLD && monhost_child_anchor != NULL) {
+		for (monhost_child_current = monhost_child_anchor;
+		     monhost_child_current != NULL;
+		     monhost_child_current = monhost_child_current->next
+		) {
+			kill(monhost_child_current->pid, signo);
+		}
+
+		if (signo != SIGHUP)
+			fatalx(EXIT_FAILURE, "Killed by user");
+	}
+}
+
 int main(int argc, char **argv)
 {
 	int	interval = 30, i, foreground = -1;
+	size_t	monhost_len = 0, mh;
 	const char	*prog = xbasename(argv[0]);
 	time_t	now, nextpoll = 0;
 	const char	*user = NULL;
@@ -407,7 +444,7 @@ int main(int argc, char **argv)
 
 	printf("Network UPS Tools %s %s\n", prog, UPS_VERSION);
 
-	while ((i = getopt(argc, argv, "+hs:l:i:f:u:Vp:FB")) != -1) {
+	while ((i = getopt(argc, argv, "+hs:l:i:f:u:Vp:FBm:")) != -1) {
 		switch(i) {
 			case 'h':
 				help(prog);
@@ -415,6 +452,33 @@ int main(int argc, char **argv)
 				break;
 #endif
 
+			case 'm': { /* var scope */
+				char *m_arg, *s;
+
+				monhost_child_prev = monhost_child_current;
+				monhost_child_current = xmalloc(sizeof(struct monhost_child));
+				if (monhost_child_anchor == NULL)
+					monhost_child_anchor = monhost_child_current;
+				else
+					monhost_child_prev->next = monhost_child_current;
+				monhost_child_current->next = NULL;
+				monhost_len++;
+
+				/* Be sure to not mangle original optarg, nor rely on its longevity */
+				s = xstrdup(optarg);
+				m_arg = s;
+				monhost_child_current->monhost = xstrdup(strsep(&m_arg, ","));
+				if (!m_arg)
+					fatalx(EXIT_FAILURE, "Argument '-m upsspec,logfile,pidfile' requires exactly 3 components in the tuple");
+				monhost_child_current->logfn = xstrdup(strsep(&m_arg, ","));
+				if (!m_arg)
+					fatalx(EXIT_FAILURE, "Argument '-m upsspec,logfile,pidfile' requires exactly 3 components in the tuple");
+				monhost_child_current->pidfilebase = xstrdup(strsep(&m_arg, ","));
+				if (m_arg) /* Had a third comma - also unexpected! */
+					fatalx(EXIT_FAILURE, "Argument '-m upsspec,logfile,pidfile' requires exactly 3 components in the tuple");
+				free(s);
+				} /* var scope */
+				break;
 			case 's':
 				monhost = optarg;
 				break;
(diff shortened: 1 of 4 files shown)


cschuber:
Original idea came from FreeBSD PR/265963 (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=265963). Certainly more elegant than what the user suggested.

jimklimov (changes requested):
Change looks good. Probably will need adaptation to Windows branch (no signals and possibly forking, etc), but that's a separate story not just yet merged.

Fix the help text please - not `-s ups` but `-m 'ups,logfile,port'`? And `make spellcheck-interactive` (`fomrat...`) to fix wording or update `nut.duct` as applicable.

Commuting now, not sure if you updated in-program usage (`--help`) text?

jimklimov:
Per your comment, threads might indeed be better (also easier to track for the likes of systemd that struggles with dual-process upsmon). Also systems-dependent though. Or a select loop to read FDs and dispatch writes...

For instance, nut-scanner can serve as in-codebase example of threading vs looping depending on build env.

cschuber:
> Per your comment, threads might indeed be better (also easier to track for the likes of systemd that struggles with dual-process upsmon). Also systems-dependent though. Or a select loop to read FDs and dispatch writes...
> 
> For instance, nut-scanner can serve as in-codebase example of threading vs looping depending on build env.

A select() loop would be better an no threads would be required then.

Unfortunately, or fortunately ;), I'm not a Windows programmer.

ATM I don't have the time to rework the patch to a select() loop though that could be done at a later date as I already have other more pressing issues (wifi driver and Heimdal) which should really take precedence. This approach was a quick win to get that PR off my plate.

I won't mind if this isn't accepted. I plan on putting the FreeBSD port patches on the back burner for now anyway.

jimklimov:
No problem, as soon as CI has no complaints about this PR it can become acceptable (I'll add a comment for refactoring and an issue to track yet another missing feature in Windows port for now).

clients/upslog.c
```diff
@@ -393,28 +414,53 @@ static void run_flist(void)
...
+	while ((i = getopt(argc, argv, "+hs:l:i:f:u:Vp:FBm:")) != -1) {
 		switch(i) {
 			case 'h':
 				help(prog);
 #ifndef HAVE___ATTRIBUTE__NORETURN
 				break;
 #endif
 
+			case 'm':
+				monhost_child_prev = monhost_child_current;
```
jimklimov, line 452: Starting value of "current" as declared on top is not "NULL". Care was taken to predefine others :)

clients/upslog.c
```diff
@@ -393,28 +414,53 @@ static void run_flist(void)
...
+				monhost_child_current = malloc(sizeof(struct monhost_child));
```
jimklimov, line 453: Perhaps NUT `xmalloc` to consistently catch errors

clients/upslog.c
```diff
@@ -479,6 +525,29 @@ int main(int argc, char **argv)
...
+		     monhost_child_current = monhost_child_current->next) {
+			if ((monhost_child_current->pid = fork()) == 0) {
+				monhost = monhost_child_current->monhost;
+				logfn = monhost_child_current->logfn;
+				pidfilebase = monhost_child_current->pidfilebase;
+				break;
+			}
+		}
+		if (monhost_child_anchor->pid) {
+			for (i = 0; i++; i < sizeof(trapped_signals)/sizeof(trapped_signals[0])) {
```
jimklimov, line 540: `for` clause (`i<...`) and op (`i++`) misplaced

clients/upslog.c
```diff
@@ -479,6 +525,29 @@ int main(int argc, char **argv)
...
+				upslog_sigaction.sa_handler = &term_handler;
+				sigfillset(&upslog_sigaction.sa_mask);
+				upslog_sigaction.sa_flags = SA_NOCLDSTOP | SA_NOCLDWAIT;
+				sigaction(trapped_signals[i], &upslog_sigaction, NULL);			}
```
jimklimov, line 544: `}` to next line please :)

docs/man/upslog.txt
```diff
@@ -78,6 +78,11 @@ upslog will run in the background, regardless of logging target.
 Monitor this UPS.  The format for this option is
 +upsname[@hostname[:port]]+.  The default hostname is "localhost".
 
+*-m* 'ups'::
```
jimklimov, line 81: I guess `-m 'tuple'` sits best here, to keep it short and in sync with updated `usage()`

jimklimov (changes requested):

jimklimov:
Updated the PR with fixes to complaints I had so far, waiting for CI's next round of opinions :)

jimklimov:
Needs even more love...

````
[01m[Kupslog.c:[m[K In function '[01m[Kmain[m[K':
[01m[Kupslog.c:439:22:[m[K [01;31m[Kerror: [m[Kinitialization discards '[01m[Kconst[m[K' qualifier from pointer target type [[01;31m[K-Werror=discarded-qualifiers[m[K]
  439 |  char *pidfilebase = [01;31m[Kprog[m[K;
      |                      [01;31m[K^~~~[m[K
[01m[Kupslog.c:553:15:[m[K [01;31m[Kerror: [m[Kpassing argument 1 of '[01m[Kwait[m[K' makes pointer from integer without a cast [[01;31m[K-Werror=int-conversion[m[K]
  553 |    while(wait([01;31m[Kwait_status[m[K));
      |               [01;31m[K^~~~~~~~~~~[m[K
      |               [01;31m[K|[m[K
      |               [01;31m[Kint[m[K
In file included from [01m[Kupslog.c:38[m[K:
[01m[K/usr/include/x86_64-linux-gnu/sys/wait.h:77:27:[m[K [01;36m[Knote: [m[Kexpected '[01m[Kint *[m[K' but argument is of type '[01m[Kint[m[K'
   77 | extern __pid_t wait ([01;36m[Kint *__stat_loc[m[K);
      |                      [01;36m[K~~~~~^~~~~~~~~~[m[K
At top level:
[01m[Kupslog.c:48:27:[m[K [01;31m[Kerror: [m[K'[01m[Kchild_pid[m[K' defined but not used [[01;31m[K-Werror=unused-variable[m[K]
   48 |  static  int wait_status, [01;31m[Kchild_pid[m[K;
      |                           [01;31m[K^~~~~~~~~[m[K
[01m[Kcc1:[m[K [01;36m[Knote: [m[Kunrecognized command-line option '[01m[K-Wno-unknown-warning-option[m[K' may have been intended to silence earlier diagnostics
[01m[Kcc1:[m[K [01;36m[Knote: [m[Kunrecognized command-line option '[01m[K-Wno-reserved-identifier[m[K' may have been intended to silence earlier diagnostics
````

**Comment by jimklimov** [comment]
@cschuber : I've hopefully untangled the build issues of this codebase so it might actually compile.

Have you tested it before PRing? Can you please test it in practice now, so we would know it does what you originally wanted? :)

File: clients/upslog.c [hunk 6]
```diff
@@ -407,14 +443,28 @@ int main(int argc, char **argv)
...
+				monhost_child_current = xmalloc(sizeof(struct monhost_child));
+				if (monhost_child_anchor == NULL)
+					monhost_child_anchor = monhost_child_current;
+				else
+					monhost_child_prev->next = monhost_child_current;
+				monhost_child_current->next = NULL;
+				m_arg = optarg;
+				monhost_child_current->monhost = strsep(&m_arg, ",");
+				monhost_child_current->logfn = strsep(&m_arg, ",");
+				monhost_child_current->pidfilebase = strsep(&m_arg, ",");
```
jimklimov, line 465: @cschuber : I think sanity-check is missing here, that these 3 strings are non-trivial.

Also, do I remember right that `strsep` would change the original string (`optarg`) and return pointers to its pieces? 
* Do we risk that original `optarg` would be released from RAM (e.g. is a temporary copy of `argv[X]`) on some systems?
* If it is just a pointer to `argv[X]`, are we okay to mangle original arguments array?

I think it would be more solid on all these fronts to make a copy of optarg, mangle it, `xstrdup` the pieces, free the copy.

jimklimov: Note: addressed this idea in a later PR.

cschuber: I don't think it should matter but for the sake of correctness marg = optarg could be replaced by xstrdup(). It's the correct thing to do.

BTW, argv is only freed on execv() but using xstrdup is the better approach.

cschuber:
First, my apologies for the poor quality pull request. I should have spent more time on it (though this isn't and shouldn't be my top priority ATM).

Secondly, I've extracted the patches from this branch to add them to my upslog branch of FreeBSD ports. I'll test them locally before requesting the person who submitted the FreeBSD PR (Problem Report) to test them.

jimklimov:
Thanks, and good luck on the priority job! Gotta get back to mine too ;)

cschuber:
Not quite there yet. The process attached to the terminal must itself fork() before forking children to disconnect from the terminal (or console). I'll test here before submitting an update. Something I had overlooked.

jimklimov:
Did it previously get away with this? `nohup upslog &` and init-scripts without terminals?

jimklimov:
Otherwise probably same play as in upsmon and upsd with explicit foreground/background flags.

cschuber:
I've made seven additional commits during testing and have rolled the patch into a pending FreeBSD port commit to sysutils/nut (2.8.0) and sysutils/nut-devel (tracking your master branch). https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=265963 has been updated with the proposed FreeBSD patch.

cschuber:
I'm not sure how to resolve doc/nut.dict.

jimklimov:
Do I read it right that in `foreground>0` mode you would not sub-process for the tuple-driven configurations at all? Would it log the multiple UPSes then?

I thought forking here served two purposes at two places:
* detach from controlling terminal or other parent program, or not (fore-/back-grounding)
* spawn workers that write one log each and be able to terminate them later when shutting down the parent process

Security-wise, probably in both single-log and multi-log modes it should drop privileges to `user` (if it is `root` at start; pointless otherwise).

cschuber:
In foreground mode the parent (main process) simply forks for each logger child and waits for their completion. When the parent receives a SIGINT or SIGTERM the parent terminates all children.

In background mode the parent forks first, then follows the process in the paragraph above.

The children setuid(uucp) as they did before. The parent setuid(uucp) after forking off the children -- the same as the children. Dropping privileges prior to or following the sigaction()s doesn't matter.

cschuber:
To explain fork()ing a little more,

If in background mode it forks first. The new parent will fork the children and wait for them. If in foreground mode the original parent will simply fork the children and wait.

Do you think a comment there could help?

jimklimov:
No, I guess no comment is needed - I just misunderstood the neat `if !daemon` trickery when reviewing from the phone.

jimklimov:
I use octodroid, fasthub, or mobile browsers much of the time, all of are
differently lame :) FOSS being not dayjob, it gets much time during commute.

jimklimov:
@cschuber : Do you think this PR is now feature-complete? Is it tested practically? :)

jimklimov:
Also, would you mind rebasing and amending the PR history in a way to reduce back-and-forth with small typo fixes, etc. and addition/removal of the editor swap file, so those are not in the ultimate merged history? "Meaningful" commits that increment or fix some logic, or even clarify some dead-end which we should not try again, are OK to stay :)

cschuber:
I can do a squahed commit, yes.

jimklimov:
Close enough :) And how did testing go?

cschuber:
> Close enough :) And how did testing go?

Went well. Created a FreeBSD ports patch for nut and nut-devel from this pull request. Pushed it early this afternoon.  It's in FreeBSD now. I'll yank it from the -devel port when the hash points to a nut commit post-merge. Thanks for all your help.

Works here!

cschuber:
End user refused to  test before it was committed but now reports there is a problem. I am not able to reproduce his problem here.

cschuber:
I've tested. My user has not. Let's either put this on hold or I will close this pull request and open a new one when he's finally ready to AND finally tested.

jimklimov:
With last force-push, lost back the spellcheck fix, again :\

As for "on hold", CI-wise it should suffice to mark the PR as "draft" so it would not keep rebuilding newer commits and ephemeral merges as the "master" branch would march on. When ready to continue, un-draft it and push a commit (to trigger a webhook), I think... or wait up to 12 hours for CI fallback polling :)

jimklimov:
As of currently last iteration https://ci.networkupstools.org/job/nut/job/nut/job/PR-1604/30/ CI likes this PR.

So feel free to un-draft it when end-user issues are understood and resolved.

cschuber:
My user has replied. Looks like what I'll need to do is have the children replace their argv with something that would identify them, differentiating them from one another, and from the parent, so that a ps listing will make more sense.

The other thing I notice is that upslog doesn't write to syslog. This needs to be fixed.

Thinking about this further, the individual process approach I took might be the wrong way to approach this problem. I felt that by simply forking we would have a quick win but it appears, according to my user, that some some child processes exit early without a reason why. It could be that they try to establish too many TCP sessions to upsd and the connection is refused. I'm thinking your suggestion at the start of this pull request to use select() makes a lot more sense now. Let's close this and I'll send a new pull request when I have something that my user has tested -- my testing has found no problem while he is unable to make it work.

I'll try to juggle this with the higher priority project of getting the rtl8192eu (USB WiFi dongle) to work. More users want that working than this. Let's put this on the back burner while I work through this user's problem. I can't reproduce it locally but seems the child processes are terminating prematurely (without any messages to syslog to explain why).

I'll get back  to you.

jimklimov:
Okay yjen. At least this solution attempt is bookmarked in github history now ;)

cschuber:
FYI, I have a new patch which does not use multiple processes. The FreeBSD PR (problem report) has been updated and I am waiting for the user to test. Once he's tested and it solves his problem a new PR (pull request) will be submitted. This new patch works here as well, using fewer resources.

Issue #1507, opened 2022-07-18
Revise libusb-1.0 usage in Windows branch


jimklimov:
The NUT for Windows effort up to 2.6.5-7 branch abandoned several years ago did not consider the not-yet-existing libusb-1.0 codebase, but included some changes for the libusb-0.1 codebase.

While preparing the Windows-2.8.0-1 branch, I tried to apply changes that happened in `libusb0.c` to `libusb1.c` as well, but there may be things I've missed or misunderstood (APIs do differ).

It would be beneficial if someone better versed in libusb APIs went over those changes (comparing Windows branch to master), preferably both files for good measure, so we are more confident that it should actually work.

(Per small-scale testing at the time of this codebase update, linking with libusb-compat did function enough to discover the device and get initial data from the driver, although device-closing crashed when stopping it and I am not sure that run-time measurement/status updates did happen)

UPDATE: This situation also compromises use of libmodbus with RTU USB support, as it was not coded to accept libusb-0.1 API. PR #2778 should allow to explicitly prefer libusb-1.0 by using `configure --with-modbus+usb` (but the result may be flaky or not, per this issue).

jimklimov:
Currently, both libusb-0.1-compat and libusb-1.0 builds misbehave in regard to data collection: effectively just the data points from device discovery (vendor, model...) are populated. Even status is not (remains "OB" by default, lacking other info).

jimklimov:
FWIW, data collected there also complains about resource descriptor length which is miniscule and may be technically behind seeing no data (why - is another matter):
````
   2.272043     [D2] Checking device (0463/FFFF) (001/009)
   2.557041     [D2] - VendorID: 0463
   2.558664     [D2] - ProductID: ffff
   2.559767     [D2] - Manufacturer: EATON
   2.560962     [D2] - Product: Ellipse ECO
   2.562178     [D2] - Serial Number: 000000000
   2.563512     [D2] - Bus: 001
   2.564361     [D2] - Device: 009
   2.565317     [D2] - Device release number: 0100
   2.567701     [D2] Trying to match device
   2.568923     [D2] match_function_subdriver (non-SHUT mode): matching a device...
   2.571537     [D3] match_function_regex: matching a device...
   2.573576     [D2] Device matches
   2.574590     [D3] nut_usb_set_altinterface: skipped usb_set_altinterface(udev, 0)
[...]

PR #1629, opened 2022-08-28
m4/nut_check_libusb.m4: when on Windows, prefer libusb-0.1 [#1507]

jimklimov:
Not really sure if this would help (checking) but the Windows branch was originally developed before libusb-1.0 effort started, so tweaks to code applied to what is now `libusb0.c` should be more successful.

In limited dev-testing so far, builds with plain libusb-1.0 and with it wrapped as libusb-0.1-compat (mis-)behave in a similar manner, claiming I/O errors with interrupt processing mode, and nearly no data collected with polling mode. Maybe snatching the device from "kernel" should be done differently (Windows does represent the "USB HID Battery" and probably holds the critical resources).

Have yet to try with original libusb-win32 project the branch was built against.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -321,11 +321,13 @@ AS_IF([test x"${ac_cv_func_strncasecmp}" = xyes],
     [AC_MSG_WARN([Required C library routine strncasecmp not found; try adding __EXTENSIONS__])]
     )
 
+dnl Note: this changes the original string in argument!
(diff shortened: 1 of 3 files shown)


jimklimov:
Regarding AppVeyor fail: Huh, this relies on #1628 to be there, for MSYS2 builds. Will retry after merging that one.

Issue #1624, opened 2022-08-27
usbhid-ups with pollonly never sets "Driver stale"


jimklimov:
While testing NUT for Windows builds (so might be a platform issue as well), I pulled the USB cable to disconnect the UPS completely while the driver was running (in `pollonly` mode, since interrupt mode failed with I/O errors early after stratup for this device/system combo).

It still reports regular "Quick updates" and less frequent "Full updates" even an hour since the cable yank, never noticing that the device is physically not there anymore.

PR #1630, opened 2022-08-28
drivers/usbhid-ups.c: try to detect "Driver stale" situations when in "pollonly" mode

jimklimov:
* drivers/usbhid-ups.c: hid_ups_walk(): for HU_WALKMODE_FULL_UPDATE with "pollonly" mode, re-read HU_FLAG_SEMI_STATIC and HU_FLAG_STATIC entries to detect "Driver stale" situations [#1624]

Currently the driver code does not seem to have any concept of "staleness" if it is not in interrupt mode (and failing those requests). Unplugging the USB cable has no effect on driver's ability and eagerness to report info about the device, even an hour later. I did not yet test if this is a platform issue (seen with Windows builds) or a generic one.

Not sure in practical terms however, if the trick in this PR helps - e.g. it might be prudent to actively re-request static fields known to have been served during initial startup, to make sure they are still served or the device is AWOL (assuming libusb actually requests that, and does not cache responses), but I am not quickly sure how to do that.

More expert eyes are welcome :)

### DIFF
--- a/drivers/usbhid-ups.c
+++ b/drivers/usbhid-ups.c
@@ -1422,11 +1422,23 @@ static bool_t hid_ups_walk(walkmode_t mode)
 
 		case HU_WALKMODE_FULL_UPDATE:
 			/* These don't need polling after initinfo() */
-			if (item->hidflags & (HU_FLAG_ABSENT | HU_TYPE_CMD | HU_FLAG_STATIC))
+			if (item->hidflags & (HU_FLAG_ABSENT | HU_TYPE_CMD))
 				continue;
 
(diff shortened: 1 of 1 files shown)


jimklimov:
UPDATE: PR #2010 introduces state entry timestamping, so if it gets merged - we have another approach to the staleness detection problem generally (can just test in every loop if non-(semi)static values were last written longer than SOME_TIMEOUT ago).

jimklimov:

jimklimov:
CI faults irrelevant (free space problems on an agent).

Issue #1627, opened 2022-08-27
usbhid-ups attempts to reconnect are a suicide mission (Windows builds at least)


jimklimov:
Recovery from I/O error goes wrong:

````
   1.653775     Using subdriver: MGE HID 1.46
...
   6.805866     [D1] upsdrv_updateinfo...
   6.807082     [D1] Got to reconnect!
   6.808156
Reconnecting. If you saw "nut_libusb_get_interrupt: Input/Output Error" or similar message in the log above, try setting "pollonly" flag in "ups.conf" options section for this driver!

   6.815235     [D4] Closing comm_driver previous handle
   6.817622     [D4] ===================================================================
   6.820101     [D4]  device has been disconnected, try to reconnect
   6.821994     [D4] ===================================================================
   6.825334     [D4] Opening comm_driver ...
````

Code in question (in `reconnect_ups()`):
````
	upsdebugx(4, "Opening comm_driver ...");
	ret = comm_driver->open_dev(&udev, &curDevice, subdriver_matcher, NULL);
	upsdebugx(4, "Opening comm_driver returns ret=%i", ret);
````

so it never passes `open_dev` alive.

jimklimov:
One difference from `upsdrv_initups()` is the `NULL` arg here and `&callback` there.

jimklimov:
`udev` was not invalidated when closed. Fixing this seems to work right.

Issue #1295, opened 2022-02-15
usbhid-ups: automate fallback defaults for pollonly vs interruptonly


jimklimov:
It seems that various platforms (variability covering OS, CPU, virtualization, USB hardware, hubs, libusb version and whatnot) fare with different success about interrupt vs polling backed communications. This is something users have to know to tweak if their USB UPS is not supported out of the box, and seems so far there is no known (documented) "science" behind it to predict who would need what.

It would be helpful to improve the current situation with:
* `nut-scanner` ways to check if it can communicate with the USB UPS by default, or needs further tweaks into suggested configuration;
* `usbhid-ups` (and other?) drivers to track if they were configured not in a "something-only" mode, and the method they try always fails (e.g. `usbhid-ups[3135]: [D2] nut_libusb_get_interrupt: Connection timed out` as reported) so try to enable the other method as "the only" for current run-time or until it fails.

Keep an eye out for a possibly related problem of USB driver disconnections after some hours of uptime - maybe the fix for this issue would also help address (detect, reset/reconnect/... fix) that other one.

jimklimov:
UPDATE: In a way this mirrors the situation of #1309 / #1315 that benefits from adaptive async/sync networking comms.

jimklimov:
For now, adding a soft variant of this - to post a message in driver log about reconnecting. We currently did not track explicitly if recent interrupt processing had I/O errors, and it can take several retry loops until `update_drvinfo()` decides to reconnect:
````
   4.443853     [D5] send_to_all: SETINFO driver.parameter.pollfreq "30"
   4.445986     [D1] upsdrv_updateinfo...
   4.447386     nut_libusb_get_interrupt: Input/Output Error
   4.449303     [D2] dstate_init: sock \\.\pipe\usbhid-ups-nutdev1 open on handle 0000000000000168
   4.452033     [D5] send_to_all: SETINFO driver.parameter.pollinterval "2"
   4.455017     [D5] send_to_all: SETINFO driver.parameter.synchronous "auto"
   4.457235     [D5] send_to_all: SETINFO device.mfr "EATON"
   4.459150     [D5] send_to_all: SETINFO device.model "Ellipse ECO"
   4.461171     [D5] send_to_all: SETINFO device.serial "000000000"
   4.463199     [D1] upsdrv_updateinfo...
   6.466385     [D1] upsdrv_updateinfo...
   6.468542     [D1] Got to reconnect!

[...]

PR #1632, opened 2022-08-28
usbhid-ups: do not crash due to auto-reconnecting

jimklimov:
Closes: #1627
Partially touches on #1295
Follows up from #1335

### DIFF
--- a/drivers/usbhid-ups.c
+++ b/drivers/usbhid-ups.c
@@ -833,7 +833,12 @@ void upsdrv_updateinfo(void)
 			return;
 		}
 
-		upsdebugx(1, "Got to reconnect!\n");
+		upsdebugx(1, "Got to reconnect!");
+		if (use_interrupt_pipe == TRUE) {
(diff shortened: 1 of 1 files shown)


jimklimov:
CI: Spurious error, failed to detect a busy NUT_PORT in check-NIT on centos7.
Guess there we should also check upsd stderr that it is "not listening" on all ports tried - then try another port...

drivers/usbhid-ups.c
```diff
@@ -1493,6 +1504,12 @@ static bool_t hid_ups_walk(walkmode_t mode)
 			hd = NULL;
 			return FALSE;
 
+		case LIBUSB_ERROR_IO:        /* I/O error */
+			/* Uh oh, got to reconnect, with a special suggestion! */
+			interrupt_pipe_EIO_count++;
```
jimklimov, line 1509: Wondering if the count should go down/zero if there were successful reads later?.. So far this is mostly informative (suggest pollonly if there were any interrupt errors)...

Issue #1429, opened 2022-05-17
APC Smart UPS 1000 with latest firmware 04.3 will not work with usbhid-ups anymore


johnynfulleffect:
Hi,

I have an APC Smart UPS 1000 (SMC1000-2UC) and was using the [usbhid-ups](https://networkupstools.org/docs/man/usbhid-ups.html) driver fine.  APC released a firmware update with security fixes which I made sure to upgrade.  [UPS SMT 04.6 & SMC 04.3 – Firmware Release Notes](https://download.schneider-electric.com/files?p_File_Name=SMT1015UPS_v04.6_SMC1018UPS_v04.3_Firmware_Release_Notes.pdf&p_Doc_Ref=SPD_CLII-AXYUKZ_EN&p_enDocType=User+guide)

After upgrading, the nut-driver service fails to start with error: No matching HID UPS found

Details on my UPS

```
  idVendor           0x051d American Power Conversion
  idProduct          0x0004 
  bcdDevice            0.01
  iManufacturer           1 American Power Conversion 
  iProduct                2 Smart-UPS_1000 FW:UPS 04.6 / ID=1018
```

jimklimov:
Just in case, which NUT version is it - 2.7.4 or recent 2.8.0 or custom build from git (master)?

My guess would be the FW update changed some identification values used by NUT and/or your ups.conf section (productid? vendorid?) and so NUT driver start-up does not find a good match.

Can you run the driver stand-alone with more debug e.g. `usbhid-ups -a apc -DDDDDD` to check what it fails to see?

Also, make sure other processes (including another copy of the NUT driver, or default kernel handler which udev rules should have taken care of) have not grabbed the device already - this also precludes new ones from seeing it.

johnynfulleffect:
> 

Hi @jimklimov  thanks for the reply.

Current nut version is `2.7.4-13` and was installed via apt.  Before the upgrade, my productid was 3, and was changed to 4 after the FW upgrade, so I updated the ups.conf appropriately.

Question about running the driver.  I do not have that installed, in other words, usbhid-ups is not found.  Can you help me with that step?  I can provide those debug messages.

In the meantime, I am going to try again and revert the productid back to 3 and see if anything changes.

jimklimov:
Please see the packaging metadata for your distribution (also, often packages can be opened as archives in UIs like Midnight Commander).

[...]

PR #1655, opened 2022-09-12
drivers/apc-hid.c, NEWS: Support VID:PID = 0x051d:0x0004 [#1429]

jimklimov:
What the tin says.

Closes: #1429


jimklimov:
Testing stalled on user's side; however it does not seem this change would bring regressions so merging.
One thing to look out for in future reports is whether `disable interrupt pipe` applies to this firmware revision too.

And it seems unfortunate that firmware change modifies the device product ID... but if it *does* behave differently than original release, it may make sense to differentiate identification too.

Issue #1650, opened 2022-09-04
Compiling issues due to strptime.c


Anutrix:
I tried compiling on WSL2 using [llvm-mingw](https://github.com/mstorsjo/llvm-mingw) project.

Build failed with following error on `make`:
```
...
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I../include -I../include -isystem /usr/local/include -g -O2 -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Wno-system-headers -Wall -Wextra -Wsign-compare -pedantic -Wno-error -MT strptime.lo -MD -MP -MF .deps/strptime.Tpo -c strptime.c  -fPIC -DPIC -o .libs/strptime.o
strptime.c:59:25: error: expected ';' before 'uint64_t'
   59 | typedef unsigned __int64 uint64_t;
      |                         ^~~~~~~~~
      |                         ;
strptime.c:59:1: warning: useless type name in empty declaration
   59 | typedef unsigned __int64 uint64_t;
      | ^~~~~~~
strptime.c: In function 'strptime':
strptime.c:389:25: warning: implicit declaration of function '_tzset'; did you mean 'tzset'? [-Wimplicit-function-declaration]
  389 |                         _tzset();
      |                         ^~~~~~
      |                         tzset
At top level:
cc1: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics
cc1: note: unrecognized command-line option '-Wno-reserved-identifier' may have been intended to silence earlier diagnostics
make[2]: *** [Makefile:603: strptime.lo] Error 1
...
```

I could make it finish without crashing with following changes:
```diff
diff --git a/common/strptime.c b/common/strptime.c
index 1672ded8a..5adb81ebb 100644
--- a/common/strptime.c
+++ b/common/strptime.c
@@ -56,7 +56,7 @@ __weak_alias(strptime,_strptime)
 */
 typedef unsigned char u_char;
 typedef unsigned int uint;
-typedef unsigned __int64 uint64_t;
+// typedef unsigned __int64 uint64_t;
 
 #define	_ctloc(x)		(_CurrentTimeLocale->x)
 
@@ -386,7 +386,7 @@ recurse:
 			continue;
 
 		case 'Z':
-			_tzset();
+			tzset();
 			if (strncasecmp((const char *)bp, gmt, 3) == 0
           || strncasecmp((const char *)bp, utc, 3) == 0) {
[...]

PR #1658, opened 2022-09-14
Document LXC setup for NUT builds with Arch Linux

jimklimov:
Part of investigation for #1279 - needed a platform to reproduce user build failure. That is tangentially related to #1650 (also exposes `strptime` issues).


jimklimov:
Resulting Arch Linux setup passes NUT build and self-check, except Python 3.10 PyNUT tests (claim to not connect to the localhost server).

PR #1662, opened 2022-09-15
Hide 'Init SSL without certificate database' message for upsc

tofurky:
This can clutter up logs on things that call 'upsc', such as Munin plugins. Switch it to upsdebugx() so it's not always output to stderr.

### DIFF
--- a/clients/upsclient.c
+++ b/clients/upsclient.c
@@ -405,10 +405,10 @@ int upscli_init(int certverify, const char *certpath,
 	PK11_SetPasswordFunc(nss_password_callback);
 
 	if (certpath) {
-		upslogx(LOG_INFO, "Init SSL with cerificate database located at %s", certpath);
+		upslogx(LOG_INFO, "Init SSL with certificate database located at %s", certpath);
 		status = NSS_Init(certpath);
 	} else {
-		upslogx(LOG_NOTICE, "Init SSL without certificate database");
+		upsdebugx(1, "Init SSL without certificate database");
 		status = NSS_NoDB_Init(NULL);
 	}
 	if (status != SECSuccess) {


tofurky:
unsure if both the Init SSL messages should go to debugx (or neither, really).

jimklimov:
One alternative is to use debug level 0 so it always pops up on stderr like now, but does not hit the syslog.

jimklimov:
Asked community on mailing lists - if insecurity reminders were deemed useful.

jimklimov:
Question: which logs does it clutter - syslog (thanks to `upslogx` used) or some unhandled stderr in munin?

tofurky:
munin writes unhandled stderr to syslog (*/5):

```
Sep 16 09:45:11 aquos munin-node[2781871]: 2022/09/16-09:45:11 [2781871] Error output from nut_misc:
Sep 16 09:45:11 aquos munin-node[2781871]: 2022/09/16-09:45:11 [2781871] #011Init SSL without certificate database
Sep 16 09:45:12 aquos munin-node[2781871]: 2022/09/16-09:45:12 [2781871] Error output from nut_volts:
Sep 16 09:45:12 aquos munin-node[2781871]: 2022/09/16-09:45:12 [2781871] #011Init SSL without certificate database
```

i suppose the munin plugin could be modified to redirect stderr to /dev/null, but that'd also remove useful errors. my reasoning was that adding a grep -v pipe would add extra overhead. i think it'd also require redirecting stderr to stdout which would mess with the parser if other errors were emitted.

thanks for putting a feeler out.

jimklimov:
FWIW, a link to stirred-up discussion : https://alioth-lists.debian.net/pipermail/nut-upsdev/2022-September/007761.html

jimklimov:
So far it seems the proposition is reasonable, *but* this quiescing should be non-default. Might be activated by an envvar and/or command-line option. since `upsc` (or for that matter other clients -- consumers of this library -- like `upsrw` and `upscmd`) do not have any config files.

tofurky:
i think an env variable might work. adding a new cli flag wouldn't really be usable if it gave a syntax error when it's unsupported.

i did look pretty closely at how NSS is initialized, and the conclusion i came to is that simply changing the verbosity of the message was the most straightforward thing to do.

jimklimov:
Thanks for raising the concern about "a syntax error when it's unsupported", so I posted a PR for an envvar-based solution. Can you check if that looks good for your purposes?

tofurky:
yep, it should work. left a note there as well. thanks!

PR #1667, opened 2022-09-25
Support NUT_QUIET_INIT_SSL envvar to hide "Init SSL without certificate database" message

jimklimov:
Follows up from discussion for PR #1662


tofurky:
gave this a quick build+test and it works as advertised, thanks!

PR #1676, opened 2022-10-11
snmp-ups: Add support for Eaton Network-M2

zakx:
I recently acquired a couple Eaton 9PX UPSes that came bundled with the [Eaton Gigabit Network Card](https://www.eaton.com/us/en-us/catalog/backup-power-ups-surge-it-power-distribution/eaton-gigabit-network-card---na.html) (Network-M2). While that device does have support for some proprietary Eaton management solutions, it also has SNMP support based on the XUPS-MIB.

This PR introduces the eaton-network-m2-mib driver which enables basic support for the Eaton Gigabit Network Card (Network-M2), a modern and recent monitoring card for UPSes such as the Eaton 9PX.

I've left a lot of commented mibs that I couldn't immediately map to nut names, but might be of help for someone who tries to use that data in the future.


jimklimov:
Thanks, I'd ask @aquette @eric3873 @EatonEKia to double-check this so we do not end up with duplicate functionality: there was quite a bit of Eaton/MGE/Powerware collaborated support in NUT.

In particular, SYSOID ".1.3.6.1.4.1.534.1" pops up in https://github.com/networkupstools/nut/blob/master/drivers/powerware-mib.c so I am not sure if they would conflict if built together - and did you make sure the existing sub-drivers did not handle something for your devices? It may be a matter of updating an existing one with lacking entries (if a data point exists, it would be mapped; otherwise ignored; for same NUT data point names, first hit wins) - rather than adding a new one.

There is also NM2-specific support in https://github.com/networkupstools/nut/blob/master/drivers/eaton-ats16-nm2-mib.c vs. older NMC https://github.com/networkupstools/nut/blob/master/drivers/eaton-ats16-nmc-mib.c - but it may be limited to ATS devices (and at different SysOID).

jimklimov:
@zakx : a couple more ideas catching up:

1) Which version of NUT were you running that did not seem to support these UPSes? Many distributions still tend to ship the older NUT release 2.7.4 (~6 years old), although 2.8.0 is out since this spring. There were many updates for Eaton-branded devices between these releases, so if you began this as a fix for 2.7.4 - maybe it is already solved in the current master branch.

2) For the future PRs, it is recommended to make uniquely named branches as the PR source for each feature, in particular to avoid clumsy resynchronization and spurious PR changes whenever you would want to update against "real" upstream branch contents (master or otherwise).

zakx:
Thanks for your input; it looks like I did not actually test with master before. I thought I had a recent-enough version (using some recent Debian builds somebody in another GitHub issue on this project had provided) and that certainly had no support for the card.

I've closed this request and instead offer you a documentation update at #1678.

aquette:
@zakx you should still check upsc* outputs to see if nothing is missing 😉
Thanks again for your contrib

PR #1678, opened 2022-10-12
Documentation update regarding Network-M2 card

zakx:
The Eaton Gigabit Network Card (Network-M2) is already well-supported by the powerware driver. This patch updates the documentation to reflect this.

(Replaces #1676.)


aquette (approved):
Lgtm, thanks again

PR #1682, opened 2022-10-18
Fix walk issue with snmpv3 (** add type error exception **)

eric3873:
Fix walk issue with snmpv3 by adding the treatment of type error exception in loop reading.

### DIFF
--- a/drivers/snmp-ups.c
+++ b/drivers/snmp-ups.c
@@ -1156,7 +1156,20 @@ static struct snmp_pdu **nut_snmp_walk(const char *OID, int max_iteration)
 			snmp_free_pdu(response);
 			break;
 		} else {
-			numerr = 0;
+			/* Checked the "type" field of the returned varbind if it is a type error
+			 * exception (only applicable with SNMPv2 or SNMPv3 protocol) */
+			if (response->variables->type == SNMP_NOSUCHOBJECT ||
+			    response->variables->type == SNMP_NOSUCHINSTANCE ||
+			    response->variables->type == SNMP_ENDOFMIBVIEW) {
+				upslogx(LOG_WARNING, "[%s] Warning: type error exception (OID = %s)",
+						upsname?upsname:device_name, OID);
+				snmp_free_pdu(response);
+				break;
+			}
+			else {
+				/* no error */
+				numerr = 0;
+			}
 		}
 
 		nb_iteration++;


drivers/snmp-ups.c
```diff
@@ -1156,7 +1156,20 @@ static struct snmp_pdu **nut_snmp_walk(const char *OID, int max_iteration)
...
-			numerr = 0;
+			/* Checked the "type" field of the returned varbind if it is a type error
+			 * exception (only applicable with SNMPv2 or SNMPv3 protocol) */
+			if (response->variables->type == SNMP_NOSUCHOBJECT ||
+			    response->variables->type == SNMP_NOSUCHINSTANCE ||
+			    response->variables->type == SNMP_ENDOFMIBVIEW) {
+				upslogx(LOG_WARNING, "[%s] Warning: type error exception (OID = %s)",
+						upsname?upsname:device_name, OID);
+				snmp_free_pdu(response);
+				break;
```
jimklimov, line 1167: Should it set some numerr here?

drivers/snmp-ups.c
```diff
@@ -1156,7 +1156,20 @@ static struct snmp_pdu **nut_snmp_walk(const char *OID, int max_iteration)
 			snmp_free_pdu(response);
 			break;
 		} else {
-			numerr = 0;
+			/* Checked the "type" field of the returned varbind if it is a type error
+			 * exception (only applicable with SNMPv2 or SNMPv3 protocol) */
```
jimklimov, line 1160: Should it check for the protocol actually used (2/3) here?

drivers/snmp-ups.c
eric3873: For me, it is not really an error, it just that the OID is not implemented. For large request, it should find the next available OID in the list without stopping loop treatment.

drivers/snmp-ups.c
eric3873: I think it is not necessary to check the protocol here, it will add a test for nothing.
 Just with snmpv1, this test will never occur.

drivers/snmp-ups.c
aquette: FYI, this numerr is a mess.
it's supposed to catch transient fetch error, while in fact this is generally a missing OID.
I once had a patch to fix that, that got lost...

aquette (approved):

jimklimov (approved):
Thanks for clarifications!

jimklimov:
There was some hiccup on CI, will replay it before merging

PR #1710, opened 2022-11-21
Fix sysOID retrieval on non-compliant device

aquette:
Some SNMP agent (device) wrongly return the sysOID value as a string instead of an OID.

This breaks the initial sysOID matching system of the driver


jimklimov:
Cute. I wonder who is so creative reading (or writing?) the spec?.. :)

PR #1713, opened 2022-11-22
snmp-ups: add support for HPE G2 Metered & Switched PDU

aquette:
This support is unitary (no daisychain support yet). Also note that, due to SNMP v1 implementation limitations on this device, you should prefer SNMP v3 to get both read and write rights


jimklimov:
Thanks, letting CI chew this for a while, in case of typos. Special thanks for the recent stream of such updates, keep them coming :)

And lest it be forgotten by release time, bumping `NEWS` during revision iterations became a good habit (though benefits from knowing an issue/PR number before writing the entry).

jimklimov:
hm, `#if WITH_UNMAPPED_DATA_POINTS` was a bit too eager, addressing... :)

jimklimov (approved):
Generally LGTM, waiting for CI nuances to un-wrinkle...

Issue #743, opened 2019-12-07
Warning: excessive poll failures adding APC NMC card


ssbarnea:
I tried adding the UPS and got:
```
[root@n0 ~]# snmp-ups -a apc
Network UPS Tools - Generic SNMP UPS driver 0.97 (2.7.4)
Duplicate driver instance detected! Terminating other driver!
No matching MIB found for sysOID '.1.3.6.1.4.1.318.1.3.27'!
Please report it to NUT developers, with an 'upsc' output for your device.
Going back to the classic MIB detection method.
Detected Smart-UPS RT 3000 XL on host 192.168.33.9 (mib: apcc 1.2)
[apc] Warning: excessive poll failures, limiting error reporting (OID = .1.3.6.1.4.1.318.1.1.1.9.2.3.1.5.1.1.3)
[apc] Warning: excessive poll failures, limiting error reporting (OID = .1.3.6.1.4.1.318.1.1.1.9.3.3.1.6.1.1.1)
```

How can I provide more info so we avoid the `No matching MIB found for sysOID '.1.3.6.1.4.1.318.1.3.27'!` line?

What is the purpose of these warnings, can we avoid it?

I would be happy to provide the output of upsc for my device but sadly that error line does not tel how to obtain this output.

```
cat /etc/ups/ups.conf
[apc]
driver = snmp-ups
port = 192.168.33.9
```

I tried to list output but all I got was:
```
upsc -L
Error: Connection failure: No route to host
```

PS. Excuse my low quality issue as today is my first day with `nut` package, so I may lack some background knowledge.

milkotodorov:
I have the same issue...

aquette:
Fyi, i never finished the fix but it's simply due to non existent OIDs. The reporting of your non supported sysoid is still valid though, to connect this fingerprint to the right OID map in nut

milkotodorov:
Thanks! It's an APC750I UPS with APC AP9630 NMC. This is not supported?

stevehop:
Same here with a SMT1500I and AP9631 NMC, OID is '.1.3.6.1.4.1.318.1.3.27'

Salvora:
same here with srt5kxli

linnemannr:
[...]

PR #1716, opened 2022-11-25
snmp-ups: fix "Warning: excessive poll failures"

aquette:
Fix that long standing issue, which was tied to non-existent OIDs, not well handled in some part of the driver

Closes: #743


jimklimov (approved):
Great! :)

PR #1720, opened 2022-11-27
C99 fixes for the configure script

fweimer-rh:
Future C compilers will likely default to language modes which do not support implicit function declarations. Rebuilding nut with an [instrumented compiler](https://fedoraproject.org/wiki/Toolchain/PortingToModernC#Discovering_Packages_Which_Need_Fixing) revealed a few issues in the configure script.

### DIFF
--- a/m4/ax_c___attribute__.m4
+++ b/m4/ax_c___attribute__.m4
@@ -67,7 +67,7 @@ AC_DEFUN([AX_C___ATTRIBUTE__], [
 	  foo(__attribute__ ((unused)) int i) {
 	      return;
 	  }
-        ]], [func(1);])],
+        ]], [foo(1);])],
       [ax_cv___attribute__unused_arg=yes],
       [ax_cv___attribute__unused_arg=no]
     )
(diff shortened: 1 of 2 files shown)


jimklimov:
Decent catches, in line with the #823 and related issues of the "fightwarn effort" that took years to brush up the codebase - thanks!

configure.ac
```diff
@@ -475,7 +475,6 @@ AC_CACHE_CHECK([for strdup(s)],
 char *t = "Test";
 char *s = strdup(t);
 if (!s || !(s[0]=='T'&&s[1]=='e'&&s[2]=='s'&&s[3]=='t'&&s[4]=='\0') || s == t) res = 1;
-if (s) free (s);
```
jimklimov, line 478: This looks odd -- IIRC when this test program was PoCed, at least in some build systems it warned about leaks so `free()` was added. What did your instrumentation complain about - lack of header for it? Maybe better fix it from that vector :)

jimklimov: Sorry for the delay - it would be OK to add another commit. I'm currently not at a computer so can't check quickly, but hope the `HAVE_STDLIB_H` is discovered in the `configure.ac` (or the m4's it calls) before this check. If not, feel free to move it until after that test is done, so you can safely `ifdef ... include` it :)

Thanks!

jimklimov: (and then maybe `ifdef ... free()` to be pedantic?..)

PR #1722, opened 2022-11-28
Update mge-hid.c

Tiph:
fixed conversion issue

small issue I've spotted while playing around.

The issue: 

We can't override the powerfactor (override.output.powerfactor = 0.xx) because of the atoi function that change the type to Integer


jimklimov:
CC @aquette

* Are values served by device ints or floats? If ints - are they scaled to become a real factor? (e.g. 100:1?)
* Should nearby string-number conversions be handled similarly (nominal...)?

Tiph:
> CC @aquette
> 
> * Are values served by device ints or floats? If ints - are they scaled to become a real factor? (e.g. 100:1?)
> * Should nearby string-number conversions be handled similarly (nominal...)?

Issue #1721, opened 2022-11-28
Missing pid-file upsmon.pid and no browsing possible


ullix:
So far I am failing in installing nut 2.8.0 on my "Linux Mint 21" computer. This may be one of the issue, when trying a test-shutdown:

```
$ upsmon -c fsd
Network UPS Tools upsmon 2.8.0
fopen /run/nut/upsmon.pid: No such file or directory
```
and there is no shutdown. The content of /run/nut is this:

![image](https://user-images.githubusercontent.com/9335726/204285889-1e77d3b7-c01b-47f3-88b3-8de08e7cf62f.png)

Something else is working, though, so there is hope:
```
$ upsc eaton@10.0.0.51 ups.model
Init SSL without certificate database
Ellipse PRO 650 
```
Trying access with a webbrowser has completely failed. Apache is installed, and when using the url: `http://10.0.0.51:3493/`, it akes 1 ... 2 minutes(!) before I get back:
```
ERR INVALID-ARGUMENT
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
ERR UNKNOWN-COMMAND
```
Same outcome when I try e.g. `10.0.0.51:3493/upsstats.cgi` or variations of it.

jimklimov:
Ugh, did you read the docs on NUT architecture and setup? :) Some hints:
* port 3493 is for NUT protocol, for data-server and its clients (like upsc, nut-cgi and upsmon) to talk to the server, localhost or remote. So this is not for the browsers. You can set up Apache, NGinx, etc. to serve CGI "scripts" and represent the nut-cgi programs as such scripts.
* The `upsmon` error seems to imply that you do not currently have the `upsmon` daemon running locally. The `-c` is a way for the newly launched binary to effectively find its copy already running (via PID file) on the "master"/"primary" system and send it a signal to trigger the shutdown (tell "secondary" systems to go down, then shut down itself, and if the late-shutdown integration works for your distro, tell the UPS to power off in the end). If the service is not running, then there's nobody to signal to :)

ullix:
[...]

PR #1724, opened 2022-12-01
Suggest in docs and daemon messages how command-signals vs. PID files vs. systemd units interact

jimklimov:
Follow-up from #1721 discussions, mostly to clarify docs and messages to be more actionable with regard to changes in NUT 2.8.0 and systemd unit wrapping (e.g. that by default it starts `upsd -F` and so does not save PID files).

Also clarifies some example texts to be more consistent between files (e.g. NUT user for upsmon).

### DIFF
--- a/clients/upsmon.c
+++ b/clients/upsmon.c
@@ -2370,100 +2370,117 @@ int main(int argc, char *argv[])
 		}
 	}
 
-	if (cmd) {
-		int reported = 0;
+	/* Note: "cmd" may be non-trivial to command that instance by
+	 * explicit PID number or lookup in PID file (error if absent).
+	 * Otherwise, we are being asked to start and "cmd" is 0/NULL -
+	 * for probing whether a competing older instance of this program
+	 * is running (error if it is).
+	 */
 #ifndef WIN32
-		if (oldpid < 0) {
-			cmdret = sendsignal(prog, cmd);
-		} else {
-			cmdret = sendsignalpid(oldpid, cmd);
-		}
-
-# ifdef HAVE_SYSTEMD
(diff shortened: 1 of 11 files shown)


jimklimov:
For builds with systemd:
````
$ ./server/upsd -c reload
Network UPS Tools upsd 2.8.0-175-g3d5537378
fopen /var/state/ups/upsd.pid: No such file or directory
Try 'systemctl reload nut-server.service' or add '-P $PID' argument
````

For builds on other systems:
````
$ ./server/upsd -c reload
Network UPS Tools upsd 2.8.0-182-gdbe5e39e6
fopen /var/state/ups/upsd.pid: No such file or directory
Failed to signal the currently running daemon
````

jimklimov:
Refactoring the code for detection and reporting of probe/signal outcomes into one block:
````
$ ./server/upsd -c reload
Network UPS Tools upsd 2.8.0-182-gdbe5e39e6
fopen /var/state/ups/upsd.pid: No such file or directory
Could not find PID file '/var/state/ups/upsd.pid' to see if previous upsd instance is already running!
Failed to signal the currently running daemon (if any)
Try to add '-P $PID' argument

$ ./server/upsd -c reload -P1
Network UPS Tools upsd 2.8.0-182-gdbe5e39e6
kill: Operation not permitted
Failed to signal the currently running daemon (if any)

$ ./server/upsd -c reload -P123123
Network UPS Tools upsd 2.8.0-182-gdbe5e39e6
kill: No such process
Failed to signal the currently running daemon (if any)
````

jimklimov:
See some more testing of commands vs. starts with and without a competing daemon in #1728 research trail.

PR #1734, opened 2022-12-20
Fix discovery issue with Eaton pdu nlogic

eric3873:
Use generic name in mib2 instead of the real device name in eaton mib.


aquette (approved):
mea culpa :D

aquette:
side note: the (not so visible) impact of this is that the driver can consider any SNMP device as matching, due to the device.model data (used as a 2nd step matching) being the systemDescription from MIB2, and thus monitor unhandled devices.
This can prove to be an issue when integrated into devices discovery processes

Issue #1737, opened 2022-12-22
On some platforms we can not safely pass NULL parameters to printf and relatives


jimklimov:
For example, `upsdebugx()` is used all over the place and `printf("%s", NULL)` segfaults on some platforms/libc implementations.

If this bites in the field, we would have to use `NUT_STRARG()` macro intensively, to ensure strings are never NULL in such parameter lists. Ideally the varargs handling of `upsdebugx()` and related definitions might take care of this centrally and substitute all NULLs referred by string tokens?.. probably that would be too complex and/or error-prone (e.g. need to differentiate printing null pointers as such, or even zero numbers).

See #1736 for more practical context.

Issue #1736, opened 2022-12-22
Problems building recent NUT on Solaris 8


jimklimov:
Following up from mailing list discussion at https://alioth-lists.debian.net/pipermail/nut-upsuser/2022-December/013051.html

* lacking python (so need to disable augeas to run `autogen.sh`)
* older perl does not support newer `open()` syntax we use (fixes in helper scripts needed at least to run `autogen.sh`)
* default shell interpreter issues, and no easy way to choose another for `configure` script
* after getting something built, NUT clients seem to not actually `connect()` to a data server

Original issue was reported with a SPARC build of the OS, but many points were reproduced and are being fixed in an x86 (32-bit) VM prepared per https://github.com/mac-65/Solaris_8_x86_VM/issues/1 notes, using an add-on TGCware set of packages for compiler and some modern tools that are absent in default OS installation.

jimklimov:
For lack of python - can follow the suggestion (tweaked for older shell syntax):
````
WITHOUT_NUT_AUGEAS=true; export WITHOUT_NUT_AUGEAS; ./autogen.sh
````

jimklimov:
So far one problem remains: `nutlogtest` fails when libc is dumb about `printf("%s", NULL)` -- this segfaults on Solaris 8 and likely would kill NUT if it logs a NULL somewhere; but otherwise with this branch NUT can be built and self-tested (including runs of `upsd` with a dummy driver and client connections talking to it).

Some tweaks were imminent:

````
:; git clean -fffdddxxx ; CONFIG_SHELL=/usr/tgcware/bin/bash WITHOUT_NUT_AUGEAS=true PATH="/usr/xpg4/bin:$PATH" /usr/tgcware/bin/bash ./ci_build.sh
````

jimklimov:
Builds are fixed by PR above, as far passing configuration, self-tests including NIT, and a `make distcheck`.

Bugs due to `printf` handling of NULL as detailed above are still possible, maybe focus of a different effort.

PR #1738, opened 2022-12-22
Fix some problems building recent NUT on Solaris 8

jimklimov:
That was an interesting trip down the memory lane... see #1736 (and #1737 to an extent - not addressed much here except handling the situation in tests)

Now let CI chew at this, whether dumbed-down shell and perl scripts for syntax of the past millenium are compatible with all the new kids on the block? ;)


jimklimov:
TODO: need a better test for availability of `round()` (and some others seeked nearby in math.h)

jimklimov:
Oddly, several platforms failed a C++ test that this did not intentionally touch on:

````
cpputest-client.cpp:320:Assertion
Test name: nut::NutActiveClientTest::test_auth_user
assertion failed
- Expression: !s1.empty()
- Did not expect empty value here

Failures !!!
Run: 5   Failure total: 1   Failures: 1   Errors: 0
D: Got to the end of test suite with code 'false'
[ERROR] cppnit complained, check above
````

jimklimov:
As another clue, it seems to fail this way in distcheck - so maybe something misbehaves for out-of-tree builds/tests?..

jimklimov:
In the end, somehow `sed -i` avoidance was complicit. Another way to skin the cat worked though...

Issue #1699, opened 2022-11-10
Allow optional builds with "unmapped.x.y" data point names


jimklimov:
Subdrivers generated with SNMP and USB HID helper scripts initially provide unclassified data mappings as `unmapped.x.y` patterns with generated names derived from collected reports. These blocks of definitions are normally hidden with `#if 0` so they are never compiled unless local workspace with sources is edited to comment that away for experiments. This status-quo was recently documented in commit 2b9ca931d4e84e10bbf603a87dd942e45d60c00e

A more sustainable approach for Git merges etc. would be to replace that with `#if WITH_UNMAPPED_DATA_POINTS` or something of the sort, selectable with a `configure` script option, so that development/testing builds would have those data points collected and exposed for practical purposes (e.g. identify better what they are about, so a proper "NUT name" can be assigned or negotiated with community).

* Maybe subdrivers built with this option should automatically expose that it is in force (e.g. by suffixing their version) so there may be whatever different behaviors from "standard" driver builds.

@aquette @clepple : WDYT?

aquette:
Good idea. May help users to help developers

PR #1742, opened 2022-12-25
Introduce WITH_UNMAPPED_DATA_POINTS for USB-HID and SNMP subdrivers

jimklimov:
Troubleshooting aid, to help developers help users.

Closes: #1699


jimklimov:
Checked by usual builds (`unmapped` not seen in produced binaries) and a build with `./configure --with-unmapped-data-points --with-all=auto` where `drivers/snmp-ups` had a lot of those.

PR #1758, opened 2023-01-02
COPYING: update to acknowledge codebase picked up from BSD, curl and CC BY-SA sources over time

jimklimov:
Question to the gallery: is this the right thing to do, and the right way to do it? Please do reply :)

CC @bigon @bagder @kleink


jimklimov:
(Acknowledged as "good and sane" by @aquette offline)

Issue #1767, opened 2023-01-04
Want a general "driver.state" machine reporting for fun and profit


jimklimov:
On profit side, this would help code "know" if the device is currently (re?)initializing or doing a regular data walk. This in turn can impact some logical behavior (error-handling, reporting, etc.) such as for #1756

On the fun side, we would see if the driver is currently responsive or e.g. doing I/O, via clients as well as debug-logs for troubleshooting.

UPDATE: Seems to duplicate #1388 to an extent

Issue #1388, opened 2022-04-23
Update standard dstate list with a driver (re?-)initialization status


jimklimov:
Use-cases:
* snmp-ups with large devices (daisy-chains etc.) can take literally minutes to walk the MIB and initialize the first state after which it would regularly do smaller updates
    * more so if the device is not accessible

* dummy-ups can take a while to parse the data file, interrupting for TIMER lines etc. and then would stop reading it (with dummy-once mode since #1385) - hence initialization complete

Currently in automated testing we have to rely on timing or casual guess to decide if we can interrogate the driver and have it some expected state. To an extent, the "Error: driver not connected" handles the bare initialization but does not tell us if the driver is ready to give stable reports. In a way, that is the moment where a "dump mode" with `-d 1` option would exit the driver.

A similar (but probably different) useful dstate option could reflect if the driver is currently parsing external data (began/finished the regular update), also during non-initializing loops.

PR #1770, opened 2023-01-05
Introduce an `allow_duplicates` flag for common USB matching options

jimklimov:
Closes: #1756 

Initially also addressed #1767 (seen in "screenshots" below) but that change was offloaded to PR #1771

Also does minor codebase maintenance, and partially follows up from recent work in PRs:
* #1766
* #1763

May help in situations raised in issues such as:
* #1744
* #1754 
* #1414
* #1273
* #1174

Note that use of USB device matching by a unique combination of fields is much more preferable to `allow_duplicates` (if possible).

-------

Example with use of the new feature (while the driver is running as another process):

````
# NUT_STATEPATH=/tmp NUT_CONFPATH=/etc/nut ./drivers/usbhid-ups -DDDDDD -a eco650 -u nut -x allow_duplicates
Network UPS Tools - Generic HID driver 0.49 (2.8.0-Windows-299-g5e8e40cca)
USB communication driver (libusb 1.0) 0.43
   0.000000     [D3] main_arg: var='driver' val='usbhid-ups'
   0.000015     [D3] main_arg: var='port' val='auto'
   0.000022     [D5] send_to_all: SETINFO driver.parameter.port "auto"
   0.000027     [D3] main_arg: var='vendorid' val='0463'
   0.000033     [D5] send_to_all: SETINFO driver.parameter.vendorid "0463"
   0.000037     [D3] main_arg: var='productid' val='FFFF'
   0.000041     [D5] send_to_all: SETINFO driver.parameter.productid "FFFF"
   0.000046     [D3] main_arg: var='product' val='Ellipse ECO'
   0.000050     [D5] send_to_all: SETINFO driver.parameter.product "Ellipse ECO"
   0.000055     [D3] main_arg: var='desc' val='Eaton Ellipse ECO 650'
   0.000059     [D3] main_arg: var='serial' val='000000000'
   0.000064     [D5] send_to_all: SETINFO driver.parameter.serial "000000000"
   0.000068     [D3] main_arg: var='vendor' val='EATON'
   0.000075     [D5] send_to_all: SETINFO driver.parameter.vendor "EATON"
   0.000079     [D3] main_arg: var='bus' val='003'
   0.000083     [D5] send_to_all: SETINFO driver.parameter.bus "003"
   0.000092     [D1] Built-in default or configured user for drivers 'nobody' was ignored due to 'nut' specified on command line
   0.000098     [D3] main_arg: var='allow_duplicates' val='<null>'
[...]

PR #1771, opened 2023-01-05
Introduce `driver.state` tracking

jimklimov:
Closes: #1767
Closes: #1388

Split off from PR #1770 as a separate feature, and at that - one for NUT community to agree upon (new standard nut-names entry).

This change allows drivers to uniformly report (to NUT clients as well as a dstate entry that can be internally consulted by driver logic) whether they are at this moment initializing, reconnecting, updating info from the device, or just being quiet between work bursts.


jimklimov:
Got no comments (nor "vetoes") here or on mailing lust during the week.

PR #1774, opened 2023-01-05
Set curDevice->Device during USB enumeration

efuss:
During USB enumeration, set `curDevice->Device`.
Try to use _libusb1_'s port number obtained by `libusb_get_port_number()`, but that seems useless at least on NetBSD. So, use `devnum` (the enumeration loop variable), which at least makes two identically looking USB UPSen distinguishable again.

Maybe we want to add a `Port` property set to what `libusb_get_port_number()` reports?

Previously missed during refactoring to libusb1.

### DIFF
--- a/drivers/libusb1.c
+++ b/drivers/libusb1.c
@@ -234,18 +234,36 @@ static int nut_libusb_open(libusb_device_handle **udevp,
 			libusb_free_device_list(devlist, 1);
 			fatal_with_errno(EXIT_FAILURE, "Out of memory");
 		}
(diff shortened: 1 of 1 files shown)


jimklimov:
Good catch! I'm unsure about `if 0` never firing. Can the device/port ever be validly 0? Maybe we can check for that instead (at runtime) - if port>0 use it else use devnum (and upsdebugx that).

jimklimov:
> Maybe we want to add a Port property set to what libusb_get_port_number() reports?

Looking at docs, it was known in libusb0 as a "path"...

PR #1778, opened 2023-01-06
huawei-ups2000.c: log error code/string for Modbus failures

biergaizi:
Currently, failures in `modbus_read_registers()` or `modbus_write_registers()` do not show any error code or message in the log, making it difficult to debug. This commit logs the error code, errno, and `modbus_strerror()` if an error occurs. Furthermore, the log level of a fatal warning has been changed from `LOG_WARNING` to `LOG_ERR` (`LOG_WARNING` was previously used due to an oversight).


jimklimov (approved):
Thanks, LGTM :)

Issue #1787, opened 2023-01-09
configure behavior about python is hard for packaging


gdt:
I am using nut git master as of today, building in pkgsrc.

(From my viewpoint python 2.7 is now history and we might as well gc the support.  But that's off point.)

pkgsrc chooses a python version to depend on and ensures it is there, and sets variables in our Makefile that I can pass in the env or to configure, in order to use that version (and only that version).  So I don't want nut's configure to look, I want to pass the specific version.

It's unsound to use env or python3 because we are installing in some particular versions site-packages, so we can't have that just change.  It has to match.

I am already passing -I/-L CFLAGS/LDFLAGS for python3.10 but I think that's not important.

The python configure code appears not to take  a version as input, and it looks for python3 before python3.N.  It doesn't look for 3.10.   It finds it without a pass, and testclient ends up with "#!/usr/bin/env python3", and this is not allowed (not bound to the dependency precisely).   I want "#!/usr/pkg/bin/python3.10" after I pass in --with-python=/usr/pkg/bin/python3.10 or PYTHON= (whatever works is fine in env or configure cmd line).

My suggestions (severable :-) are:
  * rip out python2 stuff
  * take a --with-python=/path/to/python3.N arg
  * if not, for python3, look for python3.10 through python3.7 or maybe lower by default.  Skip python3 as dangerous and people who want that can use --with-python.

jimklimov:
Please take a look at `ci_build.sh` use of `PYTHON` envvar which does optionally set `--with-python=/some/path` (and is directly used by `autogen.sh` to run helper scripts that provide includes for `configure.ac`. Indeed, documentation missed that flag, will update. IIRC the envvar (and `PYTHON2` and `PYTHON3` for targeted activity) also acts as default for `configure` if a CLI option is not used.

So far there is no plan to "rip out" python2 since NUT project aims to do its best to build and run on all systems released this millenium, as much as is reasonably feasible (e.g. where older versions ran okay). Being off upstream support does not mean there are no boxes still running that obsoleted version and depending on it with no (non-disruptive) way forward. To that extent, PyNUT module, NUT-Monitor GUI app, and build helper scripts were either changed syntactically to work with at least python-2.7 and several version of python-3.x that I could get in various CI-tested distros, or in case of GUI app (and changes of GUI backend involved) - split into two that can be alternately installed.

The NUT CI farm does test several implementations in its build matrix, so the mechanism is known to work :)

[...]

PR #1788, opened 2023-01-09
Document configure options regarding Python (module, app)

jimklimov:
Closes: #1787

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2561,6 +2561,7 @@ if test x"$solarissmf" = xauto ; then
 fi
 AC_MSG_RESULT([${solarissmf}])
 AM_CONDITIONAL(WITH_SOLARIS_SMF, test x"$solarissmf" = x"yes")
+NUT_REPORT([consider basic SMF support], [${solarissmf}])
 
 
 AC_MSG_CHECKING(whether to install Solaris SVR4 (legacy) init-script files)
--- a/configure.ac
+++ b/configure.ac
@@ -1812,13 +1812,13 @@ AS_IF([test "${nut_with_nut_monitor}" != no -o "${nut_with_pynut}" != no], [
     NUT_REPORT([use specific Python3 interpreter],   [${PYTHON3}])
 
     NUT_REPORT_PATH_INTEGRATIONS([Default  Python  interpreter site-packages],
-        ["${PYTHON_SITE_PACKAGES}"])
+        [${PYTHON_SITE_PACKAGES}])
 
     NUT_REPORT_PATH_INTEGRATIONS([Specific Python2 interpreter site-packages],
-        ["${PYTHON2_SITE_PACKAGES}"])
+        [${PYTHON2_SITE_PACKAGES}])
 
     NUT_REPORT_PATH_INTEGRATIONS([Specific Python3 interpreter site-packages],
-        ["${PYTHON3_SITE_PACKAGES}"])
+        [${PYTHON3_SITE_PACKAGES}])
         dnl # Python site-packages installation path for specific Python3 interpreter
 ])
 
@@ -2645,7 +2645,7 @@ if test "${systemdsystemunitdir}" = "auto" ; then systemdsystemunitdir=""; fi
 if test -n "${systemdsystemunitdir}"; then
 	have_systemd="yes"
 	AC_MSG_RESULT(using ${systemdsystemunitdir})
-	NUT_REPORT_PATH_INTEGRATIONS([Service units for systemd], ["${systemdsystemunitdir}"])
+	NUT_REPORT_PATH_INTEGRATIONS([Service units for systemd], [${systemdsystemunitdir}])
 else
 	have_systemd="no"
 	AC_MSG_RESULT(no)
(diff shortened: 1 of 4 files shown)


gdt:
In general this feels too complicated. 
  * If python is used as a build tool that should be separate from the target.
  * It seems nut should find and select a single python (to be used after nut is installed)
  * The default should steer to python3.N first, and bare python3 (which by definition is a pointer that might get changed)
  * There is code that works with 3 as the standard approach.  If there's py27 code, fine, install that when 2 is selected.  If that's done optionally *also*, make that opt in.   I can't imagine anyone wanting that unless their system is python3-deficient.
  * I see zero need for dispatching at runtime.   That just seems like extra complexity that solves a problem people shouldn't have.

gdt:
You mentioned `realpath` and I am not sure you meant that literally.
POSIX specifies `realpth(3)` but there is no requirement for a `realpath(1)` and NetBSD (and likely others) does not have one.

A thought is that given a `--with-python=FOO`, one can just run `FOO-config --prefix`` to find the prefix, avoiding a lot of complexity.

jimklimov:
Well, for systems that have it - literally indeed. For others, fall back to the "FOO" pathname (which may be a symlink so less useful).

Good point about `FOO-config`, quite worth trying it as a fallback too. Not ubiquitous though, fewer and fewer projects and/or packages ship a `FOO-config` script/program.

gdt:
I thought that e.g. python3.10-config was the standard approach in python's install -- that's not the kind of thing pkgsrc makes up.   But fair point that if others don't follow that, you can't relay on it.

But, I only really care what happens when one provides the full path to a real program, since that's how my world is.

gdt:
Thanks for working on this.  I tested this and it's mostly ok.  However, with `--with-python=/usr/pkg/bin/python3.10` I still ended up with files installed to python2.7.   I'm rebuilding with `--without-python2` which is easy enough, but I don't think that should be necessary after specifying a python version.

jimklimov:
As a counter-offer, I'll update the docs with the suggestion to use the option :)

jimklimov:
Heh, `pythonX-config` is (always?..) a python script itself, so has a useful shebang...

jimklimov:
Neat:
````
:; ./autogen.sh && ./configure --without-all --with-pynut
...
checking for python... python
configure: WARNING: A python program name without a specific version number was requested (may be a symlink prone to change over time): /usr/bin/python
checking python interpeter to call... /usr/bin/python (sys.version_info(major=3, minor=8, micro=10, releaselevel='final', serial=0))
checking python site-packages location... /usr/local/lib/python3.8/dist-packages
checking for python2... python2
checking python2 interpeter to call... /usr/bin/python2 (sys.version_info(major=2, minor=7, micro=18, releaselevel='final', serial=0))
checking python2 site-packages location... /usr/local/lib/python2.7/dist-packages
configure: WARNING: A python3 program name was not specified during configuration, will default to '/usr/bin/python3.8' (derived from --with-python setting which has a suitable version)
checking python3 interpeter to call... /usr/bin/python3.8 (sys.version_info(major=3, minor=8, micro=10, releaselevel='final', serial=0))
checking python3 site-packages location... /usr/local/lib/python3.8/dist-packages
...
````

So here the `python` value as defaulted was used almost as such -- just resolved to `/usr/bin/python` (with warnings about the uncertainty). Similarly for `python2`.

However for `python3` defaulted to try the value of `PYTHON` first because it has `major=3`, the complete `/usr/bin/python3.8` path got discovered.

Wondering how well MSYS2/mingw would play with that, considering its resolution to windows paths rather than POSIXy shims:
````
$ which python ; python -c 'import sys; print(sys.executable);'
/mingw64/bin/python
C:/msys64/mingw64/bin/python.exe
````

And ouch, this does not help build confidence (gotta filter discovered names after all):
````
$ cat /mingw64/bin/python-config
#!/bin/sh
````

jimklimov:
Hm, seems --without-python2 now misbehaves

gdt:
Let me know when you want me to retest (i.e., when you have come to a stopping place and think it's ok).

jimklimov:
> Let me know when you want me to retest (i.e., when you have come to a stopping place and think it's ok).

Hopefully found the responsible typo, fixed.

````
:; rm -f Makefile configure ; ./autogen.sh && ./configure --without-all --with-pynut --without-python2
...
checking for python... python
configure: WARNING: A python program name without a specific version number was requested (may be a symlink prone to change over time): /usr/bin/python
checking python interpeter to call... /usr/bin/python (sys.version_info(major=3, minor=8, micro=10, releaselevel='final', serial=0))
checking python site-packages location... /usr/local/lib/python3.8/dist-packages

checking python2 interpeter to call... no

configure: WARNING: A python3 program name was not specified during configuration, will default to '/usr/bin/python3.8' (derived from --with-python setting which has a suitable version)
checking python3 interpeter to call... /usr/bin/python3.8 (sys.version_info(major=3, minor=8, micro=10, releaselevel='final', serial=0))
checking python3 site-packages location... /usr/local/lib/python3.8/dist-packages

...

* install NUT-Monitor desktop application:      no
* install PyNUT binding module: yes
* Default  Python  interpreter  :       /usr/bin/python
* Default  Python  site-packages:       /usr/local/lib/python3.8/dist-packages
* Specific Python2 interpreter  :
* Specific Python2 site-packages:
* Specific Python3 interpreter  :       /usr/bin/python3.8
* Specific Python3 site-packages:       /usr/local/lib/python3.8/dist-packages
````

gdt:
I made a distfile from this and am able to build it in pkgsrc with only the chosen python.  So WORKSFORME and thanks for hacking on this.

jimklimov:
Super! Thanks for the report!

PR #1814, opened 2023-01-15
CI: add NIT tests for nut-scanner/libupsclient

jimklimov:
While the test is rudimentary, it allows to check that `nut-scanner` tool basically works, and the C client library talks to upsd well. In practice during tests in build tree it often depends a lot on libtool wrapper preparing the environment that allows (or not) to find the library at run-time search, so the tests are skipped (not failed) when it says the library could not be found.

Also addresses a poorly-understood message on console when "NewNUT" (avahi) discovery fails.

Note: initially hoped this would be a test for sanity-checker from #1811, with dummy-ups configs providing some bogus or duplicate data, but NUT discovery only reports the client/relay connection point, not further details of the already running driver's config or data.

Closes: #1815

### DIFF
--- a/tests/NIT/nit.sh
+++ b/tests/NIT/nit.sh
@@ -1127,36 +1127,38 @@ testcase_sandbox_nutscanner_list() {
     log_info "Call libupsclient test suite: nut-scanner on localhost:${NUT_PORT}"
 
     # NOTE: Currently mask mode is IPv4 only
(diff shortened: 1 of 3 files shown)


jimklimov:
AppVeyor: `Cannot load NUT library (libupsclient.dll) : The specified module could not be found.. NUT search disabled.` ugh :\

PR #1816, opened 2023-01-16
nut-monitor.appdata.xml: add launchable=nut-monitor.desktop

jimklimov:
Inspired by https://salsa.debian.org/debian/nut/-/blob/a009ca10a327d7db383f5d7726091050ceb4d60f/debian/patches/appdata-launchable.patch from @bigon but adjusted for general version-agnostic dispatcher (py2 or py3, whatever is installed for this end-user deployment)


jimklimov:
CC @bigon : debian packaging may need a change to use this (or override still with explicit Py3)

PR #1819, opened 2023-01-17
Use libusb_get_device_address(), not libusb_get_port_number()

efuss:
In USB device enumeration, when setting `curDevice->Device`, use `libusb_get_device_address()`, not `libusb_get_port_number()`.

While there, rename `bus` to `bus_num` to avoid ambiguity.

When `printf()`ing `devnum` (a `size_t`), use `%zd`, which is POSIX, not `PRIuSIZE`, which is not.


jimklimov:
`PRIuSIZE` is set by `nut_stdint.h` according to platform. Not all systems we compile for care about POSIX, alas.

jimklimov:
For that matter, `%zd` is signed and `size_t` is not (`ssize_t` is).

jimklimov:
...and original `%i` for port number also (`PRIu8`)

efuss:
> `PRIuSIZE` is set by `nut_stdint.h` according to platform. Not all systems we compile for care about POSIX, alas.
Ah, sorry, that seems to be new; I'm still on 2.8.0.

efuss:
> For that matter, `%zd` is signed and `size_t` is not (`ssize_t` is).
Stupid me. I intended `%zu` -- I remember having looked up whether `devnum` was `size_t` or `ssize_t` and then typed the wrong thing.

jimklimov:
@efuss : I bumped the PR with printing changes, and with a name-sync of other libusb1 code to minimize diffs later. And updated nut-scanner to use same facility as drivers would.

To me the core question would be if `libusb_get_device_address()` fares better on your platform than `libusb_get_port_number()` (I suppose the latter is prone to return 0 where knowing physical topology is not supported, per your older posted issues?)

efuss:
> To me the core question would be if `libusb_get_device_address()` fares better on your platform than `libusb_get_port_number()` (I suppose the latter is prone to return 0 where knowing physical topology is not supported, per your older posted issues?)
Yes (I'm actually running a patched version using `libusb_get_device_address()` now). The NetBSD implementation of libusb (1.x) doesn't set the port field because no such information is available from the OS's USB interface.

As far as I understand, `libusb_get_bus_number()`/`libusb_get_device_address()` and `libusb_get_port_number()` refer to two distinct name spaces.
It may be worth adding a `Port` field using the value returned by `libusb_get_port_number()` if available, but I can't test that.

jimklimov:
Ok, thanks!

PR #1820, opened 2023-01-18
Improve Best Fortress driver

### LINKED ISSUE #1817: bestfortress shutdown does not respect configured delay - initups not called?
nut from git, Best Fortress LI660 ("old Fortress"), driver bestfortress

Doing "upsdrvctl shutdown" results in a near-immediate power off, instead of the 10 s one would expect from reading the code.  My copy does (20 instead of 10 in git).
```
        dstate_setinfo("ups.delay.shutdown", "20");
```
in `ups_initinfo()`.   `upsc foo` includes `ups.delay.shutdown: 20`, so this does get set.

Then in `upsdrv_shutdown`:
```
        grace = dstate_getinfo("ups.delay.shutdown");

        if (!grace) {
                upsdebugx(1, "upsdrv_shutdown: ups.delay.shutdown is NULL!");
                grace = "1"; /* apparently, OFF0 does not work */
        }
```
and the debug log said it was sending (in angle brackets)
```
<OFF1
>
```
So the bug is either
  * bestfortress is confused and setting that variable doesn't belong in `ups_initinfo`
  * `upsdrvctl shutdown` doesn't lead to calling `ups_initinfo` but it should

Regardless, I'm going to prepare a PR which is far more defensive, but the larger question is interesting and should probably be addressed.

We can close this when shutdown is sane and open a fresh issue once we can say what's wrong.

gdt:
This PR does two main things:
  - rationalize and improve logging, and add/update comments
  - improve shutdown process, fixing a serious bug where the intended 10s delay was not honored, resulting in a way-too-short 1s delay.  Bump driver version because of this.  Fixes #1817.

I have tested a functionally equivalent earlier version on a Fortress LI660 including `upsmom -c fsd`, and this version with `-D -D -D -D -D`.

Changes are entirely in `drivers/bestfortress.c`; interfaces are not changed.
 
Things that I think can be done later (not in this PR):
  - More substantial comment improvements, including about the protocol.
  - Enable setting more parameters (there are 10).
  - Enable setting the shutdown delay (internal to driver; not a UPS parameter).
  - More logging rototilling.
  - Reorder functions to match order of being called from driver/main.c.
  - Cleanup of ssize_t/unsigned/int in various code (this PR does no harm, I believe).

### DIFF
--- a/drivers/bestfortress.c
+++ b/drivers/bestfortress.c
@@ -47,10 +47,29 @@ upsdrv_info_t   upsdrv_info = {
 	{ NULL }
 };
 
+/*
+ * Logging plan:
+ * CRIT: executing poweroff
+ * ERR: indicates a bug in nut
+ * WARNING: indicates a serious misconfiguration that will cause nut
+ *           to fail to function
+ * NOTICE: rare situations without a planned reporting path (none so
+ *         far)
+ * INFO: One message each for startup, first establishment of comms
+ * DEBUG: Loss/restore of comms, and perhaps things that are normally
+ *        reported via other daemons (on battery, low battery)
+ * D1: Things that happen rarely (startup/shutdown, param set, instant
+ *     commands)
+ * D2: ?
+ * D3: trace of communications
+ * D4: more io than necessary to see working communications
+ * D5: log even succesful acks
+ */
+
 static int instcmd (const char *cmdname, const char *extra);
 static int upsdrv_setvar (const char *varname, const char *val);
 
-/* rated VA load if known */
+/* Rated maximum VA output as configured by the user. */
 static int maxload = 0;
 
 void upsdrv_initinfo(void)
@@ -64,6 +83,8 @@ void upsdrv_initinfo(void)
 	if (maxload)
 		dstate_setinfo("ups.load", "0");
 	dstate_setinfo("output.voltamps", "0");
+
+	upsdebugx(1, "bestfortress: arbitrarily setting ups.delay.shutdown to 10");
 	dstate_setinfo("ups.delay.shutdown", "10");	/* write only */
 
 	/* tunable via front panel: (european voltage level)
@@ -197,8 +218,12 @@ static int upssend(const char *fmt,...) {
 #endif
 	va_end(ap);
 
+	/* Why do we not upsflushin here? */
+
+	upsdebugx(3, "upssend: sending %d bytes <%s>", ret, buf);
+
 	if ((ret < 1) || (ret >= (int) sizeof(buf)))
-		upslogx(LOG_WARNING, "ser_send_pace: vsnprintf needed more "
+		upslogx(LOG_ERR, "upssend: vsnprintf needed more "
 				"than %d bytes", (int)sizeof(buf));
 
 	for (p = buf; *p && sent < INT_MAX - 1; p++) {
@@ -224,8 +249,12 @@ static int upssend(const char *fmt,...) {
 
 		sent++;
 		if (sent >= INT_MAX) {
-			upslogx(LOG_WARNING, "ser_send_pace: sent more than INT_MAX, aborting");
+			upslogx(LOG_ERR, "upssend: sent more than INT_MAX, aborting");
 		}
+	}
+
+	if (ret != (int) sent) {
+		upsdebugx(1, "upssend: ret %d != sent %u", ret, sent);
 	}
 
 	return (int)sent;
@@ -233,13 +262,21 @@ static int upssend(const char *fmt,...) {
 
 static ssize_t upsrecv(char *buf,size_t bufsize,char ec,const char *ic)
 {
-	return ser_get_line(upsfd, buf, bufsize - 1, ec, ic,
-	                    SER_WAIT_SEC, SER_WAIT_USEC);
+	ssize_t nread;
+
+	nread = ser_get_line(upsfd, buf, bufsize - 1, ec, ic,
+			     SER_WAIT_SEC, SER_WAIT_USEC);
+
+	/* \todo is buf null terminated? */
+	upsdebugx(4, "upsrecv read %d <%s>", (int) nread, buf);
+
+	return nread;
 }
 
 static ssize_t upsflushin(int f, int verbose, const char *ignset)
 {
 	NUT_UNUSED_VARIABLE(f);
+	upsdebugx(4, "upsflushin: begin");
 	return ser_flush_in(upsfd, ignset, verbose);
 }
 
@@ -255,7 +292,7 @@ void upsdrv_updateinfo(void)
 	char ch;
 	int checksum_ok = -1, is_online = 1, is_off, low_batt, trimming, boosting;
 
-	upsdebugx(1, "upsdrv_updateinfo");
+	upsdebugx(2, "upsdrv_updateinfo: begin");
 
 	for (retry = 0; retry < 5; ++retry) {
 		upsflushin (0, 0, "\r ");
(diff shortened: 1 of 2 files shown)


jimklimov:
Looks promising, thanks!

Keep in mind that logging with `__func__` may be less cumbersome and future-proof to define and maintain, especially as code gets copy-pasted, moved, split, or methods are renamed.

jimklimov:
````
bestfortress.c:535:14: error: unused variable 'p' [-Werror,-Wunused-variable]
        const char *p;
                    ^
bestfortress.c:611:39: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]
        dstate_setinfo("ups.delay.shutdown", shutdown_delay);
                                             ^~~~~~~~~~~~~~
bestfortress.c:611:39: note: treat the string as an argument to avoid this
        dstate_setinfo("ups.delay.shutdown", shutdown_delay);
                                             ^
                                             "%s", 
2 errors generated.
*** Error code 1 (continuing)
`all' not remade because of errors.
````

The latter would be treated well with e.g.:
````
dstate_setinfo("ups.delay.shutdown", "%s", shutdown_delay);
````

Also, please verify (ensure if needed) that if `ups.conf` supports an override setting for this, or some explicit `shutdown_delay` option, etc., then the driver init does not overwrite an existing value. I don't remember OTOH if it does for that driver...

**Comment by gdt** [comment]
The "not a literal" error is fair but didn't happen for me, and those ilnes are just moved.  But will fix as the point of this PR is to generally fix up this driver.  ack func; will switch to that too.

**Comment by gdt** [comment]
bestfortress is documented not to support `ups.delay.shutdown`.   That should be fixed and is on the future work list.

**Comment by gdt** [comment]
There's a warning in the windows build, but it's because the type for file descriptors changes on Windows.  That changing type feels like a bug in the windows port.   I can only think of:
  - table from ints to windows stuff
  - changing ser_foo to use a void * which is really a struct, globally
  - 
My usual bias is not to muck things up for POSIX to accomodate a failure to follow POSIX...

**Comment by jimklimov** [comment]
For the FILE type, try `if (INVALID_FD(upsfd))` - it should take care of platform-dependent issue like this:

````
bestfortress.c: In function 'upsdrv_initups':
bestfortress.c:605:19: warning: ordered comparison of pointer with integer zero [-Wpedantic]
  605 |         if (upsfd < 0) {
````

**Comment by jimklimov** [comment]
Probably should update `developers.txt` about such nuances :\

**Comment by gdt** [comment]
I can't figure out the bigger dynamatrix failure; looks sort of like a test issue, not a real problem.   The details link is behaving badly in firefox; it takes many seconds to display a screen and  then seems to regenerate on scrolling.

Please let me know if something needs to be changed.

**Comment by jimklimov** [comment]
Odd, possibly GitHub kept a verdict from an older build and did not get a notification about the latest. Currently-latest https://ci.networkupstools.org/job/nut/job/nut/job/PR-1820/9 ended up green.

**Comment by gdt** [comment]
OK, let me know if you are ok to merge or you want changes.

File: drivers/bestfortress.c [hunk 1]
```diff
@@ -301,23 +356,29 @@ void upsdrv_updateinfo(void)
...
-		upsdebugx(2, "checksum corruption");
+		upsdebugx(2, "%s: checksum corruption", __func__);
 		upsdebug_hex(3, "buffer", temp, (size_t)len);
 		dstate_datastale();
 		return;
 	}
 
-	/* upslogx(LOG_INFO, "updateinfo: %s", p); */
+	/* Log assuming ASCII which it 99% is.  \todo Improve. */
+	upsdebugx(3, "%s: %s", __func__, p);
```
jimklimov, line 381: Do we surely expect this is a zero-byte terminated string at least, or should impose a sanity check (or some snprintf) above? The "99%" makes me uneasy :)

drivers/bestfortress.c
```diff
@@ -420,42 +499,60 @@ static int upsdrv_setvar (const char *var, const char * data) {
...
 {
 	const	char	*grace;
 
+	upsdebugx(2, "%s: begin", __func__);
+
 	grace = dstate_getinfo("ups.delay.shutdown");
+	if (!grace) {
+		upsdebugx(1, "%s: ups.delay.shutdown is NULL!", __func__);
+		/* Pick a different value than 20 so we can see it in the logs. */
+		grace = "30";
```
jimklimov, line 525: Not sure about losing `grace = "1"; /* apparently, OFF0 does not work */` here for eternity.

Probably the original intent was "if not set, go down immediately" which then got confused by a default setting (20) which comes later in process lifecycle when not `-k`. Maybe it is that default that should go away, eventually, and users would be expected to give a setting in `ups.conf`?..

jimklimov:
So for that CI fault, it somehow was not diagnozed well by the matrix, but was a fault in build 9 after all:
````
[2023-01-22T13:14:24.548Z] ERROR: Error cloning remote repo 'origin'

[2023-01-22T13:14:24.548Z] hudson.plugins.git.GitException: Command "git fetch --tags --force --progress -- https://github.com/networkupstools/nut.git +refs/pull/1820/head:refs/remotes/origin/PR-1820 +refs/heads/master:refs/remotes/origin/master" executed in workdir "/home/abuild/jenkins-nutci-debian-11-amd64/workspace/nut_nut_PR-1820" returned status code 128:

[2023-01-22T13:14:24.548Z] stdout: 

[2023-01-22T13:14:24.548Z] stderr: fatal: unable to access 'https://github.com/networkupstools/nut.git/': Failed to connect to github.com port 443: No route to host
````

Posted some more questions/concerns, and a semi-cosmetic fix commit.

File: drivers/bestfortress.c (hunk 2 above)
Comment (gdt) [reply drivers/bestfortress.c:525]: I moved the comment about OFF0 not working to higher up where I explain the protocol.   That's where it belongs, rather than as a side comment about something else.

Now, the code uses the configured delay because I fixed the ordering bug you found for me.  The 30 is just  a "if the impossible try to cope" and I picked 30 vs 20 so that it might be noticeable.

IMHO, it *never* made sense to use 1 when the variable was empty.  That's essentially always too hasty.

File: drivers/bestfortress.c (hunk 1 above)
Comment (gdt) [reply drivers/bestfortress.c:381]: It is zero-terminated and it is ASCII every time I look at it.  The worst I'd expect is an embedded \r.
But, I want to later add escapedascii to do this right in general.  I can reword the comment.

Comment (gdt) [reply drivers/bestfortress.c:381]: Also I can force-terminate the string to take that concern off the table.

File: drivers/bestfortress.c (hunk 2 above)
Comment (jimklimov) [reply drivers/bestfortress.c:525]: Ok, makes sense. Just making sure it is a conscious decision :)

File: drivers/bestfortress.c (hunk 1 above)
Comment (jimklimov) [reply drivers/bestfortress.c:381]: If you can force it (or if some `calloc()` already does), could be nice to be on the safe side.

Comment (jimklimov) [reply drivers/bestfortress.c:381]: So after some looking around, I guess it can be assumed zero-terminated, with `temp[2]==0` and/or `strlen(temp+2)` not being an infinite loop :) and `p` being either `temp` or `temp+2`.

File: drivers/bestfortress.c [hunk 3]
```diff
@@ -500,14 +602,25 @@ void upsdrv_initups(void)
...
+		/* \todo: Deal with the failure */
+	}
+
+	/* ser_set_speed returns int 0 always; fatal if ioctl fails */
 	ser_set_speed(upsfd, device_path, speed);
-	/* TODO: probe ups type */
 
-	/* the upsh handlers can't be done here, as they get initialized
-	 * shortly after upsdrv_initups returns to main.
-	 */
```
jimklimov, line 508: Should these comments have disappeared? Do they no longer apply?

jimklimov:
@gdt: I guess one question remains (posted above, about some comments gone missing). I've also updated this branch with recent `master` state so you have a chance to test together all the replies to your different concerns discussed recently :D and in a way that it will end up published.

drivers/bestfortress.c
gdt: I removed them on purpose because I don't think they make sense.   The docs define what each function is supposed to do and explaining that we don't do something not documented doesn't seem sensible to me.  These things are set in initinfo which seems like the right place.   I don't want to add comments for why we aren't setting everything else that is set in initinfo.  Is that ok?

jimklimov: Sounds reasonable from this PoV, thanks :)

gdt:
I have rebased and am going to run live tests over the next ~day, making a distfile and building that in pkgsrc and installing it.  I'll comment here when that is ok, and sounds like you are then reading to hit merge. ( My preference is to merge without squashing -- I tried to keep commits separated logically -- and it looks like that's how you do things.  I didn't try to rebase the PRI changes since it seemed ilke a bunch of work for little gained clarity for some future diff reader.)

gdt:
It's been stable overnight.  I'll certainly notice any future issues, and there's more to fix anyway.  So I'm good with you merging.

PR #1823, opened 2023-01-18
scripts/python/*: import definition and use-case of CheckUPSAvailable()…

jimklimov:
…from Fedora packaging patches:

* Py2: https://src.fedoraproject.org/rpms/nut/blob/3c9fea3a359db592e3c7d25ff5c8050325b21dfe/f/nut-2.6.5-unreachable.patch
* Py3: https://src.fedoraproject.org/rpms/nut/blob/48685366900985696be8371d8deea6494ae36c43/f/nut-2.6.5-unreachable.patch

originally by Michal Hlavinka @mhlavink


jimklimov:
Did not test the GUI app change, so at the moment can only hope there are no fatal py2/py3 syntax nuances (e.g. involving ascii encoding or not)

jimklimov:
Verified working for both pythons with respective NUT-Monitor UI implementations, and a driver defined but not started, so `upsc` says this:
````
# upsc dummy
Error: Driver not connected
````

PR #1832, opened 2023-01-22
tools/Makefile.am: Add missing dependency

### LINKED ISSUE #1831: build of `scripts/hotplug/libhid.usermap` is unsound (`make -j2` fails, confusion between autogen time and make time)
NetBSD 9, system make, recent git (actually bestfortress PR).   Building with -j should enable parallel builds, but apparently there is a missing dependency.
```
--- upscode2 ---
libtool: link: gcc -I../../include -I/usr/pkg/include/libusb-1.0 -I/usr/pkg/include/neon -isystem /usr/pkg/include -g -O2 -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Wno-system-headers -Wall -Wextra -Wsign-compare -pedantic -Wno-error -Wl,-R/usr/pkg/lib -o upscode2 upscode2.o  -L/usr/pkg/lib ./.libs/libdummy_serial.a ./.libs/libdummy.a ../common/.libs/libcommon.a ../common/.libs/libparseconf.a -lm -lpthread
--- victronups ---
libtool: link: gcc -I../../include -I/usr/pkg/include/libusb-1.0 -I/usr/pkg/include/neon -isystem /usr/pkg/include -g -O2 -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Wno-system-headers -Wall -Wextra -Wsign-compare -pedantic -Wno-error -Wl,-R/usr/pkg/lib -o victronups victronups.o  -L/usr/pkg/lib ./.libs/libdummy_serial.a ./.libs/libdummy.a ../common/.libs/libcommon.a ../common/.libs/libparseconf.a -lpthread
Making all in tools
--- all-recursive ---
make[1]: make[1]: don't know how to make ../scripts/hotplug/libhid.usermap. Stop

make[1]: stopped in /home/n0/gdt/SOFTWARE/IOT/nut/BUILD/tools
*** [all-recursive] Error code 1
```

The problem is likely in tools/Makefile.am.

The build seems ok with GNU make, but that's not documented to be necessary (and isn't without -j) and it's  a bug to need GNU make anyway.

I'll see if I can figure this out.

gdt:
The files in GENERATED_USB_OS_FILES are created by nut-usbinfo.pl, but this was not expressed.

Fixes #1831, a make -j2 failure.  My guess is that because serial order requires GENERATED_USB_FILES first, that this rule typically fired before the GENERATED_USB_OS_FILES rules fired, avoiding the problem.  I have no theory why it worked with GNU make.

### DIFF
--- a/tools/Makefile.am
+++ b/tools/Makefile.am
@@ -60,7 +60,7 @@ $(GENERATED_SNMP_FILES): $(top_srcdir)/drivers/*-mib.c
 		echo "----------------------------------------------------------------------"; \
 	fi
 
-$(GENERATED_USB_FILES): $(top_srcdir)/drivers/*-hid.c $(top_srcdir)/drivers/*usb*.c $(top_srcdir)/drivers/nutdrv_qx.c
+$(GENERATED_USB_FILES) $(GENERATED_USB_OS_FILES): $(top_srcdir)/drivers/*-hid.c $(top_srcdir)/drivers/*usb*.c $(top_srcdir)/drivers/nutdrv_qx.c
 	@if perl -e 1; then \
 		echo "Regenerating the USB helper files in SRC dir."; \
 		TOP_SRCDIR="$(top_srcdir)" ; export TOP_SRCDIR; \


jimklimov:
FWIW, there were a few changes to the perl script during the past years, but mostly syntactic - not "functional" like adding new generated files: https://github.com/networkupstools/nut/commits/master/tools/nut-usbinfo.pl

Sometimes parallel makes have a sort of raci-ness between threads, so depending on CPU counts and other factors, issues can be exposed by running more and more `make -j X` threads, especially odd ones with an even amount of CPUs. At least this approach helped with finding earlier issues. Two threads however were not an expected problem for a long time :)

jimklimov:
Actually the core problem *here* seems to be different (even if your fix helps band-aid around it): I think the `../scripts/hotplug/libhid.usermap` (and some others) is generated by same perl script when you `autogen.sh` originally, and since for the most-general case NUT release tarballs should not use as many dependencies as builds from scratch do (in same vein as not requiring autoconf), these build products should be `EXTRA_DIST`'ed but this one was not.

Maybe that is the bit that somehow implicitly differed for non-GNU makes after generation with GNU automake/autoconf suite, and they added an implicit addition to the list here but not there. At least, `make distcheck` purpose is specifically to weed out such things, and it succeeds on a lot of platforms and make implementations, with parallelism in play too.

Am I right that the `make -j2` failed for your second build - from the "dist tarball", NOT for the initial build which prepared the tarball?

jimklimov:
FWIW, what versions of autoconf and automake are involved? I suspect that files used as templates from `configure.ac` (which `libhid.usermap` is not, but many of the generated others are) are `dist`'ed automatically. Maybe starting from certain toolkit versions though...

I'll take a blanket approach here to `EXTRA_DIST` everything that `git status --ignored` currently highlights after `make maintainer-clean; ./autogen.sh`...

gdt:
``
$ autoconf --version
autoconf (GNU Autoconf) 2.71
$ automake --version
automake (GNU automake) 1.16.5
```
I think these are pretty recent, and nowhere near too old to care about.

I got the same failure when building from the tarball I generated.

I don't follow it being generated by autogen, but I see autogen does more than is normal.  But:
```
if [ ! -f scripts/udev/nut-usbups.rules.in -o \
     ! -f scripts/devd/nut-usb.conf.in ]
then
        if perl -e 1; then
                echo "Regenerating the USB helper files..."
                cd tools && {
                        ./nut-usbinfo.pl || exit 1
                        cd ..
```

Regardless, I think we need a clear plan of autogen/dist vs build time as there is the srcdir/objdir difference.

jimklimov:
> I got the same failure when building from the tarball I generated.

Given PR #1834 this "was suddenly expected" :\

Did it misbehave for initial build however? (git checkout, autogen, configure, make all)?

gdt:
Yes, the initial autogen/configure/make failed.

PR #1834, opened 2023-01-22
Explicitly distribute all files generated by `autogen.sh` into dist tarballs

jimklimov:
Could be behind the problems in issue #1831 although still strange (probably toolkits differ in what rules they generate and/or how they are handled).

CC @gdt : can you please test this branch without the recipe change you proposed per #1832? Does this fix alone address the problem in practice?

UPDATE:
Closes: #1832
Closes: #1831


gdt:
If something is needed for a correct distfile, then make dist should fail if configured without.  I mean the conditonal EXTRA_DIST based on hotplug.    And, if it just takes perl to make it, then it shouldn't be conditional on hotplug which is about use??

jimklimov:
Probably can be done unconditionally like in other files with more complex rules, but here a simple `if/else` is sufficient - either the file is explicitly part of build or install target (which should `dist` it), or it is explicitly dist'ed.

gdt:
What I meant is if you try to make dist on  a machine that shouldn't build hotplug, these files need to be present in the distfile anyway, because it could be used on some other system.   Basically, assuming same autoconf/automake, the distfile should be exactly the same (modulo timestamps etc.) no matter what machine it was built on.    So if perl needs to be run to make these, then it should just be an error at autogen time if not found.

gdt:
Running autogen (with perl present :-), `make -j6` succeeded (where make is BSD make).

I see this as a different workaround for incorrect make rules.   If the perl script is run at autogen time it should be removed from being run by make.  Or the other way around, making perl a build dep.   Or better yet converted to python which already is.

jimklimov:
> What I meant is if you try to make dist on a machine that shouldn't build hotplug, these files need to be present in the distfile anyway, because it could be used on some other system.

big yes on that :)

so really wondering how this was not caught before (with the file in question being part of `make all` scope, I suppose), and if it is a `qmake`-specific deficiency of sorts, or just was not triggered by others?..

gdt:
I wonder if the old code didn't generate these because it was conditioned on hotplug.  But the script makes things that are used for other than hotplug, I think.

gdt:
I think the bug needed  a combination of all of
  - autogen on system w/o hotplug
  - j2
  - BSD make
  - perhaps luck since -j bugs are nondeterministic

and that's just rare.

jimklimov:
Maybe... if tarballs happened to be made on systems that tried to install or otherwise reference these files, they got dist'ed regardless. But still, all `make all`'s should have requested and so got them, during `make distcheck`.

Chances remain that a `distcheck` build requesting the other generated files for nut-scanner got these present (at least, or completed at best).

One bit I remain uneasy about, especially for parallel builds, is that many build products depending on the same recipe (which generates all at once) can end up running the perl script several times in parallel to write them over and over. Lucky if consistently and not appending to same inode. A `NOPARALLEL` section is IIRC not a portable-POSIX concept?.. For some targets concerns like these were gated by either using dummy/flag files to run a tool once (all these files need the touch-file, touch-file needs the original SNMP/USB deps and runs the perl script, and probably should be dist'ed - but in this particular case one of the "real" files can be used in this role), or by chaining dependencies artificially without a real dependency between them.

jimklimov:
Except for the `hwdb` file, all others were in tarballs created by `make dist` on OpenBSD 6.5, recent Linux, and even Solaris 8.

gdt:
I see that the perl script is run at configure time.  I wonder if it is also necessary to do an objdir build to trigger this?  Is there confusion between source vs object in terms of where they live in the makefile?

gdt:
> One bit I remain uneasy about, especially for parallel builds, is that many build products depending on the same recipe (which generates all at once) can end up running the perl script several times in parallel to write them over and over. Lucky if consistently and not appending to same inode. A `NOPARALLEL` section is IIRC not a portable-POSIX concept?.. For some targets concerns like these were gated by either using dummy/flag files to run a tool once (all these files need the touch-file, touch-file needs the original SNMP/USB deps and runs the perl script, and probably should be dist'ed - but in this particular case one of the "real" files can be used in this role), or by chaining dependencies artificially without a real dependency between them.

NetBSD make has `.NOTPARALLEL`.  But I do no assert that it is truly portable or required by POSIX.

I would think that this script should be invoked by exactly one rule, and that there shouldn't be recursive make invocations that can run make twice at once in any directory.

jimklimov:
> I see that the perl script is run at configure time

If you mean the configure script - it should not be, at least not explicitly and intentionally (`git grep -i perl *.ac '*.m4'` is essentially empty, also for `pl` hits). The `autogen.sh` run is not a "configure time" for this context, to be clear.

Unfortunately it was not easy to make `autogen.sh` less loaded (it is indeed more complex than in most projects), because it provides certain `.in` templates that `configure.ac` generation into a shell script requires to be available, and which in turn are effectively built products dependent on other sources so not "kosher" to keep in Git... And as mentioned, "configure time" alone (for builds from tarballs) has fewer dependencies than a build from Git.

BTW, beside the files argued about above, one more big difference of tarballs vs. git is the docs -- e.g. man pages are pre-built and embedded into dist tarballs, so systems lacking `asciidoc` (and much of X11 that its dependencies might pull in) can still package NUT from a dist tarball.

gdt:
Sorry, I meant perl run at autogen time.   I'm fine with doing it at autogen time.   What I don't like is it being optional - I think autogen should just hard require tools it needs, because nobody should be able to make a bad distfile (without really trying).  perl and python are not a big deal for people building from git.

jimklimov:
> autogen should just hard require tools it needs

It does abort when python/perl are absent or not usable (bail-outs for developers are however possible with envvars it suggests, to make partially functional builds feasible even on "incomplete" platforms, by touching empty files), or fail to generate the files. It would not be suitable for general-purpose dist tarballs, but often they are not the goal of building from scratch. Think embedded systems that need to use busybox as system shell, to get an `upsmon` and do not care for USB/SNMP drivers per se, etc.

jimklimov:
@gdt : can you please check again if the current state of this PR (without yours) still works for both your "initial" and "dist" builds, and generally looks more acceptable? (Disregarding the matter of possibly pre-building same files as part of `autogen.sh` and later re-built with `make` runs, at least as developers iterate?)

gdt:
```
Making install in data
--- install-recursive ---
Making install in html
--- install-am ---
--- install-dist_htmlDATA ---
--- install-nodist_htmlDATA ---
--- install-am ---
--- install-dist_dataDATA ---
--- install-nodist_dataDATA ---
--- install-dist_dataDATA ---
 ../.././install-sh -c -d '/home/gdt/SOFTWARE/IOT/nut/DESTDIR/usr/pkg/share'
--- install-nodist_dataDATA ---
 ../.././install-sh -c -d '/home/gdt/SOFTWARE/IOT/nut/DESTDIR/usr/pkg/share'
--- install-dist_dataDATA ---
mkdir: /home/gdt/SOFTWARE/IOT/nut/DESTDIR/usr/pkg/share: File exists
*** [install-dist_dataDATA] Error code 1

```

should be mkdir -p, probably (and I need to page out nut for a day, ish)

jimklimov:
Probably... but `install-sh` is part of autotools and so used by its generated rules. Black box as far as we are concerned or can impact...

jimklimov:
Got a NetBSD 9.2 VM from earlier NUT CI tests, spun it up. Can't find a `qmake` locally not as a package by such name (pointers for `pkg_add` or `pkgin install` are welcome); googling suggests it could be Qt Make?

Anyhow, reproduced your latest problem with parallel installation and directories with the available `make` which does not tell its name, but `strings` show a lot of `$NetBSD` RSC tags in it, as well as with `gmake`. The issue is indeed in `install-sh` delivered by the system (RCS-tagged `$NetBSD: install-sh.in,v 1.6 2012/01/11...`) and its `mkdirprog="${MKDIRPROG-mkdir}"` is used without any `-p` arguments. Similar scripts I have on e.g. Linux and other builders seem very loosely related (varnames match but everything looks different, possibly decades of coding apart :) )

Still, running `rm -rf /tmp/nut; make DESTDIR=/tmp/nut MKDIRPROG="mkdir -p" install -j 5` succeeded several times in a row; and without `MKDIRPROG` failed every time.

jimklimov:
Somehow these attempts upset many builders. Needs more attention then...

gdt:
> qmake



> Got a NetBSD 9.2 VM from earlier NUT CI tests, spun it up. Can't find a `qmake` locally not as a package by such name (pointers for `pkg_add` or `pkgin install` are welcome); googling suggests it could be Qt Make?

I don't follow about `qmake`.  That doesn't make sense and I'm not using it.   I'm talking about the BSD make that is part of NetBSD (which is /usr/bin/make) and GNU make (which is /usr/pkg/bin/gmake).

Sounds ilke the autoconf mkdir to use -p is working.  (The -p option is required by POSIX.)

jimklimov:
Maybe was a typo then - #1831 title says "build of scripts/hotplug/libhid.usermap is unsound (qmake -j2` fails, confusion between autogen time and make time)" so I went looking for it. Looking at the dangling backtick now, probably "q" was a typo ;)

In that case, no big deal - and the problem (or solution) is more reproducible on my side to match your situation :)

jimklimov:
On NetBSD:
````
...
* CPPFLAGS      :
* CONFIG_FLAGS  :
configure: WARNING: =====================================================
configure: WARNING: Your system provided (or NUT tarball included) an
configure: WARNING: 'install-sh' implementation which is not safe for
configure: WARNING: parallel installs; export MKDIRPROG='mkdir -p'
configure: WARNING: may help, otherwise run 'make install' sequentially.
configure: WARNING: This should not impact parallel builds.
configure: WARNING: =====================================================
````

gdt:
I see what you mean about install-sh as part of pkgsrc automake.  That's a bug on our end, apparently to accomodate Interix, but it seems time to gc that support.  I'll see about that.

I just tested master post merge, and it builds/installs with -j6, so that's great.

PR #1855, opened 2023-02-22
Adding GPIO interface

modrisb:
Attempt to add support for UPSes using only open collector interface to report device state by connecting open collector pins to GPIO lines. Pins statuses are transformed to NUT states by configuration rules and is is device independent. Testing done on 
CyberPower CyberShield CSN27U12V connected to Orange Pi Zero. Implementation uses libgpiod library to read GPIO lines statuses and does not change any other NUT components except build related.

## General points

- [X] Described the changes in the PR submission or a separate issue, e.g.

- [X] There may be multiple commits in the PR, aligned and commented with
  comments or text documents.

- [X] Revised existing driver families and added a sub-driver if applicable
  case.

- [X] Did not extend obsoleted drivers with new hardware support features

- [X] For updated existing device drivers, bumped the `DRIVER_VERSION` macro
  or its equivalent.

- [X] Proposed NUT data mapping is aligned with existing `docs/nut-names.txt`
  concept.

- [X] Did not "blindly assume" default integer type sizes and value ranges,

- [X] Progress and errors are handled with `upsdebugx()`, `upslogx()`,

- [X] Coding style (including whitespace for indentations) follows precedent

- [X] For newly added files, the `Makefile.am` recipes were updated and the

- [X] Added or updated manual page information in `docs/man/*.txt` files

  the changed codebase.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1170,9 +1170,16 @@ AC_ARG_WITH(drivers,
 		if test -z "${with_powerman}"; then with_powerman="yes"; fi
 		if test -z "${with_modbus}"; then with_modbus="yes"; fi
 		if test -z "${with_ipmi}"; then with_ipmi="yes"; fi
-		if test -z "${with_gpio}"; then with_gpio="yes"; fi
 
 		dnl Platform-dependent snowflakes that are required or auto:
+		if test -z "${with_gpio}"; then
+			dnl NOTE: Currently we only support a Linux libgpiod
+			dnl backend for GPIO; eventually there could be more.
+			case ${target_os} in
+				linux*) with_gpio="yes";;
+				*) with_gpio="auto";;
+			esac
+		fi
 		if test -z "${with_linux_i2c}"; then
 			case ${target_os} in
 				linux*) with_linux_i2c="yes";;
@@ -1378,9 +1385,17 @@ AC_ARG_WITH(all,
 		if test -z "${with_powerman}"; then with_powerman="${withval}"; fi
 		if test -z "${with_modbus}"; then with_modbus="${withval}"; fi
 		if test -z "${with_ipmi}"; then with_ipmi="${withval}"; fi
-		if test -z "${with_gpio}"; then with_gpio="${withval}"; fi
 
 		dnl Platform-dependent snowflakes that are required or auto:
+		if test -z "${with_gpio}"; then
+			dnl NOTE: Currently we only support a Linux libgpiod
+			dnl backend for GPIO; eventually there could be more.
+			with_gpio="${withval}"
+			case ${target_os} in
+				linux*) ;;
+				*) test x"${withval}" = xno || with_gpio="auto" ;;
+			esac
+		fi
 		if test -z "${with_linux_i2c}"; then
 			with_linux_i2c="${withval}"
 			case ${target_os} in
--- a/configure.ac
+++ b/configure.ac
@@ -1541,6 +1541,8 @@ nut_gpio_lib=""
 
 NUT_ARG_WITH([gpio], [build and install GPIO driver], [auto])
 
+NUT_CHECK_LIBGPIO
+
 dnl ${nut_with_gpio}: any value except "yes" or "no" is treated as "auto".
 if test "${nut_with_gpio}" != "yes"; then
       nut_with_gpio="no"
@@ -3641,6 +3643,8 @@ AC_SUBST(LIBMODBUS_CFLAGS)
 AC_SUBST(LIBMODBUS_LIBS)
 AC_SUBST(LIBIPMI_CFLAGS)
 AC_SUBST(LIBIPMI_LIBS)
+AC_SUBST(LIBGPIO_CFLAGS)
+AC_SUBST(LIBGPIO_LIBS)
 AC_SUBST(LIBI2C_LIBS)
 AC_SUBST(DOC_BUILD_LIST)
 AC_SUBST(DOC_CHECK_LIST)
--- a/configure.ac
+++ b/configure.ac
@@ -1501,6 +1501,10 @@ NUT_CHECK_LIBPOWERMAN
 NUT_ARG_WITH([modbus], [build and install modbus drivers], [auto])
 NUT_CHECK_LIBMODBUS
 
+NUT_ARG_WITH([gpio], [build and install GPIO driver], [auto])
+nut_gpio_lib=""
+NUT_CHECK_LIBGPIO
+
 NUT_ARG_WITH([avahi], [build and install Avahi support], [auto])
 NUT_CHECK_LIBAVAHI
 
@@ -1745,11 +1749,6 @@ dnl AM_CONDITIONAL(WITH_OPENIPMI, test "${nut_with_openipmi}" = "yes")
 dnl ----------------------------------------------------------------------
 dnl Check for with-gpio
 
-nut_gpio_lib=""
-
-NUT_ARG_WITH([gpio], [build and install GPIO driver], [auto])
-
-NUT_CHECK_LIBGPIO
 
 dnl ${nut_with_gpio}: any value except "yes" or "no" is treated as "auto".
 if test "${nut_with_gpio}" != "yes"; then
--- a/configure.ac
+++ b/configure.ac
@@ -1507,6 +1507,16 @@ NUT_CHECK_LIBGPIO
 NUT_ARG_WITH([avahi], [build and install Avahi support], [auto])
 NUT_CHECK_LIBAVAHI
 
+NUT_ARG_WITH([ipmi], [build and install IPMI PSU driver], [auto])
+NUT_ARG_WITH([freeipmi], [enable IPMI support using FreeIPMI], [auto])
+dnl NUT_ARG_WITH([openipmi], [enable IPMI support using OpenIPMI], [auto])
+
+dnl Platform-dependent drivers, currently their detection code is directly
+dnl spelled out in configure.ac
+NUT_ARG_WITH([macosx_ups], [build and install Mac OS X Power Sources meta-driver], [auto])
+
+NUT_ARG_WITH([linux_i2c], [build and install i2c drivers], [auto])
+
 dnl A Python GUI client application for the sysadmin desktop
 dnl (not necessarily on the NUT server itself):
 NUT_ARG_WITH([nut_monitor], [install the NUT-Monitor GUI files], [auto])
@@ -1685,10 +1695,6 @@ dnl if both are available (since it is the only one supported ATM!!)
 
 nut_ipmi_lib=""
 
-NUT_ARG_WITH([ipmi], [build and install IPMI PSU driver], [auto])
-NUT_ARG_WITH([freeipmi], [enable IPMI support using FreeIPMI], [auto])
-dnl NUT_ARG_WITH([openipmi], [enable IPMI support using OpenIPMI], [auto])
-
 dnl ${nut_with_ipmi}: any value except "yes" or "no" is treated as "auto".
 if test "${nut_with_ipmi}" != "no"; then
    dnl check if FreeIPMI (and maybe later OpenIPMI) was explicitly requested
@@ -1766,7 +1772,6 @@ dnl the internal USB UPS driver.
 dnl
 dnl FIXME: be slightly more clever here:
 
-NUT_ARG_WITH([macosx_ups], [build and install Mac OS X Power Sources meta-driver], [auto])
 if test "${nut_with_macosx_ups}" != no; then
    if test -d /System/Library/Frameworks/IOKit.framework/ ; then
       nut_with_macosx_ups="yes"
@@ -1785,7 +1790,6 @@ NUT_REPORT_FEATURE([build Mac OS X meta-driver],
 dnl ----------------------------------------------------------------------
 dnl checks related to --with_linux_i2c
 dnl Check for i2c header on Linux, used for ASEM UPS driver
-NUT_ARG_WITH([linux_i2c], [build and install i2c drivers], [auto])
 LIBI2C_LIBS=""
 if test "${nut_with_linux_i2c}" != no; then
     case ${target_os} in
--- a/configure.ac
+++ b/configure.ac
@@ -1176,7 +1176,7 @@ AC_ARG_WITH(drivers,
 			dnl NOTE: Currently we only support a Linux libgpiod
 			dnl backend for GPIO; eventually there could be more.
 			case ${target_os} in
-				linux*) with_gpio="yes";;
+				linux*) with_gpio="auto";; dnl # TODO: Detect 2018+ distros?
 				*) with_gpio="auto";;
 			esac
 		fi
@@ -1371,7 +1371,7 @@ dnl check for --with-all (or --without-all, or --with-all=auto) flag
 
 AC_MSG_CHECKING(for --with-all)
 AC_ARG_WITH(all,
-	AS_HELP_STRING([--with-all], [enable serial, usb, snmp, neon, ipmi, powerman, modbus, linux_i2c (on Linux), macosx-ups (on MacOS), cgi, dev, avahi, nut-scanner, pynut]),
+	AS_HELP_STRING([--with-all], [enable serial, usb, snmp, neon, ipmi, powerman, modbus, gpio (currently on Linux released after ~2018), linux_i2c (on Linux), macosx-ups (on MacOS), cgi, dev, avahi, nut-scanner, pynut]),
 [
 	if test -n "${withval}"; then
 		dnl Note: we allow "no" as a positive value, because
--- a/configure.ac
+++ b/configure.ac
@@ -12,8 +12,17 @@ dnl primarily for better messaging in the script itself.
 dnl If the NUT codebase in this workspace is being developed and rebuilt
 dnl without reconfiguration, detailed version in the binaries will differ.
 (command -v git >/dev/null 2>/dev/null) \
-&& GITREV="`git describe --tags 2>/dev/null | sed -e 's/^v\([0-9]\)/\1/' -e 's,^.*/,,'`" \
-|| GITREV=""
+&& NUT_SOURCE_GITREV="`git describe --tags 2>/dev/null | sed -e 's/^v\([0-9]\)/\1/' -e 's,^.*/,,'`" \
+|| NUT_SOURCE_GITREV=""
+
+dnl Note: except for experiments, do not pass this into config.h - use
+dnl the NUT_VERSION_MACRO from nut_version.h instead. Developers may
+dnl want to disable their "force-nut-version-header" configure option
+dnl to reduce rebuild scope and so speed up the iterations with ccache.
+dnl New "config.h" contents for every reconfig would defeat that purpose!
+dnl ### AS_IF([test -n "${NUT_SOURCE_GITREV-}"],
+dnl ###     [AC_DEFINE_UNQUOTED([NUT_SOURCE_GITREV], ["${NUT_SOURCE_GITREV}"], [NUT source revision when the build was configured])],
+dnl ###     [AC_DEFINE([NUT_SOURCE_GITREV], [NULL], [NUT source revision when the build was configured: not detected])])
 
 dnl Keep track of command-line options passed to this script:
 AC_MSG_CHECKING([for CONFIG_FLAGS])
@@ -37,15 +46,16 @@ for F in "$@" ; do
 done
 dnl Keep it "known" that we tried to re-use at least some local configuration
 dnl so unspecified options are not necessarily at default values for this script
+AS_IF([test x"${NUT_VERSION_DEPLOYED-}" = x"<reenter>"], [_flag_enable_inplace_runtime="reenter"])
 F=""
-AS_CASE(["${_flag_enable_inplace_runtime}"],
-    [no], [ dnl Explicitly disabled... did we re-enter? Save the value to be seen:
-        AS_IF([test x"${NUT_VERSION_DEPLOYED-}" != x], [
-            F="--enable-inplace-runtime='${NUT_VERSION_DEPLOYED}'"
-        ])
-    ],
-    [""], [],
-    [F="${_flag_enable_inplace_runtime}"])
+AS_IF([test x"${NUT_VERSION_DEPLOYED-}" != x && test x"${NUT_VERSION_DEPLOYED-}" != x"<reenter>"], [
+     dnl Even if explicitly disabled... did we re-enter? Save the value to be seen:
+     F="--enable-inplace-runtime='${NUT_VERSION_DEPLOYED}'"
+    ],[
+     AS_CASE(["${_flag_enable_inplace_runtime}"],
+        [no|""|reenter], [],
+        [F="${_flag_enable_inplace_runtime}"])
+    ])
 AS_IF([test x"$F" != x], [
     test -z "${CONFIG_FLAGS}" \
     && CONFIG_FLAGS="$F" \
@@ -62,9 +72,9 @@ AC_SUBST(CONFIG_FLAGS)
 AC_CONFIG_AUX_DIR([.])
 AC_CONFIG_SRCDIR(server/upsd.c)
 AC_CONFIG_MACRO_DIR([m4])
-AS_IF([test x"${GITREV}" = x],
+AS_IF([test x"${NUT_SOURCE_GITREV}" = x],
     [echo "Network UPS Tools version ${PACKAGE_VERSION}"],
-    [echo "Network UPS Tools version ${PACKAGE_VERSION} (${GITREV})"])
+    [echo "Network UPS Tools version ${PACKAGE_VERSION} (${NUT_SOURCE_GITREV})"])
 AC_CANONICAL_TARGET
 NUT_CHECK_OS
 NUT_STASH_WARNINGS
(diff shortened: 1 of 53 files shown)


docs/man/gpio.txt
```diff
@@ -0,0 +1,118 @@
...
+------------
+
+This driver is not built by default. You can build it by installing
+libgpiod and running `configure --with-gpio=yes`.
+
+You also need to give proper permissions on the local serial device
+file (/dev/gpiochip0 for example) to allow the run-time NUT driver user
+account to access it, like by adding the following rule to rules.d directory:
+SUBSYSTEM=="gpio*", PROGRAM="/bin/sh -c '\
+        chown -R nut:nut /dev/gpiochip0 && chmod -R 700 /dev/gpiochip0\
```
jimklimov, line 99: This code example should be indented, and possibly preceded by a blank line.

docs/man/gpio.txt
```diff
@@ -0,0 +1,118 @@
...
+~~~~~~~~~~~~~~
+
+*rules*='value'::
+A string consisting of sub-strings. Each sub-string describes GPIO line
+states conversion formula to specific NUT state, like
+nut_state=[^]line_num[logical_operation.line_num]. Not (^) , and (&) , or
+(|)operations are supported for now. nut_state should correspond to NUT
+state, line_num to GPIO line number connected to UPS open collector pin.
+CyberShield CSN27U12V describes pins as:
+ON BATTERY			Low when operating from utility line
```
jimklimov, line 47: I think asciidoc table markup (with pipe characters) would do well here, and this may need blank lines around to differentiate from other paragraphs. If possible, build with HTML docs enabled to iterate the result-oriented markup more easily (or use an IDE with an asciidoc plugin to render it real-time as you type) :)

docs/man/Makefile.am
```diff
@@ -669,6 +669,17 @@ endif
 
 HTML_LINUX_I2C_MANS = asem.html pijuice.html
 
+SRC_LINUX_GPIO_PAGES = gpio.txt
```
jimklimov, line 672: Just in case: is it right to call this a LINUX_GPIO? The currently used libgpiod "backend" does say it is for linux, but original discussion implied this driver could be more generic with other platforms logically supporting the interface.

configure.ac
```diff
@@ -1526,6 +1534,20 @@ dnl and AM_CONDITIONALs (see below)...
...
 dnl AM_CONDITIONAL(WITH_OPENIPMI, test "${nut_with_openipmi}" = "yes")
 
+dnl ----------------------------------------------------------------------
+dnl Check for with-gpio
+
+nut_gpio_lib=""
+
+NUT_ARG_WITH([gpio], [build and install GPIO driver], [auto])
+
+dnl ${nut_with_gpio}: any value except "yes" or "no" is treated as "auto".
```
jimklimov, line 1544: I think somewhere here it should check for libgpiod (potentially other equivalents later). 

Typically the check is done as a new method under `m4` directory (copy and adapt an existing file), setting the `LIBGPIO_CFLAGS`, `LIBGPIO_LDFLAGS`, etc. as appropriate, manually or from `pkg-config` data where available.

The `configure.ac` code calls such method, and if things are found and requested mode was "auto" it becomes "yes", otherwise "no". If the request was "yes" but library was not found, it should cause a fatal `AC_MSG_ERROR`.

drivers/Makefile.am
```diff
@@ -56,12 +59,13 @@ MODBUS_DRIVERLIST = phoenixcontact_modbus generic_modbus huawei-ups2000 socomec_
...
 POWERMAN_DRIVERLIST = powerman-pdu
 IPMI_DRIVERLIST = nut-ipmipsu
+GPIO_DRIVERLIST = generic_gpio_libgpiod
 
 # distribute all drivers, even ones that are not built by default
 EXTRA_PROGRAMS  = $(SERIAL_DRIVERLIST) $(USB_DRIVERLIST) $(SERIAL_USB_DRIVERLIST)
 EXTRA_PROGRAMS += $(SNMP_DRIVERLIST) $(NEONXML_DRIVERLIST) $(MACOSX_DRIVERLIST)
 EXTRA_PROGRAMS += $(LINUX_I2C_DRIVERLIST)
-EXTRA_PROGRAMS += $(NUTSW_DRIVERLIST)
+EXTRA_PROGRAMS += $(GPIO_DRIVERLIST)
```
jimklimov, line 68: This was supposed to add a new line, not forget about software-only drivers, right? ;)

drivers/generic_gpio_common.c
```diff
@@ -0,0 +1,444 @@
...
+	* docs/developers.txt
+	* docs/new-drivers.txt
+	* docs/new-names.txt
+
+	and possibly also to:
+	* docs/hid-subdrivers.txt for USB/HID devices
+	* or docs/snmp-subdrivers.txt for SNMP devices
+*/
+/* ./configure --with-pidpath=/run/nut --with-altpidpath=/run/nut --with-statepath=/run/nut --sysconfdir=/etc/nut --with-gpio --with-user=nut --with-group=nut	*/
+#pragma GCC optimize("O0")
```
jimklimov, line 15: By release time this should go away (or hidden by `#if 0`) :)

Is this to ease debugging, or is there some compile/run-time issue that the driver won't work because of optimizations? Then a comment to keep this pragma long-term would help.

drivers/generic_gpio_common.c
```diff
@@ -0,0 +1,444 @@
+/*
```
jimklimov, line 1: Add the skeleton GPL header with your copyrights into the new source files, please.

drivers/generic_gpio_common.c
```diff
@@ -0,0 +1,444 @@
...
+				if(upsfd->upsLines[k]==upsfd->rules[i]->cRules[j]) {
+					pinOnList=1;
+				}
+			}
+			if(!pinOnList) {
+				if(upsfd->rules[i]->cRules[j]>=0) {
+					upsfd->upsLinesCount++;
+					upsfd->upsLines=xrealloc(upsfd->upsLines, sizeof(upsfd->upsLines[0])*upsfd->upsLinesCount);
+					upsfd->upsLines[upsfd->upsLinesCount-1]=upsfd->rules[i]->cRules[j];
+					if(upsfd->upsLines[upsfd->upsLinesCount-1]>upsfd->upsMaxLine) {
```
jimklimov, line 241: General coding style (all around the file) - this would be much more readable with spaces around operators like `<`, `>`, `=`, `==` etc., and we have spaces between `if/for` and the `(` parenthesis after it :)

drivers/generic_gpio_common.h
```diff
@@ -0,0 +1,48 @@
+#ifndef GENERIC_GPIO_COMMON_H
+#define GENERIC_GPIO_COMMON_H
+
+#include <stdlib.h>
+#include <gpiod.h>
```
jimklimov, line 5: This is technically `#if HAVE_GPIOD_H` or `#if WITH_LIBGPIOD` - right?
The file claims to be common, but hardcodes a dependency on (currently the only) one backend here.

drivers/generic_gpio_common.h
```diff
@@ -0,0 +1,48 @@
+#ifndef GENERIC_GPIO_COMMON_H
```
jimklimov, line 1: GPL header

drivers/generic_gpio_common.h
```diff
@@ -0,0 +1,48 @@
...
+#define RULES_CMD_CBR   -6
+#define RULES_CMD_LAST  RULES_CMD_CBR
+
+/*  run option definitions      */
+#define ROPT_REQRES 0x00000001  /* reserve GPIO lines only during request processing    */
+#define ROPT_EVMODE 0x00000002  /* event driven run                                     */
+
+typedef struct rulesint_t { /* structure to store processed rules configuration per each state */
+    char    stateName[12];  /* NUT state name for rules in cRules */
+    int     archVal;        /* previous state value */
```
jimklimov, line 23: Just in case: does this deal with architecture-dependent sizes of `int` or some wire-protocol specific `(u)int32_t` etc.? The `nut_stdint.h` takes care of providing those for different platforms.

drivers/generic_gpio_libgpiod.c
```diff
@@ -0,0 +1,296 @@
...
+	* docs/developers.txt
+	* docs/new-drivers.txt
+	* docs/new-names.txt
+
+	and possibly also to:
+	* docs/hid-subdrivers.txt for USB/HID devices
+	* or docs/snmp-subdrivers.txt for SNMP devices
+*/
+/* ./configure --with-pidpath=/run/nut --with-altpidpath=/run/nut --with-statepath=/run/nut --sysconfdir=/etc/nut --with-gpio --with-user=nut --with-group=nut	*/
+#pragma GCC optimize("O0")
```
jimklimov, line 15: Similar comments to above - GPL header, this pragma, spaces in clauses (code style/readability).

drivers/Makefile.am
```diff
@@ -164,6 +171,11 @@ upscode2_LDADD = $(LDADD) -lm
 victronups_SOURCES = victronups.c
 riello_ser_SOURCES = riello.c riello_ser.c
 riello_ser_LDADD = $(LDADD) -lm
+#gpio_SOURCES = generic_gpio_common.c generic_gpio_libgpiod.c
+generic_gpio_libgpiod_SOURCES = generic_gpio_common.c generic_gpio_libgpiod.c
+#gpio_CFLAGS = $(AM_CFLAGS)
+#gpio_LDADD = $(LDADD_DRIVERS) -lgpiod
+generic_gpio_libgpiod_LDADD = $(LDADD_DRIVERS) -lgpiod
```
jimklimov, line 178: I'm a bit split on whether explicit `-lgpiod` belongs here. Consistently, a `LIBGPIOD_LIBS` as detected by `configure.ac` and `m4` snippet, would be better (especially if some more custom flags come later or on certain platforms).

Also, I think this block belongs lower in the file, among "non-serial drivers" - e.g. near I2C as the added lines neighbor elsewhere.

jimklimov (changes requested):
Thanks, quite a solid contribution - but some changes are desired, as commented in details.

* [ ] GPL headers in sources
* [ ] Make sure ASCIIDOC markup (man page) renders reasonably
* [ ] Complete the `configure.ac`/`m4`/`Makefile.am` integration (and a few bugs in it)
* [ ] Coding style can be made friendlier :)
* [ ] Probably something else here and there

jimklimov:
By the way, does GPIO availability on systemd-aware OSes rely on some certain units? Should `nut-driver-enumerator` be updated to generate a special dependency for these drivers to start as soon as (but not sooner than) they are able to talk to hardware?

modrisb:
As far as I know GPIO support is implemented via kernel driver - should 
not depend systemd. Permissions may cause issues - will have to check. I 
remember permission issues some long time ago, but have not seen 
anything similar recently. Remember some changes in rules.d and then 
stable start, but need to check once more.

modrisb:
Clear on next steps.

published part of fixes: m4 will follow tomorrow.

Coding style - 1st round.

On 2/26/23 13:55, Jim Klimov wrote:
>
> ***@***.**** requested changes on this pull request.
>
> Thanks, quite a solid contribution - but some changes are desired, as 
> commented in details.
>
>   * GPL headers in sources
>   * Make sure ASCIIDOC markup (man page) renders reasonably
>   * Complete the |configure.ac|/|m4|/|Makefile.am| integration (and a
>     few bugs in it)
>   * Coding style can be made friendlier :)
>   * Probably something else here and there
>
> ------------------------------------------------------------------------
>
> In docs/man/gpio.txt 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118066606>:
>
> > +SHUTDOWN COMMAND
> +----------------
> +
> +This driver does not support shutdown command.
> +
> +INSTALLATION
> +------------
> +
> +This driver is not built by default. You can build it by installing
> +libgpiod and running `configure --with-gpio=yes`.
> +
> +You also need to give proper permissions on the local serial device
> +file (/dev/gpiochip0 for example) to allow the run-time NUT driver user
> +account to access it, like by adding the following rule to rules.d directory:
> +SUBSYSTEM=="gpio*", PROGRAM="/bin/sh -c '\
> +        chown -R nut:nut /dev/gpiochip0 && chmod -R 700 /dev/gpiochip0\
>
> This code example should be indented, and possibly preceded by a blank 
> line.
>
> ------------------------------------------------------------------------
>
> In docs/man/gpio.txt 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118066897>:
>
> > +---------------
> +
> +This driver supports the following optional settings in the
> +linkman:ups.conf[5] file:
> +
> +Driver control:
> +~~~~~~~~~~~~~~
> +
> +*rules*='value'::
> +A string consisting of sub-strings. Each sub-string describes GPIO line
> +states conversion formula to specific NUT state, like
> +nut_state=[^]line_num[logical_operation.line_num]. Not (^) , and (&) , or
> +(|)operations are supported for now. nut_state should correspond to NUT
> +state, line_num to GPIO line number connected to UPS open collector pin.
> +CyberShield CSN27U12V describes pins as:
> +ON BATTERY			Low when operating from utility line
>
> I think asciidoc table markup (with pipe characters) would do well 
> here, and this may need blank lines around to differentiate from other 
> paragraphs. If possible, build with HTML docs enabled to iterate the 
> result-oriented markup more easily (or use an IDE with an asciidoc 
> plugin to render it real-time as you type) :)
>
> ------------------------------------------------------------------------
>
> In docs/man/Makefile.am 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118067120>:
>
> > @@ -669,6 +669,17 @@ endif
>   
>   HTML_LINUX_I2C_MANS = asem.html pijuice.html
>   
> +SRC_LINUX_GPIO_PAGES = gpio.txt
>
> Just in case: is it right to call this a LINUX_GPIO? The currently 
> used libgpiod "backend" does say it is for linux, but original 
> discussion implied this driver could be more generic with other 
> platforms logically supporting the interface.
>
Removed _LINUX to match agreement
> ------------------------------------------------------------------------
>
> In configure.ac 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118067542>:
>
> > @@ -1526,6 +1534,20 @@ dnl and AM_CONDITIONALs (see below)...
>   AM_CONDITIONAL(WITH_FREEIPMI, test "${nut_with_freeipmi}" = "yes")
>   dnl AM_CONDITIONAL(WITH_OPENIPMI, test "${nut_with_openipmi}" = "yes")
>   
> +dnl ----------------------------------------------------------------------
> +dnl Check for with-gpio
> +
> +nut_gpio_lib=""
> +
> +NUT_ARG_WITH([gpio], [build and install GPIO driver], [auto])
> +
> +dnl ${nut_with_gpio}: any value except "yes" or "no" is treated as "auto".
>
> I think somewhere here it should check for libgpiod (potentially other 
> equivalents later).
>
> Typically the check is done as a new method under |m4| directory (copy 
> and adapt an existing file), setting the |LIBGPIO_CFLAGS|, 
> |LIBGPIO_LDFLAGS|, etc. as appropriate, manually or from |pkg-config| 
> data where available.
>
> The |configure.ac| code calls such method, and if things are found and 
> requested mode was "auto" it becomes "yes", otherwise "no". If the 
> request was "yes" but library was not found, it should cause a fatal 
> |AC_MSG_ERROR|.
>
> ------------------------------------------------------------------------
>
> In drivers/Makefile.am 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118067825>:
>
> >  
>   # distribute all drivers, even ones that are not built by default
>   EXTRA_PROGRAMS  = $(SERIAL_DRIVERLIST) $(USB_DRIVERLIST) $(SERIAL_USB_DRIVERLIST)
>   EXTRA_PROGRAMS += $(SNMP_DRIVERLIST) $(NEONXML_DRIVERLIST) $(MACOSX_DRIVERLIST)
>   EXTRA_PROGRAMS += $(LINUX_I2C_DRIVERLIST)
> -EXTRA_PROGRAMS += $(NUTSW_DRIVERLIST)
> +EXTRA_PROGRAMS += $(GPIO_DRIVERLIST)
>
> This was supposed to add a new line, not forget about software-only 
> drivers, right? ;)
>
> ------------------------------------------------------------------------
>
> In drivers/generic_gpio_common.c 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118068253>:
>
> > @@ -0,0 +1,444 @@
> +/*
> +	anything commented is optional
> +	anything else is mandatory
> +
> +	for more information, refer to:
> +	* docs/developers.txt
> +	* docs/new-drivers.txt
> +	* docs/new-names.txt
> +
> +	and possibly also to:
> +	* docs/hid-subdrivers.txt for USB/HID devices
> +	* or docs/snmp-subdrivers.txt for SNMP devices
> +*/
> +/* ./configure --with-pidpath=/run/nut --with-altpidpath=/run/nut --with-statepath=/run/nut --sysconfdir=/etc/nut --with-gpio --with-user=nut --with-group=nut	*/
> +#pragma GCC optimize("O0")
>
> By release time this should go away (or hidden by |#if 0|) :)
>
> Is this to ease debugging, or is there some compile/run-time issue 
> that the driver won't work because of optimizations? Then a comment to 
> keep this pragma long-term would help.
>
Removed - debugger did not stop at certain statements as they were 
removed because of optimization.
> ------------------------------------------------------------------------
>
> In drivers/generic_gpio_common.c 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118068290>:
>
> > @@ -0,0 +1,444 @@
> +/*
>
> Add the skeleton GPL header with your copyrights into the new source 
> files, please.
>
> ------------------------------------------------------------------------
>
> In drivers/generic_gpio_common.c 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118068815>:
>
> > +	upsfd->upsLinesCount=0;
> +	upsfd->upsMaxLine=0;
> +	for(i=0; i<upsfd->rulesCount; i++) {
> +		for(j=0; j<upsfd->rules[i]->subCount; j++) {
> +			int pinOnList=0;
> +			for(k=0; k<upsfd->upsLinesCount && !pinOnList; k++) {
> +				if(upsfd->upsLines[k]==upsfd->rules[i]->cRules[j]) {
> +					pinOnList=1;
> +				}
> +			}
> +			if(!pinOnList) {
> +				if(upsfd->rules[i]->cRules[j]>=0) {
> +					upsfd->upsLinesCount++;
> +					upsfd->upsLines=xrealloc(upsfd->upsLines, sizeof(upsfd->upsLines[0])*upsfd->upsLinesCount);
> +					upsfd->upsLines[upsfd->upsLinesCount-1]=upsfd->rules[i]->cRules[j];
> +					if(upsfd->upsLines[upsfd->upsLinesCount-1]>upsfd->upsMaxLine) {
>
> General coding style (all around the file) - this would be much more 
> readable with spaces around operators like |<|, |>|, |=|, |==| etc., 
> and we have spaces between |if/for| and the |(| parenthesis after it :)
>
> ------------------------------------------------------------------------
>
> In drivers/generic_gpio_common.h 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118069343>:
>
> > @@ -0,0 +1,48 @@
> +#ifndef GENERIC_GPIO_COMMON_H
> +#define GENERIC_GPIO_COMMON_H
> +
> +#include <stdlib.h>
> +#include <gpiod.h>
>
> This is technically |#if HAVE_GPIOD_H| or |#if WITH_LIBGPIOD| - right?
> The file claims to be common, but hardcodes a dependency on (currently 
> the only) one backend here.
>
Moved to generic_gpio_libgpiod.h - missed change during refactoring and 
now common part is fully library independent
> ------------------------------------------------------------------------
>
> In drivers/generic_gpio_common.h 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118069380>:
>
> > @@ -0,0 +1,48 @@
> +#ifndef GENERIC_GPIO_COMMON_H
>
> GPL header
>
> ------------------------------------------------------------------------
>
> In drivers/generic_gpio_common.h 
> <https://github.com/networkupstools/nut/pull/1855#discussion_r1118069544>:
>
> > +
> +/*  rules commands definition    */
> +#define RULES_CMD_NOT   -2
> +#define RULES_CMD_AND   -3
> +#define RULES_CMD_OR    -4
> +#define RULES_CMD_OBR   -5
> +#define RULES_CMD_CBR   -6
> +#define RULES_CMD_LAST  RULES_CMD_CBR
> +
> +/*  run option definitions      */
> +#define ROPT_REQRES 0x00000001  /* reserve GPIO lines only during request processing    */
> +#define ROPT_EVMODE 0x00000002  /* event driven run                                     */
> +
> +typedef struct rulesint_t { /* structure to store processed rules configuration per each state */
> +    char    stateName[12];  /* NUT state name for rules in cRules */
> +    int     archVal;        /* previous state value */
>
> Just in case: does this deal with architecture-dependent sizes of 
> |int| or some wire-protocol specific |(u)int32_t| etc.? The 
> |nut_stdint.h| takes care of providing those for different platforms.
>
No, plain integers, not related to wire or h/w.

modrisb:
m4 is in too, tested only on singe local machine.

modrisb:
I checked logs on running system: and don't see recent cases with 
permission failures that might be caused slow rules.d execution. Looks I 
did safety change in systemd
configuration for nut-server.service by adding restart option for it, like
[Unit]
Description=Network UPS Tools - power devices information server
After=local-fs.target network.target nut-driver.service
# We don't Require drivers to be successfully started! This would be
# a change of behavior compared to init SysV, and could prevent from
# accessing successfully started, at least to audit a system.
Wants=nut-driver.service
Before=nut-monitor.service

[Service]
ExecStart=/sbin/upsd
Type=forking
Restart=on-failure
RestartSec=10

[Install]
WantedBy=multi-user.target

This ensures that driver is reloaded in case of chip open failure.

Modris

jimklimov:
Confused about latest comment: the `nut-server.service.in` in NUT sources includes the `Restart` option and a very different text (more comments, etc.) Is the version you have deployed coming from a recent build, or from NUT 2.7.4 packages by distro?

Also, you should not restart the data server (upsd) for hiccups with a driver. Driver may die, restart and reconnect to upsd.

modrisb:
Code is finalized from my side.

What about unit test code? I don't see how to store/run related code. 
Wanted to add some cases for rules processing test...

jimklimov:
Unfortunately so far tests are rather nascent. There are a few one-off test programs for some aspects and libraries, but no sturdy framework to rule them all. Check under `tests/` dir to see if any precedent makes sense for your test code or add another :)

modrisb:
Jim,

I added some tests for gpio: files + makefile changes to my branch. Test 
code is pretty simple and not as clean as driver's code.

There is 1 more commit for driver code too - overlooked these changes.

Modris

jimklimov:
Revised recipes to not fail on OSes without `libgpiod` present ;)

jimklimov:
Hm, includes of C source files?.. that was a very unorthodox approach :)

jimklimov:
Well, actually this needed a lot of chizeling around the rough edges to work and not spew warnings, pass distcheck, etc. ;)

Normally C sources are built standalone and then linked together into a binary, not included directly - more so for testing of what should be library code. Was there a particular need to use much of `drivers/main.c`, and the `dstate.c` directly in the test, as if part of its source, rather than via linking? Was it due to testing of rules support?

I think something better is possible along the lines of `drivers/Makefile.am:libdummy_la_SOURCES = main.c dstate.c`, perhaps instrumenting `drivers/main.c` to wrap `main()` into some `#ifndef MAIN_WITHOUT_MAIN` sort of flag to build an object file that has almost everything.

jimklimov:
At least on one system, these recent commits got rid of unidiomatic `#include "somefile.c"` and passed for in-tree and out-of-tree (distcheck) builds. Hoping CI likes it :)

modrisb:
gpio test code is more close to unit testing rather than integration, 
therefore supposing to have special code to execute selected portions of 
code.

gpio tests runs into 3 separate areas:
1) generic driver test which includes rules testing and rules evaluation 
parts;
2) library interface test for libgpiod implementation. This includes 
mock library for libgpiod and requires part of drivers/main.c and 
dstate.c. These tests verifies how update_ups_states , upsdrv_initups, 
upsdrv_cleanup, upsdrv_makevartable, upsdrv_initinfo, upsdrv_updateinfo, 
upsdrv_cleanup works. There are reason to run these tests just once per 
source code changes (both driver/main.c&Co and gpio). I don't see 
benefit to have separate library without main unless there are plans to 
add some similar as gpio tests for other types of UPSes. Approach with 
includes takes recent sources and tests them, no reason to keep interim 
results or rely on delivery packages. We may run them even on any commit 
just to prevent code brakes.

modrisb:
there could be issue with access to
vartab_free(); vartab_h = NULL;
needed to clean up variables between test cases, both are static and 
accessible when include

jimklimov:
Take a look in the recent commits' proposal of `ifdef DRIVERS_MAIN_WITHOUT_MAIN` - I got to a similar conclusion about `static` so use of the keyword became conditional, depending on build circumstances. This macro is defined for the test-program builds, and so these test builds declare the methods and variables non-static (and the new helper library exposes them for the linker).

The Makefile changes, now explicitly referring to sources that were previously included, take care of building the test once per change of any source involved (previously this was not guaranteed - only that a change of timestamp of `generic_gpio_utest.c` would trigger rebuild of the test program) as well as building the same source into different object files with different settings (with the macro for test, without it for driver).

And indeed, having the `libdummy_mockdrv.la` now allows easier creation of more test programs that behave similarly to a real driver.

modrisb:
Changes look fine. Tests pass for me on PC, will check on PI too, but 
this will take couple hours.

modrisb:
Orange PI looks good too.

jimklimov:
@modrisb : Thank you very much for this improvement! :)

modrisb:
It was interesting activity, and you did a lot of improvements in making 
it ready for release, thanks!

Issue #1858, opened 2023-02-24
UPSSCHED doesn't work correct with multiple UPSs


shgubar:
HI,
My configuration consists of two UPS.

System: **Debian Buster**
NUT: **2.8.0.1** (_installed from source_)

**upsmon.conf**
```
MONITOR riello1000@localhost 1 user pass primary
MONITOR rps1000@localhost 0 user pass secondary
```

**upssched.conf**
```
AT ONLINE riello1000@localhost START-TIMER upsgone 60
AT ONBATT riello1000@localhost CANCEL-TIMER upsgone
```

With this configuration UPSSCHED doesn't work, if I change to:
```
AT ONLINE * START-TIMER upsgone 60
AT ONBATT * CANCEL-TIMER upsgone
```
working perfectly.

Why I can't use specific UPS for this task?

jimklimov:
Hi, looked through code and it boiled down to `upssched.c::parse_at()` expecting exactly the upsname from local configuration.
* https://github.com/networkupstools/nut/blob/master/clients/upssched.c#L1173

Try dropping the `@localhost` in `AT` lines. It is a bit inconsistent with `upsmon` sort of requiring a hostname (since it does talk TCP) and guessing a `@localhost` if there is no hostname part like `upsc` does; actually the `upscli_splitname()` library method does that. On the other hand, `upssched` supposedly only manages operations on local UPSes, so this makes sense.

jimklimov:
Are you in position to build NUT locally and test if PR #1861 fixes the original issue (allows to use e.g. `AT riello1000@localhost ...`)? Would be nice to know it works and does not crash before merging it ;) Thanks.

shgubar:
> Hi, looked through code and it boiled down to `upssched.c::parse_at()` expecting exactly the upsname from local configuration.
> 
> * https://github.com/networkupstools/nut/blob/master/clients/upssched.c#L1173
> 
> Try dropping the `@localhost` in `AT` lines. It is a bit inconsistent with `upsmon` sort of requiring a hostname (since it does talk TCP) and guessing a `@localhost` if there is no hostname part like `upsc` does; actually the `upscli_splitname()` library method does that. On the other hand, `upssched` supposedly only manages operations on local UPSes, so this makes sense.

I tried as you said:
`AT ONLINE riello1000 START-TIMER upsgone 60`
 but the command was not executed. It didn't help.

> Are you in position to build NUT locally .....

What do you mean download brunch again and install?
[...]

PR #1861, opened 2023-02-28
upssched: let `UPSNAME@localhost` specs copied from upsmon config "just work"

jimklimov:
Closes: #1858


jimklimov:
After more looking at code, not sure this is needed or correct.

Issue #1883, opened 2023-03-30
Cross build on i386 to armv7


idabsent:
Hello. Sorry for my English. This text is written with a web translation tool and with my knowledge of English.

I have a Debian (i386) based system built with debootstrap. Try to generate a configuration(**configure** executable script) file for sources (I build binaries and libraries from sources) on this system  **autom4te**, **aclocal**, **autoreconf** tools give errors. 

```
(chroot wheezy_deb) root@pop-os:/projects/nut# ./autogen.sh
Regenerating Augeas ups.conf lens with 'python'...
Regenerating the USB helper files...
Calling autoreconf...
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal -I m4
/usr/bin/m4:configure.ac:89: Warning: too few arguments to builtin `ifdef'
autom4te: /usr/bin/m4 failed with exit status: 1
aclocal: /usr/bin/autom4te failed with exit status: 1
autoreconf: aclocal failed with exit status: 1
FAILED: did not generate an executable configure script!

# Note: on some systems "autoreconf", "automake" et al are dispatcher
# scripts, and need you to explicitly say which version you want, e.g.
#    export AUTOCONF_VERSION=2.65 AUTOMAKE_VERSION=1.13
# If you get issues with AC_DISABLE_STATIC make sure you have libtool.

(chroot wheezy_deb) root@pop-os:/projects/nut# m4 --version
m4 (GNU M4) 1.4.16
Copyright (C) 2011 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Rene' Seindal.

(chroot wheezy_deb) root@pop-os:/projects/nut# autom4te --version
autom4te (GNU Autoconf) 2.69
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+/Autoconf: GNU GPL version 3 or later
<http://gnu.org/licenses/gpl.html>, <http://gnu.org/licenses/exceptions.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

(chroot wheezy_deb) root@pop-os:/projects/nut# aclocal --version
aclocal (GNU automake) 1.11.6
Copyright (C) 2011 Free Software Foundation, Inc.
[...]

PR #1885, opened 2023-03-30
C++11 support tighter check

jimklimov:
Older GCC-4.7.2 (in Debian 7 Wheezy, per #1883 discussion) claims to support the language version we need, but in fact has numerous missing features that a build hits (the whole 4.x line added C++11 support step by step, one minor release after another). This PR fixes a warning that seems useful (about `noexcept` in destructors) -- CI will show if newer systems have a problem with that change.

Primarily this PR extends the `configure.ac` check to bail out on some infractions easier to test (lack of `std::map -> emplace()` for example).

NOTE: A few earlier documentation changes went to master branch without a PR, to follow up on setup of the Debian 7 LXC container, etc.

### DIFF
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -194,12 +194,12 @@ std::string SystemException::err()
  * https://lgtm.com/rules/2165180572/ like:
  *   NutException& operator=(NutException& rhs) = default;
  */
-NutException::~NutException() {}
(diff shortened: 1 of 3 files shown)


jimklimov:
Checked to not enable C++11 on Debian 7 builds (despite its GCC claiming to support the language version by macro) anymore, and to still enable it with newer environments by default.

Issue #1688, opened 2022-10-28
blazer_usb and other cheap usb devices won't always start


deajan:
Hello,

I have a couple of setups with some cheaper USB controlled UPS with NUT and have sporadic nut-driver startup.
Sometimes after a reboot, nut-driver service won't start.
I've tried 4 different UPS, all with the same USB interface, being `0665:5161 Cypress Semiconductor USB to Serial`

I've traced down the issue manually running `blazer_usb` driver in different scenarios.
If I happen to unplug / replug the UPS, and restart nut-driver, everything works great.

Since I cannot just change all the UPS that don't behave correctly, I've built a "well written" script that resets USB hub for given device, or allows resetting device only, allowing to make sure the device will function.

I've modified the systemd service file `nut-driver.service` to include the following:
```
ExecStartPre=/usr/local/bin/usb_reset.py --reset-hub --device 0665:5161
```

With the above, I am assured that nut-driver will start correctly, since the UPS's USB device is reinitialized just before starting the nut driver.

I am aware that this fix is too specific in order to have it included into NUT sources, but perhaps:
- Is there some documentation where I could do a write up about this problem
- Maybe The `ExecStartPre` line could be added but commented out, so it's possible to enable it for users that have USB UPSes ?

jimklimov:
I suppose Wiki can be a good place. NUT docs include a FAQ that could mention this too, and the scripts area often used for systems integration stuff, if you'd go the PR way.

One thing to look out for is that many USB vendors cut corners, so a command to reset one port might recycle the whole hub and cause issues for other devices.

deajan:
I've actually anticipated that.

My script can list hubs per devices with `usb_reset.py --list-hubs --device xxxx:xxxx`, so it's easy to know which devices would get reset when resetting a hub. On my test hardware, I was lucky and have one device per hub.
On "non lucky" hardware, one must plug the UPS in a way to be "alone" on the hub.

I also have a reset device only command with `usb_reset.py --reset-device --device xxxx:xxxx` which basically sends a USBDEVFS_RESET signal to the device, but that one works less well (success rate is 3/4, I have no idea why when resetting device before loading blazer_usb driver).

> and the scripts area often used for systems integration stuff, if you'd go the PR way.*

I'll be happy to make a PR, but I have no idea where to put the file, and where to make a PR for the documentation. Mind to show me ?

jimklimov:
[...]

PR #1887, opened 2023-03-31
Add usb_resetter documentation for unreliable USB UPS interfaces (Linux, Python)

deajan:
This PR adds documentation for a workaround solution for unreliable USB interfaces, as stated in #1688 
I use this on a couple of UPS for months now, without any major issue.

## General points

- [X] Described the changes in the PR submission or a separate issue, e.g.

- [X] There may be multiple commits in the PR, aligned and commented with
  comments or text documents.

- [N/A] Updated `docs/acknowledgements.txt` (for vendor-backed device support)

- [N/A] Added or updated manual page information in `docs/man/*.txt` files

- [N/A] Passed `make spellcheck`, updated spell-checking dictionary in the

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -209,6 +209,11 @@ https://github.com/networkupstools/nut/milestone/8
  - upslog: Added support for logging multiple devices with one call to the
    program [#1604]
 
+ - Documentation to integrate NUT USB driver startup with `usb_resetter` script
+   has been contributed to `scripts/usb_resetter` (the script itself is tracked
+   externally on GitHub), along with a configuration example for Linux+systemd
+   [#1887]
+
  - Some fixes applied to Solaris/illumos packaging and SMF service support
    [#1554, #1564]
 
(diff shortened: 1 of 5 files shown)


deajan:
> I think it is worth noting (maybe in the document too) that nut-driver.service provided here is an example

Indeed, I'll have another PR for the README file

> The call to usb_resetter can be a systemd drop-in file for the particular device

I have looked at drop-in files, but i don't know how to limit to one instance of a `@.service file`

> Does the script only exit when it is sure the device is usable, or should there be an artificial delay added?

No, the script's task is only to reset devices / controllers.  
Nevertheless, I did not have to set any arbitrary wait time to make use of my USB UPSes with NUT.

jimklimov:
> don't know how to limit to one instance of a @.service file

Just name its directory for the expected instance (NDE with NUT v2.8.0+ will create some drop-ins), e.g.

/etc/systemd/system/nut-driver@myups.service.d/dropin.conf

jimklimov:
As a curious data point, similar issue (and also with Innotech chips) was addressed for Solaris/illumos earlier, see https://github.com/networkupstools/nut/blob/master/docs/solaris-usb.txt#L233 and https://github.com/networkupstools/nut/blob/master/scripts/Solaris/reset-ups-usb-solaris.sh.sample

jimklimov:
Thanks!

PR #1910, opened 2023-04-19
[TO REMOVE] Acquisition failed with daisychain epdu G3 with sensors

ericclappier-eaton:
Hot fix for infinite loop during acquisition of daisychain epdu G3 with sensors (snmp-ups driver).


arnaudquette-eaton:
replaced by https://github.com/networkupstools/nut/pull/1911

PR #1911, opened 2023-04-19
Acquisition failed with daisychain epdu G3 with sensors

ericclappier-eaton:
Hot fix for infinite loop during acquisition of daisychain epdu G3 with sensors (snmp-ups driver).


jimklimov:
Thanks a lot guys! Waiting for test result notice :)

aquette:
positive testing results received, proceeding to merge

Issue #1920, opened 2023-04-21
Extend driver-server socket protocol with verdicts


jimklimov:
Follow-up from research in issue #1914 per implementation experience in PR #1918; possibly a solution vector for bug #1901.

Currently `dstate::sock_arg()` in many cases quietly does:
````
/* The command was handled, status is a separate consideration */
return 1;
````

Which does not tell the direct client (`upsd`) or eventually the user's client (`upscmd`, `upsrw`...) how that handling ended up.

At least some (new) cases might instantly report whether they succeeded or not, similar to how `DATADUMP` handling ends with `DATAOK` (as opposed to `DATASTALE`), or `PING` ends with `PONG`, but currently do not due to lack of protocol keywords for that at the moment.

The return values from main/driver-specific handlers (per enums in `drivers/upshandler.h`) might be reported in a similar fashion at least on the socket protocol. Just to know the command was acknowledged and perhaps how it ended up.

In the bigger picture, may the socket protocol be treated as a private experience between current versions of `upsd` and NUT drivers - e.g. if someone built/scripted interactions around that, and those third-party tools fail due to unexpected key words, we commiserate but it is not a NUT problem as such?

To be clearer, this is not immediately a question about on-line NUT protocol (data server to clients), although I would not be surprised if it sprouts into that too (or practical handling of that, to more relevantly report OK/ERR for client requests).

CC @aquette @clepple @NUT-RogerPrice

gdt:
I think it's good to allow a reasonable path to maintaining a driver not in the main nut source tree, although that could be with the obligation to adapt to changes.   I thus mean with deprecation/etc. more than flag days.   Of course, complexity is bad too.

It sounds like the main issue is to add some sort of async "command is really done" after having an "in progress" return.   Once could either
  - document this now and then add it in 6 months, which would result in feeling righteous with no real gain
  - add it to nut git, and release it after a while, expecting private driver authors to adapt in the meantime
  - add a "enable async" command that drivers that understand the new way would issue at startup, as a transititon mechanism, perhaps with that being default and a no-op a year after the formal release of a nut version with it.
  -

jimklimov:
Thanks for help with the brainstorming session :)

Interesting point about out-of-tree drivers. I'd expect those to be built against NUT `dirvers/main.c` and `drivers/dstate.c` (of the files relevant to socket protocol handling - with `main.c` being a recent addition to such list in these recent PRs, adding "shared" commands/settings like changing debug verbosity on the fly).

[...]

PR #1918, opened 2023-04-20
Extend server-driver socket protocol capabilities

jimklimov:
Address a number of recent ideas and issues for driver reload and other activities.

* Added instant commands and variable-setting "shared" by all drivers, covering config reload, killpower, and debug level changes;
* Added protocol commands which allow a socket client to unsubscribe from broadcast messages (`dstate::send_to_all()` e.g. driver state quiet/updating in the main loop) and only see PtP replies to queries it posted (`dstate::send_to_one()`);
* Added support for `NUT_DEBUG_LEVEL` envvar in several daemons;
* Updated docs on these and related subjects.

Touches on issues and proceeds from PRs including:
* Closes: #1285
* #1903
* #1906
* #1912
* #1914
* Closes: #1915
* Closes: #1917

Issue #1903, opened 2023-04-17
Support `reload` operation for drivers (like we have for `upsd` and `upsmon`), at least to change `debug_min` setting


jimklimov:
This idea follows up from:
* #1202
* PR #683
* Note caveats like #1761

The general idea is to let drivers define which variables (see `addvar()`) can be reloaded at all, and to facilitate (re-)reading of the `ups.conf` file knowing which values we may safely change on the fly (warn about other changes that require a driver restart to apply). The same new internal flag can be used for NUT protocol setting of `driver.parameter.*` values per #1781 (to discern RW vs. RO values).

For `debug_min` in particular, see `upsd` and `upsmon` approach about tracking `nut_debug_level_global` vs. `nut_debug_level_args` to ultimately set the `nut_debug_level` used by `upsdebugx()` and friends, and remember that the setting can be in both the global and driver sections of `ups.conf`.

While the initial implementation is easy to tie into `SIGHUP` handling for power-user troubleshooting, more work awaits for "proper" integration for end-user convenience:
* [x] `drivername -c reload` CLI option
* [x] systemd/SMF/... `reload` action support
* [x] `upsdrvctl` and `upsdrvsvcctl` CLI options
* [x] WIN32 pipe command shim instead of POSIX signals
* [x] `nut-driver-enumerator` support to just reload and not restart drivers, where possible

At the moment the concern is both about changing debug verbosity setting on the fly, and about making a generic approach that can be useful for other dynamic activities. This currently leaves a lot of points to decide later, e.g.:
* Should we limit to "warn about other changes that require a driver restart" or do something more active?
    * report a failure to reload so systemd `reload-or-restart` action can restart the driver (possibly use a dedicated CLI command for that, and not a signal)? => currently reported to log; the signals are async (and a CLI command would at best shoot one and wait for some time if the target process dies or not).
    * crash and do the restart (relying on systemd/SMF/... to resuscitate)? => implemented as `reload-or-exit` action
    * for some values, like change of poll/interrupt, locking, SNMP credentials, etc. we might process `upsdrv_cleanup()` and re-open the media/vendor protocol connection to effectively re-init the driver to err on the safe side -- we do similar things to reconnect upon failure in at least some drivers.
    * note that for some changes like the run-time user/group, probably "port=...", and of course "driver=..." a complete restart would be needed.

jimklimov:
[...]

PR #1922, opened 2023-04-22
Teach `upsdrvctl` and drivers to talk socket protocol to their other selves

jimklimov:
As a follow-up to issue #1903, PR #1918 and discussion #1920, this PR adds a limited socket-protocol client to drivers (main.c) and upsdrvctl, so they can request `reload-or-error` on command-line to query the running instance of the driver. If it fails to reload some changed values, exit-code 2 can be seen (technically `STAT_INSTCMD_INVALID`), so e.g. nut-driver-enumerator can disable and redefine the service carefully.


jimklimov:
Note: mostly inherits PR #1918, the new feature are the last two commits as of now.

jimklimov:
Need to follow up on https://github.com/networkupstools/nut/issues/1920#issuecomment-1519043742 to add `TRACKING` support to `(NO)BROADCAST` and use that instead of indiscriminate waits to quiet down for 5 seconds. Also change the `upsdrvquery` client with a method to loop reading (and possibly ignoring what it saw) until it sees a line with expected tracking ID, so the main single-shot logic would be to (maybe) request NOBROADCAST and (maybe) wait for it to be acknowledged, then issue the SET/INSTCMD request and wait for that. Or just go on with the request and ignore the rest.

Issue #1928, opened 2023-04-24
Add support for TRACKING to (NO)BROADCAST on driver-upsd socket protocol


jimklimov:
Follow-up from #1920 discussion

Issue #1914, opened 2023-04-20
Explore additional communications with driver by Unix socket (and Windows pipe?)


jimklimov:
Following up on smarter work ideas for reload-or-exit (#1903) especially with nut-driver-enumerator, as well as requesting a driver state dump that could be visible by caller and regardless of its `stdout` attachment (#1907), we could benefit from introducing a new socket/pipe based dialogue (using new entities? can the same ones be shared on all platforms?)

Specifically for configuration reloads using NDE, the script (as well as `upsdrvctl`, as well as a `drivername -c something`) could use the dialogue to ask if the currently running driver can just reload the changed configs or if it would rather restart -- with the other side of the talk ensuring further actions then (not the driver just aborting, as implemented in `reload-or-exit` mode available now).

Somewhat borders on idea #1703 to use SHMFS where available.

jimklimov:
Implementation proposed as part of #1918 to provide new methods in main.c shared by all drivers which can handle the common instant commands and variable setting, as well as new commands to toggle broadcast visibility (vs. unicast replies to queries) on this specific connection.

Testing with command-line `socat` was fruitful.

jimklimov:
Existing code worked sufficiently for the desired features on server side; new `drivers/upsdrvquery.{c,h}` were added as a simple client for driver-to-driver socket protocol chats, with #1922.

PR #1936, opened 2023-05-10
Optionally install `sockdebug`, provide a man page

jimklimov:
Also update `upsdrvquery` preparation using `NOBROADCAST` with a `PING/PONG` confirmation.

Follows-up from #1914

Touches upon #1928

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -212,6 +212,9 @@ cppcheck:
 	@echo "CPPCHECK analysis not available since 'cppcheck' was not found."
 endif !HAVE_CPPCHECK
 
+sockdebug:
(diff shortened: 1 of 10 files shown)


jimklimov:
Upon a second look at `upsdrvquery` mechanism, found a number of embarrassing typos (notably mixup of 10e3 and 10e6 in `tv_usec` handling) and misconceptions (`time()` returning a whole number of seconds as `time_t`, so unsuitable for checking if we had honoured sub-second delays). Introduced a `difftimeval()` method to produce floating-point diffs of `struct timeval`, roughly following GNU Libc docs example for a similar task.

jimklimov:
Tying up some more loose ends - now about macros instead of numbers for stdin/out/err, missed one file.

Issue #1967, opened 2023-06-16
PowerWalker VI 1200 SHL regressions in NUT 2.8.0


blazzakrajsek:
I have noticed that upgrade of Debian 11 (bullseye) to Debian 12 (bookworm) upgraded the `nut` package from 2.7.4 to 2.8.0.
The support for PowerWalker VI 1200 SHL is now worse, see my previous report #1270. I have noticed that the driver has changed from `MGE HID 1.40` to `Phoenixtec/Liebert HID 0.41`.

Current status under NUT 2.8.0 for PowerWalker VI 1200 SHL is as follows...
`ups.conf`:
```
[ups0]
    driver = "usbhid-ups"
    port = "auto"
    vendorid = "06DA"
    productid = "FFFF"
```

`upsc ups0`:
```
Init SSL without certificate database
battery.charge: 100
battery.runtime: 7200
battery.type: PbAc
battery.voltage: 27.00
device.mfr: PPC
device.model: Offline UPS
device.serial: 000000000   
device.type: ups
driver.name: usbhid-ups
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.port: auto
driver.parameter.productid: FFFF
driver.parameter.synchronous: auto
driver.parameter.vendorid: 06DA
driver.version: 2.8.0
driver.version.data: Phoenixtec/Liebert HID 0.41
driver.version.internal: 0.47
driver.version.usb: libusb-1.0.26 (API: 0x1000109)
input.voltage: 237.0
output.frequency: 49.80
output.voltage: 237.0
ups.load: 0
ups.mfr: PPC
ups.model: Offline UPS
ups.productid: ffff
ups.serial: 000000000   
ups.status: OL CHRG
ups.vendorid: 06da
```

`upsrw -l ups0`:
```
(empty)
```

`upscmd -l ups0`:
```
Instant commands supported on UPS [ups0]:
(empty)
```

Commands and variables are now not available any more. For example, with the `shutdown.return` command missing, UPS shutdown is not possible.

I guess (as I did not test) that the fixes implemented by #1357 made support for Phoenixtec 0x06DA:0xFFFF devices in general poorer.

jimklimov:
Thanks for the report! it would really be ibteresting to know if that PR is the culprit - e.g. if its disablement for non-AEG branded devices should be expanded to more names, or made a config toggle...

blazzakrajsek:
[...]

PR #1968, opened 2023-06-16
Improved Phoenixtec 0x06DA:0xFFFF support

blazzakrajsek:
- Add additional properties by analyzing the HID descriptor of the PowerWalker VI 1200 SHL
- Existing properties were kept as they are to avoid breakage with other devices using the same driver
- See #1967 

## General points

  comments or text documents.

  case.

  or its equivalent.

  concept.

- [x] Propose a PR for NUT DDL with detailed device data dumps from tests

  the changed codebase.

### DIFF
--- a/drivers/liebert-hid.c
+++ b/drivers/liebert-hid.c
@@ -86,8 +86,8 @@ static hid_info_t liebert_hid2nut[] = {
   /* Battery page */
   { "battery.voltage", 0, 0, "UPS.PowerSummary.Voltage", NULL, "%.2f", 0, NULL },
   { "battery.charge", 0, 0, "UPS.PowerSummary.RemainingCapacity", NULL, "%.0f", 0, NULL },
-  { "battery.charge.capacity.design", 0, 0, "UPS.PowerSummary.DesignCapacity", NULL, "%.0f", 0, NULL },
-  { "battery.charge.capacity.remaining", 0, 0, "UPS.PowerSummary.FullChargeCapacity", NULL, "%.0f", 0, NULL },
+  { "experimental.battery.capacity", 0, 0, "UPS.PowerSummary.FullChargeCapacity", NULL, "%.0f", 0, NULL },
+  { "experimental.battery.capacity.nominal", 0, 0, "UPS.PowerSummary.DesignCapacity", NULL, "%.0f", 0, NULL },
   { "battery.runtime", 0, 0, "UPS.PowerSummary.RunTimeToEmpty", NULL, "%.0f", 0, NULL },
   { "battery.type", 0, 0, "UPS.PowerSummary.iDeviceChemistry", NULL, "%s", 0, stringid_conversion },
 
--- a/drivers/liebert-hid.c
+++ b/drivers/liebert-hid.c
@@ -5,6 +5,7 @@
  *  2005 - 2006 Peter Selinger <selinger@users.sourceforge.net>
  *  2007        Charles Lepple <clepple@gmail.com>
  *  2018        Markus "Links2004" <help.markus+gitk@gmail.com>
+ *  2023        Blaž Zakrajšek <zakrajsek.blaz@gmail.com>
  *
  *  This program is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


drivers/liebert-hid.c
```diff
@@ -71,28 +72,43 @@ static hid_info_t liebert_hid2nut[] = {
...
+  { "unmapped.ups.powersummary.iserialnumber", 0, 0, "UPS.PowerSummary.iSerialNumber", NULL, "%.0f", 0, NULL },
+  { "unmapped.ups.powersummary.serialnumber", 0, 0, "UPS.PowerSummary.iSerialNumber", NULL, "%s", 0, stringid_conversion },
 #endif	/* if WITH_UNMAPPED_DATA_POINTS */
 
+  /* Battery page */
   { "battery.voltage", 0, 0, "UPS.PowerSummary.Voltage", NULL, "%.2f", 0, NULL },
-  { "battery.voltage.nominal", 0, 0, "UPS.PowerSummary.ConfigVoltage", NULL, "%.2f", HU_FLAG_STATIC, NULL },
   { "battery.charge", 0, 0, "UPS.PowerSummary.RemainingCapacity", NULL, "%.0f", 0, NULL },
+  { "battery.charge.capacity.design", 0, 0, "UPS.PowerSummary.DesignCapacity", NULL, "%.0f", 0, NULL },
+  { "battery.charge.capacity.remaining", 0, 0, "UPS.PowerSummary.FullChargeCapacity", NULL, "%.0f", 0, NULL },
```
jimklimov, line 90: Names in the standardly-built mapping table should match `docs/nut-names.txt` - please check if there is a suitable concept there, or go to nut-upsdev mailing list to discuss with the community a standardization of new naming patterns.

Probably the "design" points are "nominal" in Standard Speak. 

I am not sure a "FullChargeCapacity" (per name) would be a "capacity.remaining", would rather see it as another sort of "nominal", probably :\

Note there may be several mappings for one NUT name, since not all devices serve all the data points: first-come-first-serve wins.

jimklimov (changes requested):

drivers/liebert-hid.c
blazzakrajsek: `docs/nut-names.txt` contains the `battery.capacity` and `battery.capacity.nominal` entries, but these are in `Ah` units. UPS in my case returns data for those two entries in `%` units (`UPS.PowerSummary.CapacityMode, Type: Feature, ReportID: 0x0b, Offset: 24, Size: 8, Value: 2`). Should I still use these names although the units don't match (`Ah` and `%`)?

jimklimov: Feeling confused here... does it return *charge* (percentage) or *capacity* (Ah) for these queries? Maybe re-post explicitly, which numbers it says for which of these reports? :'\

I mean, "design capacity = 100%, old battery capacity = 70%" is sort of meaningful for maintenance like battery replacement, but moderately useless for run-time decisions - especially for the former which should not be not-100% I suppose :)

jimklimov: Also note that for some situations the support for conversion functions may be applicable, see e.g. `mge_battery_capacity` in `mge-hid.c`.

Looking at precedents, a few drivers define `battery.capacity` as a result of either of those queries (first hit wins IIRC), although use of `.nominal` as posted in your update seems proper here.

blazzakrajsek: Sorry if I wasn't clear, I will try better...

`nut-names.txt` has this:
```
...
| battery.charge               | Battery charge (percent)            | 100.0
...
| battery.capacity             | Battery capacity (Ah)               | 7.2
| battery.capacity.nominal     | Nominal battery capacity (Ah)       | 8.0
...
```

Extract of the relevant data (#1270 has this information in more detail):
```
Path: UPS.PowerSummary.DesignCapacity, Type: Feature, ReportID: 0x0b, Offset: 0, Size: 8, Value: 100
Path: UPS.PowerSummary.CapacityGranularity1, Type: Feature, ReportID: 0x0b, Offset: 8, Size: 8, Value: 1
Path: UPS.PowerSummary.FullChargeCapacity, Type: Feature, ReportID: 0x0b, Offset: 16, Size: 8, Value: 100
Path: UPS.PowerSummary.CapacityMode, Type: Feature, ReportID: 0x0b, Offset: 24, Size: 8, Value: 2
Path: UPS.PowerSummary.RemainingCapacity, Type: Feature, ReportID: 0x06, Offset: 0, Size: 8, Value: 100
```

`CapacityMode` has value `2`, meaning that `DesignCapacity`, `CapacityGranularity1`, `FullChargeCapacity` and `RemainingCapacity` use units `%`. I will explain further...

`Capacity Mode` can have these values[1]:
```
Capacity Mode DV Battery capacity units are as follows:
• 0: maH, (used in SMB)
• 1: mwH (used in SMB)
• 2: %
• 3: Boolean support only (OK or failed)
```

`Design Capacity` description[1]:
```
Design Capacity SV The theoretical capacity of a new pack. (See CapacityMode for units.)
```
... so this would be `battery.capacity.nominal`, but since it's in `%`, battery designed capacity is always `100 %` :)
It's not of much use in `%` units, but would be useful, for example in other units like `maH`?

`Capacity Granularity 1` description[1]:
```
Capacity Granularity 1 SV Battery capacity granularity between low and warning. (Units are
defined by CapacityMode.)
```
... is a step value of all the properties mentioned. Currently I have not made it an entry, but would be needed if the calculation is done via a function. Above from the extract it has a value of `1 %`.

`Full Charge Capacity` description[1]:
```
Full Charge Capacity DV The predicted pack capacity when it is fully charged. (See CapacityMode for units.)
```
... so this would be `battery.capacity`. With the new battery it should be `100 %`, but with an older battery I would expect for that value to drop. 

`Remaining Capacity` description[1]:
```
Remaining Capacity DV The predicted remaining capacity. (See CapacityMode for units.)
```
... so this would be `battery.charge`. Fully charged battery has `100 %`, depleted `0 %`, so this does show the exact state of charge of the battery in `%`.

So all in all, probably I can try to write the calculation functions for the correct presentation of the Capacity and Charge according to the Capacity Mode and Granularity.
But that would be a new merge request, or should I put it here as well?

[1] https://www.usb.org/sites/default/files/hut1_4.pdf , page 375 and onwards, there are contained links to descriptions of the separate entries

jimklimov: Great thanks for digging and detailing this!

I think yes - to move forward quickly, it is better to put these entries into experimental for now so the "less questionable" ones can be merged.

I am not in fact sure if there would then be a good place for these percentages in the NUT data model (unless we find elsewhere the "true nominal" Ah values (or mW or other real energy/power measurements); however with this practical device it seems like it would be useful to recognize the `CapacityMode` to convert or ignore the readings.

Also wondering: can it be *set*? Can we actively ask your UPS to report `Ah` instead of `%`? That might be an even better twist :)

blazzakrajsek: Agreed, I will move those in question under experimental namespace for now.

I agree it's not _that_ useful having `%` units, but (as you said earlier) it might be useful for indication / maintenance if at least these data are provided by the UPS. This is by all standards a quite cheap UPS, in the specifications it's noted that it has 2 x 7Ah batteries, but I don't believe it has a capability to detect this or have a value programmed in firmware. The manufacturer did not even bother to write the serial number properly (it's all zeros). :/

Tried something like...
```
{ "experimental.battery.capacity.mode", ST_FLAG_RW | ST_FLAG_NUMBER, 1, "UPS.PowerSummary.CapacityMode", NULL, "%.0f", 0, NULL },
 ```
And issuing the command:
```
upsrw -u <user> -p <password> -s experimental.battery.capacity.mode=1 -w -t 10 ups0
ERR UNKNOWN
```
Shall I try anything differently?

Comment (jimklimov) [reply drivers/liebert-hid.c]: Regarding battery replacement, I guess a (user-configurable?) low enough percentage could be an reason to raise the `RB` => `REPLACEBATT`=> `"replacebatt"` flag, see e.g. belkin-hid.c for an example, although there a direct report from the UPS is considered - but still, a conversion function that makes a decision, so might as well compare to threshold %:
````
  { "BOOL", 0, 0, "UPS.PowerSummary.PresentStatus.NeedReplacement", NULL, NULL, HU_FLAG_QUICK_POLL, liebert_replacebatt_info },
````

There are a few other hits to this string in other sources too, if you care :)

Regarding the `upsrw`=>UNKNOWN, it seems that the (running, daemonized) driver did not register the value with `upsd` so it does not know the setting is supported there. Is this a driver (and/or server) running from the build workspace, or installed into the system?

For running in the workspace, see e.g. how `tests/NIT/nit.sh` sets up the freshly-built drivers to talk with a freshly-built `upsd` for the test suite (preparing configs and pointing to them via env vars). In fact, I commonly hijack those after a custom `make check-NIT` to add some more driver definitions to test something other than a `dummy-ups`.

Comment (blazzakrajsek) [reply drivers/liebert-hid.c]: Thanks for the pointers, much appreciated! :)

I have NUT installed (but not running and systemd units disabled) as well as running from the workspace (for building and testing).
When testing this it could be that I did not restart the `upsd` after building anew, however after testing this now, `upsrw -l ups0` does report the command as supported:
```
...
[experimental.battery.capacity.mode]
Description unavailable
Type: NUMBER
Value: 2
...
```

I have also tested the variable with `HU_FLAG_ABSENT`, in that case I get a `SUCCESS`, but I believe nothing was actually set on the UPS. Also, currently defined as a `ST_FLAG_NUMBER`, I also did test with `ST_FLAG_STRING`, but the results from `upsrw` are the same.

PR #1977, opened 2023-06-30
Use `msgfmt` to "compile" translations for NUT-Monitor (Python GUI app)

jimklimov:
Previously these files were somehow generated by someone and git-committed, and recently happened to be "just changed" by https://github.com/networkupstools/nut/pull/1976/commits/0bea0af6a4d59ef657b8703d61635415ebffdddd which presumably makes them incorrect binaries.

This PR adds an ability to `make` the translation binary `*.mo` files from source `*.po` files, and updates the existing ones.

For now they remain in git, but maybe to avoid surprises across CI farm on nodes that have `msgfmt` - they should go away and only be `EXTRA_DIST`ed into tarballs (should then require `msgfmt` unlike what this PR offers as a first shot).


jimklimov:
It seems that the files would have to go away from Git... different builds/implems of `msgfmt` (e.g. GNU vs. those on Sun-heritage systems) produce different binaries.

jimklimov:
Taking https://github.com/sphinx-doc/sphinx/issues/3443 and https://github.com/sphinx-doc/sphinx/pull/3490 into account...

PR #1979, opened 2023-07-04
Add notes about systemd-hwdb subsystem docs

jimklimov:
Follow-up to #1342


jimklimov:
See also http://who-t.blogspot.com/2019/02/adding-entries-to-udev-hwdb.html about more details on editing HWDB in a live system, and possibly more tricks to apply it.

jimklimov:
Posted https://github.com/networkupstools/nut/wiki/Maintaining-systemd-udev-HWDB-for-NUT

jimklimov:
UPDATE: Reviving discussion in #1342 - not convinced now that *this* is the file we want edited about systemd udev/hwdb notes. It is about UPower configuration, which is similar and also from FreeDesktop project, but is not udev, right?..

aquette:
UPower is the integrated power manager, descendant of the HAL, and relating to the HAL NUT work. Only USB/HID, DBus and udev for privileges...

marcelodotnet:
@jimklimov I took another look at this PR against the current `master` and the current NUT documentation layout.

One thing that seems worth clarifying is that the UPower HWDB and NUT's USB permission rules still serve different purposes: `scripts/upower/95-upower-hid.hwdb` identifies supported HID UPS devices for UPower, while `scripts/udev/nut-usbups.rules.in` is still used for NUT device permissions.

Since the HWDB file is generated by `tools/nut-usbinfo.pl` and can also be consumed downstream by UPower, would it make sense to keep its generated header relatively short and put the detailed installation/update notes into a new `scripts/upower/README.adoc`, similar to `scripts/udev/README.adoc`?

That README could document the HWDB purpose, `/usr/lib/udev/hwdb.d` vs `/etc/udev/hwdb.d`, local overrides, `systemd-hwdb update`, and re-triggering already-connected devices when needed.

This would also avoid implying that the UPower HWDB replaced NUT's USB permission rules. If this direction makes sense, I'd be happy to prepare the documentation update.

jimklimov:
Sounds like a good idea, one thing that worries me are minimized installations that forgo documentation files. The reduced header should probably still refer to at least the NUT website (or github URL to the script) and people might pick their way from there...

PR #1982, opened 2023-07-07
Added more variables from the CyberPower MIB

cholyoak:
Added some more fields from the CyberPower MIB (v2.11) including:

* `ups.id`
* `ups.status` (adding hardware status from `CPS-MIB::upsStatus.0`
* `ups.temperature`
* `battery.date`
* `input.transfer.reason`
* `ups.test.date`
* `ups.test.result`


jimklimov:
Thanks! Can you please check against docs/nut-names.txt for standard naming to common concepts (and bring up on nut-upsdev mailing list if more need to be introduced)?

The `ups.status` in particular has a fixed set of values or combinations (OL, OB, CHRG etc.)

jimklimov:
Hi, somehow the question in mailing list did not gain traction, however your idea there about putting these status values (overheat, HW fault) into alarms/alerts instead sounds interesting.

While https://github.com/networkupstools/nut/blob/master/docs/new-drivers.txt#L223-L239 defines quite a lot of `ups.state` field values, they are almost all about power and less so about the device. A few that are on the edge by wording are "The battery needs to be replaced" and "UPS bypass circuit is active -- no battery protection is available"...

The next chapter there about UPS Alarms having a less pre-defined vocabulary, and particularly the example in https://github.com/networkupstools/nut/blob/3a11558e66334510309758f1728dd55f572620e5/docs/nut-names.txt#L128 (literally `OVERHEAT` as a sample `ups.alarm` value), suggest that this is a better route to follow.

cholyoak:
I've moved the `ups.status` changes to `ups.alarm`. Not I haven't used the `OVERHEAT` value as other drivers seem to use more textual values. Let me know if this is a problem.

jimklimov:
Thanks, looks reasonable. I refreshed some maintenance cosmetics, waiting for CI give its verdict just in case... :)

Issue #1994, opened 2023-07-18
CI/Legalese: Require DCO (`Signed-off-by:` git tags) in commits


jimklimov:
For a long time many (but not all, myself included) NUT contributors have signed off their commits to conform to the legal practice of shared development. I've seen a GitHub app https://github.com/apps/dco which allows to enforce the presence of such signatures, worth adding it to the CI checks.

* https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for

From https://developercertificate.org/:

````
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
````

Sign-off can be marked by use of `-s` (note also PGP signing for those with a key, with `-S`):
````
:; git commit -s -m 'DIR/FILENAME: something about this change'
````

Note that while commit hook tricks like those discussed in https://stackoverflow.com/questions/15015894/git-add-signed-off-by-line-using-format-signoff-not-working are available to automatically sign off all commits, these signatures are intended to be a conscious (legally meaningful) act -- hence not automated in git core. A `format.signoff` setting is for something else.

For PGP signing, see more at:
* https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
* https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key

jimklimov:
App installed into NUT organization.

jimklimov:
Also of interest for (custom) CI farms may be https://github.com/christophebedard/dco-check

[...]

PR #1995, opened 2023-07-18
CI/Legalese: Require DCO (`Signed-off-by:` git tags) in commits

jimklimov:
Closes: #1994

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,6 +4,11 @@
   as for which text documents to update. See also docs/developer-guide.txt
   for general points on NUT architecture and design.
 
+* Please note that we require "Signed-Off-By" tags in each Git Commit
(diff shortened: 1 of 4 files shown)


jimklimov:
Thanks for the question :)

This is not exactly an overall codebase license as such (a better filename is welcome), rather an acknowledge license grant by particular patch/PR author(s). They "sign off" an agreement that they contribute their piece of code to be used under the project's license (GPLv2+ in case of most of NUT; some other files and directories were borrowed elsewhere with their compatible licenses - see https://github.com/networkupstools/nut/blob/master/COPYING for details).

aquette (approved):

jimklimov:
Anyhow, the GitHub org configuration was changed back when this was posted - so the PR just clarifies the rules in text. Merging of contributions is still up to maintainers, whether CI checks passed or not.

PR #1239, opened 2021-12-23
generic_modbus driver - handle connection errors (broken pipe), configurable modbus timeouts

dtsecon:
generic_modbus driver now handles connection errors (broken pipe) with modbus tcp server devices.  Response time out between requests and between frame bytes are now configurable via ups.conf to handle "invalid data" errors. The new code have been tested.

configuration options for modbus response and byte timeouts in ups.conf
`# Modbus timeouts`
`mod_resp_to_s =  2`
`mod_resp_to_us = 0`
`mod_byte_to_s = 0`
`mode_byte_to_us = 20000`

Issue #1238, opened 2021-12-23
I'd like to add support for Armac UPS


blaa:
Hello,
  I have http://armac.pl/oferta/ups-rack-line-interactive-r2000ipsw/ and should be able to add support for it, but I wonder how to do it best and if there's some place to talk/consult the approach. I've sniffed and reveng a bit it's "native" 2005-old, x86, hideous driver. 

Bus 002 Device 008: ID 0925:1234 Lakeview Research UPS USB MON V1.4

It has USB port with "emulated serial" over control + interrupt transfers. It generally supports Q1-like protocol (I found Q1\r, I, F, Q for alarm, T01 for test commands).  Its driver seems to support the richcomm protocol but it doesn't seem to be used with this model, and richcomm_usb doesn't work with it either (connects, but decodes the status badly).

Commands are 4 byte long, whereas answers are *6* byte long. As far as I understand both Blazer and nutdrv_qx drivers use 8 byte long transfers which is(?) the main problem.

What should I do? Copy some matching subdriver of nutdrv_qx and try to code it there? Use blazer? Or maybe I'm missing something obvious?

Cheers,

jimklimov:
I hope a little that @zykh would chime in with more details, but haven't heard from our resident Qx protocol expert lately :\

In general, the best way forward would be to add a subdriver for nutdrv_qx, others may have been there first, but duplicate a lot of code and should be eventually obsoleted. That said, many of those older drivers for Qx devices did become parts of nutdrv_qx.c or specific subdrivers, so I suppose if you make it work by injecting new logic into an older driver, you would be able to import it into the new one as well.

blaa:
Linked driver from the producer site is http://armac.pl/wp-content/uploads/2018/08/PowerManagerII-EN.7z which doesn't seem to be really that Armac-specific. It's rather old (2005-2009) too. According google-found https://www.pythontr.com/makale/debian-wheezy-icin-ups-kurulumu-409 this might work with Digitech UPS. Vendor/product and partially protocol matches richcomm_usb (which is a limited experimental driver).

Grepping those names doesn't help finding easier solution than implementing the subdriver though, but maybe after implementation wider set of UPSes could be supported.

I'll try to do a PoC based on "fabula" if no better idea appears. ;)

blaa:
PoC:
```c
int armac_command(const char *cmd, char *buf, size_t buflen)
{
	int	ret = 0;
	size_t	i;
  int index = -1;

	struct {
      const char *str;
[...]

PR #2003, opened 2023-07-31
Problem with NUTDRV_QX, subdriver armac (add EOL support)

jimklimov:
Closes: #1978

Thanks @blaa for proposing the fix, and @convicte for testing it.

### DIFF
--- a/drivers/nutdrv_qx.c
+++ b/drivers/nutdrv_qx.c
@@ -1838,14 +1838,20 @@ static int	armac_command(const char *cmd, char *buf, size_t buflen)
 			tmpbuf[0], tmpbuf[1], tmpbuf[2], tmpbuf[3], tmpbuf[4], tmpbuf[5],
 			tmpbuf[1], tmpbuf[2], tmpbuf[3], tmpbuf[4], tmpbuf[5]);
 
+		/* This includes status/length byte on R/3000I/PF1 */
 		bytes_available = (unsigned char)tmpbuf[0] & 0x0f;
 		if (bytes_available == 0) {
 			/* End of transfer */
 			break;
 		}
 
-		memcpy(buf + bufpos, tmpbuf + 1, bytes_available);
-		bufpos += bytes_available;
+		memcpy(buf + bufpos, tmpbuf + 1, bytes_available - 1);
+		bufpos += bytes_available - 1;
+
+		if (tmpbuf[bytes_available - 1] == 0x0d) {
+			/* End of line is an end of the message as well */
+			break;
+		}
 
 		if (bytes_available <= 2) {
 			/* Slow down, let the UPS buffer more bytes */
--- a/NEWS
+++ b/NEWS
@@ -99,6 +99,8 @@ as part of https://github.com/networkupstools/nut/issues/1410 solution.
  - nutdrv_qx updates:
    * the `voltronic_qs_protocol` should now accept both "V" (as before)
      and newly "H" dialects, which otherwise seem interchangeable [#1623]
+   * the `armac` subdriver was enhanced to support devices with a different
+     response pattern than previously expected per initial contribution [#1978]
 
  - usbhid-ups updates:
    * cps-hid subdriver now applies same report descriptor fixing logic to


blaa:
Hey, waaaait a bit with this. ;-) As I've wrote in the issue - this might break compatibility with other Armac/powermanagerii devices though. 

The end of line test is certainly SAFE, but the -1 added to the copying look rather suspicious. I want to test it with my UPS (on my vacation and can do this in september), or get a traffic dump from the @bdkacz for whom the previous version worked. 

So, blindly merging that can break compatibility. ;s

jimklimov:
@blaa: I have a question about the logic in this proposed change:
* previously, you received the `bytes_available` amount of bytes via `tmpbuf` populated by `usb_interrupt_read()`, and either copied them all to `buf` or saw a zero-sized reply and aborted.
* now you copy one byte less to `buf` and drop out if the remaining byte is `0x0d`; however you do not copy it into `buf` if it is something else. And then I think you loop to `usb_interrupt_read()` again so lose the old `tmpbuf` contents.

Am I right with a gut feeling that you would lose one character in the middle of a reply from each loop here under the "right" conditions? Should there be an `else` to that comparison to `0x0d` clause?

jimklimov:
Converted to "draft" as asked by @blaa (to avoid merge before regression testing) ;)

drivers/nutdrv_qx.c
```diff
@@ -1838,14 +1838,20 @@ static int	armac_command(const char *cmd, char *buf, size_t buflen)
...
 			/* End of transfer */
 			break;
 		}
 
-		memcpy(buf + bufpos, tmpbuf + 1, bytes_available);
-		bufpos += bytes_available;
+		memcpy(buf + bufpos, tmpbuf + 1, bytes_available - 1);
+		bufpos += bytes_available - 1;
+
+		if (tmpbuf[bytes_available - 1] == 0x0d) {
```
blaa, line 1851: This should be safe in all cases... (although it uses the bytes_available as well and it's suspicious that it worked without -1 previously).

jimklimov: At least, it is unsigned and non-zero (per previous checks), so array-wise the expression is safe :)

drivers/nutdrv_qx.c
```diff
@@ -1838,14 +1838,20 @@ static int	armac_command(const char *cmd, char *buf, size_t buflen)
...
+		/* This includes status/length byte on R/3000I/PF1 */
 		bytes_available = (unsigned char)tmpbuf[0] & 0x0f;
 		if (bytes_available == 0) {
 			/* End of transfer */
 			break;
 		}
 
-		memcpy(buf + bufpos, tmpbuf + 1, bytes_available);
-		bufpos += bytes_available;
+		memcpy(buf + bufpos, tmpbuf + 1, bytes_available - 1);
```
blaa, line 1848: This doesn't look like a simple mistake in dropping `-1`. My UPS rather wouldn't work with it. Maybe it needs some kind of a detection on how this byte is coded. I need traces from both UPSes to make certain it will cover both cases.


Maaaybe the original issue were I did this driver has one...

jimklimov: That would be #1238 and #1239 I believe.

jimklimov: Otherwise, initially copying all of `bytes_available` in the original loop seemed reasonable. Say the usb method returned one new byte, here now you'd copy `1 - 1 == 0` bytes (and check if that one is end of line)...

blaa: there seem to be only a trace where UPS send 1 byte at a time during each interrupt transfer:
```
                      [D3] armac_command: Got a known command 'Q1\r'                         
   2.121903     [D3] read: ret 6 buf c1: 28 30 30 30 30  >(0000<                  
   2.126943     [D3] read: ret 6 buf 81: 32 30 30 30 30  >20000<
   2.130978     [D3] read: ret 6 buf c1: 33 30 30 30 30  >30000<
   2.133998     [D3] read: ret 6 buf 81: 38 30 30 30 30  >80000<      
(...)
```

Here bytes available (low nibble: tmpbuf[0] & 0x0f)  seems to be equal to 1 and convey a 1 byte. Interrupt transmits 6 bytes at a time, but the rest is filled with literal ASCII 0. (would be cool if they used ZERO 0x00).

Current trace:
```
   0.083301     [D4] armac command Q1
   0.164847     [D4] read: ret 6 buf a6: 28 32 34 31 2e  >(241.<
   0.184839     [D4] read: ret 6 buf 86: 35 20 30 30 30  >5 000<
   0.205851     [D4] read: ret 6 buf a6: 2e 30 20 32 33  >.0 23<
   0.226849     [D4] read: ret 6 buf 86: 30 2e 33 20 30  >0.3 0<
```
The first nibble oscillates between `a` and `8` - previously `c` and `8`. 


But TBH, that seems like an easy fix. I'd assume this `6` means this UPS has available 6 bytes in a buffer. This interrupt transfer transferred 5... it can't transfer more. It's just FASTER than the previous model.


Instead of `-1` one could:
```
bytes_available = (unsigned char)tmpbuf[0] & 0x0f;
if (bytes_available == 0) {
    /* End of transfer */
    break;
}
if (bytes_available > 5) {
    /* More bytes in UPS buffer, but this transfer only handles up to 5. */
    bytes_available = 5;
}
```

Or even use the 'ret' from usb interrupt in case it can transfer more later?

```
if (bytes_available > ret - 1) {
    /* More bytes in UPS buffer, but this transfer only handles up to 5. */
    bytes_available = ret - 1;
}
```

blaa: My current take is this one:
```
diff --git a/drivers/nutdrv_qx.c b/drivers/nutdrv_qx.c
index 350c555a0..550eb4071 100644
--- a/drivers/nutdrv_qx.c
+++ b/drivers/nutdrv_qx.c
@@ -1825,6 +1825,9 @@ static int        armac_command(const char *cmd, char *buf, size_t buflen)
                 * (which will happen after successfully writing a command
                 * to the UPS) */
                if (ret != 6) {
+                       /* NOTE: If end condition is invalid for particular UPS we might make one
+                        * request more and get this error. If bufpos > (say) 10 this could be ignored
+                        * and the reply correctly read. */
                        upsdebugx(1,
                                "interrupt read error: %s (%d)",
                                ret ? nut_usb_strerror(ret) : "timeout",
@@ -1838,15 +1841,32 @@ static int      armac_command(const char *cmd, char *buf, size_t buflen)
                        tmpbuf[0], tmpbuf[1], tmpbuf[2], tmpbuf[3], tmpbuf[4], tmpbuf[5],
                        tmpbuf[1], tmpbuf[2], tmpbuf[3], tmpbuf[4], tmpbuf[5]);
 
+               /*
+                * On most tested devices (including R/2000I/PSW) this was equal to the number of
+                * bytes returned in the buffer, but on some newer UPS (R/3000I/PF1) it was 1 more
+                * (1 control + 5 bytes transferred and bytes_available equal to 6 instead of 5).
+                *
+                * Current assumption is that this is number of bytes available on the UPS side
+                * with up to 5 (ret - 1) transferred.
+                */
                bytes_available = (unsigned char)tmpbuf[0] & 0x0f;
                if (bytes_available == 0) {
                        /* End of transfer */
                        break;
                }
+               if (bytes_available > 5) {
+                       /* Single interrupt transfer has 1 control + 5 data bytes */
+                       bytes_available = 5;
+               }
 
                memcpy(buf + bufpos, tmpbuf + 1, bytes_available);
                bufpos += bytes_available;
 
+               if (tmpbuf[bytes_available] == 0x0d) {
+                       /* End of line is an end of the message as well */
+                       break;
+               }
+
                if (bytes_available <= 2) {
                        /* Slow down, let the UPS buffer more bytes */
                        usleep(15000);
```

It compiles, but could use some testing. ;-)

jimklimov: I'm still a bit worried about losing some data in reads:
* What if if it does not return exactly 6 bytes? Here the `ret - 1` idea sounded better than hard-coded `5`... (although we do hard-code a `6` above too).
  * Side idea: should we deduct 5 maybe? `if (bytes_available > 5) bytes_available -= 5;` to read up the remainder? Or can/should we rely that the next call to `usb_interrupt_read()` returns a suitable number (perhaps if it buffered some info, so the remainder after deducting 5 is worthless anyway)... Even more: *can* we rely on that if it returned a "bogus" value now?
* What if the last byte in current `tmpbuf` was not `0x0d`, per my initial question - should it be copied?
* And for good measure: is there a chance that the UPS returns several lines (so a buffer does not end with an `0x0d` but has it in the middle, and so the next line of same or different reply starts in the next byte)?

jimklimov: Hm, looking at the data dumps in https://github.com/networkupstools/nut/issues/1978#issuecomment-1652657354 :

````
   0.414746     [D4] read: ret 6 buf 86: 2e 30 20 30 30  >.0 00<
   0.435756     [D4] read: ret 6 buf a6: 30 30 30 30 30  >00000<
000<.939819     [D4] read: ret 6 buf 83: 31 0d 30 30 30  >1
   1.940207     [D1] interrupt read error: Operation timed out (-7)
````

So for a line where we have lower nibble (first byte `& 0x0f`) equal to `6` and `ret==6`, we have 5 protocol content bytes. However in the last practical line, with the 0x0d, the nibble is `3` and there are 2 content bytes (and the UPS stops sending more, hence the timeout).

I believe, what the driver did originally, with your UPS where its nibble was `1` it copied one byte into the buffer. Where the nibble was `6` it probably copied 6 bytes (starting from `tmpbuf+1` so overflowing the `tmpbuf` defined with size 6 exactly... wondering why compilers or run-times did not complain/segfault - not a write so don't care?) and probably adding garbage into that extra byte of the resulting string for each loop :\

blaa: Yeah, I've dropped ret - 1 because 6 is hardcoded above. (Original driver does reads always by 6 bytes). But that's equivalent - if ret != 6 we bail out anyway.

In general we can introduce a constant: `const int READ_SIZE = 6;`
Use it in usb_interrupt_read, in following if and later instead of `ret` - it will be cleaner and I can prepare a patch like that.

Architecture-wise, I believe the UPS works like this:
`([controller] --uart or rs232-- [usb bridge with buffer]) --usb-- [PC]`
And this USB interface is a really poor USB Serial implementation. ;) 

So the `bytes_available` is not a length of a whole serial buffer, just the amount that currently is stored within the usb bridge. MAYBE. As far as we can tell. It's important to differentiate a TRUE data from a fill (0, 0x30), when there's less data then the read size certainly. Would not be necessary if the fill was 0x00. I think deducting 5 is worthless, but we could store info like `more_available = true` if it's larger than `READ_SIZE - 1`. 

Unsure what to do with this information though. If we find `\x0d` but `more_available == true` that's probably an error and we could print a message - not much more though.

blaa: > Hm, looking at the data dumps in [#1978 (comment)](https://github.com/networkupstools/nut/issues/1978#issuecomment-1652657354) :
> 
> ```
>    0.414746     [D4] read: ret 6 buf 86: 2e 30 20 30 30  >.0 00<
>    0.435756     [D4] read: ret 6 buf a6: 30 30 30 30 30  >00000<
> 000<.939819     [D4] read: ret 6 buf 83: 31 0d 30 30 30  >1
>    1.940207     [D1] interrupt read error: Operation timed out (-7)
> ```
> 
> So for a line where we have lower nibble (first byte `& 0x0f`) equal to `6` and `ret==6`, we have 5 protocol content bytes. However in the last practical line, with the 0x0d, the nibble is `3` and there are 2 content bytes (and the UPS stops sending more, hence the timeout).

Watchout for the \r that shifts the 0 in the output in those lines. But yes, nibble is 3 and there are pretty much 2 bytes sent. I forgot about it today - and it suggest I should scan for \d and not just check the last byte.

> 
> I believe, what the driver did originally, with your UPS where its nibble was `1` it copied one byte into the buffer. Where the nibble was `6` it probably copied 6 bytes (starting from `tmpbuf+1` so overflowing the `tmpbuf` defined with size 6 exactly... wondering why compilers or run-times did not complain/segfault - not a write so don't care?) and probably adding garbage into that extra byte of the resulting string for each loop :\

Driver when the nibble was 6 could be reading 1 byte over tmpbuf. C compilers and run-times are quirky. :S In Rust that would certainly be a runtime Panic, maybe compilation error. tmpbuf is on the stack, so there's a big chance it was reading 1 byte more from the stack just.

Certainly adding a garbage - that changed the meaning of this nibble in a way not observed in other UPSes.

I'll take a look at this final byte too, maybe it needs special handling to find \r correctly.

jimklimov: I think it will be more productive for you to cherry-pick the changes I've proposed so far and start another PR from your fork to override this one, to avoid me posting your patches from discussion into git :) WDYT?

Issue #2007, opened 2023-08-03
apcupsd-ups driver race condition


tbsky:
Hi:
   we have more and more new APC ups which need to use nut apcupsd-ups driver (with serial modbus supported in apcupsd).
they work fine under nut 2.7.4, but it is not good under nut 2.8.0.

   the difference is that the default nut apcupsd-ups driver polling interval is 60 seconds under 2.7.4 and change to 2 seconds under 2.8.0. the default normal polling interval settings for apcupsd is also 60 seconds.  so nut 2.7.4 seems co-exist with apcupsd well.

  we need to record the ups data like "ups.load" history so we use command "upsc myups" to read the data. and we found sometimes there are no "ups.load" data under 2.8.0. I write a little script and try to execute "upsc myups" every second and found it will fail to read the full data about 60 seconds interval. the normal full data for "upsc myups" is like below:

````
battery.charge: 100.0
battery.charge.low: 5
battery.date: 2021-11-10
battery.runtime: 1440.0
battery.runtime.low: 180
battery.voltage: 26.2
device.mfr: APC
device.type: ups
driver.name: apcupsd-ups
driver.parameter.pollinterval: 2
driver.parameter.port: localhost
driver.parameter.synchronous: auto
driver.version: 2.8.0
driver.version.internal: 0.6
input.frequency: 59.8
input.transfer.reason: Line voltage notch or spike
input.voltage: 105.7
output.current: 4.03
output.frequency: 59.8
output.voltage: 121.5
output.voltage.nominal: 120
power.percent: 33.9
ups.date: 2023-08-02
ups.delay.shutdown: 0
ups.delay.start: 0
ups.firmware: UPS 04.4
ups.firmware.aux: 00.5
ups.id: APC UPS
ups.load: 47.1
ups.mfr: APC
ups.mfr.date: 2021-11-10
ups.model: Smart-UPS 1500
ups.power.nominal: 1440
ups.realpower.nominal: 1000.0
ups.serial: AS2145160441
ups.status: BOOST OL
ups.temperature: 34.0
ups.test.result: OK
ups.time: 17:46:18
````
===================================================

when there is racing, "upsc myups" returns partial data like below:

````
battery.charge: 100.0
battery.runtime: 1440.0
device.mfr: APC
device.type: ups
driver.name: apcupsd-ups
driver.parameter.pollinterval: 2
driver.parameter.port: localhost
driver.parameter.synchronous: auto
driver.version: 2.8.0
[...]

PR #2010, opened 2023-08-03
Introduce state entry timestamping and use it to fix `apcupsd-ups`

jimklimov:
CC @aquette @clepple : This is a fairly noticeable change to the deep internals, state info trees are at the core of everything in NUT. Would welcome another set of eyeballs that I got this conceptually right :)

Should I be concerned about computational overheads to keep reading and comparing time values? Should this behavior become a toggle (e.g. for programs that know they intend to use this feature at all)?

In context of issue #2007 there is a need to somehow avoid the current behavior with deletion of nearly all info about the device and then re-querying `apcupsd` to populate the dstate again. With this change we hopefully (un-tested so far) have a way to determine if any entries were not updated in the last read, and so should be cleaned away AFTER talking to the device (apcupsd server in this case).

While this is a first use for such tech, I can see it proliferating into other drivers to support "Data stale" decisions (e.g. attempt to re-initialize the hardware connection), etc.

Closes: #2007
Fixes: #797 fallout

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -788,22 +788,22 @@ Release notes for NUT 2.7.4 - what's new since 2.7.3:
      input, output, outlet and outlet.group)
 
  - support for new devices:
-   AEG PROTECT B / NAS
-   APC ATS AP7724 (should be supported)
-   Asium P700
-   Eaton ATS
-   Eaton 5E 1100iUSB
-   Eaton E Series DX UPS 1-20 kVA
-   Eaton Powerware 9125-5000g
-   Electrys UPS 2500
-   Fideltronic INIGO Viper 1200
-   Legrand Keor Multiplug
-   LYONN CTB-800V
-   Micropower LCD 1000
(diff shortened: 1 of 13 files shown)


clepple:
Benchmarking by eyeball is always dubious at best. But given that the values going into the state info trees are converted to C strings, I have a feeling that the timings for int-to-double conversion of the timestamps will be dwarfed by the conversion to strings (at least on modern platforms).

I definitely don't claim to be an expert on the state trees, so I wouldn't trust myself to notice whether there are subtle bugs being introduced.

jimklimov:
Checked locally, does not regress with an Eaton UPS (not impacted directly like `apcupsd-ups` which was deliberately changed by this PR, but impacted indirectly by change of state/dstate tech). Seems to work still quickly and report same data :)

jimklimov:
Also checked in original issue #2007 to actually fix the problem the user came with :)

PR #2029, opened 2023-08-30
Revert "cyberpower-mib: fix instant command declarations"

### LINKED ISSUE #2028: cyberpower-mib.c: Instant commands don't work
UPS: Cyberpower OLS1500ERT2U + RMCARD205
OS: Rocky Linux 9.2
nut-2.8.0 patched with newest cyberpower-mib (1.54) and snmp-ups (1.30) drivers

Instant commands don't work,

I found that the problem is in the instant commands declarations in **cyberpower-mib.c**.
OID values are stored in the third parameter but should be in place of the fifth parameter (default value) which is NULL.
If I place OID values to default value parameter, the commands work.

Looking to the cyberpower-mib.c history I don't understand this commit
https://github.com/networkupstools/nut/commit/8bcf255b624e6ba6c89a2a3eeab37658d9cb8cb4

From **snmp-ups.c**

```
/* Common function for setting OIDs, from a NUT variable name,
 * used by su_setvar() and su_instcmd()
 * Params:
 * @mode: SU_MODE_INSTCMD for instant commands, SU_MODE_SETVAR for settings
 * @varname: name of variable or command to set the OID from
 * @val: value for settings, NULL for commands

 * Returns
 *   STAT_SET_HANDLED if OK,
 *   STAT_SET_INVALID or STAT_SET_UNKNOWN if the command / setting is not supported
 *   STAT_SET_FAILED otherwise
 */
static int su_setOID(int mode, const char *varname, const char *val)
{
. . .
```
So for SU_MODE_INSTCMD val is NULL and OID value is taken from default value.

overhacked:
This reverts commit 8bcf255b624e6ba6c89a2a3eeab37658d9cb8cb4 from PR #686.

Cyberpower `snmp_info_t` got updated to conform with pre 5d97d07 instant
commands layout, but got merged after it. Reverting 8bcf255 so that command
values are correctly in `su_info_t->dfl`.

Fixes #2028.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -66,8 +66,9 @@ as part of https://github.com/networkupstools/nut/issues/1410 solution.
      fixed for NUT v2.8.1.
    * A table in `cyberpower-mib` (for `snmp-ups` driver) sources was
      arranged in NUT v2.8.0 release in a way that precluded the driver
-     logic from looking at all of its entries. Regression fixed for NUT
-     v2.8.1 [#1432]
+     logic from looking at all of its entries. Also a fix for instant
+     command definitions had in fact broken them due to other development.
+     Regressions fixed for NUT v2.8.1 [#1432, #2029]
    * A change for file-change detection in `dummy-ups` driver for NUT
      v2.8.0 release misfired on some platforms. Regression fixed for NUT
      v2.8.1 [#1420]
--- a/drivers/cyberpower-mib.c
+++ b/drivers/cyberpower-mib.c
@@ -195,22 +195,22 @@ static snmp_info_t cyberpower_mib[] = {
 		SU_FLAG_OK | SU_TYPE_TIME, NULL },
 	/* instant commands. */
 	/* upsAdvanceControlUpsOff */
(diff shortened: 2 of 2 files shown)


Bomorav:
But this revert also rolled back an old command for _shutdown.return_ which had since been changed. Please change lines 203, 204

```
    /* upsBaseControlConserveBattery */
    { "shutdown.return", 0, 1, ".1.3.6.1.4.1.3808.1.1.1.6.1.1.0", "2", SU_TYPE_CMD | SU_FLAG_OK, NULL },
```
 to

 ```
   /* upsAdvanceControlUpsSleep */
    { "shutdown.return", 0, 1, ".1.3.6.1.4.1.3808.1.1.1.6.2.3.0", "3", SU_TYPE_CMD | SU_FLAG_OK, NULL },
```
_upsBaseControlConserveBattery_ command is not suitable here because it puts ups to sleep only in battery mode. If power is restored during the shutdown process, the command is not executed by ups.

jimklimov:
Oh, thanks, missed that bit!

Issue #2046, opened 2023-09-10
Docs: update rendered document metadata with actual release history


jimklimov:
Currently the (PDF) rendered documentation includes a revision history which lists just two entries:
````
2.8.0.1 | 2023-09-10 | Current release of Network UPS Tools (NUT).
2.6.0   | 2011-01-14 | First release of AsciiDoc documentation for Network UPS Tools (NUT).
````
(where "2.8.0.1" is a placeholder for "build-du-jour from master or feature branch after 2.8.0 before 2.8.1")

When building from Git sources (at least) the tag-based history can be produced on the fly.

Such file should be `EXTRA_DIST`ed (maybe also Git-tracked?) so offline builds would also have the list.

While at it, the author name list can be extended with myself as well as unnamed "and NUT project community contributors"

jimklimov:
This information seems to be tracked in `docs/docinfo.xml(.in)` following from #150

jimklimov:
Result of the PR above:

![image](https://github.com/networkupstools/nut/assets/1636756/2c371ccb-c0f7-4bb9-8dc5-cd367c7dcd96)

There are a few side effects (e.g. the revision list appearing in ALL PDFs, even if their source files did not exist back then), and the `revremark` comments are not populated yet... refer to Changelog or NEWS file per release? But these can be another PR another time...

PR #2047, opened 2023-09-10
Refer to existing releases from rendered PDF docs

jimklimov:
Update the way we maintain releases (and list of releases in docs, in particular).

Closes: #2046


jimklimov:
Referencing the NEWS files for each release proved problematic: the `revremark` tag does not expect nested tags (like `<a href>` or even `<p>`).

PR #2056, opened 2023-09-17
Fightwarn: numerous complaints from clang-15 [#823]

jimklimov:
Surprised that so many of these were not found sooner. Almost all fall into a few simple categories:
* method(void) declared without a "void"
* declarations after code
    * NUT_UNUSED_VARIABLE() is code
    * scoping in switch() cases
* a few smartly detected unused variables => added reasonable use-cases (logging or better)


jimklimov:
NOTE: Additionally there is a category of warnings from clang-16 about unsafe buffer access concerns (not checking for bounds - we often iterate until NULL; doing pointer arithmetic without some checks, etc.)

PR #2057, opened 2023-09-17
Feature/add support sms brazil

alexwbaule:
PR For support on SMS Brazil single phase.
Add on docs, instructions about the protocol.
This is for now the driver, maybe in a near future i will migrate it to nutdrv_qx.


jimklimov:
Cheers, I've taken the liberty of updating this PR a bit, almost-cosmetically :)

Some remaining things that pop out are spaces (not TABs) used here for indentation, adding a driver man page (maybe the current protocol file can be referenced from there or added as a chapter) and a NEWS entry eventually.

Generally documentation files are welcome to be listed as at least spellchecked sources (so probably that would add to `docs/nut.dict` after enablement and a `make spellcheck-interactive` to add new terminology.

drivers/sms_ser.c
```diff
@@ -0,0 +1,698 @@
...
+            snprintfcat(results->model, 24, "%c", rawvalues[i]);
+        } else {
+            snprintfcat(results->version, 6, "%c", rawvalues[i]);
+        }
+    }
+}
+
+void sms_parse_results(uint8_t *rawvalues, SmsData *results) {
+    char buf[BUFFER_SIZE];
+    unsigned char byte, mask, p;
```
jimklimov, line 96: Wondering if `uint8_t` would be better than `unsigned char` here, with bitmask maths involved... just to be sure on different platforms.

drivers/sms_ser.c
```diff
@@ -0,0 +1,698 @@
...
+    upslogx(LOG_INFO, "bypass: %d", results->bypass);
+    upslogx(LOG_INFO, "lowbattery: %d", results->lowbattery);
+    upslogx(LOG_INFO, "onbattery: %d", results->onbattery);
+}
+
+static int get_ups_nominal() {
+    uint8_t length;
+    ssize_t ret;
+
+    upslogx(LOG_INFO, "get_ups_nominal");
```
jimklimov, line 168: Here and in other places, `upslogx` may be too loud (and going to syslog etc.) Consider changing to `upsdebugx` with a non-zero verbosity level, at least where it is "just" debug tracing.

drivers/sms_ser.c
```diff
@@ -0,0 +1,698 @@
...
+        return;
+    }
+    if (get_ups_information() != 0) {
+        upslogx(LOG_ERR, "Short read from UPS");
+        dstate_datastale();
+        return;
+    }
+
+    if (get_ups_nominal() == 0) {
+        dstate_setinfo("device.model", "%s", (unsigned char *)DeviceData.model);
```
jimklimov, line 380: Was casting to `(unsigned char *)` required? I think `%s` deals with `char*` just fine...

drivers/sms_ser.c
```diff
@@ -0,0 +1,702 @@
...
+
+void upsdrv_cleanup(void) {
+    upslogx(LOG_INFO, "upsdrv_cleanup");
+    /* free(dynamic_mem); */
+    ser_close(upsfd, device_path);
+}
+
+uint8_t sms_prepare_get_status(uint8_t *buffer) {
+    buffer[0] = 'Q';
+    buffer[1] = -1;
```
jimklimov, line 586: Here and in subsequent methods, for an `uint8_t` array there may be no `-1` value. Maybe better set it as `0xFF` or `255` explicitly?

Revise also the (checksum/CRC?) calculation with `(...) * -1` patterns.

jimklimov:
Thanks a lot for bringing it into such a good shape, and the constructive discussions in #1987 earlier!

alexwbaule:
Tks @jimklimov  !!  And feel free to update this PR !

drivers/sms_ser.c
alexwbaule: the "-1" im just following the protocol, and maybe its OK for me, to set as 0xFF, 255 or NULL. Maybe if change to number or HEX, change all to the same (like change the 'Q' command to a HEX to).

drivers/sms_ser.c
alexwbaule: Thats a good question! i think that it not necessary to do this cast. i will fix.

drivers/sms_ser.c
alexwbaule: Ok ! A lot of things was to "ok, i get to this part", i will review all logs.

drivers/sms_ser.c
alexwbaule: For me its ok, i only has Linux and freebsd to test it.

drivers/sms_ser.c
jimklimov: Well, to me the actual letters like `'Q'` are okay, human-readable as the actual protocol posted to the device.

alexwbaule:
@jimklimov , i need to signed-off my commits ?

jimklimov:
> i need to signed-off my commits ?

Yes, that would be welcome - see https://github.com/networkupstools/nut/wiki/Code-contributions,-PRs,-PGP-and-DCO#troubleshooting for suggestions. It can be worked-around in some cases, but sort of defeats the legalese purpose of having the sign-offs :)

jimklimov:
Hi, how close do you think this work is to completion? If I were to cut a next NUT release soon (say during October), would this contribution make it and get included, to benefit packaged users? :)

alexwbaule:
Hi @jimklimov  sorry about the delay, to busy on work in this week.

Today i will update this PR to be able to merge it.

jimklimov:
Super, thanks for the update! :)

drivers/sms_ser.c
```diff
@@ -152,25 +152,25 @@ void sms_parse_results(uint8_t *rawvalues, SmsData *results) {
...
 
-    upslogx(LOG_INFO, "beepon: %d", results->beepon);
-    upslogx(LOG_INFO, "shutdown: %d", results->shutdown);
-    upslogx(LOG_INFO, "test: %d", results->test);
-    upslogx(LOG_INFO, "upsok: %d", results->upsok);
-    upslogx(LOG_INFO, "bypass: %d", results->bypass);
-    upslogx(LOG_INFO, "lowbattery: %d", results->lowbattery);
-    upslogx(LOG_INFO, "onbattery: %d", results->onbattery);
+    upsdebugx(LOG_DEBUG, "beepon: %d", results->beepon);
+    upsdebugx(LOG_DEBUG, "shutdown: %d", results->shutdown);
```
jimklimov, line 156: `upsdebugx()` does not deal with symbolic names (maybe should have, but...), but rather with relatively arbitrarily assigned verbosity level numbers (with debug setting "X or greater" the message would be seen), with nuances like this probably suitable for level 3 or more.

jimklimov (approved):

drivers/sms_ser.c
alexwbaule: So i will return to upslogx with LOG_DEBUG ? or you think is not necessary to show this ? Maybe a command can retrieve this.

alexwbaule: @jimklimov i cant signed my commits because has a lot of merges from master. Doing a `git rebase --signoff` its almost impossible to edit all commits, to sign it.

There is some tip for this ?

alexwbaule: oh.. i see, higher level to print, not debug.. 

i will see the script..

jimklimov:
Well, you can return your workspace to be as it was in case of big mishaps - run `git reflog` to see changes of local index, e.g. as you started a rebase session, etc. and you can `git reset` to that commit just before.

PR #2065, opened 2023-09-18
CI: Fix `aspell` use on MinGW x64 and MSYS2 platforms

jimklimov:
Thanks to reports in https://github.com/msys2/MSYS2-packages/issues/2088 and following up from #2060, a path was found how to un-break the uses of `aspell` in NUT for Windows builds. Hopefully this change does not break it on other platforms ;)

Closes: #2067


jimklimov:
Ok, so it did a little breakage, fixed with latest commit so now it caters to both layouts seen in the wild:
* Debain(ish) Linux: `/usr/lib/aspell/x86_64-linux-gnu/tex.amf` (and `*.so` files nearby)
* MSYS2: `/usr/lib/aspell-0.60/tex.amf`
* OpenIndiana: `/usr/lib/aspell-0.60/tex.amf`

jimklimov:
Eh, and then there's CentOS 7 with `/usr/lib64/aspell-0.60/tex.amf`

jimklimov:
One of the CentOS 7 agents missed `aspell-en` package which is among requirements. Adding it fixed the issue for default builds...

````
  ASPELL   Spell checking on ./../tools/nut-scanner/README.adoc
FAILED : Aspell reported errors here:
----- vvv
Error: The file "/usr/lib64/aspell-0.60/en" can not be opened for reading.
----- ^^^
````

It might be more "correct" perhaps to detect not only presence of `aspell`, but also its ability to run with a trivial text (not unlike we do for `asciidoc`/`a2x` in the `configure` script) in the long run.

PR #2089, opened 2023-10-04
Correct voltage and battery capacity for Opti-UPS 230VAC and/or 24VDC models

moonbuggy:
The _optiups_ driver only gave accurate voltage information with 120VAC models and assumed a 12V battery when calculating capacity.

There's an `FV` command that gives a (fixed) voltage that correlates with the voltage selection DIP switches on the back of the UPS, taking into account whether ti's a 120 or 240VAC model. So it's easy enough to do `if (FV > 180) then voltage = voltage * 2`, still using the `NV` and `OV` commands to pull the live in/out voltage data.

I suspect this would apply to all UPS using this protocol, under the assumption that the driver was developed on 120VAC devices where the scaling issue wasn't apparent.  However, as I can only test on a PS-1440RM at 230VAC, I've only applied the change to "PS-*" series UPS through a new `_pollv_ps[]` struct.

It would make sense to apply the change by default via the `_pollv[]` struct, if it can be confirmed it doesn't break things for other models at 120VAC.

The battery capacity fix _is_ applied globally., based on whether or not the battery voltage is greater than 20V.

I don't work with C very often, and due to my lack of familiarity there's some inefficiency with how the `FV` command is handled. It really should only need to be run once, during init. I have it being run on every update because I wasn't sure how best to set it during init and then get it for the updates. So that's not ideal, but it works and a few extra bytes each update doesn't seem like the end of  the world.  :) I'd be happy to fix the inefficiency if someone wants to point me in the right direction though.

### DIFF
--- a/drivers/optiups.c
+++ b/drivers/optiups.c
@@ -488,22 +488,22 @@ void upsdrv_updateinfo(void)
 		optifill( _pollv_ps, sizeof(_pollv_ps)/sizeof(_pollv_ps[0]) );
 
 		r = optiquery( "NV" );
-		float inV = strtol ( _buf, NULL, 10);
+		short inV = atoi ( _buf );
 		r = optiquery( "OV" );
-		float outV = strtol ( _buf, NULL, 10);
+		short outV = atoi ( _buf );
 
 		r = optiquery( "FV" );
 		if ( r >= 1 )
 		{
-			float f = strtol ( _buf, NULL, 10 );
+			short f = atoi ( _buf );
 			if ( f > 180 )
 			{
 				inV = inV * 2;
 				outV = outV * 2;
 			}
 		}
-		dstate_setinfo( "input.voltage", "%.1f", inV );
-		dstate_setinfo( "output.voltage", "%.1f", outV );
+		dstate_setinfo( "input.voltage", "%d", inV );
+		dstate_setinfo( "output.voltage", "%d", outV );
 	}
 	else
 		optifill( _pollv, sizeof(_pollv)/sizeof(_pollv[0]) );
--- a/drivers/optiups.c
+++ b/drivers/optiups.c
@@ -485,9 +485,9 @@ void upsdrv_updateinfo(void)
 	if ( optimodel == OPTIMODEL_ZINTO )
 		optifill( _pollv_zinto, sizeof(_pollv_zinto)/sizeof(_pollv_zinto[0]) );
 	else if ( optimodel == OPTIMODEL_PS ) {
-		optifill( _pollv_ps, sizeof(_pollv_ps)/sizeof(_pollv_ps[0]) );
-
 		short inV, outV, fV;
+
+		optifill( _pollv_ps, sizeof(_pollv_ps)/sizeof(_pollv_ps[0]) );
 
 		r = optiquery( "NV" );
 		str_to_short ( _buf, &inV, 10 );
--- a/drivers/optiups.c
+++ b/drivers/optiups.c
@@ -487,16 +487,18 @@ void upsdrv_updateinfo(void)
 	else if ( optimodel == OPTIMODEL_PS ) {
 		optifill( _pollv_ps, sizeof(_pollv_ps)/sizeof(_pollv_ps[0]) );
 
+		short inV, outV, fV;
+
 		r = optiquery( "NV" );
-		short inV = atoi ( _buf );
+		str_to_short ( _buf, &inV, 10 );
 		r = optiquery( "OV" );
-		short outV = atoi ( _buf );
+		str_to_short ( _buf, &outV, 10 );
 
 		r = optiquery( "FV" );
 		if ( r >= 1 )
 		{
-			short f = atoi ( _buf );
-			if ( f > 180 )
+			str_to_short ( _buf, &fV, 10 );
+			if ( fV > 180 )
 			{
 				inV = inV * 2;
 				outV = outV * 2;
--- a/drivers/optiups.c
+++ b/drivers/optiups.c
@@ -28,7 +28,7 @@
 #include "nut_stdint.h"
 
 #define DRIVER_NAME	"Opti-UPS driver"
-#define DRIVER_VERSION "1.03"
+#define DRIVER_VERSION "1.04"
 
 /* driver description structure */
 upsdrv_info_t upsdrv_info = {
@@ -83,7 +83,8 @@ static char _buf[256];
 static int optimodel = 0;
 enum {
 	OPTIMODEL_DEFAULT = 0,
-	OPTIMODEL_ZINTO = 1
+	OPTIMODEL_ZINTO = 1,
+	OPTIMODEL_PS = 2
 };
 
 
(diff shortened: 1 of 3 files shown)


drivers/optiups.c
```diff
@@ -463,6 +484,27 @@ void upsdrv_updateinfo(void)
 	/* read some easy settings */
 	if ( optimodel == OPTIMODEL_ZINTO )
 		optifill( _pollv_zinto, sizeof(_pollv_zinto)/sizeof(_pollv_zinto[0]) );
+	else if ( optimodel == OPTIMODEL_PS ) {
+		optifill( _pollv_ps, sizeof(_pollv_ps)/sizeof(_pollv_ps[0]) );
+
+		r = optiquery( "NV" );
+		float inV = strtol ( _buf, NULL, 10);
```
jimklimov, line 491: Does the call return an integer or a floating point number in the string? Why `strtol()` and not `long` but `float` to hold the result?

jimklimov (approved):

jimklimov:
Thanks for the contribution, seems right to do this in the update loop - wall voltages can change over time...

One question posted above; another suggestion is to add a note to `NEWS.adoc` about this update. And perhaps an entry to `data/driver.list.in` (HCL) to mark the device model as known supported.

A post of data etc. dump for the DDL would also be welcome, see `tools/nut-ddl-dump.sh` :)

moonbuggy:
Sorry, I possibly didn't explain it well enough. 'FV' always seems to respond with the same fixed voltage, depending on how the DIP switches on the back of thing are set. So for my 230V model, I expect 'FV' will only ever be 200, 220, 230 or 240V, regardless of the actual line voltage. 

At least, in my testing it changed from 240V to 230V when I flipped the relevant switch, and stayed at that value while the voltage reported by NV/OV happily wobbled around at 117-120V (before scaling, obviously). I assume it should behave the same for 120V models, just locked to the lower voltages.

Anyway, that's why it should ideally be requested during init rather than during an update. (Which is what the official Opti-UPS software does, ask for it at init and then just uses NV/OV after that.)

Here's the section from the manual about the DIP switches, just for clarity:
Opti-DIP-switches

I've just done a pull request with a data dump here: https://github.com/networkupstools/nut-ddl/pull/39

And I've updated the device list.

drivers/optiups.c
moonbuggy: Why strtol() and not long but float to hold the result?

I'll refer you to the "I don't work with C very often" comment in the PR. :) I just looked at the existing code and emulated what I saw.

The strings are integers.

Ir's quite late here, but I can sort it out in the morning. It's obviously not the right way to do things now you've pointed it out. :)

moonbuggy: So, looking at the code properly, neither _float_ nor _long_ seem warranted. Unless I'm missing something, a short integer from `atoi()` should be fine..? At least until someone plugs a UPS into a supply greater than 32kV, but I think if that happens a NUT driver is going to be the least of their worries. :)

As I say though, I'm not particularly fluent in C, so it's possible I've misunderstood something fundamental.

BTW, I'm not entirely clear on the difference between the `%i` and `%d` formats, both seemed to work. I went with `%d` arbitrarily.

jimklimov: Well, it depends on strings returned and their character counts, mostly... I remember in some driver there was a problem of a protocol byte-sized voltage not sufficing for the 280V or so reality (which IIRC the LCD panel did reflect)...

jimklimov: While here, it may make sense to use NUT's `str_to_*()` methods with error-checking, something #677/#966 stories were long about.

jimklimov: As for the latest force-push - `atoi()` is specified as returning an `int`, so static checkers gotta complain about a `short int` there. Better safe than sorry :)

moonbuggy: Sorry if forced pushes caused any problems..? I keep forgetting to `--signoff`, so end up rebasing. :)

Using NUT's `str_to_*()` does make sense. Sorry, I would have done that earlier If I'd realised you had your own string library.

I've gone with `str_to_short()`. I wasn't clear if you were suggesting I should use `str_to_int()` specifically. The UPS returns a 4-character string for the voltages, with a leading zero, as such (from  `upsdrvctl -DDD`):

```
   4.715086     [D2] SEND: "NV"
   4.868114     [D2] READ: "0119"
   5.028188     [D2] SEND: "OV"
   5.187680     [D2] READ: "0119"
```

The raw output from a serial monitor in Windows 2000 running the official Opti-UPS software shows it split across two (or four) lines, but I don't know if that's relevant at all..?

```
 752	0.00147421	UPSCOMM.EXE	IRP_MJ_WRITE	Serial2	SUCCESS	Length 1: N	
 753	0.00197511	UPSCOMM.EXE	IRP_MJ_WRITE	Serial2	SUCCESS	Length 1: V	
 754	0.00194131	UPSCOMM.EXE	IRP_MJ_WRITE	Serial2	SUCCESS	Length 1: .	
 755	0.00481067	UPSCOMM.EXE	IRP_MJ_WRITE	Serial2	SUCCESS	Length 1: .	
 756	0.00000587	UPSCOMM.EXE	IRP_MJ_READ	Serial2	SUCCESS	Length 0: 	
 757	0.00005084	UPSCOMM.EXE	IRP_MJ_READ	Serial2	SUCCESS	Length 2: 01	
 758	0.00000615	UPSCOMM.EXE	IRP_MJ_READ	Serial2	SUCCESS	Length 3: 19.	
 759	0.00000587	UPSCOMM.EXE	IRP_MJ_READ	Serial2	SUCCESS	Length 1: .	
```

This is about where I hit my limit of really knowing what's going on The driver works with my UPS with `str_to_short()`. I'm unable to judge if I should expect it to have issues if a voltage hits "0256" or something, you're going to be a better judge of that than I am.

I'm happy to change it to an unsigned short or an integer, if you think either of those would be better.

(I went with a signed variable because I thought the UPS might actually want to send negative voltages if the power cable is dodgy and the live/neutral wires are swapped. I didn't want to make up a bad IEC cable to test that though. :))

jimklimov: Hopefully, as long as it matches the method declaration, `short` should be good enough for this use-case (as long as C standards are followed on the platform and it is at least 16 bits wide).

jimklimov: And yes, better not test it that way (with broken cable). I had a lot of sparks and one device less back in the day, when a computer was plugged into one wall outlet, its display into another, and the wall wires in the ancient building were attached to leads randomly...

jimklimov:
Thanks, seems good. If CI likes it too, will merge :)

drivers/optiups.c
moonbuggy: I don't even know if many people still run these old Opti-UPS devices..? So it's possible the use-case is quite narrow and I'm it. :)

In terms of platforms, btw, all my NUT testing has been on an Orange Pi Zero (as part of this work-in-progress: https://github.com/moonbuggy/Opti-UPS-expansion-slot-SBC).

So I can confirm it works on ARMv7 to '0240'. Which is really as high as it should ever get since that's the limit of the DIP switch setting and the reported live in/out voltages are half the real voltages. (I'd need an actual mains voltage of 480V to get anything higher than '0240' back from any of the three commands involved.)

Anyway, thanks for your help with the datatypes. :) Apologies if my limited C ability was annoying. :)

jimklimov: Not at all, and thanks to you! :)

PR #2090, opened 2023-10-04
add support to SMS 1phase brazil from #1987

alexwbaule:
PR For support on SMS Brazil single phase.
Add on docs, instructions about the protocol.
This is for now the driver, maybe in a near future i will migrate it to nutdrv_qx.

### DIFF
--- a/drivers/sms_ser.c
+++ b/drivers/sms_ser.c
@@ -51,7 +51,7 @@ upsdrv_info_t upsdrv_info = {
     DRV_BETA,
     {NULL}};
 
-void sms_parse_features(uint8_t *rawvalues, SmsData *results) {
+static void sms_parse_features(uint8_t *rawvalues, SmsData *results) {
     char tbattery[6];
     char frequency[4];
 
@@ -76,7 +76,7 @@ void sms_parse_features(uint8_t *rawvalues, SmsData *results) {
     results->frequency = atoi(frequency);
 }
 
-void sms_parse_information(uint8_t *rawvalues, SmsData *results) {
+static void sms_parse_information(uint8_t *rawvalues, SmsData *results) {
     /* Count from 1 to ignore first char and remove 2 from BUFFER_SIZE
      *  to compensate the start and ignore '\r' from end. */
     memset(results->model, 0, sizeof(results->model));
--- a/drivers/sms_ser.c
+++ b/drivers/sms_ser.c
@@ -27,6 +27,7 @@
 #include "sms_ser.h"
 #include "main.h"
 #include "serial.h"
+#include "nut_stdint.h"
 
 #define ENDCHAR '\r'
 
@@ -173,7 +174,7 @@ static int get_ups_nominal() {
         return -1;
     }
 
-    upsdebugx(3, "Get nominal Ok: received byte %lu", ret);
+    upsdebugx(3, "Get nominal Ok: received byte %" PRIiSIZE, ret);
 
     if (bufIn[0] == '=' || bufIn[0] == '<' || bufIn[0] == '>') {
         sms_parse_results(&bufIn[0], &DeviceData);
@@ -205,7 +206,7 @@ static int get_ups_information() {
         return -1;
     }
 
-    upsdebugx(3, "Get information Ok: received byte %lu", ret);
+    upsdebugx(3, "Get information Ok: received byte %" PRIiSIZE, ret);
 
     if (bufIn[0] == ';' || bufIn[0] == ':') {
         sms_parse_information(&bufIn[0], &DeviceData);
@@ -237,7 +238,7 @@ static int get_ups_features() {
         return -1;
     }
 
-    upsdebugx(LOG_DEBUG, "Get features Ok: received byte %lu", ret);
+    upsdebugx(LOG_DEBUG, "Get features Ok: received byte %" PRIiSIZE, ret);
 
     if (bufIn[0] == ';' || bufIn[0] == ':') {
         sms_parse_features(&bufIn[0], &DeviceData);
--- a/drivers/sms_ser.c
+++ b/drivers/sms_ser.c
@@ -1,0 +1,693 @@
+/*
+ * sms_ser.c: code for mono protocol for SMS Brazil UPSes
+ *
+ * Copyright (C) 2023 - Alex W. Baule <alexwbaule@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ *
+ * Reference of the derivative work: riello driver
+ */
+
+#include "config.h" /* must be the first header */
+#include <string.h>
+
+#include "common.h" /* for upsdebugx() etc */
+#include "sms_ser.h"
+#include "main.h"
+#include "serial.h"
+
+#define ENDCHAR '\r'
+
+#define DRIVER_NAME "SMS Brazil UPS driver"
+#define DRIVER_VERSION "1.00"
+
+#define QUERY_SIZE 7
+#define BUFFER_SIZE 18
+#define RESULT_SIZE 18
+#define HUMAN_VALUES 7
+
+static unsigned int bootdelay = DEFAULT_BOOTDELAY;
+static uint8_t bufOut[BUFFER_SIZE];
+static uint8_t bufIn[BUFFER_SIZE];
+static SmsData DeviceData;
+
+/* driver description structure */
+upsdrv_info_t upsdrv_info = {
+    DRIVER_NAME,
+    DRIVER_VERSION,
+    "Alex W. Baulé <alexwbaule@gmail.com>",
+    DRV_BETA,
+    {NULL}};
+
+void sms_parse_features(uint8_t *rawvalues, SmsData *results) {
+    char tbattery[6];
+    char frequency[4];
+
+    memset(results->voltageRange, 0, sizeof(results->voltageRange));
+    memset(results->currentRange, 0, sizeof(results->currentRange));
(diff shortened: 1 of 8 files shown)


jimklimov:
Ok, thanks!

jimklimov:
Got some homework to do ;)

````

Making all in drivers
In file included from sms_ser.c:26:0:
sms_ser.c: In function 'get_ups_nominal':
sms_ser.c:176:18: error: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'ssize_t {aka int}' [-Werror=format=]
     upsdebugx(3, "Get nominal Ok: received byte %lu", ret);
                  ^
../include/common.h:321:62: note: in definition of macro 'upsdebugx'
  do { if (nut_debug_level >= (level)) { s_upsdebugx((level), __VA_ARGS__); } } while(0)
                                                              ^~~~~~~~~~~
sms_ser.c: In function 'get_ups_information':
sms_ser.c:208:18: error: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'ssize_t {aka int}' [-Werror=format=]
     upsdebugx(3, "Get information Ok: received byte %lu", ret);
                  ^
../include/common.h:321:62: note: in definition of macro 'upsdebugx'
  do { if (nut_debug_level >= (level)) { s_upsdebugx((level), __VA_ARGS__); } } while(0)
                                                              ^~~~~~~~~~~
sms_ser.c: In function 'get_ups_features':
sms_ser.c:240:26: error: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'ssize_t {aka int}' [-Werror=format=]
     upsdebugx(LOG_DEBUG, "Get features Ok: received byte %lu", ret);
                          ^
../include/common.h:321:62: note: in definition of macro 'upsdebugx'
  do { if (nut_debug_level >= (level)) { s_upsdebugx((level), __VA_ARGS__); } } while(0)
                                                              ^~~~~~~~~~~
sms_ser.c: In function 'sms_instcmd':
sms_ser.c:315:61: error: signed and unsigned type in conditional expression [-Werror=sign-compare]
         long delay = extra ? strtol(extra, NULL, bootdelay) : bootdelay;
                                                             ^
sms_ser.c: At top level:
cc1: error: unrecognized command line option '-Wno-unknown-warning-option' [-Werror]
cc1: error: unrecognized command line option '-Wno-reserved-identifier' [-Werror]
gmake[1]: *** [Makefile:1974: sms_ser.o] Error 1
gmake[1]: Target 'all' not remade because of errors.
````

**Comment by alexwbaule** [comment]
Hi @jimklimov , 

Ok, I will fix it today. 

Tks !

**Comment by jimklimov** [comment]
Actually taking a stab at a couple of these that block passing Jenkins... :)

**Comment by jimklimov** [comment]
Ok, should be quiet on my side for now.

Did not yet look much into the changed code itself, but if it works for you and others as an actual driver (as IIRC reported before), that is a solid vouch for it :)

How are you about putting this into `nutdrv_qx` family, and/or adding the man page?

File: drivers/sms_ser.c [hunk 1]
```diff
@@ -0,0 +1,701 @@
...
+void sms_parse_features(uint8_t *rawvalues, SmsData *results) {
+    char tbattery[6];
+    char frequency[4];
+
+    memset(results->voltageRange, 0, sizeof(results->voltageRange));
+    memset(results->currentRange, 0, sizeof(results->currentRange));
+    memset(tbattery, 0, sizeof(tbattery));
+    memset(frequency, 0, sizeof(frequency));
+
+    for (int i = 1; i < BUFFER_SIZE - 2; i++) {
```
jimklimov, line 63: Here and in other loops, I think some of the compiler portability checks would complain about first declaring a variable inside the loop. Clang tends to do so even if compiling for C99 or newer (which are okay with this), when pedantic warnings are enabled... No worry if CI does not complain with default settings though. Becoming "clean" against more and more warnings types is a separate effort :)

drivers/sms_ser.c
```diff
@@ -0,0 +1,701 @@
...
+}
+
+void sms_parse_results(uint8_t *rawvalues, SmsData *results) {
+    char buf[BUFFER_SIZE];
+    uint8_t byte, mask;
+    long v;
+    double h;
+
+    memset(buf, 0, BUFFER_SIZE);
+    sprintf(buf, "0x%02x%02x", (unsigned char)rawvalues[1], (unsigned char)rawvalues[2]);
```
jimklimov, line 101: I think CI would complain about not setting constraints (e.g. using `snprintf(buf, sizeof(buf), ...)`), and makes sense in practice to do so anyway, to err on the safe side.

drivers/sms_ser.c
```diff
@@ -0,0 +1,701 @@
...
+    mask = 1;
+
+    results->beepon = ((byte & (mask << 0)) != 0) ? true : false;
+    results->shutdown = ((byte & (mask << 1)) != 0) ? true : false;
+    results->test = ((byte & (mask << 2)) != 0) ? true : false;
+    results->upsok = ((byte & (mask << 3)) != 0) ? true : false;
+    results->boost = ((byte & (mask << 4)) != 0) ? true : false;
+    results->bypass = ((byte & (mask << 5)) != 0) ? true : false;
+    results->lowbattery = ((byte & (mask << 6)) != 0) ? true : false;
+    results->onbattery = ((byte & (mask << 7)) != 0) ? true : false;
```
jimklimov, line 152: Wouldn't it be better to define macros for bit mask meanings in the header, and just use `if (rawvalue & SMS_UPSOK) ...` etc.? At least is computationally quicker, uses less memory, and has lots of precedents in NUT code :)

At least, something to consider if/when refactoring this codebase further e.g. to move under `nutdrv_qx` umbrella.

drivers/sms_ser.c
```diff
@@ -0,0 +1,701 @@
...
+    upsdebugx(LOG_DEBUG, "get_ups_nominal");
+
+    length = sms_prepare_get_status(&bufOut[0]);
+
+    if (ser_send_buf(upsfd, bufOut, length) == 0) {
+        upsdebugx(LOG_ERR, "Communication error while writing to port");
+        return -1;
+    }
+    memset(bufIn, 0, BUFFER_SIZE);
+    ret = ser_get_buf_len(upsfd, &bufIn[0], BUFFER_SIZE, 3, 1000);
```
jimklimov, line 168: As a bit of coding style for future, it is more future-proof to pass size constraints as `sizeof(arrayname)` than as the specific number (or macro) that happened to initialize that size in some iteration of code.

drivers/sms_ser.c
```diff
@@ -0,0 +1,701 @@
...
+        dstate_setinfo("ups.delay.reboot", "%u", bootdelay);
+        return STAT_SET_HANDLED;
+    }
+    return STAT_SET_UNKNOWN;
+}
+
+void upsdrv_initinfo(void) {
+    char *battery_status;
+
+	upsdebugx(LOG_DEBUG, "upsdrv_initinfo");
```
jimklimov, line 365: Here and elsewhere, `upsdebugx()` first argument is a number - relatively arbitrarily assigned by driver authors, larger numbers are less visible (need higher verbosity level). Method entry is usually 2-3 or higher, and traces inside methods may be same or +1 usually...

drivers/sms_ser.c
```diff
@@ -0,0 +1,701 @@
...
+    buffer[0] = 'Q';
+    buffer[1] = 255;
+    buffer[2] = 255;
+    buffer[3] = 255;
+    buffer[4] = 255;
+    buffer[5] = (buffer[0] + buffer[1] + buffer[2] + buffer[3] + buffer[4]) * 255;
+    buffer[6] = ENDCHAR;
+
+    return 7;
+}
```
jimklimov, line 593: Feels a bit weird to see code that just pokes bytes in memory, with no internal (e.g. arg) way of knowing how far it can go. Externally we see that the bufOut is 18 bytes etc., but the method itself has no way of checking if it overflows some allocation.

Also, would it help to set a `buffer[7] = '\0';` to be safe (instead of relying on `memset(0...)` outside)? Is this chain of bytes eventually used like a (zero-terminated) string?

jimklimov (approved):
Seems workable, commented some things I'd improve eventually - but none seem like pressing matters right now.

alexwbaule:
> Ok, should be quiet on my side for now. At least, I'll be away from computers for the evening :)
> 
> Did not yet look much into the changed code itself, but if it works for you and others as an actual driver (as IIRC reported before), that is a solid vouch for it :)
> 
> How are you about putting this into `nutdrv_qx` family, and/or adding the man page?

Hi @jimklimov , to man page i can wrote one... i will see another man as example.

About the `nutdrv_qx` we need to talk about the multiple return on "ok command" that this driver can receive vs "only one valid" on `nutdrv_qx`. (look at the docs/sms-brazil-protocol.txt file, to understand).

alexwbaule:
Hi @jimklimov 

The Man page was added.

drivers/sms_ser.c
alexwbaule: In the doc there is no null terminator. its the "\r" the "ENDCHAR" (is strange). But there is a good idea to fixed all [7] index to NULL, i will do it.

drivers/sms_ser.c
alexwbaule: The LOG_DEBUG is not a const to a number ?  like:
LOG_DEBUG = 3 ?

alexwbaule: Captura de tela de 2023-10-08 11-56-59

drivers/sms_ser.c
alexwbaule: All itens has fixed size, and calling the "ser_send_buf" we inform the length... so, its safe for me.
buffOut and buffIn has a fixed size too.
all "writes and reads" relying on size, send and receive has a fixed size, and i dont know why, but the "terminator char" is "\r" for this driver. Im just following the documentation in the XML.

drivers/sms_ser.c
jimklimov: Well, it is, but in a different "dimension" or unit of measurement. These `LOG*` definitions originate from syslog, I think (e.g. `/usr/include/x86_64-linux-gnu/sys/syslog.h` or our shims for WIN32 builds), not from NUT verbosity... traditions.

I agree it would have been useful to have symbolic names for those levels in NUT, but so far they fared well enough with the inconsistent rogue use like what we have, for a couple of decades and counting.

alexwbaule:
Hi @jimklimov , whats is needed to merge this ? Something is missing, can i do something ? Or just something from CI need to run...? Its my first PR here...

jimklimov:
Seems good, thanks. If moving this into nutdrv_qx is possible and makes sense (would benefit from some shared code patterns?) it can be pursued as another PR and will be welcome :)

jimklimov:
(With CI - it canceled runs because newer code was pushed, but i guess event IDs changed so the newer run did not overwrite the "check" verdict when it succeeded.

PR #2093, opened 2023-10-07
tripplite_usb.c add device matching based on ups.id

maximilanna:
Implemented a function: match_ups_id to read the upsid from config file, read the ups.id from the device and compare them such that the driver can match based on the ups.id (unit id) which is user configureable using upsrw. 

We need to add this to driver documentation.
Probably bump driver version.

Not sure if this code actually works. I didn't compile... theoretically though it should.

### DIFF
--- a/drivers/tripplite_usb.c
+++ b/drivers/tripplite_usb.c
@@ -331,7 +331,7 @@ static int reconnect_ups(void)
 	upsdebugx(2, "= device has been disconnected, try to reconnect =");
 	upsdebugx(2, "==================================================");
 
-	ret = comm_driver->open_dev(&udev, &curDevice, reopen_matcher, NULL);
+	ret = comm_driver->open_dev(&udev, &curDevice, reopen_matcher, match_ups_id);
 	if (ret < 1) {
 		upslogx(LOG_INFO, "Reconnecting to UPS failed; will retry later...");
 		dstate_datastale();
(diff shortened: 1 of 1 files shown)


drivers/tripplite_usb.c
```diff
@@ -331,7 +331,7 @@ static int reconnect_ups(void)
 	upsdebugx(2, "= device has been disconnected, try to reconnect =");
 	upsdebugx(2, "==================================================");
 
-	ret = comm_driver->open_dev(&udev, &curDevice, reopen_matcher, NULL);
+	ret = comm_driver->open_dev(&udev, &curDevice, reopen_matcher, match_ups_id);
```
jimklimov, line 334: You mention the method in reconnect, but declare it later - either move the definition even further up, or add a prototype early in the file (with one, it probably will not like being "static" anymore - similar like when you have prototypes in a header first).

drivers/tripplite_usb.c
```diff
@@ -1582,8 +1609,9 @@ void upsdrv_initups(void)
 	regex_array[4] = getval("serial"); /* probably won't see this */
 	regex_array[5] = getval("bus");
 	regex_array[6] = getval("device");
+	regex_array[7] = getval("upsid");  /* this is a unique number that is settable with upsrw */
 #if (defined WITH_USB_BUSPORT) && (WITH_USB_BUSPORT)
-	regex_array[7] = getval("busport");
+	regex_array[8] = getval("busport");
 #endif
```
jimklimov, line 1615: The `regex_array` is originally sized `USBMATCHER_REGEXP_ARRAY_LIMIT` and I think the common matching code expects certain values to be there in certain positions. For this code to have a chance at working, an `upsid` regex would have to follow after these initial items, and array definition a few lines above extended as `[USBMATCHER_REGEXP_ARRAY_LIMIT+1]`. Not sure if other code won't explode on this sort of hack though :)

In fact with `getval("upsid")` used in `match_ups_id()` directly (which would get called as a callback per comments earlier), and the comparison there done for an exact match, this item may be left out of `regex_array` completely.

What is lacking though is an `addvar()` for the driver to know support such configuration toggle (and save its value if seen in configs into the data arrays where `getval()` will find it later). A suitable spot is in `upsdrv_makevartable()` method, near `nut_usb_addvars()` line - so it is close in the help message.

jimklimov (approved):

drivers/tripplite_usb.c
maximilanna: I fixed all the above in the new pull request; thank you so much for bearing with me. Learning as i go here. I stayed off the regex array, and added the variable to upsdrv_makevartable(), I think the last commit might just work.

PR #2102, opened 2023-10-11
tripplite_usb.c Enhanced Device Identification in the tripplite_usb Driver for Tripp Lite USB UPSes

### LINKED ISSUE #2075: tripplite_usb driver bugs
The man page for the tripplite usb driver states the driver supports extra arguments which it doesn't. Specifically:

1. device, the driver accepts a bus argument but not a device, which becomes a problem when you have two of the same devices
2. serial, the serial isn't available to the driver instead what is passed is "ups.id" which is a user settable identifier with a range of 1 - 65535 using 'upsrw' utility, by default it is set to 65535. i propose this should be utilized instead of serial so that one can monitor multiple units of the same type and model.
3. allow_duplicates, doesn't work, the driver rejects this argument stating incompatibility. 
--
@jimklimov @seanm @sshimko @mbastiaan

sapireli:
@jimklimov My apologies, I screwed up the pull request with a rebase. I just started a new pull request to have a clean slate scenario. Added the NUT_UNUSED_VARIABLE  per your recommendation.

background:
tripplite_usb.c Added UPS ID (upsid) Support: The code now includes support for matching and uniquely identifying UPS devices using the UPS ID (upsid). The upsid configuration option has been added, which accepts a regular expression to match the UPS ID string. This was done by adding a new function called match_by_unitid, that reads the upsid from the config file, then reads the upsid from the device, and checks if they match. This function is now passed to the comm_driver matcher on first connect and reconnects.

Added the upsid to the upsdrv_makevartable function so it is read and stored from the config file.

tripplite_usb.txt Updated Documentation: The driver documentation was updated to reflect the new functionality. The "EXTRA ARGUMENTS" section now includes information about the upsid configuration option, and how to set it.

thank you to @jimklimov for pointing me in the right direction and hand holding.
closes #2075

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -178,6 +178,9 @@ as part of https://github.com/networkupstools/nut/issues/1410 solution.
 
  - powercom driver should now try harder to refresh data from device [#356]
 
+ - tripplite_usb driver now supports configuration of `upsid` to match the
+   specific device (not all firmware/hardware models support this) [#2075]
+
  - apcupsd-ups:
    * improvement for `POLL_INTERVAL_MIN` from PR #797 was buggy [#2007]
    * fix to clean obsoleted readings (if any) AFTER getting new info from an
--- a/docs/man/tripplite_usb.txt
+++ b/docs/man/tripplite_usb.txt
@@ -63,6 +63,14 @@ file (or with '-x' on the command line):
 
 include::nut_usb_addvars.txt[]
 
+- `-x upsid="12345"`
+
+Select a specific UPS by its unique UPS ID. The argument is a regular expression 
+that must match the UPS ID string. This allows for precise identification of UPS 
+devices when multiple devices of the same make and model are connected.
+See below regarding how to read and write the ups id (unit id) using linkman:upsrw[8].
+
+
 [NOTE]
 .Notes for `tripplite_usb` driver handling of common USB matching settings:
 ======
@@ -110,7 +118,7 @@ changed at runtime by linkman:upsrw[8].
 
 *ups.id*::
 
-Some SMARTPRO models feature an ID that can be set and retrieved.
+Some SMARTPRO models feature an Unit ID (ups.id) that can be set and retrieved.
 If your UPS supports this feature, this variable will be listed in
 the output of linkman:upsrw[8].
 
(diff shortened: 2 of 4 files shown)


sapireli:
We have success. I think we're good to go finally. The only thing i was thinking about is nut-scanner, how difficult would it be to add the ups.id to it?

jimklimov:
Regarding rebase blowing up - oh well, that happens :)

Some hints for future, if you would generally continue with git-based development:
* `git reflog` is your friend to review the local git index history, so among other things - you can `git reset` to a pre-rebase state, `git checkout -f` to force-checkout to reset the actual file contents, and try again
* `git push --force-with-lease` is a relatively safe variant of force-push - it checks that the remote repository is in a state compatible with changesets you modified locally; e.g. if I were to push to your remote (github) master branch to fix some bug, your usual force-push would forget this fix, but the "lease" one will abort the push attempt and require you to fetch the current remote state, merge/rebase it in your local workspace, and then push again
* do not develop on your copy of a master/main/trunk named branch - this complicates updates vs. same-named upstream :)

Regarding "good to go": There are several CI's involved across many platforms, waiting for all of them to green-light the changes :)

Regarding nut-scanner: this is probably the harder piece of the puzzle. There is a bit of precedent for the nut-scanner library using code from some drivers during the linking procedure. So there is impact on makefiles, possibly on the set of exported methods or variables from libnutscan (at least, there was recently for debug verbosity methods to behave; this one might be an internal detail of the scan-usb.c link), static or not for the match method... Related approach seen in the codebase is to use ifdef's to copy/symlink use the driver source file to nut-scanner sources during build and compile it "as is" but to not actually build the majority of its code (which is ifdef'ed away), leaving just the methods you want in nut-scanner. This sounds complicated, but in fact simplifies maintenance to only have one implementation of that method for all build variants. A cleaner approach could be to extract it into a helper C source file and build/link it for the driver and for nut-scanner. Effectively same as the ifdef arcane magic, but easier to understand :)

jimklimov:
I think it makes sense to do nut-scanner in a different PR. So once we stabilize and merge this change, you can build the next one on top. Can privately prepare though (in a separate branch) :)

sapireli:
the git reset is what i was missing, thank you!

jimklimov:
Seems good for CI, merged. Awaiting experiments regarding `nut-scanner` (hopefully as one PR and made in a non-"master" branch this time ;) )

Issue #2109, opened 2023-10-16
compiler warning: printf allows null string args... no (new/broken check?)


desertwitch:
Compiling on my usual (unchanged) build environment on Slackware 15.0, gcc 11.2.0:

**last build from 3bf526e72332d01a6242a87188ba21c2fdc628ab**
```
checking for practical support to printf("%s", NULL)... yes
```
```
checking whether printf allows null string args... (null)ok
```

**new build from current master (after 3bf526e72332d01a6242a87188ba21c2fdc628ab)**
```
checking for practical support to printf("%s", NULL)... yes
```
```
checking whether printf allows null string args... no
configure: WARNING: Your C library requires workarounds to print NULL values; if something crashes with a segmentation fault (especially with verbose debug) - that may be why
```

**Attempt to re-create the check on my build system:**
```
#include <stdlib.h>  
#include <stdio.h>
int main() 
{ 
   printf("%s", NULL); 
   return 0;
} 
```
```
(null)
```
I was trying to pin-point this from the commits since, but I couldn't figure out what has changed that could cause this.

jimklimov:
No, nothing was changed recently that should have impacted this... 

Is it a cross-build or native (so executables can run locally)? Can something like an antivirus get in the way of running test programs, having their newly written binaries locked for its own checks - breaks a lot of active logic on Windows, for example.

The autotools `config.log` should have a detailed log of what it ran and what errors it got. It is usually helpful...

desertwitch:
> No, nothing was changed recently that should have impacted this...
> 
> Is it a cross-build or native (so executables can run locally)? Can something like an antivirus get in the way of running test programs, having their newly written binaries locked for its own checks - breaks a lot of active logic on Windows, for example.
> 
> The autotools `config.log` should have a detailed log of what it ran and what errors it got. It is usually helpful...

It's a native build on Slackware 15.0 with gcc 11.2.0
No antivirus or the likes, the build environment is entirely unchanged.

Actually I've checked-out 3bf526e72332d01a6242a87188ba21c2fdc628ab again just now and built it with the null arg check passing.
Afterwards I've checked-out the latest master state and built with the null arg check failing.

desertwitch:
Here goes **config.log**:
```
configure:14136: checking whether printf allows null string args
[...]

PR #2111, opened 2023-10-16
Fix loss of configure-phase detection due to recent enhancements

jimklimov:
Closes: #2109
Follows up from: #2096, #823

As now posted, the PR should allow configure script to gauge practical support of `printf("%s", NULL)` again, and refactors the two detections of this ability, that have evolved over time, into one.

Also includes a little fix to same feature that broke detection of C++ CPPUNIT testability support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1205,8 +1205,8 @@ AC_CACHE_CHECK([for inet_pton() with IPv4 and IPv6 support],
             [[/* int inet_pton(int af, const char *src, char *dst); */
 struct in_addr  ipv4;
 struct in6_addr ipv6;
-printf("%i", inet_pton(AF_INET, "1.2.3.4", &ipv4));
-printf("%i", inet_pton(AF_INET6, "::1", &ipv6))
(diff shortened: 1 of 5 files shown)


desertwitch:
Thanks, this compiles nicely with my Slackware 15.0/gcc 11.2.0 build environment now:
```
checking for practical support to printf("%s", NULL)... (null)yes
```

```
configure:12389: checking for practical support to printf("%s", NULL)
configure:12508: x86_64-slackware-linux-gcc -o conftest -isystem /usr/local/include -g -O2 -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Werror -Werror=implicit-function-declaration -Wno-format-truncation -Wno-format-overflow
       conftest.c  >&5
configure:12508: $? = 0
configure:12508: ./conftest
SUCCESS: RETURNED a string that contains something like 'null' from snprintf() with a variable NULL string argument: '(null)'
configure:12508: $? = 0
```

desertwitch:
Also compiles without problems on my other build environment Slackware 15.1/gcc 13.2.0

PR #2117, opened 2023-10-18
Follow-up for `apc_modbus` driver

jimklimov:
NUT scaffolding add-ons for `apc_modbus` driver introduced with #2063

CC @EchterAgo - LGTY?


jimklimov:
Parking as draft until merge of #2063 - so CI does not spend cycles on updated rebuilds as the master branch tip changes over time with other development.

EchterAgo:
Is there an easy way to setup the CI on a private host? I have a fast host here that mostly idles on which I could run tests before pushing.

PR #2120, opened 2023-10-18
nut_usbinfo.pl: Lowercase the USB ids before inserting

EchterAgo:
If we don't normalize the case before inserting VID/PID/Driver into `$vendor` we can have duplicate entries.


EchterAgo:
Maybe this should just use `lc()` because C also accepts hex numbers with `0X` prefix.

PR #2133, opened 2023-10-27
clients/upsmon.c et al: introduce SHUTDOWNEXIT boolean/delay-length option [#2133]

jimklimov:
As suggested by mailing list: "secondary" systems that take long to stop their workloads can get a power-cut too early because their `upsmon` exits just after calling `SHUTDOWNCMD` and so allow the primary to stop and `killpower`. This PR makes such behavior configurable - adding an option for `upsmon` to linger after telling its system to shut down.

TODO/Idea for further improvement: Handle numbers as the time to linger, not as a boolean?

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -428,25 +428,33 @@ as part of https://github.com/networkupstools/nut/issues/1410 solution.
  - Clarified documentation in codebase according to end-user feedback [#1721,
    #1750 and others over time]
 
- - Several fixes for `upsmon` behavior [#1761, #1680...], including new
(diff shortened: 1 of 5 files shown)


jimklimov:
Ah, the pedantic strictness... now the "noreturn" annotation is no longer valid - the method does not ALWAYS `exit()` the program :D

jimklimov:
Yay, finally, issues in jenkins-dynamatrix got ironed out and the commit status backref URLs work - pointing to the log of the failed matrix cell sub-stage right from github! :)

PR #2136, opened 2023-10-29
Introduce `NUT_QUIET_INIT_UPSNOTIFY=true` support

jimklimov:
Issue raised in mailing list, follow-up to #1590 and related issues: messages about *inability* to notify a framework are considered noise on OSes without any framework to notify. They can be important troubleshooting aid on systems *with* frameworks however.

* https://alioth-lists.debian.net/pipermail/nut-upsuser/2023-October/013457.html


desertwitch:
Nice, this'll also come in handy for me, just tested and works here... :-)

Issue #2137, opened 2023-10-29
Incorrect CRC calculation in Riello driver.


cool0nik:
In file nut/drivers/riello.c

This line:
					pom = (pom << 8) ^ (pom << 3) ^ (pom >> 4);
was replaced by:
					pom  = (uint16_t)(pom << 8);
					pom ^= (pom << 3);
					pom ^= (pom >> 4);
					
This is wrong changes.
The result of CRC calculation was changed.
UPS does not working with this change and change should be reverted.

This "optimization" was introduced here:
https://github.com/networkupstools/nut/commit/6a2e042e3e0fa6a821a9edb444f042fbc81723c3

Issue was found during upgrade Debein 11 -> 12.
from 
Network UPS Tools - Riello serial driver 0.03 (2.7.4)
to
Network UPS Tools - Riello serial driver 0.07 (2.8.0)

jimklimov:
Thanks for the report. Would you be able to check the PR above, with e.g. https://github.com/networkupstools/nut/wiki/Building-NUT-for-in%E2%80%90place-upgrades-or-non%E2%80%90disruptive-tests ?

PR #2138, opened 2023-10-30
drivers/riello*.c: fix back bit maths changed with #1106

jimklimov:
Closes: #2137

Follows up from [#1106 int-type changes to riello.c (among many others)](https://github.com/networkupstools/nut/pull/1106/files#diff-547696e27242b307a42b56fef3bc671cdeb0266794d735e2bb40a16f6a32db7f)

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -114,6 +114,7 @@ as part of https://github.com/networkupstools/nut/issues/1410 solution.
      (without specifying the single device) [#1759, #1806, #1875]
    * The `apcsmart` and `apcsmart-old` handled invalid data too zealously
      and aborted instead of skipping over it, like they did before [#2015]
+   * A bit maths optimization in `riello_ser` and `riello_usb` misfired [#2137]
    * Something about compile-time macros or other warnings-related refactoring
      seems to have confused the MGE SHUT (Serial HID UPS Transfer) driver
      support [#2022]
--- a/drivers/riello.c
+++ b/drivers/riello.c
@@ -67,14 +67,12 @@ uint16_t riello_calc_CRC(uint8_t type, uint8_t *buff, uint16_t size, uint8_t che
 				size--;
 				CRC_Word = 0x554D;
 				while(size--) {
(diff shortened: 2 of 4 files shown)


jimklimov:
Hoping for end-user testing via #2137 but not holding out for it: the reported used packaged NUT on one hand, and this reverts the bit of code broken by optimizations to what single-liner it was before (and allegedly working in 2.7.4).

The crux of the issue was that the original (and now restored) code picked several bits by shifting the original `pom` value left and right, and assigned into new `pom` all together in one operation. Optimization split that into several bit-shifts and picks, reassigning `pom` every time - so picking probably zeroes always for the subsequent bits. It could have worked if some `pom_original` was stashed for the several shifting selections, but that did not happen in practice.

jimklimov:
Per `git diff` (excerpt) these code lines are restored to previous glory:
````
:; git diff 6a2e042~1..origin/issue-2137 drivers/riello.c
diff --git a/drivers/riello.c b/drivers/riello.c
index 8b5a78e87..ccbaee6ab 100644
--- a/drivers/riello.c
+++ b/drivers/riello.c
...
@@ -68,8 +67,11 @@ uint16_t riello_calc_CRC(uint8_t type, uint8_t *buff, uint16_t size, uint8_t che
                                size--;
                                CRC_Word = 0x554D;
                                while(size--) {
-                                       pom = (CRC_Word ^ *buff) & 0xff;
-                                       pom = (pom ^ (pom << 4)) & 0xff;
+                                       pom = (CRC_Word ^ *buff) & 0x00ff;
+                                       pom = (pom ^ (pom << 4)) & 0x00ff;
+                                       /* Thanks to &0x00ff above, pom is at most 255 --
+                                        * so shifted by 8 bits is still uint16_t range:
+                                        */
                                        pom = (pom << 8) ^ (pom << 3) ^ (pom >> 4);
                                        CRC_Word = (CRC_Word >> 8) ^ pom;
                                        buff++;
...
````

PR #2146, opened 2023-11-03
m4/nut_compiler_family.m4: fix cross-compilation

ffontaine:
Do not use isystem with host paths when cross-compiling


jimklimov:
Thanks, looks reasonable. Assuming `cross_compiling` gets set when applicable? Did you check it works for actual cross builds?

So far one I had experience with was linux+mingw to build for windows, and as much as codified in those helper scripts - it worked...

ffontaine:
Yes, it works under [buildroot](https://buildroot.org) and allows to drop an "ugly" hack

PR #2147, opened 2023-11-05
m4/nut_check_python.m4: allow overriding of PYTHON{2,3}_SITE_PACKAGES

ffontaine:
Allow the end-user to override `PYTHON{2,3}_SITE_PACKAGES` as this can be useful when cross-compiling

### DIFF
--- a/m4/nut_check_python.m4
+++ b/m4/nut_check_python.m4
@@ -116,17 +116,17 @@ AC_DEFUN([NUT_CHECK_PYTHON],
         AM_CONDITIONAL([HAVE_PYTHON], [test -n "${PYTHON}" && test "${PYTHON}" != "no"])
         AS_IF([test -n "${PYTHON}" && test "${PYTHON}" != "no"], [
             export PYTHON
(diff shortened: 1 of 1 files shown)


jimklimov:
Thanks, seems reasonable :)

Let's see how CI likes it. Eventually, it would be great to also get a PR for updates to docs/config-prereqs.txt and/or docs/ci-farm-lxc-setup.txt about setting up a worker for cross-build scenarios in the NUT CI farm.

So far there are very occasional (manually triggered) runs in whole containers running under qemu for this or that platform - which is prohibitively slow to be done regularly, and sometimes blows up with unimplemented CPU instructions etc.

If there could be some mix, e.g. using a native (amd64) gcc or clang binary with appropriate resource file trees to produce ARM etc. binaries, and only use qemu-static to actually run them for `make check check-NIT` tests, that would probably be a workable balance. For best results, we
 may also have to teach our `AX_RUN_OR_LINK_IFELSE` m4 macro about some caller-defined option that this cross-builder can also run the binaries it has just built (which is not the default case).

PR #2148, opened 2023-11-06
Add usb_config_index to usb_communication_subdriver_s, default to 0.

kellybyrd:
The tactical goal of this change is to change
                ret = libusb_get_config_descriptor(device,
                        (uint8_t)usb_subdriver.hid_rep_index,
                        &conf_desc);
to
                ret = libusb_get_config_descriptor(device,
                        (uint8_t)usb_subdriver.usb_config_index,
                        &conf_desc);

Before this change, libusb1.c did libusb_get_config_descriptor() with a config index equal to the interface number. For composite devices using an interface index > 0, this is usally the wrong choice. Concretely, I'm using an Arduino for a DIY UPS project and these are composite devices with multiple interfaces under the first (and only) config descriptor.

In the USB descriptor heirarchy, deivce descriptors have config descriptors which have interface descriptors. Also, nearly all USB devices have a single configuration (index 0).

In order to do this, I added a new member alongside the existing hid_rep_idex and hid_desc_index. I chose to do this instead of using the add_var method because this member is used in places in very similar ways to how hid_rep_index and hid_desc_index is used. This new member defaults to 0 which covers the majority of USB devices. Any future subdriver is able to use this if a device requires it. For existing subdrives, we'll just use an index of 0.

I also changed some debug logging to print out the config index where the code was already printing the interface index.

## General points

  comments or text documents.

  case.

  or its equivalent.

  concept.

  the changed codebase.

### DIFF
--- a/drivers/usb-common.h
+++ b/drivers/usb-common.h
@@ -122,6 +122,10 @@
  #define USB_CTRL_MSGVALUE_MIN	0
  #define USB_CTRL_MSGVALUE_MAX	UINT16_MAX
 
+ typedef uint8_t usb_ctrl_cfgindex;
+ #define USB_CTRL_CFGINDEX_MIN	0
+ #define USB_CTRL_CFGINDEX_MAX	UINT8_MAX
+
  typedef uint16_t usb_ctrl_repindex;
  #define USB_CTRL_REPINDEX_MIN	0
  #define USB_CTRL_REPINDEX_MAX	UINT16_MAX
@@ -410,6 +414,10 @@
  typedef int usb_ctrl_msgvalue;
  #define USB_CTRL_MSGVALUE_MIN	INT_MIN
  #define USB_CTRL_MSGVALUE_MAX	INT_MAX
+
+ typedef uint8_t usb_ctrl_cfgindex;
+ #define USB_CTRL_CFGINDEX_MIN	0
+ #define USB_CTRL_CFGINDEX_MAX	UINT8_MAX
 
  typedef int usb_ctrl_repindex;
  #define USB_CTRL_REPINDEX_MIN	INT_MIN
--- a/drivers/libusb0.c
+++ b/drivers/libusb0.c
@@ -435,8 +435,9 @@ static int libusb_open(usb_dev_handle **udevp,
 				}
 
 				fatalx(EXIT_FAILURE,
-					"Can't claim USB device [%04x:%04x]@%d/%d: %s",
+					"Can't claim USB device [%04x:%04x]@%d/%d/%d: %s",
 					curDevice->VendorID, curDevice->ProductID,
+					usb_subdriver.usb_config_index,
 					usb_subdriver.hid_rep_index,
 					usb_subdriver.hid_desc_index,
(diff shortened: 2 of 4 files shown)


jimklimov:
For clarification, the `addvar()` idea (raised in the mailing list) regards the way for end-users to set a specific number value via configuration variables their `ups.conf` sections, if for some reason they don't want to default to 0 but to something else.

drivers/usb-common.h
```diff
@@ -122,6 +122,10 @@
  #define USB_CTRL_MSGVALUE_MIN	0
  #define USB_CTRL_MSGVALUE_MAX	UINT16_MAX
 
+ typedef uint8_t usb_ctrl_cfgindex;
```
jimklimov, line 125: May need a copy around line 420 for libusb-0.1 compatibility.

jimklimov: Yeah, Appveyor noted this too:
````
n file included from nutdrv_qx.c:50:
nut_libusb.h:84:9: error: unknown type name 'usb_ctrl_cfgindex'
   84 |         usb_ctrl_cfgindex usb_config_index;             /* index of the device config we use. Almost always 0; see comments above */
      |         ^~~~~~~~~~~~~~~~~
cc1.exe: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics
cc1.exe: note: unrecognized command-line option '-Wno-reserved-identifier' may have been intended to silence earlier diagnostics
In file included from libusb0.c:34:
nut_libusb.h:84:9: error: unknown type name 'usb_ctrl_cfgindex'
   84 |         usb_ctrl_cfgindex usb_config_index;             /* index of the device config we use. Almost always 0; see comments above */
      |         ^~~~~~~~~~~~~~~~~
````

kellybyrd:
> For clarification, the `addvar()` idea (raised in the mailing list) regards the way for end-users to set a specific number value via configuration variables their `ups.conf` sections, if for some reason they don't want to default to 0 but to something else.

Ahh, gotcha! IMO, it's really only useful to have this one be a config var, if the others are too. So that would be
usb_config_index
usb_ctrl_repindex
hid_desc_index
hid_ep_in
hid_ep_out

 
> Such user-friendly integration can be addressed separately (more so if this bites someone); so far it is great to have separate (potentially-)manageable values with recognized semantic difference :)

Cool, I'll leave this PR without that then. Doing the above set of value might let someone handle odd hardware without a code change. I *think* `lsusb` can be made to output enough information for a user to set those values in a config.

drivers/usb-common.h
kellybyrd: Fixed (I hope) and pushed. I'll check in on this in the morning. 

I also need to get the rest of NUT setup and get this working with my DIY project!

jimklimov:
So, that fault for shellcheck was irrelevant (a github connection issue). Otherwise, CI seems mostly satisfied (not all scenarios finished yet, others green).

jimklimov:
Thanks! Tested with my USB UPS, still works. Not all zeroes were created equal ;)

PR #2155, opened 2023-11-08
Misc fixes

dougnazar:
A couple small one-line fixes:

* Fix SIGSEGV while preforming INSTCMD
* Fix passing of errors up from libusb 1.x

### DIFF
--- a/drivers/dstate.c
+++ b/drivers/dstate.c
@@ -780,7 +780,7 @@ static int sock_arg(conn_t *conn, size_t numarg, char **arg)
 			upsdebugx(3, "%s: TRACKING = %s", __func__, cmdid);
 
 		/* try the handler shared by all drivers first */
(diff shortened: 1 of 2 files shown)


jimklimov:
Thanks! The SIGSEGV one seems like this-year's addition, so not a part of earlier releases and I suppose too few people ran the new code with actual commands to notice it :( and neither did static analyzers :\

I suppose this misfires with all single-token commands without an argument?

Out of curiosity, in what sort of workflow did you detect it?

dougnazar:
I'd assume so. It happened due to my monthly battery test. Just got lucky that it happened while I was working on the driver.

jimklimov:
Note for self: latest commit checks "failure" was due to a master-branch mischief (not git-ignored touch-file).

PR #2163, opened 2023-11-09
python: Do not install setup.py

gdt:
The code as is installs the python module's setup.py in the site-packages directory; this is incorrect.

This commit removes setup.py from an automake variable, which resolves the issue.  I do not know if it impacts any other kind of module publicaition process.  We may need a more complicated fix.

### DIFF
--- a/scripts/python/Makefile.am
+++ b/scripts/python/Makefile.am
@@ -86,7 +86,6 @@ PYNUT_TEMPLATE = \
 	module/test_nutclient.py.in
 
 PYNUT_GENERATED_NOEXEC = \
-	module/setup.py \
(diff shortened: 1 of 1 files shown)


jimklimov:
Yeah, thanks, that was probably an overkill, and a long night :)

Can you please check if `setup.py.in` comes into dist archives?

gdt:
make distcheck resulted in PyNUT.py being installed.
Building the generated tarball in pkgsrc succeeded, with PyNUT.py installed.
Yes, setup.py.in is in the distfile.

jimklimov:
Thanks!

PR #2164, opened 2023-11-10
Revise NUT USB id list files generation (nutscan-usb.h et al)

jimklimov:
Following up on mail comments.

Also help ensure that `autogen.sh` would re-generate the files if relevant sources got updated.

CC @gdt


autogen.sh
```diff
@@ -90,12 +90,19 @@ fi
 #    grep -i '">' tools/nut-usbinfo.pl
 # * List the names involved:
 #    grep -E 'output.*=' tools/nut-usbinfo.pl
+# Also check that the last re-generation is newer than the sources involved
+# (stay on top of CI rebuilds, development, Git branch switching...)
+# Someone please tell me why GNU `find dir -newer X -name Y -o -name Z` does
+# not filter away layer by layer, but rather finds the names Z and beyond
```
gdt, line 100: I think that means newer than X and Y, or called Z (with no newer).  And binds tighter.  You need () maybe.   But what you have seems sensible and I don't suggest you redo it.

tools/Makefile.am
```diff
@@ -101,7 +101,7 @@ $(GENERATED_USB_OS_FILES): $(GENERATED_USB_FILES)
 	 fi
 	@touch "$@"
 
-$(GENERATED_USB_FILES): $(top_srcdir)/drivers/*-hid.c $(top_srcdir)/drivers/*usb*.c $(top_srcdir)/drivers/nutdrv_qx.c
+$(GENERATED_USB_FILES): $(top_srcdir)/drivers/*-hid.c $(top_srcdir)/drivers/*usb*.c $(top_srcdir)/drivers/nutdrv_qx.c $(top_srcdir)/tools/nut-usbinfo.pl
```
gdt, line 104: Is the .pl file really *generated*?

tools/Makefile.am
```diff
@@ -60,7 +60,7 @@ nut-scanner-deps-usb: $(GENERATED_USB_FILES)
 
 # The distributed nut-snmpinfo.py.in template is assumed to only differ from
 # a generated nut-snmpinfo.py by the @PYTHON@ shebang.
-$(GENERATED_SNMP_FILES): $(top_srcdir)/drivers/*-mib.c
+$(GENERATED_SNMP_FILES): $(top_srcdir)/drivers/*-mib.c $(top_srcdir)/tools/nut-snmpinfo.py.in
```
gdt, line 63: Is that really generated?  It is checked in to git.

jimklimov: That's not assignment, that's dependencies to trigger a build of generated stuff. Also above.

autogen.sh
jimklimov: I guessed so, but no sort of explicit `-and`/`-or` magic helped. Not sure how portable parentheses are, so went for plain `-newer` as something all find's I've seen can do  :)

tools/Makefile.am
gdt: ENOTENOUGHCOFEE!

gdt:
I have not tried to run this, but it looks good and I'm fine with merging it.

Issue #2168, opened 2023-11-11
is_ups_critical: seems that UPS [xxx] is OB+LB now, but it is also calibrating : but it's not (longer) true (from 2.8.0)


GreyWolfBW:
Hi everyone,

During a power cut long enough to bring my UPS batteries to a critical level, I noticed that my server was not shut down by NUT as I expected.

I think this is a bug in UPSMON since version 2.8.0, here is what I noticed :

First of all, I'm using an **ONLINE-USV XANTO 2000R UPS**, connected to my server via a USB cable and therefore managed by NUT with the **nutdrv_qs** driver with **cypress voltronic** subdriver protocol. The server's operating system is Debian 12 Bookworm.

I've been using this model of UPS since Debian 11 or Debian 10 and NUT without encountering this problem.

On Debian 11 Bullseye, the NUT version supplied is 2.7.4.

I've already used this UPS on this version and management is very satisfactory. There's no relevant battery runtime feedback. So, I prefer to use the "_ignorelb_" flag and set a low battery threshold via "_override.battery.charge.low_" to trigger server shutdown at a predefined battery level.

With Debian 12 Bookworm, the **NUT version is 2.8.0**.

Unfortunately, it didn't work with this version. So, I tried to figure out what was going on.

The log clearly shows this line :
`is_ups_critical: seems that UPS [Onduleur] is OB+LB now, but it is also calibrating - not declaring a critical state`

At first glance, among the changes between these two versions, UPSMON can now manage the "_calibration_" mode of an UPS.

I think it's something universal, but every UPS I know of performs a ten-second battery self-test every one or two weeks. During a self-test, a call to UPSC indicates the status "_OL CAL_" with this model of UPS. UPSMON notices this because a log entry reads "_**UPS Onduleur: calibration in progress**_".

If I understand the UPSMON source code correctly, it records the calibration status in a variable ( `setflag(&ups->status, ST_CAL);` [on line 864 in nut/clients/upsmon.c](https://github.com/networkupstools/nut/blob/ff16dabca191e5fd8ddc20137317bdebee554d8d/clients/upsmon.c#L852C1-L865C2) ). **But at no time do I see it deleting this status ?** And I think this is a problem.

So, when there's a power cut, the UPS switches to battery power and when the battery is low, the _is_ups_critical_ function thinks that the UPS is in calibration and that the server shouldn't be shut down.

Here's an example :

Version tested = v2.8.0 supplied by Debian 12 Bookworm :

> apt policy nut
[...]

PR #2169, opened 2023-11-11
It's important to reset the "CAL" status. This will be define later if necessary.

GreyWolfBW:
See issue 2168 :
https://github.com/networkupstools/nut/issues/2168

Be careful I'm a beginner


jimklimov:
Built a bit on top of the original PR, to align with the current (not earlier release) codebase, and to add notifications for leaving the calibration state.

PR #2184, opened 2023-11-14
apc_modbus: Updates, command, writable variables and outlet group support

### LINKED ISSUE #2119: apc_modbus: Add support for commands
This is for after merging #2063 .

We need to add support for UPS commands. This is what we can add:

- UPSCommand_BF
  - RestoreFactorySettings

- OutletCommand_BF
  - Cancel
  - OutputOn
  - OutputOff
  - OutputShutdown
  - OutputReboot

  - args:
    - ColdBootAllowed
    - UseOnDelay
    - UseOffDelay

  - target:
    - UnswitchedOutletGroup
    - SwitchedOutletGroup0
    - SwitchedOutletGroup1
    - SwitchedOutletGroup2

- SimpleSignalingCommand_BF
  - RequestShutdown
  - RemoteOff
  - RemoteOn

- ReplaceBatteryTestCommand_BF (incorrectly named UPSStatus_BF in docs)
  - start test

- RunTimeCalibrationCommand_BF
  - start calibration

- UserInterfaceCommand_BF
  - mute
  - cancel mute
  - acknowledge battery alarm

EchterAgo:
This adds support for commands, writable variables and outlet groups.

- For commands there is a new table called `apc_modbus_command_map` which defines the supported commands as a tuple of command name, register offset and value to write. This also adds a new instcmd status called `STAT_INSTCMD_CONVERSION_FAILED` for when conversion of values fails. On startup all the commands are registered using `dstate_addcmd`. This also adds support for the `upsdrv_shutdown` function.
- For writable variables we added a new flag called `APC_VF_RW` to the existing variables that indicates a writable variable. We added code to convert from a string to UINT/INT/STRING variables with output in APCs register format. There is a new `_apc_modbus_setvar` function that handles setting variables and rereading them from the device. This also adds a new setvar status called `STAT_SET_CONVERSION_FAILED` for when conversion of values fails. Variables are now correctly set as `ST_FLAG_STRING` and `ST_FLAG_RW` and we call `dstate_setaux` to give a maximum length for strings.
- For outlet groups, we now have names, configurable delays and commands per outlet group. Devices have an outlet group called MOG (Main outlet group) that switches all the outputs of the UPS and 1-3 SOGs (Switched outlet groups) that can be controlled independently. Note that MOG is `outlet.group.0` and the SOGs start at `outlet.group.1`. The outlet groups should have markings with the same index at the back of the unit.
- This also reduces the length of some of the defines to make the variable maps more readable.
- It also adds a comment to when the reopen matcher is created that clarifies why we create it.

fixes #2119

## General points

  comments or text documents.

  case.

- [x] For updated existing device drivers, bumped the `DRIVER_VERSION` macro
  or its equivalent.

  concept.

- [x] Updated `data/driver.list.in` if applicable (new tested device info)

- [x] Updated `docs/acknowledgements.txt` (for vendor-backed device support)

  the changed codebase.


EchterAgo:
Please give this some more testing. I did test a bit already but was unable to test controlling the load on the UPS.

@PrplHaz4

EchterAgo:
@jimklimov I'd still like to know what to do regarding outlet group commands, see https://github.com/networkupstools/nut/issues/2119#issuecomment-1810230174

EchterAgo:
What I also need to know if the `load.*` and `shutdown.*` commands work or if we need to specify the MOG as a target.

EchterAgo:
@jimklimov thanks for the header change, I forgot that.

jimklimov:
@EchterAgo : sorry, I'm a lot on the road and in meetings over this week, could not give much thought to your questions :\

CC @aquette : any ideas regarding outlets (individual/groups) management here?

EchterAgo:
I noticed `eaton-pdu-marlin-mib.c` already uses `outlet.group.*` for commands.

jimklimov:
@EchterAgo : I think your force-push removed `apc_modbus.h` from the `Makefile.am` again?..

EchterAgo:
@jimklimov readded

jimklimov:
@EchterAgo : got a detailed testing report in the mailing list : https://alioth-lists.debian.net/pipermail/nut-upsuser/2023-November/013486.html

Seems there's some work still needed here and there... :)

EchterAgo:
> @EchterAgo : got a detailed testing report in the mailing list : https://alioth-lists.debian.net/pipermail/nut-upsuser/2023-November/013486.html
> 
> Seems there's some work still needed here and there... :)

That is good to hear.

I'll look into the power switching commands, already have an idea what might be wrong. Sadly I can not test this here at the moment because I'm currently running a 200TB data migration and don't want to interrupt that.

As for the `battery.date` issue, I though I tested that, will look again.

I did test the `calibrate.start` and `calibrate.stop` commands.

I'll implement `ups.test.result`.

EchterAgo:
Rebased and added `ups.test.result`

@jimklimov DCO is not happy because of a mismatch between your emails:

> Commit sha: [6d08119](https://github.com/networkupstools/nut/pull/2184/commits/6d08119feeef422e34cae72557ab0018706dcd37), Author: Jim Klimov, Committer: Axel Gembe; Expected "Jim Klimov [jimklimov@gmail.com](mailto:jimklimov@gmail.com)", but got "Jim Klimov [jimklimov+nut@gmail.com](mailto:jimklimov+nut@gmail.com)".

EchterAgo:
I think I found the `battery.date` issue, `strptime` produces a `struct tm` with `tm_isdst`, `tm_gmtoff` and `tm_zone` based on the current time zone.

`_apc_modbus_date_from_nut` works with local time while `_apc_modbus_date_to_nut` works with UTC. I'm going to make everything work with UTC.

EchterAgo:
Pushed new commits, `setvar` issue is fixed and the `battery.date` issue too.

EchterAgo:
I also added the target outlet group to the `load.*` and `shutdown.*` commands. I think this might be what is causing those commands not to work, not sure about the `outlet.*` commands, I don't know if those were specifically tested.

EchterAgo:
@jimklimov do you want me to make the author and signed-off-by in your commit use the same email address? If so, with or without the `+nut` suffix?

EchterAgo:
mingw build:

```
apc_modbus.c: In function '_apc_modbus_date_from_nut':
apc_modbus.c:784:27: error: implicit declaration of function 'timegm'; did you mean 'time'? [-Werror=implicit-function-declaration]
  784 |         if ((epoch_time = timegm(&tm_struct)) == -1) {
      |                           ^~~~~~
      |                           time
```

looks like we need another way to fix this or implement a `timegm` alternative for Windows.

EchterAgo:
Or we work with local time, but then we have to also calculate `apc_date_start_offset` based on the current time zone.

jimklimov:
> author and signed-off-by...?

Yes, please - probably gotta make the mail in comment match the one in commit metadata (not sure if commit GPG carries over rebases somehow, but would be tied to that ID probably)

EchterAgo:
> > author and signed-off-by...?
> 
> Yes, please - probably gotta make the mail in comment match the one in commit metadata (not sure if commit GPG carries over rebases somehow, but would be tied to that ID probably)

Fixed. The commit is GPG signed with my key (1CB3AA2143109AAC) now, but if you provide a new signed commit on top of 8be037acf55df47f9a44b7b180788876d4f80f3b I'll replace it.

jimklimov:
At least, glancing over our fallback `common/strptime.c` - it seems to consider timezones as well as DST offsets. Perhaps if you read the user string with the date and append a `+0000` or whatever specifier it would use for explicit UTC, it might work just as fine (assuming the libc implementations on this or that platform behave similarly in this regard).

EchterAgo:
The reading of the timestamp string isn't the problem, the problem is that `mktime` uses the local timezone without a way to diable it, which is what `timegm` is for. 

We could instead always use the local timezone, but I think I'd need to add code to initialize `apc_date_start_offset` based on the current TZ, otherwise we'd always have +-1 days

EchterAgo:
I think I fixed the `timegm` issue by using `_mkgmtime` on Windows. Still need to rebase.

EchterAgo:
FYI, if we want an actual fallback implementation for `timegm`, there are some here:

https://stackoverflow.com/questions/16647819/timegm-cross-platform

The one from boost should be license compatible and easy to port.

jimklimov:
@EchterAgo : more testing results coming in the mail thread, many things proven working :)

One point that got me concerned was about (mis-)treating battery calibration as a "TEST" state. There are some places where we treat the "CAL" state specially (e.g. for requesting a shutdown or not when the battery charge goes low). Now that I looked for it, I saw a few more cases of `status_set("TEST")` in older drivers, but did not right now revise what they were about (battery or some other tests?)

FWIW, `docs/new-drivers.txt` does mention `CAL` but not a `TEST`...

jimklimov:
Further discussion and practical testing reports in Mailing list:
* https://alioth-lists.debian.net/pipermail/nut-upsuser/2023-November/013489.html
* https://alioth-lists.debian.net/pipermail/nut-upsuser/2023-November/013491.html
* https://alioth-lists.debian.net/pipermail/nut-upsuser/2023-November/013492.html

jimklimov:
Cheers, how is it going here? Do you plan to post some updates shortly, or is (from your viewpoint) the change OK to merge as is and follow up on specific problems and improvements later?

EchterAgo:
> Cheers, how is it going here? Do you plan to post some updates shortly, or is (from your viewpoint) the change OK to merge as is and follow up on specific problems and improvements later?

Sorry, it has been difficult for me to find time to work on this lately. I just was able to test the load commands and it seems that the outlet specific commands for outlet group 1, which is the only one I have, all work for me. What does not work are the outlet commands that specify the MOG (main outlet group). I'm not sure if that group is actually supposed to be switchable. I'll check the documentation some more.

jimklimov:
Cheers, I'm going to merge this PR as it stands today, posted an issue for follow-ups according to comments above.

I believe this change set would already benefit some users, and would expose the code for more practical experience and feedback for further steps. Thank you very much for bringing it this far, and I hope you would have time to get even more done :)

cardil:
Just for visibility's sake. There's an endianness problem with multi-register BF writes for some APC firmware families like SMT1500 or SMT750 - definitely a firmware bug. See details at https://github.com/networkupstools/nut/issues/2338#issuecomment-4142082170

This MOST likely renders the critical `shutdown.return` command non-functional for affected UPSs. During FSD testing with only the serial Modbus driver active, the UPS did not shut down. I still need to re-run FSD to confirm the exact failure path.

PR #2188, opened 2023-11-15
battery.runtime, status OL/OB and more for Arduino subdriver

kellybyrd:
Add support for the common status flags as well as time remaining, battery voltage and other variables found in other HID subdrivers. The Arduino library always supported these items, just the NUT subdriver didn't pick them up so things like status were always OB.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -80,7 +80,7 @@ https://github.com/networkupstools/nut/milestone/10
    support most devices [#2156]
 
  - usbhid-ups driver:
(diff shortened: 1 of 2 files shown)


kellybyrd:
Thanks, Jim for the NEWS.adoc update!

kellybyrd:
i didn't ask before, will this make 2.8.2? If the goal is a fast follow bug fix only from 2.8.1, this doesn't fit. But, it would be nice to have a more functional Arduino subdriver get packages up by Debian, Ubuntu, and others.

jimklimov:
Yes, it should

Issue #2187, opened 2023-11-15
PyNUTClient installs to src


mtelka:
`PyNUTClient` installs itself into the `src` directory.  This is wrong.  I assume nobody will `import src` to get `PyNUTClient` Python module.
```
$ find build/prototype/i386/ -type f
build/prototype/i386/usr/lib/python3.9/vendor-packages/src/__init__.py
build/prototype/i386/usr/lib/python3.9/vendor-packages/src/test_nutclient.py
build/prototype/i386/usr/lib/python3.9/vendor-packages/src/PyNUT.py
build/prototype/i386/usr/lib/python3.9/vendor-packages/src/__pycache__/__init__.cpython-39.pyc
build/prototype/i386/usr/lib/python3.9/vendor-packages/src/__pycache__/PyNUT.cpython-39.pyc
build/prototype/i386/usr/lib/python3.9/vendor-packages/src/__pycache__/test_nutclient.cpython-39.pyc
build/prototype/i386/usr/lib/python3.9/vendor-packages/PyNUTClient-2.8.1-py3.9.egg-info/PKG-INFO
build/prototype/i386/usr/lib/python3.9/vendor-packages/PyNUTClient-2.8.1-py3.9.egg-info/dependency_links.txt
build/prototype/i386/usr/lib/python3.9/vendor-packages/PyNUTClient-2.8.1-py3.9.egg-info/top_level.txt
build/prototype/i386/usr/lib/python3.9/vendor-packages/PyNUTClient-2.8.1-py3.9.egg-info/SOURCES.txt
$
```

jimklimov:
Thanks. All manuals on the subject had the src dir :\

mtelka:
But none installs to `src` :-). Maybe you miss something like this: https://github.com/zopefoundation/zope.location/blob/master/setup.py#L83

Issue #2186, opened 2023-11-15
PyNUTClient is missing tests


mtelka:
Please add some tests to PyNUTClient sdist package.  Thank you.

jimklimov:
There is the self-test program, should be attached, used in NIT (NUT Integration Test suite). I have a hard time imagining tests for this client that would go without a running server anyway, but PRs are welcome. Python is not my native tongue :)

mtelka:
> There is the self-test program, should be attached, used in NIT (NUT Integration Test suite).

Where?

> I have a hard time imagining tests for this client that would go without a running server anyway, but PRs are welcome.

Special requirements to run tests are okay.

> Python is not my native tongue :)

My neither. :-)

jimklimov:
The `test_nutclient.py` is in `src` (other bug) of https://test-files.pythonhosted.org/packages/c3/6d/f058cf24da6e3f45d8043219b4a17d3e31ba2d637a2c75bb5d04b04b4ceb/PyNUTClient-2.8.1.139.1.tar.gz per https://test.pypi.org/project/PyNUTClient/2.8.1.139.1/#files

The PyPI repo does not really allow to fix old releases, so whatever is uploaded is there forever. Files may be deleted, but names can not be reused...

mtelka:
Could you please suggest some complete (dummy) configuration to see tests passed?  I started with this in `ups.conf`:
```
[dummy] 
        driver = dummy-ups
        desc = "Test device"
        port = /src/nut/data/evolution500.seq
```
but I see this failure:
```
[DEBUG] Class initialization...
[DEBUG]  -> Host  = 127.0.0.1 (port 3493)
[DEBUG]  -> Login = 'None' / 'None'
[DEBUG] Connecting to host
--------------------------------------------------------------------------------
Testing 'GetUPSList' :
[DEBUG] GetUPSList from server
{b'dummy': b'Test device'}

--------------------------------------------------------------------------------
Testing 'GetUPSVars' for 'dummy' (should be registered in upsd.conf) :
[DEBUG] GetUPSVars called...
Traceback (most recent call last):
  File "$(BUILD_DIR)/src/test_nutclient.py", line 34, in <module>
    result = nut.GetUPSVars( "dummy" )
  File "$(BUILD_DIR)/src/PyNUT.py", line 199, in GetUPSVars
    raise PyNUTError( result.replace( b"\n", b"" ).decode('ascii') )
[...]

PR #2191, opened 2023-11-18
Fix contents and recipes for Python packaging

jimklimov:
Closes: #2186
Closes: #2187

* Deliver package payload into `PyNUTClient` not `src` dir
* Deliver test script and `tox.ini` to run it (still needs a NUT setup running to test python client against)
* Re-use the Makefile logic also for the github action implementation (DRY coding)


jimklimov:
Merged to see it run in github workflow...

PR #2198, opened 2023-11-24
drivers/libusb1.c: Remove unnecessary validation for bus_num

delphij:
The libusb_get_bus_number() function serves as a simple accessor for the bus number of the opaque device struct. Unlike libusb_get_port_number(), it does not have the ability to convey errors back to the caller, and a bus number of 0 is a valid value. Therefore, any validation around bus_num is redundant.

This commit removes the unnecessary validation code related to bus_num.

Signed-Off-By:	Xin LI <delphij@FreeBSD.org>


jimklimov:
Thanks! I've now re-checked that this issue does not apply to our `libusb0.c` and `nut-scanner/scan_usb.c` sources.

PR #2238, opened 2023-12-20
Added support for (APC) MGE Galaxy 5500

mateusz-jastrzebski:
I am working with (APC) MGE Galaxy 5500 UPS and the default way of grabbing ups.status is different from other APC drivers sadly (talking about drivers/apc-mib.c). OID that works: [1.3.6.1.4.1.318.1.1.1.11.1.1](http://oidref.com/1.3.6.1.4.1.318.1.1.1.11.1.1)

Easiest way I found out to make my ups.status work (as far as OL OB go and also not block other APC/snmp-ups devices's functionality) was to make a custom snmp-ups driver with special treatment; If I have used custom functions for it, I'd still need to modify macro (and compile whole project) that allows custom lookup functions to work but I might be wrong
The new driver is a modified clone of the apc-mib.c
It is in _usable_ state; should get the job done for turning off devices.

Feel free to leave comments.


mateusz-jastrzebski:
Gosh I am being clueless with PRs right now, give me some time

PR #2239, opened 2023-12-20
Added support for APC Galaxy 5500 UPS

mateusz-jastrzebski:
I am working with _(APC) MGE Galaxy 5500 UPS_ and the default way of grabbing ups.status is different from other APC drivers sadly (talking about drivers/apc-mib.c). OID that works: [1.3.6.1.4.1.318.1.1.1.11.1.1](http://oidref.com/1.3.6.1.4.1.318.1.1.1.11.1.1)

Easiest way I found out to make my ups.status work (as far as OL OB go and also not block other APC/snmp-ups devices's functionality) was to make a custom snmp-ups driver with special treatment; If I have used custom functions for it, I'd still need to modify macro (and compile whole project) that allows custom lookup functions to work but I might be wrong
The new driver is a modified clone of the apc-mib.c
It is in _usable_ state; should get the job done for turning off devices.

Feel free to leave some comments - I am still learning coding.


jimklimov:
Hello, thanks for this! Just in case - we have many apc-something subdrivers; did you check that none of them fits your device?

If not, and the problem is about just a few mapping-table data points, it may work to just add them into an existing subdriver (not sure OTOH if the first or last initially probed working match would be used in fact). They are flexible enough to support vendor variant MIBs - so you can see a few duplicate names with different OIDs already.

jimklimov:
From CI:

````
In file included from snmp-ups.c:45:
./apcgalaxy5500-mib.h:10:33: error: no newline at end of file [-Werror,-Wnewline-eof]
#endif /* APCGALAXY5500_MIB_H */
                                ^
snmp-ups.c:3512:9: error: mixing declarations and code is incompatible with standards before C99 [-Werror,-Wdeclaration-after-statement]
                        char pwr_status = buf[1];
                             ^
2 errors generated.
make[1]: *** [Makefile:2389: snmp_ups-snmp-ups.o] Error 1
In file included from apcgalaxy5500-mib.c:30:
./apcgalaxy5500-mib.h:10:33: error: no newline at end of file [-Werror,-Wnewline-eof]
#endif /* APCGALAXY5500_MIB_H */
                                ^
apcgalaxy5500-mib.c:303:3: error: no newline at end of file [-Werror,-Wnewline-eof]
*/
  ^
2 errors generated.
````

File: drivers/snmp-ups.c [hunk 1]
```diff
@@ -3504,19 +3506,26 @@ bool_t su_ups_get(snmp_info_t *su_info_p)
...
-			"ups.status, with%s daisy template originally",
-			__func__, (format_char!=NULL ? "" : "out"));
-		status = nut_snmp_get_int(su_info_p->OID, &value);
-		if (status == TRUE)
+		if(!(strcasecmp(su_info_p->OID, ".1.3.6.1.4.1.318.1.1.1.11.1.1.0")))
 		{
-			su_status_set(su_info_p, value);
-			upsdebugx(2, "=> value: %ld", value);
+			status = nut_snmp_get_str(su_info_p->OID, buf, sizeof(buf), su_info_p->oid2info);
+			char pwr_status;
```
jimklimov, line 3512: This looks error-prone: `strtol()` deals with a (nul-terminated) string, but your copy of the first character from `buf` is not necessarily followed in RAM by a zero byte. Any random garbage can pop up here...

Trying to make sense of that MIB/OID with http://oidref.com/1.3.6.1.4.1.318.1.1.1.11.1.1.0 I see that it is a child of http://oidref.com/1.3.6.1.4.1.318.1.1.1.11.1.1 but otherwise got no clue of its own value. That parent however is a string representing 64 bit-flags (or `UNKNOWN` which you don't check for), and the second byte (your `buf[1]`?) would be the "On Battery" flag. There is a separate flag for "On Line" as well as bypass, self-test, and a united "Low Battery / On Battery" flag, so it may be prudent to check several values for this reading.

Also wondering if all this complex logic should be here or can be better served by "lookup functions" in the `*-mib.c` file (the later, generally `NULL`, fields of lookup tables like `APCC_G5500_pwr_info[]`)?.. See e.g. `drivers/eaton-pdu-marlin-mib.c` for some practical examples.

drivers/snmp-ups.c
```diff
@@ -3504,19 +3506,26 @@ bool_t su_ups_get(snmp_info_t *su_info_p)
...
-
-		free_info(tmp_info_p);
+		{
+			status = nut_snmp_get_int(su_info_p->OID, &value);
+			su_status_set(su_info_p, value);
+			if (status == TRUE)
+					upsdebugx(2, "=> value: %ld", value);
+			else
+					upsdebugx(2, "=> Failed");
+		}
```
jimklimov, line 3528: Why did this block change so drastically - a lot more than indenting generally, losing the debug log and now calling `su_status_set()` regardless of a successful `status` to fetch the value, for example? Did a local experiment seep out here?

drivers/apcgalaxy5500-mib.c
```diff
@@ -0,0 +1,299 @@
...
+    { 1, "noTransfer", NULL, NULL },
+    { 2, "highLineVoltage", NULL, NULL },
+    { 3, "brownout", NULL, NULL },
+    { 4, "blackout", NULL, NULL },
+    { 5, "smallMomentarySag", NULL, NULL },
+    { 6, "deepMomentarySag", NULL, NULL },
+    { 7, "smallMomentarySpike", NULL, NULL },
+    { 8, "largeMomentarySpike", NULL, NULL },
+    { 9, "selfTest", NULL, NULL },
+    { 10, "rateOfVoltageChange", NULL, NULL }
```
jimklimov, line 130: Here and above, lookup tables must end with a `{ 0, NULL, NULL, NULL }` entry to stop iterations.

drivers/apcgalaxy5500-mib.c
```diff
@@ -0,0 +1,299 @@
...
+	{ "ups.power", 0, 1, ".1.3.6.1.4.1.318.1.1.1.4.2.9.0", "", SU_FLAG_OK, NULL },
+	{ "ups.realpower", 0, 1, ".1.3.6.1.4.1.318.1.1.1.4.2.8.0", "", SU_FLAG_OK, NULL },
+	{ "ups.status", ST_FLAG_STRING, SU_INFOSIZE, APCC_G5500_OID_POWER_STATUS, "OFF",
+		SU_FLAG_OK | SU_STATUS_PWR, APCC_G5500_pwr_info },
+	{ "ups.status", ST_FLAG_STRING, SU_INFOSIZE, APCC_G5500_OID_BATT_STATUS, "",
+		SU_FLAG_OK | SU_STATUS_BATT, APCC_G5500_batt_info },
+	{ "ups.status", ST_FLAG_STRING, SU_INFOSIZE, APCC_G5500_OID_CAL_RESULTS, "",
+		SU_FLAG_OK | SU_STATUS_CAL, APCC_G5500_cal_info },
+	{ "ups.status", ST_FLAG_STRING, SU_INFOSIZE, APCC_G5500_OID_NEEDREPLBATT, "",
+		SU_FLAG_OK | SU_STATUS_RB, APCC_G5500_battrepl_info },
```
jimklimov, line 201: Aren't these all different statuses? Check with `docs/nut-names.txt` for suitable strings for the first column.

Also, I think the first hit (query that gets a response from a particular device for the reading name) wins, so the later lines would be effectively ignored at run time.

jimklimov (changes requested):

drivers/snmp-ups.c
mateusz-jastrzebski: I'll try to learn about `strtol()` and its functioning, thank you.

As far as I know, those lookup functions only work when (to put it simply) 
the macro `WITH_SNMP_LKP_FUN' is set to 1

drivers/snmp-ups.h  137:142
```
# if (defined WITH_DMFMIB) && (WITH_DMFMIB != 0)
#  define WITH_SNMP_LKP_FUN 0
# else
#  define WITH_SNMP_LKP_FUN 1
# endif
#endif
```

Does it mean that in best scenario I will still have to use a self-built version of NUT for this particular UPS model?
Or I might be missing something 

Yes, I acknowledge the fact that `ups.status` is more than `OL` or `OB`, things would get easier with lookup functions, I understand

drivers/snmp-ups.c
mateusz-jastrzebski: Alright I agree I did mess this up

drivers/apcgalaxy5500-mib.c
mateusz-jastrzebski: Didn't think of that, this code snippet is copied from `drivers/apc-mib.c`

drivers/snmp-ups.c
mateusz-jastrzebski: And umm I have put child OID (not parent) there because that's what actually works when using something like `snmpget`
The parent OID does not work

`snmpget -v1 -c public 'redacted_ip' 1.3.6.1.4.1.318.1.1.1.11.1.1`
```
Error in packet
Reason: (noSuchName) There is no such variable name in this MIB.
Failed object: iso.3.6.1.4.1.318.1.1.1.11.1.1
```

aquette:
Have you checked the subpart (after .318) do not match mge-mib? These Galaxy may just the rebadged MGE.

jimklimov:
Odd, notifications here show a commit with nuked `README.adoc`... why?

mateusz-jastrzebski:
Umm, I just wanted to have a clean readme for my fork, I acknowledge the fact that this PR's not going to be merged anytime soon. I have also just found out the importance of merging a non default branch from forked repo..
I think I'm going to just close this PR "for now" as I physically don't have time and patience for fixing my code. 
If there will be next time, I'll make sure I do it with best manners, sorry!

jimklimov:
Ah, I see. Just was worried someone stole your account to spam, or something.

Issue #2232, opened 2023-12-12
Split dummy-ups between nutdrv_nutclient and nutdrv_simulation


aquette:
Part of #94, with some refresh

dummy-ups code should be split, to better separate its 2 current purposes and serve in SmartUPS:
* nutdrv_nutclient could serve as a general NUT client, which connects to a remote upsd. This can replace upsmon, to repeat devices data from a remote upsd.
* nutdrv_simulation: provide static or dynamic device simulation from a plain text file.
Useful for development, testing and troubleshooting purpose

jimklimov:
While looking at it, keep in mind the earlier discussion about the two `clone*` drivers which seem to share a lot with `dummy-ups` and among themselves. Maybe here's an opportunity to refactor a lot into one?..

aquette:
@jimklimov iirc, the clone* are local (dstate, same figth as dstate_nut2mqtt...)
still, I'll take a look. At least for future consolidation...

Good for the names otherwise? (nutdrv_nutclient & nutdrv_simulation)

jimklimov:
I'm in favor of names that fit in narrow-ish terminals and tools, so under ~16 chars of unambiguous name is a benefit. The prefix makes sense but eats 7 chars, so gotta be careful with the rest :)

PR #2245, opened 2023-12-22
2236 nut scanner reports non existent driver for nut discovery

aquette:
When discovering NUT, either with Avahi or old method, the reported driver is "nutclient", which doesn't (yet) exists.
This should be fixed to dummy-ups, which will use the repeater mode, while waiting for #2232 

the short term fix is to symlink nutclient to dummy-ups (example on Debian):

> cd /lib/nut && ln -s dummy-ups nutclient

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -88,6 +88,8 @@ https://github.com/networkupstools/nut/milestone/10
    * Added generation of FreeBSD/pfSense quirks for USB devices supported
      by NUT (may get installed to `$datadir` e.g. `/usr/local/share/nut`
      and need to be pasted into your `/boot/loader.conf.local`). [#2159]
+   * nut-scanner now report 'dummy-ups' as driver when scanning NUT with
+     Old or Avahi method.
 
  - upsd: Fixed conditions for "no listening interface available" diagnosis
    to check how many listeners we succeeded with, not whether the first one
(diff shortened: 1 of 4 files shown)


aquette:
seriously @jimklimov? 
https://ci.networkupstools.org/job/nut/job/nut/job/PR-2245/1/pipeline-console/?selected-node=252

please put such targets as optional, and revise the CI to be dev friendly. Breaking the momentum for such errors is no good ;)

jimklimov:
Can't see details too well on mobile on the road, but errors look like broken automake syntax? Did you change the makefiles to match?

aquette:
boils down to 1 single poor OpenIndiana failure among bagazillions working, and that destroys the whole castle :/
> configure.ac:2739: error: possibly undefined macro: AC_DISABLE_STATIC

jimklimov:
Well, we sort of (officially) pride ourselves on supporting builds of NUT everywhere it could be built before, hence the very diverse CI-testing matrix with C99+ and 2000'ish or newer (du'jour and beyonder) toolkit implenentations. While not part of regular builds, there were in recent past questions about Solaris 8 and CentOS 6, and a few fixes to make those happen. Developer momentum can quite quickly happen on their few systems, but non-regression for everyone has to happen on all the 180-ish os/toolkit combos.

I'll try to check that code if there will be signal later in the night, or after holidays at worst.

aquette:
for sure a pride, I've been the one porting it for more than 15 years.
Still, that should block a final release, not a dev flow!

jimklimov:
As noted elsewhere, the issue was in the end due to NUT CI farm VM migration to a new provider, and it ended up pointing to a wrong container (not set up for builds) as the OI agent.

PR #2247, opened 2023-12-23
nut-scanner: fix device-name collisions in generated conf

aquette:
When nut-scanner is called separately, to scan different buses, it will generate similar names for each of these:
* USB scan
> [nutdev1]
> ...
* NUT scan
> [nutdev1]
> ...
* ... As a result, one can't just redirect nut-scanner outputs directly to ups.conf, without a 'sed' hack workaround.

Solution: simply refine the device names, like "nutdev-<method><id>".
Ex: nutdev-usb1, nutdev-usb2, nutdev-nut1, nutdev-snmp1, nutdev-simulation1, ...

### DIFF
--- a/docs/man/nut-scanner.txt
+++ b/docs/man/nut-scanner.txt
@@ -19,6 +19,9 @@ DESCRIPTION
 *nut-scanner* scans available communication buses and displays any
 NUT-compatible devices it has found.
 
+*nut-scanner* can also display the detected devices in various formats,
+including ups.cconf, and ensures generated devices name unicity across buses.
+
 INSTALLATION
 ------------
 
@@ -90,7 +93,6 @@ This option must be requested explicitly, even for a complete scan.
 ports using 'X-Y', where X and Y are characters referring to the port number.
 - a single port name.
 - a list of ports name, coma separated, like '/dev/ttyS1,/dev/ttyS4'.
-
 
 NETWORK OPTIONS
 ---------------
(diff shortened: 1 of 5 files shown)


aquette:
@jimklimov CI issues, unrelated to the PR (trivial) changes
cc @desertwitch for review (interested in an onboarding?)

desertwitch:
Currently have no means to test (stuck on a train) but the principal idea is great and should make chaining different nut-scanner commands, while outputting into one configuration file, much easier and reduce the need to hack around the generated UPS names.

aquette:
Note that #2246, as for all my dummy-ups work, is there for when you don't have access to real devices (or for CI). I coded that when on a plane ...
Just touch some .de files in /etc/nut and try nut-scanner -n

jimklimov:
For coding on a plane or a boat, did I already suggest Termux? An Android app that is a debian-ish Linux VM, can build NUT for native ARM for example :)

aquette:
Got it on all phones and tabs 😉
But home assistant on my rpi gives me remotes access to my editable containers and has a term and file edit through web UI too, quite nice. Give it a try ...

jimklimov:
Finally back to the online world, re-read this PR and was like "how did I never hit this, aren't numbers sequential for all found devices on whatever bus?"... and then saw the part about "running separately for different buses". Makes sense and LGTM then :)

jimklimov:
@aquette : changes seem to work, NIT complains about test result mismatch (will post a fix in a sec) ;)

Now that we have a `./ci_build.sh` here, feel free to use it locally to weed out at least simple hiccups without incurring the wrath of the whole CI farm :)

Issue #2249, opened 2023-12-24
NUT master-branch builds broken after recent PRs and/or infra changes


jimklimov:
The two most-recent builds at https://ci.networkupstools.org/job/nut/job/nut/job/master/750/ and https://ci.networkupstools.org/job/nut/job/nut/job/master/751/ (after PRs #2245 and #2246) are broken in different ways.

* [ ] The first one (also discussed in PR comments) seems to have a problem with `AC_DISABLE_STATIC` in OpenIndiana builds. There are no relevant code changes, so possibly a regression on the NUT CI platform which during this month underwent (and will continue) hosting migration and upgrades. Possibly got some crippled autoconf version/half-installed package on one of the workers there?..

* The latter has numerous problems with other platforms as well, including older CentOS and cross-mingw builds, AppVeyor (mingw on windows), CircleCI (MacOS). Some are almost-benign complaints about coding style/static analysis in the newer compilers (e.g. requiring an explicit `funcname(void)` for lack of args), others are more serious with use of absent struct fields. @aquette : can you please take a look at those, if the code can be made portable or maybe `#ifdef` the platform-specific bits to they are not implemented in binary on incompatible platforms? Notably, this is what the CI-augmented PR mechanism is for, to not let these bugs creep into master ;) Alas, this time the real problem probably got obscured by the "expected-red" former issue.

[...]

PR #2250, opened 2023-12-24
Fix parts of issue 2249

jimklimov:
See #2249 

* Could not reproduce the autoconf problem on any of the workers, will look out for it more later.
* Fixed method declarations.
* Hacked around lack of `dirent->d_type` on some platforms (any non-GNU land I suppose). A better fix is desirable for fallback equivalent logic via `fstat()`, see e.g.
  * https://mingw-users.narkive.com/JpEGojiV/linux-dirent-d-type-windows-equivalent
  * https://stackoverflow.com/questions/35215109/struct-dirent-does-not-have-de-type-in-header-file/35215258#35215258
* Did not address the methods missing on mingw at this time.

### DIFF
--- a/tools/nut-scanner/nut-scan.h
+++ b/tools/nut-scanner/nut-scan.h
@@ -155,7 +155,7 @@ nutscan_device_t * nutscan_scan_xml_http_range(const char *start_ip, const char 
 
 nutscan_device_t * nutscan_scan_nut(const char * startIP, const char * stopIP, const char * port, useconds_t usec_timeout);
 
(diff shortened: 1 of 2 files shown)


aquette:
spasibo for investigation and fixing @jimklimov 
fstat PR would be appreciated, since I'lll be a little busy with the base devs :D

aquette:
Faulty code fixed by removing these non portable things. This PR can be closed, though the code may serve one day...

jimklimov:
Ref:  #2251

Issue #2259, opened 2024-01-09
Check that you've installed libusb-dev or equivalent for your distro. Without it, nutdrv_qx is serial-only.


Ariessanea1:
Check that you've installed libusb-dev or equivalent for your distro. Without it, nutdrv_qx is serial-only.

_Originally posted by @jimklimov in https://github.com/networkupstools/nut/issues/2219#issuecomment-1883591220_

jimklimov:
What is this new issue about? :)

On NUT side, we check it and report on the wall of text and later summary of `configure` script, with USB being just one of a gazillion dependencies and optional build variants. Probably yours said that auto-detection for building against libusb failed, it was not required (by some explicit `--with-usb`), so skipped.

Your build probably lacks a few drivers with `usb` in the name (`blazer_usb`, `usbhid-ups`, etc.)

FWIW though, the `nutdrv_qx` being special with multiple media support in one binary, its identification banner could reflect the build circumstances to ease trouble-shooting.

PR #2260, opened 2024-01-11
Help users about `nutdrv_qx` and similar drivers built for serial-only support but used for USB devices

jimklimov:
Closes: #2259

Currently emitted messages in this case require some sleuthing. This PR hopefully makes troubleshooting easier:
````
./nutdrv_qx -s test -x productid=1234
Network UPS Tools - Generic Q* Serial driver 0.36 (2.8.1-285-ge2bfdbbd4)

Fatal error: 'productid' is not a valid variable name for this driver.

Look in the man page or call this driver with -h for a list of
valid variable names and flags.

NOTE: for driver 'nutdrv_qx', options like 'productid' are only available
if it was built with USB support. If you are running a custom build of NUT,
please check results of the `configure` checks, and consider an explicit
`--with-usb` option. Also make sure that both libusb library and headers
are installed in your build environment.

upsnotify: notify about state 4 with libsystemd: was requested, but not running as a service unit now, will not spam more about it
upsnotify: failed to notify about state 4: no notification tech defined, will not spam more about it
````

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -104,6 +104,12 @@ https://github.com/networkupstools/nut/milestone/10
      up in the link:https://pypi.org/project/PyNUTClient[PyPI repository].
      [#2158]
 
+ - main driver core codebase:
+   * Help users of drivers that can be built to support optionally USB and
+     other media (like `nutdrv_qx` built for serial-only support), and built
+     in fact without USB support but used for USB devices, with some more
+     information to make troubleshooting easier. [#2259]
+
  - powerpanel text driver now handles status responses in any format and should
    support most devices [#2156]
 
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -109,6 +109,9 @@ https://github.com/networkupstools/nut/milestone/10
      other media (like `nutdrv_qx` built for serial-only support), and built
      in fact without USB support but used for USB devices, with some more
      information to make troubleshooting easier. [#2259]
+   * Driver programs with debug tracing support via `-D` CLI option and/or
+     the `NUT_DEBUG_LEVEL` environment variable now check those earlier in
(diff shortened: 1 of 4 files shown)


jimklimov:
For drivers built with USB support (and using the common `nut_usb_addvars()` method, which covers all but a couple of them at the moment), detailed info about the libusb involved is shown during init with debug (the "Using USB implementation..." line below):
````
$ ./nutdrv_qx -D -s test -x productid=1234 -x port=auto
Network UPS Tools - Generic Q* USB/Serial driver 0.36 (2.8.1-285-ge2bfdbbd4)
USB communication driver (libusb 1.0) 0.46
   0.000000     [D1] upsdrv_makevartable...
   0.000029     [D1] Using USB implementation: libusb-1.0.25 (API: 0x1000109)
   0.000057     [D1] Network UPS Tools version 2.8.1-285-ge2bfdbbd4 (release/snapshot of 2.8.1.1) built with gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0 and configured with flags: --enable-Wcolor --enable-warnings --enable-Werror --enable-keep_nut_report_feature --with-all=auto --with-cgi=auto --with-serial=auto --with-dev=auto --with-doc=skip --with-nut_monitor=auto --with-pynut=auto --disable-force-nut-version-header --enable-check-NIT --enable-maintainer-mode --disable-silent-rules
   0.000070     [D1] debug level is '1'
   0.000133     [D1] Can not become_user(nobody): not root initially, remaining UID=1000 GID=1000
   0.000151     Can't chdir to /var/state/ups: No such file or directory
   0.000164     upsnotify: notify about state 4 with libsystemd: was requested, but not running as a service unit now, will not spam more about it
   0.000179     upsnotify: failed to notify about state 4: no notification tech defined, will not spam more about it
````

PR #2264, opened 2024-01-15
drivers/apc-mib.c: update MIB values

moenoel:
Some of the mapped SNMP values for the APC MIB are incomplete/out of date, most importantly for the OID ".1.3.6.1.4.1.318.1.1.1.4.1.1.0" (upsBasicOutputStatus). This causes NUT to report that it lost connection to a UPS that is e.g. in the "ECOnversion" operation mode, as the corresponding unmapped value results in an empty `ups.status`.

I updated this and a few other things based on [PowerNet MIB v4.5.2](https://www.apc.com/us/en/download/document/APC_POWERNETMIB_452_EN/) and the [SNMP documentation for APC Galaxy VS/PX/VL UPSes](https://www.apc.com/us/en/download/document/SPD_CCON-GALAXYMIBREF_EN/).

Some things that probably need to be looked at before this can be merged:

 * I am unsure whether there is a better thing to return than "LB" for the `batteryInFaultCondition(4)` and `noBatteryPresent(5)` states. I put "LB" for now as a fail safe way to handle them.
 * I was unable to find documentation for the states `powerSavingMode(18)`, `activeLoad(23)`, `chargerOnly(26)` and `distributedEnergyReserve(27)`, so I left them empty for now. Seems like they're not inuse for the UPS I'm interested in here, at least.
 * I added a keyword `ECO` to the return values for `ecoMode(13)` and `eConversion(20)`. Not sure if that might break anything or be of value to anyone for monitoring purposes and whatnot.

### DIFF
--- a/drivers/apc-mib.c
+++ b/drivers/apc-mib.c
@@ -26,7 +26,7 @@
 
 #include "apc-mib.h"
 
-#define APCC_MIB_VERSION	"1.6"
+#define APCC_MIB_VERSION	"1.7"
 
 #define APC_UPS_DEVICE_MODEL	".1.3.6.1.4.1.318.1.1.1.1.1.1.0"
 /* FIXME: Find a better oid_auto_check vs sysOID for this one? */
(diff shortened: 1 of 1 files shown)


jimklimov:
Just in case, seems related to #2239 which tackled at least some more OIDs (but in a less efficient fashion, making a separate subdriver).

Issue #2271, opened 2024-01-19
Support for Liebert Corp PowerSure PST UPS


digitlman:
Need help getting driver support for this unit:

````
Bus 001 Device 096: ID 10af:0002 Liebert Corp. PowerSure PST UPS
Couldn't open device, some information will be missing
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0
  bDeviceSubClass         0
  bDeviceProtocol         0
  bMaxPacketSize0         8
  idVendor           0x10af Liebert Corp.
  idProduct          0x0002 PowerSure PST UPS
  bcdDevice            0.02
  iManufacturer           3 Vertiv Co.
  iProduct                1 Liebert PST5
  iSerial                 5 2309310073A4A3C
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x003b
    bNumInterfaces          2
    bConfigurationValue     1
    iConfiguration          0
    bmAttributes         0xa0
      (Bus Powered)
      Remote Wakeup
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0
      bInterfaceProtocol      0
      iInterface              0
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
          bCountryCode           33 US
          bNumDescriptors         1
          bDescriptorType        34 Report
          wDescriptorLength      27
         Report Descriptors:
           ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval               5
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        1
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      0
      bInterfaceProtocol      0
      iInterface              0
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.10
[...]

Issue #1840, opened 2023-01-25
nut usbhid driver unable to find HID descriptors on interfaces other than interface 0


x0ul:
Hi there, I have a UPS that seems to speak HID, but its configuration has two different interfaces under its (single) configuration. Interface 0 is small and has some vendor-specific junk, while interface 1 dumps a bunch of HID data and most of it fits into usages 0x84 and 0x85, resembling what is expected from a UPS. However, the usbhid driver in nut seems to be unable to communicate with interface 1, so it only sees the small junky data from interface 0 and doesn't think there's a UPS there.

Some data from usbhid-dump showing the two interfaces on my UPS.

Interface 0. Small, bad.
```
002:004:000:DESCRIPTOR
 06 00 FF 09 01 A1 01 09 02 15 00 26 FF 00 75 08
 95 08 81 82 09 02 95 08 91 82 C0
```

Interface 1. Big, good.
```
002:004:001:DESCRIPTOR
 05 84 09 04 A1 01 09 24 A1 00 85 28 09 FE 79 01
 75 08 95 01 15 00 26 FF 00 B1 03 85 29 09 FF 79
 05 B1 03 85 2B 09 FD 79 03 B1 03 09 1A A1 02 85
 30 09 40 67 21 D1 F0 00 55 07 B1 83 C0 85 11 09
 5A 75 08 95 01 15 01 25 03 65 00 55 00 B1 82 09
 02 A1 02 85 32 75 01 95 07 15 00 25 01 81 03 B1
 03 09 62 95 01 B1 A2 09 62 81 A2 B1 01 81 01 09
... (truncated by me)
```

I pasted both those into `hidrd-convert` and see the nice-looking 0x84, 0x85 usage pages ([more here if you care to see](https://pastebin.com/E2e8tRsc)).

Some specifics now that I've introduced the problem and my interpretation. The UPS is a Liebert PST5 made by Virtiv and was bought on Amazon in the last month. The device's vid is 10af; pid is 0002. This sort of matches the Belkin HID driver, so I started from there trying to get it to work after adding a specific pid to match in ups.conf.

I hacked nut into talking to interface 1 and getting some decent data (perhaps incomplete, I stopped once I got comms working) with the following:

```diff
diff --git a/drivers/belkin-hid.c b/drivers/belkin-hid.c
index 686dc9261..46cfddf0c 100644
--- a/drivers/belkin-hid.c
+++ b/drivers/belkin-hid.c
[...]

PR #2272, opened 2024-01-19
Update USB product IDs for Belkin/Liebert

jimklimov:
One change is a bug-fix for old contribution; another is at the moment speculative for a newly reported device in #2271 (earlier in #1840).

For new devices with non-zero default interface numbers, best effect can be expected with #2331 merged too (confirmed by testing in tickets above).


Bomberski:
Would this add support for the Vertiv Liebert PSI5-1100MT120?

jimklimov:
Not sure, if directly "yes". If that device has the IDs rectified here, and if it talks the same protocol otherwise - then maybe. You can give a shot to https://github.com/networkupstools/nut/wiki/Building-NUT-for-in%E2%80%90place-upgrades-or-non%E2%80%90disruptive-tests and check directly.

There were some recent issue discussions that more and more devices talk USB HID not on "interface 0" but a different one, and a few (sub-)drivers have ways to set the relevant variables when detecting devices, but there is no holistic solution at the moment. Should not be too complex, just needs a bit of time - PRs welcome ;)

PR #2275, opened 2024-01-23
FTY remerge from master as of 2024-03-02

jimklimov:
Bump the copy of 42ity fork contents to minimize difference with the master (as they got to differ considerably over the past 2 years of main trunk's evolution). Hopefully this would ease eventual import of features from the fork and facilitate comparisons of intentional differences.

Also letting NUT CI chew it in many more scenarios than I can run locally :)

After all, it is very possible that difftool/mergetool duplicated or ate away some lines (fixed some cases like that which did break local builds soon after initial conflict resolution).


aquette (approved):

jimklimov:
Yes, but a lot of treasures lurk there waiting to be upstreamed (and feed back to benefit FTY from the other side). I fear the day when I'd have to review the 50k lines of changes though, and extract that into specific features (DMF, nutconf, vanilla packages, etc.) and more so if trying to keep the 3K items of commit trail :D

jimklimov:
@aquette : FWIW, I've hopefully brought `nutconf` here up to speed, at least in regard to building without warnings and passing the self-test (locally so far, letting NUT CI loose on it now). It is probably behind the modern code in terms of keyword support, but that's iterable if someone were so inclined :)

Also, finally had a sort of bright idea how to avoid those ugly `ifdef WITH_...FUNC` in all the SNMP mapping and lookup tables, to add explicit NULL initializers for optional fields: made a series of macros to apply these defaults. So the `*-mib.c` files are now much more readable and somewhat smaller, similar to master-branch code. The next step is to apply similar macros on master to minimize the diff with DMF-capable codebase :)

jimklimov:
Hm, while assigning labels and looking for historic evidence, found the https://github.com/networkupstools/nut/tree/libconf branch among the dozens of loose threads right in this repo. Didn't go that deep into the cellar before :D 

At least, this gives me a chance to cleanly PR the feature separately of DMF etc. :)

jimklimov:
Hm, C++ obsoletions strike again in clang :\

jimklimov:
PR updated with master that separately included the IPSS Unix derived installer #2288, and the updated `nutconf` effort #2290 among others.

jimklimov:
Seems to converge into the greens, woohoo! This was a very long run, literally since the first iterations of the NUT CI farm (still with Travis CI) got established and some measure of warnings-less code were introduced by PR #823 and later efforts. The DMF codebase was so much of a change, that without knowing that the standard code is good, it was neigh impossible to say if the added code has issues of its own or not (I'd say, not so many as some friends feared).

On the other side, the recently rekindled interest to finally green-light FTY/DMF branches on CI was quite beneficial for the master branch codebase too. Some larger independent features initially used in 42ity fork got transplanted during this release cycle (nutconf, Eaton IPP installer) - if only to reduce the diff scope of the two codebases by thousands of lines. A couple of large blocks to port into future NUT releases would be DMF itself, and reference packaging recipes (and then maybe set up integration with OBS cloud for reference package builds of master and PRs to become available).

A number of issues were uncovered while trying to make FTY/DMF builds work which apply to the main codebase, its docs and recipes (thanks to building twice as many very different build scenarios and looking intently at each result), so maybe a dozen small PRs landed into master during these weeks too - largely improving portability to *BSD and Solaris/illumos platforms, revising some dependencies needed even on Linux distros, truly out-of-tree builds, etc.

It seems that currently one warning is poised to remain unsolved in the short term, about the linker on OpenBSD agents finding two differently named builds of libssl/libcrypto.

jimklimov:
Regarding several builds of `libssl`/`libcrypto` conflicting on FreeBSD builds...

From warnings analysis summary at https://ci.networkupstools.org/job/nut/job/nut/job/PR-2275/lastBuild/aggregated-analysis/ :
````
libssl.so.111, needed by /usr/local/lib/libneon.so, may conflict with libssl.so.11
libcrypto.so.111, needed by /usr/local/lib/libneon.so, may conflict with libcrypto.so.11
libcrypto.so.111, needed by /usr/local/lib/libnetsnmp.so, may conflict with libcrypto.so.11
````

Files are indeed differently named and sized:
````
$ ls -la /usr/lib32/libssl.so.111 /usr/local/lib/libssl.so.11 /usr/local/lib/libssl.so /usr/lib/libssl.so.111 /usr/lib/libssl.so /usr/lib32/libssl.so
lrwxr-xr-x  1 root  wheel      13 Oct 23  2020 /usr/lib/libssl.so -> libssl.so.111
-r--r--r--  1 root  wheel  609520 Oct 23  2020 /usr/lib/libssl.so.111
lrwxr-xr-x  1 root  wheel      13 Oct 23  2020 /usr/lib32/libssl.so -> libssl.so.111
-r--r--r--  1 root  wheel  512604 Oct 23  2020 /usr/lib32/libssl.so.111
lrwxr-xr-x  1 root  wheel      12 Jul  2  2021 /usr/local/lib/libssl.so -> libssl.so.11
-rw-r--r--  1 root  wheel  603904 Jul  2  2021 /usr/local/lib/libssl.so.11

$ file /usr/lib/libssl.so.111 /usr/lib32/libssl.so.111 /usr/local/lib/libssl.so.11
/usr/lib/libssl.so.111:      ELF 64-bit LSB shared object, x86-64, version 1 (FreeBSD), dynamically linked, stripped
/usr/lib32/libssl.so.111:    ELF 32-bit LSB shared object, Intel 80386, version 1 (FreeBSD), dynamically linked, stripped
/usr/local/lib/libssl.so.11: ELF 64-bit LSB shared object, x86-64, version 1 (FreeBSD), dynamically linked, stripped
````

* Why does it not pop out in standard NUT builds (probably DMF recipes hardcode something as a shortcut? or just got a different combo of libs needed together by some binary?)
* Is it possible to portably `configure` detection of library search paths to accommodate the whole chain of used libs or poke exact preferred filenames (e.g. to do use /usr/lib/libssl.so.111 for /usr/local/lib/libneon.so and ignore the nearby /usr/local/lib/libssl.so.11)
* Where did /usr/local/lib/libssl.so.11 come from? Why is libneon/libsnmp not in /usr/lib if it wants standard libraries? Is it some curiosity of the build agent and not a systemic distro problem - maybe the core fix is in some consistent use of packages vs. local builds, not in NUT recipes imagining workarounds for sysadmin errors?

jimklimov:
So... it seems that the `openssl` binary is the only consumer of that library, and the symlink is not needed:
````
# { find /usr/local . -type f -perm +111 | while read F ; do  ldd $F | grep /usr/local/lib/libssl.so && echo "^^^ $F" ; done ; }  2>/dev/null

        libssl.so.11 => /usr/local/lib/libssl.so.11 (0x8002e1000)
^^^ /usr/local/bin/openssl
````

jimklimov:
Tested, libxml2 was not the culprit (makes sense, little use for crypto there).

Found that DMF branch also includes this sort of difference:
````
+
+if WITH_SSL
+  snmp_ups_CFLAGS += $(LIBSSL_CFLAGS)
+  snmp_ups_LDADD += $(LIBSSL_LIBS)
+
+  snmp_ups_dmf_CFLAGS += $(LIBSSL_CFLAGS)
+  snmp_ups_dmf_LDADD += $(LIBSSL_LIBS)
+endif

 # NEON XML/HTTP
 netxml_ups_SOURCES = netxml-ups.c mge-xml.c
 netxml_ups_LDADD = $(LDADD_DRIVERS) $(LIBNEON_LIBS)
+netxml_ups_CFLAGS = $(AM_CFLAGS)
+
+if WITH_SSL
+  netxml_ups_CFLAGS += $(LIBSSL_CFLAGS)
+  netxml_ups_LDADD += $(LIBSSL_LIBS)
+endif
````

and the link commands in build logs reflect that (`snmp-ups` build):
````
STD: ... ../common/libparseconf.la -L/usr/local/lib -lnetsnmp -lm -lpthread
DMF: ... ../common/libparseconf.la -L/usr/local/lib -lnetsnmp -lm  -L/usr/local/lib -lssl -lcrypto -lpthread
````

PR #2279, opened 2024-01-26
clients/upsmon.c: fix offstate not being a requirement for a critical UPS on OFF

### LINKED ISSUE #2278: V2.8.1 Nearly immediate shutdown on powerloss
After the update from 2.8.0 to 2.8.1 the behaviour of nut, connected via snmp to an Eaton 5PX UPS, changed drastically.

On power loss, after about 30-60 seconds, all connected devices, master and slaves shut down immediately.
With 2.8.0, this only happened as soon as the UPS reached it's configured "left" battery time (300 seconds)

In the logfiles, there is an indication of a set FSB Flag:
Jan 25 18:33:03 upsmon 89442 Auto logout and shutdown proceeding
Jan 25 18:33:03 upsmon 89442 Executing automatic power-fail shutdown
Jan 25 18:33:03 upsd 22669 Client local-monitor@127.0.0.1 set FSD on UPS [Eaton_5PX2200]
Jan 25 18:33:03 upsmon 89442 UPS [Eaton_5PX2200] is reported as (administratively) OFF
Jan 25 18:33:03 upsmon 89442 UPS Eaton_5PX2200: administratively OFF or asleep
Jan 25 18:33:03 upsmon 89442 UPS Eaton_5PX2200 on battery

Comparing with a UPS on 2.8.0, I don't get the FSB message, also no indication of beeing "administratively OFF or asleep".
Jan 25 19:42:38 Mastermind upsmon[17548]: UPS Eaton_5PX2200@127.0.0.1 on battery
Jan 25 19:43:38 Mastermind upsmon[17548]: UPS Eaton_5PX2200@127.0.0.1 on line power

However, compared to the logdata of the UPS itself, there is no indication that this flag comes from the UPS.
2024/01/25 18:32:52 Normal AC NOK
2024/01/25 18:32:52 ABM state resting
2024/01/25 18:32:52 UPS on battery
2024/01/25 18:32:53 System shutdown in 2 h 17 mn 16 s
2024/01/25 18:32:53 Outlet group 1 shutdown in 2 h 49 mn 35 s
2024/01/25 18:32:53 Outlet group 2 shutdown in 2 h 49 mn 35 s
2024/01/25 18:32:53 ABM state Off
2024/01/25 18:32:53 Normal AC voltage out of tolerance

As logged there should be at least 2h of runtime. (which it normally has, just had an incident about 3 weeks ago which lasted more than two hours, and all boxes were still going strong). Back then, I did not have version 2.8.1 installed, I was still below. 

For testing, and for ruling out communication issues, I unplugged the UPS from mains about 5 times, every time, the box with 2.8.1 (and it's slaves) went black, every other box with 2.8.0 or 2.7.x are staying alive.
The UPS was still reachable an queryable during the time of my staged power loss tests.

Checking with "upsc", it seems like in a normal power loss situation, this type of UPS reports an "OFF" flag during power loss.
Screenshot_20240125_204857

I stumbled accross issue #2104 which seems kind of related, however, there it only happened during self-tests or calibrations, not during a "normal" power loss situation.

desertwitch:
I was able to replicate the issue #2278, it seems to be as suspected in #2278: Due to `offstate == 1` not being a required condition at present (rather an optional condition) when evaluating UPS criticality, instant FSD is triggered despite `offdurationtime` not being exceeded yet (and `offstate` still being 0) when the UPS is seen both `OB` and `OFF`. This happens due to disregarding `offstate` when `linestate == 0` in combination with the `ST_OFF` flag and `offdurationtime >= 0` is met.

However `offstate` must always be a required condition for `offdurationtime` to have an effect as `ST_OFF` flag is instantly set upon encountering an `OFF` state, regardless of how long the UPS has been in this state.

Testing the proposed change FSD is only happening when `offdurationtime` is reached, the `linestate` itself is no longer considered for OFF-induced UPS criticality as both an `OL` and `OB` device can no longer give power to the load due to being administratively `OFF`, regardless of the mains power situation. Hence it makes sense to disregard the linestate (mains power situation) and only look at - most importantly - `offstate == 1` in combination with `offdurationtime >= 0` and `ST_OFF` flag.

The current code results in the following behaviour when `OB` + `OFF` is encountered:
```
Jan 26 13:14:32 Tower upsmon[8689]: UPS ups@127.0.0.1 on battery
Jan 26 13:14:32 Tower upsmon[8689]: UPS ups@127.0.0.1: administratively OFF or asleep
***instant shutdown due to FSD***
```

The proposed change results in the following behaviour when `OB` + `OFF` is encountered (as expected):
```
Jan 26 13:18:56 Tower upsmon[8689]: UPS ups@127.0.0.1 on battery
Jan 26 13:18:56 Tower upsmon[8689]: UPS ups@127.0.0.1: administratively OFF or asleep
Jan 26 13:19:31 Tower upsmon[8689]: ups_is_off: ups@127.0.0.1 is in state OFF for 35 sec, assuming the line is not fed (if it is calibrating etc., check the upsmon 'OFFDURATION' option)
***delayed shutdown due to FSD***
```

fixes #2278

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -55,6 +55,9 @@ https://github.com/networkupstools/nut/milestone/10
      activity was completed by the hardware, which led to mis-processing of
      shutdown triggers. Also, notification was added to report "finished
      calibration". [issue #2168, PR #2169]
+   * `upsmon` recognition of `OFF` state as a trigger for FSD (forced shut
+     down) criticality considered also the input line state, which may be
+     an independently evolving circumstance. [issue #2278, PR #2279]
    * `upsmon` support for `POLLFAIL_LOG_THROTTLE_MAX` did not neuter the
      applied setting when live-reloading configuration, so commenting it
      away in `upsmon.conf` did not have the effect of resetting the logging
--- a/clients/upsmon.c
+++ b/clients/upsmon.c
@@ -1100,7 +1100,7 @@ static int is_ups_critical(utype_t *ups)
 
 	/* administratively OFF (long enough, see OFFDURATION) */
 	if (flag_isset(ups->status, ST_OFF) && offdurationtime >= 0
-	&& (ups->linestate == 0 || ups->offstate == 1)) {
+	&& ups->offstate == 1) {
 		upslogx(LOG_WARNING,
 			"UPS [%s] is reported as (administratively) OFF",
 			ups->sys);
--- a/clients/upsmon.c
+++ b/clients/upsmon.c
@@ -1104,6 +1104,7 @@ static int is_ups_critical(utype_t *ups)
 		upslogx(LOG_WARNING,
 			"UPS [%s] is reported as (administratively) OFF",
 			ups->sys);
+		upsdebugx(1, "UPS [%s] is now critical being OFF for too long. In case of persisting unwanted shutdowns, consider disabling the upsmon 'OFFDURATION' option.", ups->sys);
 		return 1;
 	}
 
--- a/docs/nut.dict
+++ b/docs/nut.dict
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 3439 utf-8
+personal_ws-1.1 en 3440 utf-8
 AAC
 AAS
 ABI
@@ -1856,6 +1856,7 @@ cpqpower
 cpsups
 cr
 crestfactor
+criticality
 crlf
 cron
 crontab


jimklimov (approved):

jimklimov:
It is also a bit odd how it shows `OB` twice, maybe it should not...

desertwitch:
> It is also a bit odd how it shows `OB` twice, maybe it should not...

That's weird indeed, hmm. But then I'm glad this issue brought to light the previously unknown problem of an instant FSD upon encountering OB OFF and unwanted circumvention of the OFFDURATION safeguard in the process - which my proposed changes should fix for now. 

Remains to be solved why the UPS reports OFF when OB and gets a double status while at that, sure never gets boring here... ;)

jimklimov (approved):

PR #2283, opened 2024-01-27
Import libconf/nutconf features

jimklimov:
This codebase lingered in side branches of NUT since 2013, and used in forked projects, providing a neat C++ parser of NUT configuration files as a library and a front-end CLI tool to read, write and "understand" them. It was further used in an automated installer for NUT packages on some platforms (to be ported later).

This PR brings back the initial coding effort from the `libconf` branch (cherry-picked over NUT v2.6.5 because there was a lot of noise, and modernized with a series of tag merges), as well as further clean-ups done in the `FTY` branch and recently in PR #2275.

Part of the aim of this PR is to integrate the feature separately from many others that evolved in the forks, with minimal intrusion into the generally used NUT codebase up till now. It should also help iterate build nuances for the many platforms now covered by NUT CI, so that these recipes succeed everywhere (cursorily looking, POSIX-like headers available on Windows can be problematic).

At this moment, while the tool would be likely usable for the config syntax of 2013 vintage, it would need further updates for evolution since NUT v2.6.5 for both common options (master/primary, USB matching, etc.) and for various new driver toggles. Probably the augeas lens file evolution would help in the next steps.

Note that this PR does not actually enable *builds* of the codebase by default (support of C++11, optionally CPPUNIT), but only on some select CI agents that claim the ability. A more general support (notably MinGW implementation or pruning) would be the focus of a follow-up PR.


jimklimov:
Superseded by #2290 which gets this codebase

jimklimov:
Effectively merged via #2290

PR #2285, opened 2024-01-27
snmp mib2nut wrapper macros

jimklimov:
Idea raised in PR #2275 allowing to juggle several codebases (main and DMF/FTY until merged) with different C structure property counts and minimal difference in comparable sources. For this, macros `snmp_info_default` and `info_lkp_default` (and a few for function pointers within) are provided.

This PR also clearly ends the tables with an entry literally named `..._sentinel` so it is role is well understood (and spelling of the entry fields is never botched), and it is better visible where someone forgot to add one. So a bonus for maintainability.

NOTE: Not updating sub-driver versions (except for a few files with mapping content changes) since the tables after C pre-processor expanding the macros should be equivalent to codebase before this PR.

PR #2286, opened 2024-01-27
Eaton: add support for EMP002 on ATS16 NM2 with SNMP

jimklimov:
This support is identical to what is in Eaton Marlin and Powerware MIBs. Future improvements should consider putting these bits in a common eaton-sensor-mib file, to avoid multiple definition of the same MIB structure!

Imported from FTY fork / DMF branch (with reductions for main trunk) from commit 40c883f42aede519fe716a6d26e7dff567996d75

Originally by @aquette

NOTE: After merge of other contemporary PRs, this code would welcome wrappings from #2285 (available in #2275 which is where these not-upstreamed lines were "discovered").

### DIFF
--- a/drivers/eaton-ats16-nm2-mib.c
+++ b/drivers/eaton-ats16-nm2-mib.c
@@ -24,8 +24,12 @@
  */
 
 #include "eaton-ats16-nm2-mib.h"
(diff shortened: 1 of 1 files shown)


arnaudquette-eaton (approved):
LGTM, though not checked deep

PR #2288, opened 2024-01-28
Import CLI installer for a NUT bundle

jimklimov:
Contributed years ago by Eaton from their free UPS companion software... and lay shelved in side branches...

Relies on `nutconf` for parts of its work, see e.g. #2283


arnaudquette-eaton (approved):
hope that it can be useful to the Community

arnaudquette-eaton:
also, I'm removing from the SmartNUT project. It's definitely not part of this effort ;)

jimklimov:
Well, I figured it could be used in automation of SmartNUT installations. Or serve as an example of doing that to be inspired from for new code :)

PR #2290, opened 2024-01-28
Enable build of libconf/nutconf features by default

jimklimov:
Follow-up to #2283 which aims to enable builds (and tests) of `(lib)nutconf` codebase by default (where C++11 is available, optionally CPPUNIT to test the results).

This iteration softens some rough edges for WIN32 codebase, and for (way too many) other complications it just disables parts of the originally POSIX-oriented implementation that would need to be remade for Windows (got quite a few precedents for named pipe use instead of Unix sockets and signals in other NUT programs).

Closes: #2295 along the way

By side effect (from making CI pass), this PR also fixes CI builds with standards other than C11/C++11 (requests were there... typos in scripting also...), and NUT Integration Tests (NIT) detection of available ports for `upsd` to listen on.


jimklimov:
Too many compilers out there with different opinions about correct C++11 :\

Also, listening on a temporary port does misfire even when it is random but picked once for the test without regard to what is busy on the system.

jimklimov:
Hm, the remaining complaints seem to mostly be about port clashes (where we take a random one of 10000...50000 on localhost - surprisingly often we can't bind), and a few about unexpected signals received in IPC tests. PRs to make these tests reliable would be most welcome.

I guess the best way forward for now would be to use some envvars to disable the tests which depend too much on the environment in CI builds.

PR #2297, opened 2024-02-01
Fix: tripplite_usb: if upsid is not specified, make any unit ID match

### LINKED ISSUE #2282: tripplite_usb: if `upsid` is not specified, any unit ID should match
It appears that between 2.8.0 and 746cb86110982920b9db8f5066b766d560d19eef (including 2.8.1), `tripplite_usb` stopped matching the default unit ID of 65535/0xffff when `upsid` is not specified in the configuration file.

Although I don't have hardware available to confirm, I suspect it is commit 106e7c6fe7, so I am tagging @sapireli.

IMHO, if `upsid` is not passed, then the matching function should still be `NULL` here: https://github.com/networkupstools/nut/commit/106e7c6fe7e0c67a7964e254ae8983f4537d8103#diff-51229908f9631058c142955b4726606c68b280194b6aeb184fdc9a17766f924eR1652

I would also recommend adding some logging during the matching process - the current debug message does not indicate that the ID doesn't match: https://github.com/networkupstools/nut/issues/2258#issuecomment-1912097611

sapireli:
small fix restoring matching when upsid is undefined in the ups.conf file closes #2282 as per the thread
cc @clepple

### DIFF
--- a/drivers/tripplite_usb.c
+++ b/drivers/tripplite_usb.c
@@ -330,7 +330,7 @@ int match_by_unitid(usb_dev_handle *argudev, USBDevice_t *arghd, usb_ctrl_charbu
 int match_by_unitid(usb_dev_handle *argudev, USBDevice_t *arghd, usb_ctrl_charbuf rdbuf, usb_ctrl_charbufsize rdlen)
 {
 	char *value = getval("upsid");
-	int config_unit_id = 0;
+	long config_unit_id = 0;
 	ssize_t ret;
 	unsigned char u_msg[] = "U";
 	unsigned char u_value[9];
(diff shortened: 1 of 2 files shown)


clepple (approved):
LGTM, thanks!

(note, I don't have hardware to test this, and I haven't looked at any build logs either)

drivers/tripplite_usb.c
```diff
@@ -340,8 +340,11 @@ int match_by_unitid(usb_dev_handle *argudev, USBDevice_t *arghd, usb_ctrl_charbu
 	NUT_UNUSED_VARIABLE(rdbuf);
 	NUT_UNUSED_VARIABLE(rdlen);
 
-	/* Read ups id from the ups.conf */
-    if (value != NULL) {
+    /* If upsid is not defined in the config, return 1 (null behavior - match any device),
+	 * otherwise read it from the device and match against what was asked in ups.conf */
+    if (value == NULL) {
+        return 1;
+    } else {
```
jimklimov, line 347: I've updated the comment in code to better describe why we `return 1` - hopefully I got it right? ;)

The original wording did not seem unambiguous to me.

jimklimov (approved):

PR #2312, opened 2024-02-15
`drivers/cps-hid.c` et al: add support for Cyber Energy branded devices (by Cyber Power Systems)

jimklimov:
Following up from https://alioth-lists.debian.net/pipermail/nut-upsdev/2024-February/007966.html - this PR adds support for VID:PID delegated to Cyber Energy:

> In the past, CyberPower UPS works well with NUT. Today, I want to use another VID/PID to fulfill ODM business rather than CyberPower brand UPS. My engineer will apply standard USB power device description like as CyberPower’s USB design. The only difference is VID/PID.

Note that the oficially registered vendor ID 0x0483 is for ST Microelectronics who then delegate certain product IDs to different OEMs.


jimklimov:
Confirmed working via mailing list.

PR #2332, opened 2024-03-01
huawei-ups2000.txt: add UPS2000-G-3KRTL to the confirmed list

biergaizi:
I received a user report that the model `UPS2000-G-3KRTL` was shown to be functional using the NUT driver `huawei-ups2000`. Add that to the documentation.

This is notable for being the first reported working "long backup time" variant of the UPS2000 with an external battery pack. Their hardware and firmware are likely identical to the "standard backup time" model, but it's good to see a confirmation of that.


jimklimov (approved):

jimklimov:
Considered updating `data/driver.list.in` - but it already lists the "A and G series" in general.

Issue #2335, opened 2024-03-01
`nut-scanner` fails due to simulation mode if `CONFPATH` does not exist


jimklimov:
Fall-out from #2246 / #2242 :
````
   0.175520     [D1] nutscan_init: succeeded to load the library for NUT Client library
   0.175529     Scanning USB bus.
   0.175577     No start IP, skipping SNMP
   0.175580     Scanning XML/HTTP bus.
   0.175583     [D1] XML/HTTP SCAN: starting pthread_create with run_xml...
   0.175605     [D1] run_usb: using library default link_detail_level settings
   0.175623     No start IP, skipping NUT bus (old connect method)
   0.175627     Scanning NUT simulation devices.
   0.175629     [D1] NUT simulation devices SCAN: starting pthread_create with run_nut_simulation...
   0.175674     Scanning NUT bus (avahi method).
   0.175678     [D1] NUT bus (avahi) SCAN: starting pthread_create with run_avahi...
   0.175677     [D1] Scanning XML/HTTP bus using broadcast.
   0.175714     [D2] nutscan_scan_xml_http_generic() : scanning connected network segment(s) with a broadcast, attempt 1 of 3 with a timeout of 5000000 usec
   0.175724     Scanning IPMI bus.
   0.175728     [D1] IPMI SCAN: starting pthread_create with run_ipmi...
   0.175723     [D1] Scanning: /usr/local/ups/etc
   0.175749     Failed to open /usr/local/ups/etc: No such file or directory
   0.175873     [D5] nutscan_scan_xml_http_generic() : sent request to <broadcast>, loop #1/3, waiting for responses
nut$
````

Probably related to `fatal_with_errno()` below:
````
    upsdebugx(1, "Scanning: %s", CONFPATH);

    if ((dp = opendir(CONFPATH)) == NULL) {
        fatal_with_errno(EXIT_FAILURE, "Failed to open %s", CONFPATH);
        return NULL;
    }
````

jimklimov:
https://github.com/networkupstools/nut/pull/2336/commits/cd4e190c0fbce003f223f01093867b6c2026b5c7

Issue #2334, opened 2024-03-01
nut-scanner default/-C vs -U usb behavior divergencies


desertwitch:
**nut-scanner**
```
root@DEVTEST:~# nut-scanner
Cannot load XML library (libneon.so) : file not found. XML search disabled.
Cannot load IPMI library (libfreeipmi.so) : file not found. IPMI search disabled.
Scanning USB bus.
No start IP, skipping SNMP
No start IP, skipping NUT bus (old connect method)
Scanning NUT simulation devices.
Scanning NUT bus (avahi method).
[nutdev-usb1]
        driver = "usbhid-ups"
        port = "auto"
        vendorid = "051D"
        productid = "0002"
        product = "Back-UPS 500 FW: 6.2.I USB FW: c1"
        serial = "XXXXXXXX"
        vendor = "American Power Conversion"
        bus = "008"
        # device = "002"
        busport = "001"
[nutdev-simulation2]
        driver = "dummy-ups"
        port = "E5.dev"
```
**nut-scanner -U**
```
root@DEVTEST:~# nut-scanner -U
Cannot load XML library (libneon.so) : file not found. XML search disabled.
Cannot load IPMI library (libfreeipmi.so) : file not found. IPMI search disabled.
Scanning USB bus.
[nutdev-usb1]
        driver = "usbhid-ups"
        port = "auto"
        vendorid = "051D"
        productid = "0002"
        product = "Back-UPS 500 FW: 6.2.I USB FW: c1"
        serial = "XXXXXXXX"
        vendor = "American Power Conversion"
        # bus = "008"
        # device = "002"
        # busport = "001"
```

Default (complete scan) USB behavior should probably follow one `-U` behavior and also comment out `busport` and `bus`.
Instead it currently comments out only `device` but leaves the other two, which I think is not intended.

Follow-up from #2221

PR #2336, opened 2024-03-01
Fix `nut-scanner` caveats about USB optional fields and simulation file discovery

jimklimov:
Closes: #2334
Closes: #2335

CC @aquette @desertwitch


desertwitch:
Confirmed working now - thanks Jim!

Issue #1726, opened 2022-12-01
CI: Migrate from LGTM.COM to GitHub Actions CodeQL


jimklimov:
LGTM.com is deprecated and soon to be offlined: https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/

Apparent old LGTM config per https://github.com/networkupstools/nut/blob/master/.lgtm.yml :
````
path_classifiers:
  template:
  - exclude: "**/*.py.in"
  - exclude: "**/NUT-Monitor.in"
queries:
- exclude: "cpp/fixme-comment"
- exclude: "python/fixme-comment"
extraction:
  python:
    after_prepare:
    - "./autogen.sh"
    - "./configure --without-docs"
    python_setup:
      version: "2"
      setup_py: false
    index:
      filters:
      - include: "**/*.py.in"
      - include: "**/NUT-Monitor.in"
      - include: "**/*.py"
      - include: "**/NUT-Monitor"
````

A first shot at https://github.com/networkupstools/nut/blob/master/.github/workflows/codeql.yml failed to find Python sources with the new action.
* https://github.com/networkupstools/nut/commit/ce7a01d7d8fbe8a7883975bf977637d662db1e53

jimklimov:
Recent notification e-mails confirm this is done well enough:
````
* new alerts:

    2 for Short global name

Heads-up: LGTM.com's PR analysis will be disabled on the 5th of December, and LGTM.com will be shut down ⏻ completely on the 16th of December 2022.
It looks like [GitHub code scanning with CodeQL is already set up for this repo](http://networkupstools/nut/blob/HEAD/.github/workflows/codeql.yml), so no further action is needed 🚀. For more information, please check out [our post on the GitHub blog](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/).
````

PR #2355, opened 2024-03-06
.github/workflows/codeql.yml: make a CodeQL matrix…

jimklimov:
… of compiler/NUT_SSL_VARIANTS/NUT_USB_VARIANTS instead of single fightwarn-all

Follow-up to #1726 and #2250

...if arbitrary matrix labels would work

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -75,7 +75,7 @@ jobs:
     - if: matrix.language == 'cpp'
       name: NUT CI Build (default-all-errors matrix)
       run: |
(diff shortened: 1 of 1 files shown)


jimklimov:
TODO: Filter out the ton of python analyses, only need one

jimklimov:
TODO: Fix `ci_build.sh` to build only one scenario with the mix of SSL and USB setting right away (currently does two, one specifying only SSL and another both).

https://github.com/networkupstools/nut/actions/runs/8179804508/job/22366575078

Issue #2024, opened 2023-08-25
Multiplier Issues with liebert-esp2


Funtime60:
My GXT2-2000RT120 is reporting a battery voltage of 560.0VDC. I know for a fact that it's actually 56VDC.

Either the GXT2-2000RT120 needs different multipliers than the GXT2-6000RT208 or it's not being assigned the correct multiplier at all.

jimklimov:
Which NUT version, please? Possibly a data dump to help identify the device and firmware in our code?..

Funtime60:
I'm pretty sure it's NUT 2.7.4, I'm not sure how to do a data dump or what firmware you're asking for. As for where in the code? I'm pretty sure it's in here https://github.com/networkupstools/nut/blob/master/drivers/liebert-esp2.c
Maybe around line #59, #260, or #261

Funtime60:
Thank you for the suggestion, I always forget that the raspbian repos are pretty out dates. I give it a try.

PR #2369, opened 2024-03-23
feat: Add support for Liebert PSI5

jrjparks:
Adding support for the Liebert PSI5 model of UPS, as discussed in #2271 initially.

Maybe addresses similar complaints posted in #2024 

Closes: #2271

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -180,6 +180,9 @@ https://github.com/networkupstools/nut/milestone/10
      useful as an UPS driver, not just a controller developer sandbox. [#2188]
    * `cps-hid` subdriver now supports devices branded as Cyber Energy and built
      by cooperation with Cyber Power Systems. [#2312]
+   * `belkin-hid` subdriver now supports Liebert PSI5 devices which have a
+     different numeric reading scale than earlier handled models. [issue #2271,
+     PR #2272, PR #2369]
    * The `onlinedischarge` configuration flag name was too ambiguous and got
      deprecated (will be supported but no longer promoted by documentation),
      introducing `onlinedischarge_onbattery` as the meaningful alias. [#2213]
--- a/drivers/belkin-hid.c
+++ b/drivers/belkin-hid.c
@@ -31,7 +31,7 @@
 
 #include <math.h>     /* for fabs() */
 
-#define BELKIN_HID_VERSION      "Belkin/Liebert HID 0.19"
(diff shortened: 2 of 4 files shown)


jimklimov (approved):
Looks quite solid to me. Is there a particular reason this is posted as a "draft" PR? Are you in local testing so far?

Compared to #2271 discussion I see you went for a separate method to "maybe apply" a correction factor (in the PoC there you added logic to an existing method). On one hand, this way you surely should not introduce errors into existing support for other devices, on another - I still wonder if some other reports about Liebert/Belkin HID devices could benefit from an exponent fix for those earlier HID tree node readings. And whether that original method did have an error about `1e7` vs. `1e9` comparisons and multiplications. At least, something to keep in mind if more bug reports come eventually...

jimklimov:
Passing to NUT CI farm to test this change...

jrjparks:
> Looks quite solid to me. Is there a particular reason this is posted as a "draft" PR? Are you in local testing so far?

I left it in draft before I went to bed to make sure it would pass the approvals before I published it.

> Compared to #2271 discussion I see you went for a separate method to "maybe apply" a correction factor (in the PoC there you added logic to an existing method). On one hand, this way you surely should not introduce errors into existing support for other devices, on another - I still wonder if some other reports about Liebert/Belkin HID devices could benefit from an exponent fix for those earlier HID tree node readings. And whether that original method did have an error about `1e7` vs. `1e9` comparisons and multiplications. At least, something to keep in mind if more bug reports come eventually...

Yeah, I thought about how to extend the existing function without possibly breaking others, but I wouldn't be able to be sure of any changes other than "it works on my UPS" :laughing:. This seemed to be the safer path as it's also using different HID lookup paths, so I assumed Liebert is using different maths for different models. I was also trying to think of a way to just set the line voltage multiplier based on the model or some factor, but I'm still not very familiar with the code and it's been a while since I last touched some C.

jimklimov:
"In your build" - was it a custom-compiled build of NUT? Make sure you use a sufficiently new NUT source (2.8.2 or newer, perhaps current master), as well as `libusb-1.0-dev` or equivalent package name for your platform to compile against (not just lib but also its headers and other developer resources).

This can help: https://github.com/networkupstools/nut/wiki/Building-NUT-for-in%E2%80%90place-upgrades-or-non%E2%80%90disruptive-tests

Issue #2370, opened 2024-03-24
Maths puzzle: revise Liebert/Belkin USB HID parsing for line voltage orders of magnitude and don't loose track of zeroes


jimklimov:
Following up from #2271 / #2369 discussions and changes (CC @jrjparks @clepple @aquette @digitlman @electroflame)

The differing maths approaches in the older and newer methods still nag me in the back of my mind, and I also wonder if these can be collapsed into `d_equal()` or related macros we have defined in NUT for comparing values that are almost equal, within an `epsilon` error apart...

Spilling it out here to think out loud :)

Looking at https://github.com/networkupstools/nut/blob/31e9979/drivers/belkin-hid.c#L168-L222 I see the two methods in question:

````
OLD> liebert_line_voltage_fun(double value) {
	if( value < 1 ) {
		if( fabs(value - 1e-7) < 1e-9 ) {
			liebert_line_voltage_mult = 1e7;
...
````

````
NEW> liebert_psi5_line_voltage_fun(double value) {
	if( value < 1 ) {
		if( fabs(value - 1e-3) < 1e-3 ) {
			liebert_line_voltage_mult = 1e5;
...
````

I believe the goal here is to find such a `liebert_line_voltage_mult` that we would multiply the miniscule `value` reading (coming from hardware) by, and get something in the hundreds range for the line voltage (typically 120/127/220/230/240V - so somewhat over an exact hundred).

I can also accept that the different HW/FW models (PSI5 vs. others) can serve their data differently and not "just" shifted by some orders of magnitude compared to other models, although posting some examples from low-level readings of different models could be useful to make sense of different use-cases (adding debug prints into the driver builds?)

Namely, what concerns me here is how:
* in the old case we compare `if abs(value - (10^-N)) < (10^-N)/100) then mult=(10^N)` (where `N=7`)
* and for the new code `if abs(value - (10^-N)) < (10^-N)) then mult=100*(10^N)` (where `N=3`)
* ... note the hundreds moving around. Is that correct?

Also, what do these `abs(diff)` achieve?

Assuming larger scaling numbers just for easier visible maths, with `N=0` (`10^0 == 1`) and reasonable values of 120V or 220V (seen *after* multiplication of `value` by `mult` when the test succeeds), we would see either:
* OLD LOGIC: want 120V/220V *after* multiplication by `mult = 10^N = 1` when the test succeeds, so `double value=120.0` and `value=220.0`
[...]

Issue #14, opened 2013-02-27
nut-scanner Gen2


aquette:
2nd generation of lbnutscanner and nut-scanner:
- plugin design:
  - each "bus" is provided through a specific library (aka "plugin"),
  - each plugin provides everything: help, init, scan, cleanup/close,
  - plugins using weak dependencies through dl_\* functions (no hard link) 
  - ip address iterator is also perfectible
- libnutscanner provides .so listing and loading (to wrap available and usable plugins)
- nut-scanner is almost empty shell

This will solve the #ifdef issue, and make nut scan library and tool more maintainable and usable!

Issue #13, opened 2013-02-26
NUT Controller (nutctld)


aquette:
A daemon to control NUT features locally or remotely:
- scan,
- configuration,
- daemons start / stop / restart / reload.

Use cases:
- mass configuration (including initial configuration) of a network of NUT systems,
- automated initial configuration of system with USB devices. 

@balooloo: **to be completed**

jimklimov:
Note: scope is partially related to the nut-driver-enumerator introduced later, and changes to reference systemd unit layout changed between NUT 2.7.4 and 2.7.5 releases, at least regarding the goals for more sane driver/daemon start/stop/restart/reload (managing Linux systemd or Solaris/illumos SMF service instances for each individual driver and its service/unit dependency chain).

Issue #12, opened 2013-02-25
USB integration on FreeBSD


clepple:
(Theoretically, this applies to NetBSD and OpenBSD as well, but FreeBSD's USB stack seems more complete at the moment.)

Some code: https://github.com/clepple/nut/tree/devd-usb

Things to do:
- [x] Generate devd.conf files for known USB devices
- [x] Install devd.conf files in the right place (includes detection of *BSD)
- [ ] Document the installation process (restarting devd, other permissions issues like scanning the bus, etc.)
- [ ] Coordinate with ports maintainer

PR #2371, opened 2024-03-25
Unit-test and fix the Belkin/Liebert HID UPS exponents math

jimklimov:
As raised on the mailing list at https://alioth-lists.debian.net/pipermail/nut-upsuser/2024-March/013626.html as well as in discussions for #2271, #2369 and #2370, there is some mess in exponents calculation of `drivers/belkin-hid.c`, where the driver tries to adjust for what seems wrongly encoded voltage readings in USB HID reports, which may be off by some orders of magnitude (3, 7...) depending on reading type and HW/FW model.

As detailed in https://github.com/networkupstools/nut/issues/2370#issuecomment-2017664722 : The original logic which purported to fix this was broken, but by coincidence seems to have taken the right code path for the majority of devices up till NUT v2.7.4. This was fixed in NUT v2.8.0 to correct logic, but the order of magnitude estimation results no longer fit the devices' outputs, so the scaling was not applied and this caused various voltages to be reported as zeroes - causing a practical regression.

As suggested in recent discussions on the matter, this PR adds a (rather ugly, but functional) way to test the helper conversion methods "in-vivo" - in the practically unmodified sub-driver code, mocking the methods and variables it refers to elsewhere. Maybe this can be done even without the currently posted change to no longer mark `static` the variables and methods we need for a test, either by `#include` of the driver source into a test program, or by shipping test methods as part of a driver/program source and/or custom-built binary (ZeroMQ style) - something to refine later. UPDATE: Went for including at the moment, was not hard after the other mock preparations.

At the moment the test is functional in terms of calling one of the 3 methods currently provided by the driver source (1 added in #2369) and finding that the original method fixed in 2.8.0 (to check those clauses as floating-point numbers and not zeroed out rounded integers) indeed failed to work for the typical values it expected to process.
````
Test #1         liebert_psi5_line_voltage_mult()        GOT value      27.3     mult 100000 PASS
Test #2         liebert_psi5_line_voltage_mult()        GOT value     121.2     mult 100000 PASS
LineVoltage exponent looks wrong, but not correcting.
Test #3         liebert_line_voltage_mult()             GOT value         0     mult      1 FAIL        EXPECTED v=   13.9      m=  1e+07       ORIGINAL (string)'1.39e-06'     => (double)1.39e-06
LineVoltage exponent looks wrong, but not correcting.
Test #4         liebert_line_voltage_mult()             GOT value         0     mult      1 FAIL        EXPECTED v=  220.1      m=  1e+07       ORIGINAL (string)'2.201e-05'    => (double)2.201e-05
Test #5         liebert_psi5_line_voltage_mult()        GOT value        12     mult      1 PASS
Test #6         liebert_psi5_line_voltage_mult()        GOT value      12.3     mult      1 PASS
Test #7         liebert_psi5_line_voltage_mult()        GOT value     232.1     mult      1 PASS
Test #8         liebert_psi5_line_voltage_mult()        GOT value       240     mult      1 PASS
Test #9         liebert_line_voltage_mult()             GOT value        12     mult      1 PASS
Test #10        liebert_line_voltage_mult()             GOT value      12.3     mult      1 PASS
Test #11        liebert_line_voltage_mult()             GOT value     232.1     mult      1 PASS
Test #12        liebert_line_voltage_mult()             GOT value       240     mult      1 PASS
Test #13        liebert_config_voltage_mult()           GOT value        24     mult      1 PASS
Test #14        liebert_config_voltage_mult()           GOT value       120     mult      1 PASS
````

From this point, some TDD can be done to adjust the implementations... and I suppose the `liebert_psi5_line_voltage_mult()` checks can be merged with `liebert_line_voltage_mult()` as was originally intended in PoC code posted in issue #2271 - there's little left to break in the original one, it seems.

(CC @jrjparks @clepple @aquette @digitlman @electroflame)

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -182,7 +182,9 @@ https://github.com/networkupstools/nut/milestone/10
      by cooperation with Cyber Power Systems. [#2312]
    * `belkin-hid` subdriver now supports Liebert PSI5 devices which have a
      different numeric reading scale than earlier handled models. [issue #2271,
-     PR #2272, PR #2369]
+     PR #2272, PR #2369] Generally the wrong-scale processing was addressed,
+     including a regression in NUT v2.8.0 which led to zero values
+     in voltage data points which NUT v2.7.4 reported well [#2371]
    * The `onlinedischarge` configuration flag name was too ambiguous and got
      deprecated (will be supported but no longer promoted by documentation),
      introducing `onlinedischarge_onbattery` as the meaningful alias. [#2213]
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -87,6 +87,10 @@ https://github.com/networkupstools/nut/milestone/10
      `configure` script option `--with-debuginfo`. Note that default autoconf
      behavior usually embeds moderate optimizations and debug information on
      its own. [PR #2310]
+   * A fix applied among clean-ups between NUT v2.7.4 and v2.8.0 releases
+     backfired for `usbhid-ups` subdriver `belkin-hid` which in practice
+     relied on the broken older behavior; more details in its entry below.
+     [PR #2371]
 
  - nut-usbinfo.pl, nut-scanner and libnutscan:
    * Library API version for `libnutscan` was bumped from 2.2.0 to 2.5.0
--- a/drivers/belkin-hid.c
+++ b/drivers/belkin-hid.c
@@ -33,7 +33,7 @@
 
 #include <math.h>     /* for fabs() */
 
-#define BELKIN_HID_VERSION      "Belkin/Liebert HID 0.21"
+#define BELKIN_HID_VERSION      "Belkin/Liebert HID 0.20"
 
 /* Belkin */
 #define BELKIN_VENDORID	0x050d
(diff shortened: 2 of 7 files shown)


jimklimov:
I think I've picked my way through that `if` clause. It further happened to work for @jrjparks thanks to 120V utility network - and the maths does break if I put `0.002456` (for 240V) into the new PSI5 method.

The original logic seems to have been like "check if we have a value 0..199: subtract 100, and if resulting module is within 100 - pick corresponding multiplier".

Actually the old original (with `fabs(value - 1e-7) < 1e-9`) went further with the likes of "check if we have a value with expected magnitude: subtract 100, and if resulting module is within 1 - pick corresponding multiplier" which IMHO hits for values "99..101", at best? so practically fails for everything.

Fixing this into "subtract 100, and if resulting module is within 300 - pick corresponding multiplier" should match practical voltages from 0 to 380 (three-pole) with the same order of magnitudes in question. Any objections? :)

jimklimov:
Fix in driver proposed, testing on HW would be very welcome :)

jimklimov:
Cheers, any chance for practical testing, to merge the PR this weekend? <3  :)

jrjparks:
I can try it out tomorrow

jimklimov:
Gentle bump, itching to release this fix :)

Issue #2399, opened 2024-04-14
logging "nut_libusb_get_string: Success" every 12 seconds


fenugrec:
journalctl log is being filled by two log messages every 12 seconds :

```
Apr 14 10:30:06 machin nut-driver@belk[476]: nut_libusb_get_string: Success
Apr 14 10:30:06 machin usbhid-ups[476]: nut_libusb_get_string: Success
Apr 14 10:30:18 machin nut-driver@belk[476]: nut_libusb_get_string: Success
Apr 14 10:30:18 machin usbhid-ups[476]: nut_libusb_get_string: Success
....
```

Looks like maybe here https://github.com/networkupstools/nut/blob/7f36d4909d2b194b2126b7e68df08046adfffe62/drivers/libusb1.c#L1022C8-L1022C42 , libusb is probably returning 0 ?  `libusb_get_string_descriptor_ascii` docs say it returns

> number of bytes returned in data, or LIBUSB_ERROR code on failure 

and LIBUSB_SUCCESS = 0, so this would explain the message.

My questions : 
- why is nut_libusb_get_string() called every 12 seconds ?
- why does the underlying call to libusb_get_string_descriptor_ascii() return 0 (I haven't verified this, just speculating)
- is it necessary to print a log message when it does return 0 ?

[EDIT] hardware info :

```
battery.charge: 100
battery.charge.low: 30
battery.charge.warning: 30
battery.runtime: 120
battery.type: PbAc
battery.voltage: 138.0
battery.voltage.nominal: 12
device.mfr: Belkin
device.model: UPS
device.serial: 
device.type: ups
driver.debug: 0
driver.flag.allow_killpower: 0
driver.name: usbhid-ups
driver.parameter.pollfreq: 10
driver.parameter.pollinterval: 2
driver.parameter.port: auto
driver.parameter.synchronous: auto
driver.state: quiet
driver.version: 2.8.1
driver.version.data: Belkin/Liebert HID 0.18
driver.version.internal: 0.52
driver.version.usb: libusb-1.0.27 (API: 0x100010a)
input.voltage: 123.5
input.voltage.nominal: 120
ups.beeper.status: enabled
ups.delay.shutdown: 20
ups.delay.start: 30
ups.firmware: 0
ups.load: 49
ups.mfr: Belkin
ups.model: UPS
ups.power.nominal: 550
ups.productid: 0551
ups.serial: 
ups.status: OL CHRG
ups.test.result: Done and passed
[...]

PR #2400, opened 2024-04-15
drivers/libusb{0,1}.c, NEWS.adoc: avoid spurious syslog of `(nut_)libusb_get_string: Success`

jimklimov:
Closes: #2399

The message may be due to `0` meaning either `LIBUSB_SUCCESS` or a zero byte count in returned string (man page says a positive value means a successful return). Core problem is likely a libusb or device issue, but this PR can help discern that.

NOTE: Not sure if an actual empty-string return is one byte long (the `\0` byte), so a zero-value reply is always a problem?

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -59,6 +59,10 @@ https://github.com/networkupstools/nut/milestone/11
      (similar to `runtimecal` in some other drivers, may be refactored
      to that configuration and logic model in later NUT releases)
 
+ - USB drivers could log `(nut_)libusb_get_string: Success` due to either
+   reading an empty string or getting a success code `0` from libusb.
+   This difference should now be better logged, and not into syslog. [#2399]
+
  - upsmon: it was realized that the `POWERDOWNFLAG` must be explicitly set in
    the configuration file, there is no built-in default in the binary program
    (the settings facilitated by the `configure` script during build "only"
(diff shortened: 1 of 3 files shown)


tormodvolden:
I am curious about this. First, if libusb_get_string_descriptor_ascii() returns 0 it means it successfully returned an empty string. See the API on https://libusb.sourceforge.io/api-1.0/group__libusb__desc.html#ga240aac96d92cb9f51e3aea79a4adbf42

The man page at e.g. https://manpages.debian.org/unstable/libusb2-dev/libusb.3.en.html says "positive" but that includes zero.

So for libusb1, if 0 is returned, buf must be an empty string (zero-terminated) and strlen(buf) will always be 0. So this looks unnecessary for libusb1. However, if nut_libusb_strerror() prints anything for LIBUSB_SUCCESS (0) it shouldn't be called unconditionally, so only calling it on ret < 0 would make sense.

The documentation for libusb0 usb_get_string_simple() (e.g. https://sources.debian.org/src/libusb/2%3A0.1.12-35/doc/functions.sgml/#L346) says
"Returns number of bytes returned in buf or < 0 on error."
so it should be the same here. AFAIK from the source (e.g. https://sources.debian.org/src/libusb/2%3A0.1.12-35/usb.c/#L230), if a zero length string is returned from the device (tbuf[0] = 2), an empty, zero-terminated string will be returned.

tormodvolden:
The question is if anyone will see this warning with anything else than "actual buf length is 0". I doubt it.

jimklimov:
Pedantically, "non-negative" includes zero; "positive" does not ;)

tormodvolden:
True :) But the authoritative documentation is the Doxygen API documentation. I don't even know where that man page originates from.

Issue #321, opened 2016-09-16
Wishlist: add a configurable timeout between seeing "OB LB" and actual shutdown


dark-penguin:
In many cases, it would be very convenient to have the system not react immediately to an "OB LB" status, but wait a little to make sure it doesn't go away in a couple of seconds. For example:
- When performing a manual periodical discharge of the battery in order to prevent sulfatation and correct the estimated on-battery time (if we had a few seconds to turn the power back on after it goes into the alarm stare, then we wouldn't have to go through all the trouble of disabling the whole NUT first);
- When the estimated on-battery time is less than "alert" time, the whole system shuts down during an automatic periodical UPS self-test. There is a workaround for that, but what if you did not expect the battery to be that bad, or if your UPS miscalculated something? Not to mention specifying a "panic timeout" is simply easier (and more reliable) than doing the procedure of overriding the UPS values described in the manuals.

See this thread in the mailing lists for more info and extremely complicated workarounds that could be solved easily:
http://lists.alioth.debian.org/pipermail/nut-upsuser/2016-September/010260.html

clepple:
For the first case, the `ignorelb` option can be added to `ups.conf`: http://networkupstools.org/docs/man/ups.conf.html#_ups_fields (only the driver would need to be restarted - if this is complicated, please request that your distribution simplify that procedure).

I'm not sure I follow the second part. If you can't trust the UPS to get through a self-test without signaling low battery, how can you be sure that there is enough power left to shut down properly in a real power failure? I don't think many UPSes reliably report that a test is in progress, so I doubt we could add generic logic to "ignore LB during a test".

In the current driver architecture, the only state that is carried over between polling cycles is the connection information (to reconnect to USB devices) and whether or not the previous poll worked (for data-stale notification). Adding a timeout, while it sounds simple, would require adding an extra history layer to drivers to keep track of when the LB flag was last seen, and to handle all of the possible transitions. As someone who expects the UPS to provide a working LB signal, I would prefer that any such changes happen outside of the driver and upsmon.

There has been talk of integrating a Lua interpreter into drivers - maybe there is room for another upsmon/upssched hybrid which uses a scripting language to capture the intricacies of situations like these.

dark-penguin:
OK, I understand that it would be hard to implement due to the current driver architecture. More thoughts about this later; first, let me explain other things so that everything else is clear.

[...]

Issue #2104, opened 2023-10-12
APC Back-UPS RS bi-monthly battery self-tests result in shutdown


desertwitch:
As the maintainer of the NUT plugin for UNRAID (which basically acts as a frontend for NUT) I have two people with the APC series **Back-UPS RS** (models 900MI and 1350MS) reporting that the bi-monthly battery self-tests result in NUT setting FSD on their systems. **This happens on the build from master (as of 06.10.2023), but does not happen on 2.8.0 or 2.7.4. stables respectively**, so it's likely an issue with changes introduced after the 2.8.0 release - the driver affected seems to be **usbhid-ups.**

The problem seems to be that during the UPS self-testing and right before NUT sets FSD, the driver thinks to receive an OFF event from the UPS, resulting in the MinSupplies falling below 1 (on a home setup with one UPS configured) and NUT proceeding to set FSD because of that. This can be re-produced by triggering a manual self-test on the affected APC series. Setting the MinSupplies to 0 does not result in FSD being set by NUT.

I only have limited logs available so far and will try to get more detailed logs with an increased debug level, but this is what I have from the users affected so far. Note that NUT does not report anything else above the first line, so the problem starts with that. 

The users reported to hear the UPS inverter clicking during self-testing, these logs appearing and the system shutting down. This can be re-produced by triggering a manual self-test on the affected APC series.

```
Oct  8 02:35:51 UNRAID upsmon[27897]: UPS qnapups@127.0.0.1: administratively OFF or asleep
Oct  8 02:35:51 UNRAID upsd[27893]: Client qnapuser@127.0.0.1 set FSD on UPS [qnapups]
Oct  8 02:35:51 UNRAID upsmon[27897]: Executing automatic power-fail shutdown
Oct  8 02:35:51 UNRAID upsmon[27897]: Auto logout and shutdown proceeding
Oct  8 02:35:56 UNRAID shutdown[3133]: shutting down for system halt
```

![image](https://github.com/networkupstools/nut/assets/24509509/5fa83232-f854-4de8-82ee-da555db15700)
Here we can see the UPS being on OL CHRG OFF, as well as FSD being set right before the system goes down.

I'll try to get more diagnostic information from the users (hopefully, they've already downgraded to 2.8.0 stable), but I thought I'd report this here nonetheless, maybe other users have the same problem and I think unexpected shutdowns are a larger issue.

jimklimov:
[...]

PR #2406, opened 2024-04-16
upsmon: introduce `OBLBDURATION` config toggle

jimklimov:
Largely follows in the footsteps of #2104 which introduced the `OFFDURATION` toggle.

This PR adds an `OBLBDURATION` toggle to delay issuing FSD (primary) or commencing a shutdown after `HOSTSYNC` (secondary without a seen FSD alert), for users that deem this desirable (some use-cases suggested in issue #321 discussion - primarily to survive short hiccups that happen at the wrong moment) at their discretion and risk of an ungraceful shutdown.

Closes: #321

CC @desertwitch @dark-penguin @clepple @aquette 

Testing in practice would be as welcome as code review :)

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -71,12 +71,14 @@ https://github.com/networkupstools/nut/milestone/11
    reading an empty string or getting a success code `0` from libusb.
    This difference should now be better logged, and not into syslog. [#2399]
 
- - upsmon: it was realized that the `POWERDOWNFLAG` must be explicitly set in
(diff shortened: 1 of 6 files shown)


desertwitch:
Hello Jim, thanks for your effort, I've tested all scenarios (OBLBDURATION -1, OBLBDURATION 0, OBLBDURATION 30, OBLBDURATION not set) and can confirm everything is working as it should. Code-review looking good and reasonable too, the increased polling frequency and countdown-style warning messages are an informative and useful touch. Rysz

Issue #2378, opened 2024-03-28
Support of linux hwmon (ina219)


jviki:
Would it be possible to integrate support for standard linux hwmon that can monitor power of the system? I am running a small system (baseboard: [CM4-POE-UPS-BASE](http://www.waveshare.com/wiki/CM4-POE-UPS-BASE)) with batteries that contains the [ina219](https://docs.kernel.org/hwmon/ina2xx.html) I2C device. I hoped that I can find some generic hwmon UPS-like support here but I didn't.

I've got a short bash script that can be used as a basis for a nut driver:

```
#! /bin/bash

find_ina219()
{
	for d in $(ls /sys/class/hwmon/); do
		path="/sys/class/hwmon/${d}/name"
		name=$(cat ${path})

		if [ "${name}" == ina219 ]; then
			echo "/sys/class/hwmon/${d}"
			return 0
		fi
	done

	return 1
}

ina219=$(find_ina219)

if [ "$?" != 0 ]; then
	echo "could not find ina219 device" >&2
	exit 1
fi

voltage=`cat ${ina219}/in1_input`
current=`cat ${ina219}/curr1_input`

echo "Status: "$(expr \( ${voltage} - 3000 \) / 12)" % (${voltage} mV, ${current} mA)"

if [ "${voltage}" -lt 3150 ] && [ "${current}" -gt 50 ]; then 
	echo "Low power"
	sync
fi
```

Quick test:
```
# status: OL, HB
$ ./check-ups.sh 
Status: 95 % (4140 mV, 0 mA)

# status: OB, DISCHRG (unplugged AC)
$ ./check-ups.sh 
Status: 90 % (4084 mV, 87 mA)
$ ./check-ups.sh 
Status: 89 % (4072 mV, 114 mA)

# status: OL, CHRG (plugged AC back)
$ ./check-ups.sh 
Status: 102 % (4224 mV, -150 mA)
```

> if the current is positive, it means the output current of the battery (which can be considered as the Raspberry Pi load current), and if the current is negative, it means the battery charging current.
> -- [Battery Level Detection](https://www.waveshare.com/wiki/CM4-POE-UPS-BASE#INA219_Battery_Level_Detection)

Where should I start when trying to integrate this?

jimklimov:
There are a few NUT drivers using the Linux i2c bus, a relatively recent one is pijuice. I suppose taking a look at those (and their integration to build system in makefiles) would be a helpful start.

jimklimov:
[...]

PR #2430, opened 2024-05-02
New driver hwmon_ina219

jviki:
Initiated as #2378.

The driver implements communication with INA219 module that is supported by the Linux Kernel via hwmon interface. It is assumed that the INA219 monitors current and voltage from batteries. The implementation is based on the INA219 installed on Raspberry PI CM4 baseboard https://www.waveshare.com/wiki/CM4-POE-UPS-BASE. The code is inspired by the section https://www.waveshare.com/wiki/CM4-POE-UPS-BASE#INA219_Battery_Level_Detection.

The driver was successfully tested and monitored by using `upslog`. The device was equipped with 3x3000 mAh Li-lon batteries. The plotted data:

image

The system runs on the battery only. At about 18:36 (see the dot in the plot), the power was plugged shortly to test whether it is correctly detected:

```
20240501 183536 70 NA NA [OB DISCHRG] NA NA
20240501 183606 72 NA NA [OB DISCHRG] NA NA
20240501 183636 86 NA NA [OL CHRG HB] NA NA
20240501 183706 72 NA NA [OB DISCHRG] NA NA
20240501 183736 72 NA NA [OB DISCHRG] NA NA
```

The nut-monitor service successfully detected 15 % of battery and initiated shutdown:

```
# May 01 18:05:07 umbrel upsmon[1359]: UPS test@localhost on battery
# May 01 18:05:07 umbrel systemd[1]: Started Network UPS Tools - power device monitor and shutdown controller.
# May 01 18:36:12 umbrel upsmon[1359]: UPS test@localhost on line power
# May 01 18:36:47 umbrel upsmon[1359]: UPS test@localhost on battery
# May 01 23:16:29 umbrel upsmon[1359]: UPS test@localhost battery is low
# May 01 23:16:29 umbrel upsmon[1359]: Executing automatic power-fail shutdown
# May 01 23:16:29 umbrel upsmon[1359]: Auto logout and shutdown proceeding
# May 01 23:16:34 umbrel systemd[1]: nut-monitor.service: Succeeded.
```

The shutdown was intentionally suppressed to see how the battery behaves until it is fully depleted:

```
20240501 231536 17 NA NA [OB DISCHRG] NA NA
20240501 231606 17 NA NA [OB DISCHRG] NA NA
20240501 231636 14 NA NA [FSD OB DISCHRG LB] NA NA
20240501 231706 13 NA NA [FSD OB DISCHRG LB] NA NA
20240501 231736 13 NA NA [FSD OB DISCHRG LB] NA NA
20240501 231806 13 NA NA [FSD OB DISCHRG LB] NA NA
20240501 231836 10 NA NA [FSD OB DISCHRG LB] NA NA
20240501 231906 11 NA NA [FSD OB DISCHRG LB] NA NA
20240501 231936 11 NA NA [FSD OB DISCHRG LB] NA NA
20240501 232006 9 NA NA [FSD OB DISCHRG LB] NA NA
20240501 232036 8 NA NA [FSD OB DISCHRG LB] NA NA
20240501 232106 7 NA NA [FSD OB DISCHRG LB] NA NA
20240501 232136 5 NA NA [FSD OB DISCHRG LB] NA NA
20240501 232206 4 NA NA [FSD OB DISCHRG LB] NA NA
20240501 232236 2 NA NA [FSD OB DISCHRG LB] NA NA
20240501 232306 0 NA NA [FSD OB DISCHRG LB] NA NA
```


jimklimov:
Looks interesting, thanks!

I am not sure about classifying it as an "I2C" driver though, as it has no actual dependency on `libi2c` and is based on essentially parsing strings from sysfs (whose backend module and hardware may be or not be using I2C - as a black box, as far as NUT is concerned).

Maybe this warrants a new category, or placement into `NUTSW_DRIVERLIST`. Technically this driver (or the likes of it) might not be constrained to specifically Linux (there are other OSes allowing to tap into kernel/module info as files) and/or INA219 chip (we essentially expect certain names for voltage/current/... readings).

I can quite envision a future with a generic driver that would have a mapping in configuration - which NUT data points can be found in which sysfs-style filename (and perhaps a scaling factor for general approach), not unlike DMF as proposed in forks earlier for SNMP and USB HID drivers, and so the same `nutdrv_hwmon` binary would handle various devices flexibly.

But I guess the current tested implementation can stay as is (notably to constrain builds and delivery to Linux out of the box at the moment), and the ideas above addressed/generalized later.

jimklimov:
Would you have a data dump (e.g. an `upsc` report) to stash for comparisons later?

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,339 @@
...
+static unsigned int low_battery = LOW_BATTERY;
+
+/**
+ * @brief Threshold for detection of HB status.
+ */
+static unsigned int high_battery = HIGH_BATTERY;
+
+/**
+ * @brief Voltage value as recently read from in1_input.
+ */
```
jimklimov, line 65: Makes sense to comment that it would (may?) be in milliVolts / milliAmps (below) so you would shift it by 3 orders of magnitude into standard units range.

In fact, is it "guaranteed" that the values from sysfs would be integers and in milliUnits? Would it make sense to parse them in `file_read_number()` as e.g. `double` and range-check (if over 1000 then shift else use as is)? Just something to think about, I have no answers here :)

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,339 @@
...
+
+void upsdrv_updateinfo(void)
+{
+	unsigned int charge = 0;
+	int stale = 0;
+
+	if (update_voltage() < 0)
+		stale = 1;
+
+	upsdebugx(1, "Battery voltage: %.3fV", voltage / 1000.0);
```
jimklimov, line 282: Consider a quieter debug level for the regular updates please. A single `-D` is not expected to be too noisy :)

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,339 @@
...
+	upsdebugx(1, "Battery current: %.3fA", current / 1000.0);
+
+	if (stale) {
+		dstate_datastale();
+		return;
+	}
+
+	status_init();
+
+	charge = (voltage - 3000) / 12;
```
jimklimov, line 296: Note earlier comments about possibly range-checking (e.g. if accounting is truly in milli-units or can vary between systems), so whether `3000` is correct here.

In fact, this seems to assume a few numbers about the battery (high/low voltages for its dynamic range) which can depend on its chemistry (model/maker), age, etc. This should be at least commented, ideally externalized as macros or variables (setting `default.battery.voltage.high` etc. is pretty common in other drivers; note that the `default.` or `override.` prefix would be chopped by NUT core code, so your init or update methods would deal with `battery.voltage.high` whether served by HW or coming from config).

You've mentioned Li-Ion batteries, so the nominal would be around 3600mV? The 3 batteries of yours, are they in parallel or sequential connection (do their voltages sum up or average out, essentially)?

I can guess your setup dealt with `(3600 - 3000) / 12 = 600/12 = 50` (%) charge assumption most of the time when the graph is almost level? Allegedly, starting points for some chemistries can be over 4V (per https://www.jackery.com/blogs/knowledge/lithium-ion-battery-voltage-chart for example) with a quick drop to ~3.5V when discharging, so that formula could be showing close to 100% when charged.

jimklimov (changes requested):
Quite good as noted before. Got a few comments and proposed changes while reading the code. Also, can you please add a `docs/man/hwmon_ina219.txt` man page, even if quite a simplistic one, for consistency and to document the few options the driver has (and perhaps to note that currently it only "sees" two readings from hardware, the momentary voltage and the current)? Your greatly detailed commit comment is a good start for large parts of the man page (especially about Raspberry OS image preparation).

Actually, with the added comment below about magic numbers for voltage=>charge conversion, when/if that becomes configurable, warrants being documented too (e.g. where to get those numbers, like `cat` this path on a charged system).

Also, a `NEWS.adoc` line to welcome the new driver would be useful :)

jviki:
> Would you have a data dump (e.g. an `upsc` report) to stash for comparisons later?

I am not sure what exactly do you mean. Simple the output?

```
battery.charge: 100
battery.charge.high: 75
battery.charge.low: 15
battery.current: 0.000
battery.voltage: 4.204
device.description: Bidirectional Current/Power Monitor With I2C Interface
device.mfr: Texas Instruments
device.model: INA219
device.type: ups
driver.debug: 0
driver.name: hwmon_ina219
driver.parameter.port: /dev/null
driver.state: quiet
driver.version: 2.8.1.1
driver.version.internal: 0.1
ups.mfr: Texas Instruments
ups.model: INA219
ups.status: OL HB
ups.type: ups
```

jviki:
> Quite good as noted before. Got a few comments and proposed changes while reading the code. Also, can you please 

Thanks for review. I'll address the comments soon.

> add a `docs/man/hwmon_ina219.txt` man page, even if quite a simplistic one, for consistency and to document the few options the driver has (and perhaps to note that currently it only "sees" two readings from hardware, the momentary voltage and the current)? Your greatly detailed commit comment is a good start for large parts of the man page (especially about Raspberry OS image preparation).

Sure, I didn't know where exactly put which information so I made the commit message longer. It should be probably moved out as you suggest.

> 
> Actually, with the added comment below about magic numbers for voltage=>charge conversion, when/if that becomes configurable, warrants being documented too (e.g. where to get those numbers, like `cat` this path on a charged system).

Well, I just copied what I found in the example code for [CM4-POE-UPS-BASE](https://www.waveshare.com/wiki/CM4-POE-UPS-BASE) :angel:. I'll try to figure out the magic behind that.

> 
> Also, a `NEWS.adoc` line to welcome the new driver would be useful :)

:+1:

jviki:
> Hey guys, sorry to bother you with a noob question: I can't get the demo script provided by Waveshare to read the INA219 on the CM4-UPS-POE-BASE board... What I2C bus do you guys reach it on? On my bus 1, I don't see ANY I2C addresses, and on bus 20 and 21 there seem to be loads (none of them accessible though)?!? Help or even hints greatly appreciated...

You have to enable the appropriate I2C controller, have you done it? It took me a while to find it out, it is not clear from the waveshare description. You can find the necessary steps (edit boot/config.txt and include device tree overlay) in the commit message 80756e1424bf89bbb943904e39dcda03d7b7896b.

3dti:
> > Hey guys, sorry to bother you with a noob question: I can't get the demo script provided by Waveshare to read the INA219 on the CM4-UPS-POE-BASE board... What I2C bus do you guys reach it on? On my bus 1, I don't see ANY I2C addresses, and on bus 20 and 21 there seem to be loads (none of them accessible though)?!? Help or even hints greatly appreciated...
> 
> You have to enable the appropriate I2C controller, have you done it? It took me a while to find it out, it is not clear from the waveshare description. You can find the necessary steps (edit boot/config.txt and include device tree overlay) in the commit message [80756e1](https://github.com/networkupstools/nut/commit/80756e1424bf89bbb943904e39dcda03d7b7896b).

This looks like life-saving intel Jan. Thx a million, will try this asap!

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,339 @@
...
+	if ((ret = detect_hwmon_ina219(hwmon_dir)) < 0) {
+		errno = -ret;
+		fatal_with_errno(EXIT_FAILURE, "detect_hwmon_ina219(%s) has failed",
+				hwmon_dir);
+	}
+
+	if (getval("lb")) {
+		const int tmp = atoi(getval("lb"));
+
+		if (tmp == 0 || tmp > 100) {
```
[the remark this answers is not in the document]
jviki: Yes...

```
$ xxd /sys/class/hwmon/hwmon2/name 
00000000: 696e 6132 3139 0a                        ina219.
$ strace -e read cat /sys/class/hwmon/hwmon2/name 
read(3, "\177ELF\2\1\1\3\0\0\0\0\0\0\0\0\3\0\267\0\1\0\0\0\360\16\2\0\0\0\0\0"..., 832) = 832
read(3, "ina219\n", 131072)             = 7
ina219
read(3, "", 131072)                     = 0
+++ exited with 0 +++
```

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,339 @@
...
+			upsdebugx(3, "skipping path %s/%s", sysfs_hwmon_dir, entry->d_name);
+			continue;
+		}
+
+		upsdebugx(3, "checking %s/%s", sysfs_hwmon_dir, entry->d_name);
+
+		snprintf(namepath, sizeof(namepath), "%s/%s/name",
+					sysfs_hwmon_dir, entry->d_name);
+
+		if (!file_contains(namepath, "ina219\n"))
```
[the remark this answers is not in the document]
jviki: Yes...

```
$ xxd /sys/class/hwmon/hwmon2/name 
00000000: 696e 6132 3139 0a                        ina219.
$ strace -e read cat /sys/class/hwmon/hwmon2/name 
read(3, "\177ELF\2\1\1\3\0\0\0\0\0\0\0\0\3\0\267\0\1\0\0\0\360\16\2\0\0\0\0\0"..., 832) = 832
read(3, "ina219\n", 131072)             = 7
ina219
read(3, "", 131072)                     = 0
+++ exited with 0 +++
```

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,339 @@
...
+	size_t len;
+
+	if ((f = fopen(path, "r")) == NULL) {
+		upslog_with_errno(LOG_ERR, "unable to open %s", path);
+		return 0;
+	}
+
+	len = fread(buf, 1, sizeof(buf), f);
+	fclose(f);
+	upsdebugx(4, "read %zu bytes", len);
```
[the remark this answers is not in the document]
jviki: Done.

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,339 @@
...
+#endif
+#ifndef STRFY
+	#define STRFY(x) STRFY0(x)
+#endif
+
+#define SYSFS_HWMON_DIR                     "/sys/class/hwmon"
+#define LOW_BATTERY                         15
+#define HIGH_BATTERY                        75
+#define DRIVER_NAME                         "hwmon-INA219 UPS driver"
+#define DRIVER_VERSION                      "0.1"
```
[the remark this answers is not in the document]
jviki: Done.

drivers/hwmon_ina219.c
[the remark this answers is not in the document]
jviki: Done.

drivers/hwmon_ina219.c
jviki: > Makes sense to comment that it would (may?) be in milliVolts / milliAmps (below) so you would shift it by 3 orders of magnitude into standard units range.

Mentioned in the doc comment.

> In fact, is it "guaranteed" that the values from sysfs would be integers and in milliUnits? Would it make sense to parse them in file_read_number() as e.g. double and range-check (if over 1000 then shift else use as is)? Just something to think about, I have no answers here :)

Yes. From kernel doc, it is clearly stated that the miliUnits are used [1]. In sysfs interface standard [2], it is stated that "All sysfs values are fixed point numbers." You can find the print format "%d" in the kernel source code [3].

[1] https://docs.kernel.org/hwmon/ina2xx.html#general-sysfs-entries
[2] https://docs.kernel.org/hwmon/sysfs-interface.html#naming-and-data-format-standards-for-sysfs-files
[3] https://github.com/torvalds/linux/blob/7367539ad4b0f8f9b396baf02110962333719a48/drivers/hwmon/ina2xx.c#L300C16-L300C33

drivers/hwmon_ina219.c
jviki: I put _3_ there.

drivers/hwmon_ina219.c
jviki: > You've mentioned Li-Ion batteries, so the nominal would be around 3600mV? The 3 batteries of yours, are they in parallel or sequential connection (do their voltages sum up or average out, essentially)?

I use 3x SONY VTC6 - 18650 Battery - 30A (e.g. here https://www.cigareta-shop.eu/sony-vtc6-18650-battery-30a/). Parallel setup, thus their capacity should sum up and voltage is the same. Yes, 3.6 V is the nominal voltage.

> I can guess your setup dealt with (3600 - 3000) / 12 = 600/12 = 50 (%) charge assumption most of the time when the graph is almost level? Allegedly, starting points for some chemistries can be over 4V (per https://www.jackery.com/blogs/knowledge/lithium-ion-battery-voltage-chart for example) with a quick drop to ~3.5V when discharging, so that formula could be showing close to 100% when charged.

The nominal voltage of the 18650 battery is 3,6 V. The fully charged voltage is 4,25 V. Thus, according to the section "What voltage should a lithium battery read?", it is depleted at 2,8-3 V. This is exactly what the code does:

* _(4.25 - 3) / 1.25_ - the actual formula, it should give quite exact charge in percent,
* _(4250 - 3000) / 12_ - the code is a bit shifted (miliV) and we do integer math (counting in 0..100 not in 0.01..1), thus it is not accurate and gives more than 100 % on fully charged

And empirically confirmed by the plot in this PR.

> Note earlier comments about possibly range-checking (e.g. if accounting is truly in milli-units or can vary between systems), so whether 3000 is correct here.

I believe it is correct.

>
> In fact, this seems to assume a few numbers about the battery (high/low voltages for its dynamic range) which can depend on its chemistry (model/maker), age, etc. This should be at least commented, ideally externalized as macros or variables (setting default.battery.voltage.high etc. is pretty common in other drivers; note that the default. or override. prefix would be chopped by NUT core code, so your init or update methods would deal with battery.voltage.high whether served by HW or coming from config).

OK, I am bit lost in those nut variables and their exact semantics and where they come from. Would it be ok to set:

* default.battery.voltage.high = 4250
* default.battery.voltage.low = 3000

and make somehow possible to override those values? Can you point me to a driver where do you think it is close to this situation?

jimklimov: Thanks for the maby other replies abd clarifications. As for battery voltage high/low, recent-ish examples on my mind are in blazer_usb and nutdrv_qx enhancements to "guesstimation".

Typically there are static variables in the driver with default values for low/high(/nominal), and an e.g. `getval("battery.voltage.high")` would yield either NULL or user's `default` or `override` setting (same in this case). Typically you'd check for this in initups method (like your already checked `lb` etc. settings) to set those static variables for each value.

3dti:
> You have to enable the appropriate I2C controller, have you done it? It took me a while to find it out, it is not clear from the waveshare description. You can find the necessary steps (edit boot/config.txt and include device tree overlay) in the commit message [80756e1](https://github.com/networkupstools/nut/commit/80756e1424bf89bbb943904e39dcda03d7b7896b).

I've gone through these instructions, created the .dts, placed into /boot/overlays, unfortunately to no avail... I'm stuck at the same situation: enabling I2C in Interface Options gets me 3 buses: 1, 20 and 21. With or without the "manual .dts", still not seeing the 0x43 address pop-up. On which bus do you guys find the 0x43 INA219?? At this point, I'd already be happy to see the component, not even talking about implementing proper UPS functionality yet... Thanks again, you seem to previously have ran into the same problem as me and clearly solved it!!

jviki:
> > You have to enable the appropriate I2C controller, have you done it? It took me a while to find it out, it is not clear from the waveshare description. You can find the necessary steps (edit boot/config.txt and include device tree overlay) in the commit message [80756e1](https://github.com/networkupstools/nut/commit/80756e1424bf89bbb943904e39dcda03d7b7896b).
> 
> I've followed these clear instructions to the letter, amended config.txt, manually created the .dts, placed it into /boot/overlays, unfortunately to no avail... I'm stuck at the same situation: enabling I2C in Interface Options gets me 3 buses: 1, 20 and 21. With or without the "manual .dts", i2cdetect is still not seeing the 0x43 address... On which bus do you guys find the 0x43 INA219?? Are the appropriate pull-up resistors present on the board? Do they need (dip)switching on somewhere? At this point, I'd already be happy to see the component, not even talking about implementing proper UPS functionality yet... Thanks again, you seem to previously have ran into the same problem as me and clearly solved it!!
> cm4_vs_ina219

Unfortunately, on my device, the i2c devices are not exposed this way and are only accessible by hwmon. And the device is in use, so I cannot run the same experiments. But, the outputs are very strange, I remember that I could clearly see few addresses on each bus.

Outputs from my system:
```
$ uname -a
Linux umbrel 5.10.103-v8+ #1529 SMP PREEMPT Tue Mar 8 12:26:46 GMT 2022 aarch64 GNU/Linux
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/
```
```
$ ls /sys/class/i2c-adapter/i2c-*
/sys/class/i2c-adapter/i2c-0:
delete_device  device  mux_device  name  new_device  of_node  power  subsystem  uevent

/sys/class/i2c-adapter/i2c-10:
10-002f  10-0043  10-0051  delete_device  device  mux_device  name  new_device  of_node  power  subsystem  uevent

/sys/class/i2c-adapter/i2c-22:
delete_device  device  i2c-0  i2c-10  name  new_device  of_node  power  subsystem  uevent
```

```
$ ls /sys/class/i2c-adapter/i2c-*/of_node
/sys/class/i2c-adapter/i2c-0/of_node:
'#address-cells'   name   phandle   reg  '#size-cells'

/sys/class/i2c-adapter/i2c-10/of_node:
'#address-cells'   emc2301@2f   ina219@43   name   pcf85063a@51   phandle   reg  '#size-cells'   status

/sys/class/i2c-adapter/i2c-22/of_node:
'#address-cells'   clock-frequency   clocks   compatible   interrupts   name   phandle   reg  '#size-cells'   status
```

On the i2c bus 10, you can find those devices as well:

* RTC (PCF85063a) on i2c-10, address is 0x51 (7-bit address)
* FAN (EMC2301) on i2c-10, address is 0x2f (7-bit address)

From https://www.waveshare.com/wiki/CM4-POE-UPS-BASE:

> Note DSI and CSI are prohibited using RTC.
> I2C-10 is used by default.
> CSI and DSI are disabled by default. When using the camera and DSI, it will occupy three I2C devices: I2C-10, I2C-11, and I2C-0.

> If you need to use the camera or DSI screen, it is recommended to switch the I2C pin of the INA219 to GPIO2/3, and then modify it in the demo:
>
> #ina219 = INA219(i2c_bus=10,addr=0x43)
> ina219 = INA219(i2c_bus=1,addr=0x43)

Your setup looks like you are using the camera or DSI screen and thus you configured for this situation. In that case, the DTS overlay I included in the commit might not be right. I do not understand, what are those i2c-20 and i2c-21.

What OS do you run on your system?

3dti:
> Unfortunately, on my device, the i2c devices are not exposed this way and are only accessible by hwmon. And the device is in use, so I cannot run the same experiments. But, the outputs are very strange, I remember that I could clearly see few addresses on each bus.
> 
> Outputs from my system:
> 
> ```
> $ uname -a
> Linux system 5.10.103-v8+ #1529 SMP PREEMPT Tue Mar 8 12:26:46 GMT 2022 aarch64 GNU/Linux
> $ cat /etc/os-release 
> PRETTY_NAME="Debian GNU/Linux 10 (buster)"
> NAME="Debian GNU/Linux"
> VERSION_ID="10"
> VERSION="10 (buster)"
> VERSION_CODENAME=buster
> ID=debian
> HOME_URL="https://www.debian.org/"
> SUPPORT_URL="https://www.debian.org/support"
> BUG_REPORT_URL="https://bugs.debian.org/
> ```
> 
> ```
> $ ls /sys/class/i2c-adapter/i2c-*
> /sys/class/i2c-adapter/i2c-0:
> delete_device  device  mux_device  name  new_device  of_node  power  subsystem  uevent
> 
> /sys/class/i2c-adapter/i2c-10:
> 10-002f  10-0043  10-0051  delete_device  device  mux_device  name  new_device  of_node  power  subsystem  uevent
> 
> /sys/class/i2c-adapter/i2c-22:
> delete_device  device  i2c-0  i2c-10  name  new_device  of_node  power  subsystem  uevent
> ```
> 
> ```
> $ ls /sys/class/i2c-adapter/i2c-*/of_node
> /sys/class/i2c-adapter/i2c-0/of_node:
> '#address-cells'   name   phandle   reg  '#size-cells'
> 
> /sys/class/i2c-adapter/i2c-10/of_node:
> '#address-cells'   emc2301@2f   ina219@43   name   pcf85063a@51   phandle   reg  '#size-cells'   status
> 
> /sys/class/i2c-adapter/i2c-22/of_node:
> '#address-cells'   clock-frequency   clocks   compatible   interrupts   name   phandle   reg  '#size-cells'   status
> ```
> 
> On the i2c bus 10, you can find those devices as well:
> 
> * RTC (PCF85063a) on i2c-10, address is 0x51 (7-bit address)
> * FAN (EMC2301) on i2c-10, address is 0x2f (7-bit address)
> 
> From https://www.waveshare.com/wiki/CM4-POE-UPS-BASE:
> 
> > Note DSI and CSI are prohibited using RTC.
> > I2C-10 is used by default.
> > CSI and DSI are disabled by default. When using the camera and DSI, it will occupy three I2C devices: I2C-10, I2C-11, and I2C-0.
> 
> > If you need to use the camera or DSI screen, it is recommended to switch the I2C pin of the INA219 to GPIO2/3, and then modify it in the demo:
> > #ina219 = INA219(i2c_bus=10,addr=0x43)
> > ina219 = INA219(i2c_bus=1,addr=0x43)
> 
> Your setup looks like you are using the camera or DSI screen and thus you configured for this situation. In that case, the DTS overlay I included in the commit might not be right. I do not understand, what are those i2c-20 and i2c-21.
> 
> What OS do you run on your system?

Thank you so much for sharing your reflections! To make sure I'm not conflicting with previous settings attempts, I have installed a fresh Raspberry Pi OS Lite (bookworm), then literally just activated I2C in raspi-config and installed i2ctools and smbus. This listed me buses 1, 20 and 21 as shown in my screenshot.
I then followed the instructions you pointed me too, and simply nothing seems to have changed. My aim is to interact with the INA219 through our own python scripts.

I saw that the demo code from Waveshare also mentioned/used address 43 on bus 10. I have no idea where that bus 10 would come from....

I've tried contacting Waveshare support, but they didn't come further than "advising me to install smbus".

Do not feel obliged to lose your time on me. I'm not expert enough to figure things out without proper documentation. My bad...

jviki:
> Thank you so much for sharing your reflections! To make sure I'm not conflicting with previous settings attempts, I have installed a fresh Raspberry Pi OS Lite (bookworm), then literally just activated I2C in raspi-config and installed i2ctools and smbus. This listed me buses 1, 20 and 21 as shown in my screenshot. I then followed the instructions you pointed me too, and simply nothing seems to have changed. My aim is to interact with the INA219 through our own python scripts.
> 
> I saw that the demo code from Waveshare also mentioned/used address 43 on bus 10. I have no idea where that bus 10 would come from....
> 
> I've tried contacting Waveshare support, but they didn't come further than "advising me to install smbus".
> 
> Do not feel obliged to lose your time on me. I'm not expert enough to figure things out without proper documentation. My bad...

Do not give that easily.

Another output that you can compare to:
```
$ cat /boot/config.txt 
# For more options and information see
# http://rpf.io/configtxt
# Some settings may impact device functionality. See link above for details

# uncomment if you get no picture on HDMI for a default "safe" mode
#hdmi_safe=1

# uncomment this if your display has a black border of unused pixels visible
# and your display can output without overscan
#disable_overscan=1

# uncomment the following to adjust overscan. Use positive numbers if console
# goes off screen, and negative if there is too much border
#overscan_left=16
#overscan_right=16
#overscan_top=16
#overscan_bottom=16

# uncomment to force a console size. By default it will be display's size minus
# overscan.
#framebuffer_width=1280
#framebuffer_height=720

# uncomment if hdmi display is not detected and composite is being output
#hdmi_force_hotplug=1

# uncomment to force a specific HDMI mode (this will force VGA)
#hdmi_group=1
#hdmi_mode=1

# uncomment to force a HDMI mode rather than DVI. This can make audio work in
# DMT (computer monitor) modes
#hdmi_drive=2

# uncomment to increase signal to HDMI, if you have interference, blanking, or
# no display
#config_hdmi_boost=4

# uncomment for composite PAL
#sdtv_mode=2

#uncomment to overclock the arm. 700 MHz is the default.
#arm_freq=800

# Uncomment some or all of these to enable the optional hardware interfaces
#dtparam=i2c_arm=on
#dtparam=i2s=on
#dtparam=spi=on
dtparam=i2c_vc=on
dtoverlay=i2c-rtc,pcf85063a,i2c_csi_dsi
dtoverlay=cm4io-fan,minrpm=500,maxrpm=4800,midtemp=52000,midtemp_hyst=2000,maxtemp=60000,maxtemp_hyst=2000
dtoverlay=i2c-ina219
#dtoverlay=i2c-fan,emc2301,i2c_csi_dsi

# Uncomment this to enable infrared communication.
#dtoverlay=gpio-ir,gpio_pin=17
#dtoverlay=gpio-ir-tx,gpio_pin=18

# Additional overlays and parameters are documented /boot/overlays/README

# Enable audio (loads snd_bcm2835)
#dtparam=audio=on

# Enable DRM VC4 V3D driver on top of the dispmanx display stack
dtoverlay=vc4-fkms-v3d
max_framebuffers=2

arm_64bit=1
```

The `dtoverlay` options refer to the DTBO file (overlays). You can always convert DTBO back to DTS source code using `dtc -I dtb -O dts`. It might be a little bit messy, but from that, you can discover, how does it connect with the device tree. You need to figure out the right nodes in the device tree. Which are enabled, which are disabled and why (property status, when it is _okay_, it is enabled, otherwise disabled). The difficult part to me seems to be using of aliases. Somewhere, it refers to i2c_csi_dsi, elsewhere it is just i2c3...

Can you post these outputs?

```
$ ls /sys/class/i2c-adapter/i2c-*
$ ls /sys/class/i2c-adapter/i2c-*/of_node
```

The appropriate DTS sources for bookworm should be located here: https://elixir.bootlin.com/linux/v6.1/source/arch/arm/boot/dts. The `of_node` can help to find out where it refers into the DTS. It is a bit of mess but this is probably the way to figure out, what is going on. I could find multiple i2c controllers at https://elixir.bootlin.com/linux/v6.1/source/arch/arm/boot/dts/bcm2711.dtsi#L214.

It is about finding out, what is enabled in the `/boot/config.txt`, how does it refer into the DTS and thus to learn how to configure it. And probably, **my notes in that commit might be incomplete** or misleading and thus I'd like to make sure it is correct.

3dti:
I noticed that my config.txt had dtoverlay=i2c_arm=on, and your has i2c_vc=on...
After Googling the difference, I tried changing to VC instead of ARM, but since rebooting the CM4 hasn't come back online. Its running at my office. I will post the results of your 2 commands tomorrow! Thank you so much. This board would be perfect for us, if we can handle the INA!

3dti:
> I noticed that my config.txt had dtoverlay=i2c_arm=on, and your has i2c_vc=on... After Googling the difference, I tried changing to VC instead of ARM, but since rebooting the CM4 hasn't come back online. Its running at my office. I will post the results of your 2 commands tomorrow! Thank you so much. This board would be perfect for us, if we can handle the INA!

It finally did come online... AND IT DID THE TRICK!!! With _vc instead of _arm, I now have:
/dev/i2c-0  /dev/i2c-10  /dev/i2c-20  /dev/i2c-21  /dev/i2c-22

... and 0x43 is present on bus 10! And my python script can read it!
$ i2cdetect -y 10
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:                         -- -- -- -- 0c -- -- --
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 2f
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
40: -- -- -- 43 -- -- -- -- -- -- -- -- -- -- -- --
50: -- 51 -- -- -- -- -- -- -- -- -- -- -- -- -- --
60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
70: -- -- -- -- -- -- -- --

Thanks a million. Haven't completely understood the difference between _vc and _arm, but comparing to your config.txt made the difference! Thanks a million!!!

jviki:
> > I noticed that my config.txt had dtoverlay=i2c_arm=on, and your has i2c_vc=on... After Googling the difference, I tried changing to VC instead of ARM, but since rebooting the CM4 hasn't come back online. Its running at my office. I will post the results of your 2 commands tomorrow! Thank you so much. This board would be perfect for us, if we can handle the INA!
> 
> It finally did come online... AND IT DID THE TRICK!!! With _vc instead of _arm, I now have: /dev/i2c-0 /dev/i2c-10 /dev/i2c-20 /dev/i2c-21 /dev/i2c-22
> 
> ... and 0x43 is present on bus 10! And my python script can read it! $ i2cdetect -y 10 0 1 2 3 4 5 6 7 8 9 a b c d e f 00: -- -- -- -- 0c -- -- -- 10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 2f 30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 40: -- -- -- 43 -- -- -- -- -- -- -- -- -- -- -- -- 50: -- 51 -- -- -- -- -- -- -- -- -- -- -- -- -- -- 60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 70: -- -- -- -- -- -- -- --
> 
> Thanks a million. Haven't completely understood the difference between _vc and _arm, but comparing to your config.txt made the difference! 

Me neither. The only place I could find it is probably here: https://www.waveshare.com/wiki/CM4-POE-UPS-BASE#RTC. I am happy that it works for you.

> Thanks a million!!!

Thank you for testing this. I have to add this note somewhere into the doc. And I hope you will enjoy this driver in the future instead of that ugly custom python script.

3dti:
> > > I noticed that my config.txt had dtoverlay=i2c_arm=on, and your has i2c_vc=on... After Googling the difference, I tried changing to VC instead of ARM, but since rebooting the CM4 hasn't come back online. Its running at my office. I will post the results of your 2 commands tomorrow! Thank you so much. This board would be perfect for us, if we can handle the INA!
> > 
> > 
> > It finally did come online... AND IT DID THE TRICK!!! With _vc instead of _arm, I now have: /dev/i2c-0 /dev/i2c-10 /dev/i2c-20 /dev/i2c-21 /dev/i2c-22
> > ... and 0x43 is present on bus 10! And my python script can read it! $ i2cdetect -y 10 0 1 2 3 4 5 6 7 8 9 a b c d e f 00: -- -- -- -- 0c -- -- -- 10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 2f 30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 40: -- -- -- 43 -- -- -- -- -- -- -- -- -- -- -- -- 50: -- 51 -- -- -- -- -- -- -- -- -- -- -- -- -- -- 60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 70: -- -- -- -- -- -- -- --
> > Thanks a million. Haven't completely understood the difference between _vc and _arm, but comparing to your config.txt made the difference!
> 
> Me neither. The only place I could find it is probably here: https://www.waveshare.com/wiki/CM4-POE-UPS-BASE#RTC. I am happy that it works for you.
> 
> > Thanks a million!!!
> 
> Thank you for testing this. I have to add this note somewhere into the doc. And I hope you will enjoy this driver in the future instead of that ugly custom python script.

I will immediately admit I focussed on the UPS function (and therefore its wiki section) before anything...
You are totally right, the RTC section does mention the correct overlay... Maybe Waveshare expects you to run through ALL settings, in the order in which they are mentioned on their wiki?!? Anyway, MEGA delighted you helped me sort this!

drivers/hwmon_ina219.c
jimklimov: Thanks! Maybe it's worth pointing out in code comments, on one hand to avoid similar questions in the future, and on another to help eventual portability (whether to some newer Linux kernel 10.x convention or perhaps to extend this driver to more OSes with concepts similar to `hwmon`).

jviki:
> > Quite good as noted before. Got a few comments and proposed changes while reading the code. Also, can you please
> 
> Thanks for review. I'll address the comments soon.
> 
> > add a `docs/man/hwmon_ina219.txt` man page, even if quite a simplistic one, for consistency and to document the few options the driver has (and perhaps to note that currently it only "sees" two readings from hardware, the momentary voltage and the current)? Your greatly detailed commit comment is a good start for large parts of the man page (especially about Raspberry OS image preparation).
> 
> Sure, I didn't know where exactly put which information so I made the commit message longer. It should be probably moved out as you suggest.
> 
> > Actually, with the added comment below about magic numbers for voltage=>charge conversion, when/if that becomes configurable, warrants being documented too (e.g. where to get those numbers, like `cat` this path on a charged system).
> 
> Well, I just copied what I found in the example code for [CM4-POE-UPS-BASE](https://www.waveshare.com/wiki/CM4-POE-UPS-BASE) 👼. I'll try to figure out the magic behind that.
> 
> > Also, a `NEWS.adoc` line to welcome the new driver would be useful :)
> 
> 👍

I moved part of the commit message into `docs/man/hwmon_ina219.txt`. It checked the generated man page and it seems to look good.

jviki:
> > Would you have a data dump (e.g. an `upsc` report) to stash for comparisons later?
> 
> I am not sure what exactly do you mean. Simple the output?
> 
> ```
> battery.charge: 100
> battery.charge.high: 75
> battery.charge.low: 15
> battery.current: 0.000
> battery.voltage: 4.204
> device.description: Bidirectional Current/Power Monitor With I2C Interface
> device.mfr: Texas Instruments
> device.model: INA219
> device.type: ups
> driver.debug: 0
> driver.flag.allow_killpower: 0
> driver.name: hwmon_ina219
> driver.parameter.pollinterval: 2
> driver.parameter.port: /dev/null
> driver.parameter.synchronous: auto
> driver.state: quiet
> driver.version: 2.8.1.1
> driver.version.internal: 0.1
> ups.mfr: Texas Instruments
> ups.model: INA219
> ups.status: OL HB
> ups.type: ups
> ```

Now when `default.battery.voltage.nominal = 3.6` is put into `/etc/ups.conf` into the driver's section:

```
battery.charge: 100
battery.charge.high: 75
battery.charge.low: 15
battery.current: -0.125
battery.voltage: 4.228
battery.voltage.nominal: 3.6
device.description: Bidirectional Current/Power Monitor With I2C Interface
device.mfr: Texas Instruments
device.model: INA219
device.type: ups
driver.debug: 0
driver.name: hwmon_ina219
driver.parameter.default.battery.voltage.nominal: 3.6
driver.parameter.port: auto
driver.state: quiet
driver.version: 2.8.1.1
driver.version.internal: 0.01
ups.mfr: Texas Instruments
ups.model: INA219
ups.status: OL CHRG HB
ups.type: ups
```

or when the particular `low` and `high` are given:

```
battery.charge: 100
battery.charge.high: 75
battery.charge.low: 15
battery.current: -0.130
battery.voltage: 4.228
battery.voltage.high: 4.25
battery.voltage.low: 3
device.description: Bidirectional Current/Power Monitor With I2C Interface
device.mfr: Texas Instruments
device.model: INA219
device.type: ups
driver.debug: 0
driver.flag.default.: enabled
driver.name: hwmon_ina219
driver.parameter.default.battery.voltage.high: 4.25
driver.parameter.default.battery.voltage.low: 3
driver.parameter.port: auto
driver.state: quiet
driver.version: 2.8.1.1
driver.version.internal: 0.01
ups.mfr: Texas Instruments
ups.model: INA219
ups.status: OL CHRG HB
ups.type: ups
```

drivers/hwmon_ina219.c
jviki: I inspired a bit and reworked the `charge` computation. It is now more generic and useful with other battry nominal voltages as well. What do you think?

drivers/hwmon_ina219.c
jviki: I added few links there.

jviki:
@jimklimov How can I fix the spell checker issues?

jimklimov:
Usually by updating `docs/nut.dict` - e.g. with `make spellcheck-interactive` if you have `aspell` installed, or just editing the file following the logged error reports. I'll try to update it in your branch here - so please take care to not force-push over it ;)

jimklimov:
Well, took a few tries to do it with minimally invasive changes to your branch (which stands on a bit "old" foundations).

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,496 @@
...
+
+/**
+ * @brief Path usually pointing to /sys/class/hwmon/hwmonX.
+ */
+static char ina219_base_path[PATH_MAX];
+
+/**
+ * @brief Threshold for detection of LB status.
+ */
+static unsigned int battery_charge_low = BATTERY_CHARGE_LOW;
```
jimklimov, line 58: Hm, per `docs/nut-names.dict` this could in fact be set as `default.battery.charge.low` (instead of a custom `lb` parameter), and in fact there currently is no reserved name for the `.high` but makes sense to add one. I only see a handful of precedents setting the `HB` flag so had to wonder a bit if it is "just informative" or a warning of potential overcharge (harmful to the device). Would ask the mailing list just in case.

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,496 @@
...
+
+	upsdebugx(3, "checking %s", ina219_dir);
+
+	snprintf(namepath, sizeof(namepath), "%s/name", ina219_dir);
+
+	if (!file_contains(namepath, "ina219\n"))
+		return -ENODEV;
+
+	upsdebugx(3, "detected ina219 at %s", ina219_dir);
+	return 0;
```
jimklimov, line 146: Hm, this (and the use below) is a bit un-idiomatic, especially for a method named like a verb (`check` something) to return a not-true value when the check is successful.  I guess it is due to a non-zero negative error as a possible return.

Not sure if there's a good way out, maybe negate the name if it's hard to negate the return value (so below you would go `if (check_not_ina219()) continue;` for non-INA cases)...

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,496 @@
...
+
+static int update_current(void)
+{
+	return update_intvar(ina219_base_path, "curr1_input", &current);
+}
+
+void upsdrv_makevartable(void)
+{
+	addvar(VAR_VALUE, "sysfs_dir",
+			"Path to sysfs dir of hwmon (" SYSFS_HWMON_DIR ")");
```
jimklimov, line 237: `... if using port=auto`

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,496 @@
...
+	addvar(VAR_VALUE, "lb",
+			"Low battery threshold (" STRFY(BATTERY_CHARGE_LOW) ")");
```
jimklimov, line 239: Note the comment above about `default.battery.charge.low` maybe being more maintainable, less surprising vs. other drivers.

drivers/hwmon_ina219.c
jviki: Ok, the goal was not to return bool but to return OK or error code. I consider names `check_*()` as non-bool and expect them to return something more informative. For bool calls, I am used to name functions as `is_*()` or `can_*()`. But, no problem to rename it.

I would suggest to rename `detect_hwmon_ina219()` to `scan_hwmon_ina219()` and this `check_ina219()` function to `detect_hwmon_ina219()`. Does it make sense?

drivers/hwmon_ina219.c
jimklimov: I think it might, just not sure about the history behind the flag - too few hits in code to make sense of it quickly - hence why I'll ask on the ML.

For that matter, I see a couple of hits to `git grep -iE 'over.*charge'` that refer to over-charged battery state (and nearby some references to over-temperature flags), and they are in different sources compared to `git grep -w HB`.

drivers/hwmon_ina219.c
jimklimov: I think so, thanks.

drivers/hwmon_ina219.c
```diff
@@ -0,0 +1,496 @@
...
+			"Bidirectional Current/Power Monitor With I2C Interface");
+
+	dstate_setinfo("battery.charge.low", "%d", battery_charge_low);
+	dstate_setinfo("battery.charge.high", "%d", battery_charge_high);
+}
+
+static unsigned int battery_charge_compute(void)
+{
+	const unsigned int divisor =
+		(battery_voltage_max - battery_voltage_min) / 100;
```
jimklimov, line 415: I wonder if rounding to `int` so early is problematic for precision. In practice I'd expect this to vary between roughly 10 to 15 or so (e.g. `(4200-3000)/100 = 12.0`), so every whole unit may be significant in terms of needless error induction.

I'd suggest changing `divisor` and `charge` to `double` and only cast to `unsigned int` in the end.

drivers/hwmon_ina219.c
jimklimov: Logic looks good, but I feel integer data types can be a problem. Commented in more detail separately.

drivers/hwmon_ina219.c
jviki: Done.

drivers/hwmon_ina219.c
jviki: Renamed.

drivers/hwmon_ina219.c
jviki: I based my high-battery threshold assumption on `pijuice.c`:

```
413     if ( batteryStatus == BATT_NORMAL ||                                         
414          batteryStatus == BATT_CHARGING_FROM_IN ||                               
415          batteryStatus == BATT_CHARGING_FROM_5V )                                
416     {                                                                            
417         get_charge_level_hi_res();                                               
418                                                                                  
419         if ( battery_charge_level <= LOW_BATTERY_THRESHOLD )                     
420         {                                                                        
421             upsdebugx( 1, "Battery Charge Status: LOW" );                        
422             snprintfcat( status_buf, ST_MAX_VALUE_LEN, "LB " );                  
423         }                                                                        
424         else if ( battery_charge_level > HIGH_BATTERY_THRESHOLD )                
425         {                                                                        
426             upsdebugx( 1, "Battery Charge Status: HIGH" );                       
427             snprintfcat( status_buf, ST_MAX_VALUE_LEN, "HB " );                  
428         }                                                                        
429     }
```

I removed the `hb` option and changed `lb` to `battery.charge.low`.

drivers/hwmon_ina219.c
jviki: Added.

drivers/hwmon_ina219.c
jviki: Changed.

jviki:
Current upsc output:
```
$ upsc test
Init SSL without certificate database
battery.charge: 91
battery.charge.low: 15
battery.current: 0.000
battery.voltage: 4.144
battery.voltage.nominal: 3.6
device.description: Bidirectional Current/Power Monitor With I2C Interface
device.mfr: Texas Instruments
device.model: INA219
device.type: ups
driver.debug: 0
driver.name: hwmon_ina219
driver.parameter.default.battery.voltage.nominal: 3.6
driver.parameter.port: auto
driver.state: quiet
driver.version: 2.8.1.1
driver.version.internal: 0.01
ups.mfr: Texas Instruments
ups.model: INA219
ups.status: OL
ups.type: ups
```

jimklimov:
Thanks a lot for the changes, will read into them tomorrow :)
For now a quick question: zero readings for current - are they correct (also what the sysfs node reports) or some parsing problem?

jviki:
> Thanks a lot for the changes, will read into them tomorrow :) For now a quick question: zero readings for current - are they correct (also what the sysfs node reports) or some parsing problem?

Is this listing more satisfying? :laughing: (I unplugged to power in and it now runs on battery only, see DISCHRG flag.)

```
$ upsc test
Init SSL without certificate database
battery.charge: 87
battery.charge.low: 15
battery.current: 0.081
battery.voltage: 4.092
battery.voltage.nominal: 3.6
device.description: Bidirectional Current/Power Monitor With I2C Interface
device.mfr: Texas Instruments
device.model: INA219
device.type: ups
driver.debug: 0
driver.name: hwmon_ina219
driver.parameter.default.battery.voltage.nominal: 3.6
driver.parameter.port: auto
driver.state: quiet
driver.version: 2.8.1.1
driver.version.internal: 0.01
ups.mfr: Texas Instruments
ups.model: INA219
ups.status: OB DISCHRG
ups.type: ups
```

jimklimov:
Hm, duplicate `sysf(s)_dir` in usage is not needed - vars/flags are auto-documented, I will drop the extra:
````
:; ./drivers/hwmon_ina219 -h | grep sys
Path to sysfs dir of hwmon (/sys/class/hwmon) : -x sysfs_dir=<value>
  sysf_dir = /sys/class/hwmon (sysfs hwmon directory)
````

desertwitch:
This seems really cool and innovative - just wanted to say thanks a lot for the efforts to everyone involved. 💯 
Looking forward to experimenting with this down the road, just need to get some hardware parts first...

jimklimov:
Synced with current codebase to minimize surprises with upcoming merge.

PR #2442, opened 2024-05-14
`mge-hid.c`: extend `mge_model_names[]` for Eaton 5SC and 5PX series

jimklimov:
Closes: #2380

Testing needed, CC @jstallard @lemeshovich

Confirmation that the added device names/models are what should be expected from hardware would be very welcome - CC @arnaudquette-eaton

Here so far I've added the basic VA numbers from model lists for 5PX and 5SC, but there are many models marketed for some of the same power ratings - no idea if they identify differently via USB.

* https://www.eaton.com/gb/en-gb/site-search.html.searchTerm$5sc.tabs$all.html
* https://www.eaton.com/gb/en-gb/catalog/backup-power-ups-surge-it-power-distribution/eaton-5px-ups-emea.html

Also note that original 5PX is discontinued; not sure if 5PX Gen2 should identify differently (and/or if it needs another "type" enum value).

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -79,6 +79,10 @@ https://github.com/networkupstools/nut/milestone/11
  - usbhid-ups updates:
    * Support of the `onlinedischarge_log_throttle_hovercharge` in the NUT
      v2.8.2 release was found to be incomplete. [#2423, follow-up to #2215]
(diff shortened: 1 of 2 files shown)


jstallard:
Just confirming here as well that this looks good on my 5PX 1500. I can now see values for battery.voltage and battery.voltage.nominal.

masterwishx:
Can be 9E model also added?

Issue #2431, opened 2024-05-03
Libraries not found (Ubuntu 24.04 Server arm)


sccfit:
Hi all... just tried to install nut on fresh install Ubuntu 24 LTS on raspberry pi, installed fine but running sudo nut-scanner -C results in following:
```
Cannot load USB library (libusb-1.0.so) : file not found. USB search disabled.
Cannot load SNMP library (libnetsnmp.so) : file not found. SNMP search disabled.
Cannot load XML library (libneon.so) : file not found. XML search disabled.
Cannot load AVAHI library (libavahi-client.so) : file not found. AVAHI search disabled.
Cannot load IPMI library (libfreeipmi.so) : file not found. IPMI search disabled.
Cannot load NUT library (libupsclient.so) : file not found. NUT search disabled.
```
I switched to Ubuntu because the package in it is closer to current than RPI OS, and I need the tools to connect to identical UPS on one device. I've confirmed libusb-1.0 is installed, but for some reason nut can't find it, do I need to create extra symlinks or something like that?

jimklimov:
Please retry with higher debug verbosity, e.g.: 
```
:; sudo nut-scanner -DDDDDD -C > /tmp/nutscan.log 2>&1
```

There would be a huge wall of text about directories it looks into and file names it sees and tries to match there.

I wonder if your system delivers e.g. `libusb-1.0.so.1.27.0` and no symlink like `libusb-1.0.so` which could be in another package.

jimklimov:
Also, just in case: how did you install NUT (packages? custom build?) and which version?

danmrossi:
Must be the OS that NUT was compiled on, I’ve just compiled and installed on Ubuntu 24.04 no issues.

sccfit:
Sorry, have been on holidays, looping back to this now. I'm not great with Linux/packages/github, I installed using apt which I believe should be version 2.8.1-3.1ubuntu2.

Have run with higher debug verbosity, the following seem to be the most pertinent lines:

```
   0.001188     [D1]    Built-in:       /usr/lib/gcc/aarch64-unknown-linux-gnu
   0.001239     [D2] Looking for lib libusb-1.0.so in directory #0 : /usr/lib/aarch64-linux-gnu

   -- snip --
   
   0.012290     [D5] Comparing lib libusb-1.0.so with dirpath entry libusb-1.0.so.0
[...]

Issue #2242, opened 2023-12-22
nut-scanner should detect and configure dummy simulation too


aquette:
provide an additional nut-scanner "-n | --nut_simulation_scan" option to detect and configure dummy simulation (.dev & .seq) from the default $sysconfdir.

This should standardize and streamline the use of nut-scanner as an automatic detection and configuration tool.

aquette:
PR  merged

PR #2449, opened 2024-05-18
`nut-scanner`: Fix "Old NUT" detection and clarify some log messages

jimklimov:
It seems that commit 3eb96fe98f9bd75813fa243832732e61ac614726 (PR #2246 for issue #2242) released in NUT v2.8.2 broke detection that `libupsclient.so` is not available (not a really practical situation breaking end-user setups, outside of some weird packaging choices on constrained systems, or builds on systems where NUT is not installed into common paths - it did fire up in test cases investigated for another issue #2431).

Also this PR takes a moment to clarify what "old nut" is about in `nut-scanner` context (as opposed to "new" avahi-based detection where available) so `help()` and codebase reading is more comprehensible.

Finally, this PR changes all flags about detected scanning methods to be zeroed initially, only set to `1` (even if always) in the `nutscan_init()` method, to be consistent.


aquette (approved):
Good catch @jimklimov !

PR #2459, opened 2024-05-28
bicker_ser: allow changing of non-volatile variables with `upsrw`

ntd:
I exposed all parameters and mapped them 1:1 to NUT names. I used `nut-names.txt` whenever possible and mapped the rest to other (hopefully logical) names. I then added the possibility to change everything with `upsrw` and added documentation. I don't know the policy on driver versioning, so I bumped the version in a separate commit: if not needed it can be easily dropped.

My Bicker UPS is still broken but it funnily responds to any request "almost" properly, so (apart from the disabling of parameters with an empty value) I tested everything. With "almost" I mean the returned parameter always has 0xFFFF as min/max and 0xFF as enabled.

One thing that could be a source of problems is parameter `0x01`, mapped to NUT `output.current.low`, is expressed in mA. Not sure if I need to convert it to A, as that name is not "official" (i.e. not present in `nut-names.txt`).

### DIFF
--- a/docs/man/bicker_ser.txt
+++ b/docs/man/bicker_ser.txt
@@ -82,11 +82,14 @@ Currently this setting is only valid for relay signaling. Cannot be disabled.
 *experimental.ups.relay.mode* (default `0x01`)::
 This parameter controls the behavior of the relay in case of different events.
 Cannot be disabled.
-- `0x01` On power fail (normally closed)
-- `0x02` On power fail (normally opened)
-- `0x03` Shutdown impulse (1 second)
-- `0x04` Battery low signal (normally closed)
-- `0x05` Battery defect signal (normally closed)
++
+Available relay modes:
+[horizontal]
+`0x01`::: On power fail (normally closed)
+`0x02`::: On power fail (normally opened)
+`0x03`::: Shutdown impulse (1 second)
+`0x04`::: Battery low signal (normally closed)
+`0x05`::: Battery defect signal (normally closed)
 
 INSTANT COMMANDS
 ----------------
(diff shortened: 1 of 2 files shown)


jimklimov (commented):
Looks reasonable, thanks.

Regarding your question on values not (yet) promoted in `nut-names.txt`, we have defined an `experimental.` prefix for such values to highlight on one hand for users that this is subject to change if some more common name appears after community discussion on the mailing list, and on another to maintainers/developers/community - that such discussion should be raised (perhaps several experimental names here and there map to the same concept).

docs/man/bicker_ser.txt
```diff
@@ -34,6 +34,60 @@ EXTRA ARGUMENTS
...
+
+*ups.relay.mode* (default `0x01`)::
+This parameter controls the behavior of the relay in case of different events.
+Cannot be disabled.
+- `0x01` On power fail (normally closed)
+- `0x02` On power fail (normally opened)
+- `0x03` Shutdown impulse (1 second)
+- `0x04` Battery low signal (normally closed)
+- `0x05` Battery defect signal (normally closed)
+
```
jimklimov, line 90: Can you please check this block in man/html rendition? I am afraid `a2x` would not make a bulleted list of these lines, but would place them into one long line. It may need a blank before start of list (or do I confuse that bit with markdown?) In case of "blank line as part of block", to keep the indentation, a single "plus" character on the blank line is used.

ntd: According to [asciidoctor](https://docs.asciidoctor.org/asciidoc/latest/lists/description/#mixing-lists):

> Description lists are quite lenient about whitespace [...]

so it shoud be fine with or without extra lines. Anyway I ended up converting that into a nested (horizontal) description list in 9c3dedff16d96cbc79e26e839a9747a0dec51241 as it seems more appropriate.

ntd:
> Regarding your question on values not (yet) promoted in `nut-names.txt`, we have defined an `experimental.` prefix [...]

That makes a lot of sense. I moved all non official variables into that namespace with 804323ea38578d7cd25045478225c7d9397aec5d.

jimklimov (approved):

Issue #2450, opened 2024-05-22
For the few cases where we use variables as formatting strings, find a way to ensure it is safe


jimklimov:
In a few cases we use formatting strings as variables (e.g. coming from some tables or even constructed at run-time) which is error-prone with regard to interpretation of subsequent memory stack when calling a `printf`-related method. While this is done only for strings defined in NUT codebase, it has a potential to regress if someone modifies the table value in some later revision. Currently we hush a warning like `format not a string literal and no format arguments` and a few similar others with pragmas, but there gotta be some better way.

Some first ideas (more welcome):
* create a macro (one with a number argument, or a set of macros with different amount of args) just to pass a hint about amount of expected formatting arguments, and somehow parse (`pycparser` as used elsewhere? simpler `perl` magic?) to statically check that the amount of unescaped percent characters in the resolved first argument (at least when fixed from tables) matches the macro hint amount;
* do same in runtime to at least throw `fatalx()` or similar and not do dangerous things - at least in our methods like `upsdebugx()` or `dstate_setinfo()`, we can at least control the `varargc` vs. amount of percents in the actual formatting string.

Either way, I think we lose the facility of modern compilers to also statically check the types (that a `%i` refers to an `int`-sized number, and not a `long` or `char*`, etc.) in these cases, so some error-proneness remains even if the amount of args remains but their type changes.

Maybe the solution to get the best of all worlds could be in fact to specify the runtime method to return a string (so callers would go like `dstate_setinfo("ups.model", "%s", checked_format(variableFormat, checkingFormat, ...));` and avoid hushing pragmas altogether) with `checkingFormat` being a real formatting string like `"%s%s%i%"PRIuSIZE"%f"` according to the types of subsequent vararg parameters, and the preceding `variableFormat` argument specifying the actual formatting string (expected/checked to mention exactly the same set of percent-formats in same order).

This way we could have compile-time checks that varargs conform in amount and type to some contrived formatting string, and run-time assertions that whatever variable string we *actually* use to produce the `checked_format()` string is compatible with those expectations.

Is there some (library?) method to strip non-formatting characters (plain text, format beautification with sizes/alignments like `%.01f` => `%f` or whatever) so we could directly `strcmp()` the expected pattern vs. the stripped dynamic formatting string? If not, we have a fallback `printf()` implementation that I guess could be wrangled into such a helper method...

ntd:
[...]

PR #2460, opened 2024-06-02
Harden NUT work with strings where dynamic formatting strings are used

jimklimov:
Closes: #2450

Builds upon the idea suggested in the issue above.

Impacts quite a lot of drivers and other code, so I initially plan on mile-stoning for a later release and asking the community about testing it in practice (after this seems okay with all systems of the NUT CI farm).

One way this can mis-fire is if some mapping tables (potentially) use different formatting strings like `%d` and `%ld` for some value stored in a `long` -- one of the sources deals with that by checking if the dynamic formatting string normalized to this or that and casts accordingly. Maybe this can be generalized. Test code is prepared for TDD-style development, if it comes to that. (Was already helpful for some scenarios)

Some locations were not modified, where varargs are used (at least currently) with fixed formatting strings, and static compilers do not smell anything fishy about this. A few cases can be currently simplified, where error messages etc. are thrown with only the "fixed" formatting string and no varargs. Not sure it *should* be done though (less flexibility later).

CC @arnaudquette-eaton : `snmp-ups` (primarily around daisy-chain support) is one of the impacted code-bases, testing with some ePDUs would be beneficial.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -121,6 +121,10 @@ https://github.com/networkupstools/nut/milestone/11
    `NUT_PIDPATH` environment variable in certain use-cases (such as tests).
    [#2407]
 
+ - common code hardening: added sanity-checking for dynamically constructed or
+   selected formatting strings with variable-argument list methods (typically
+   used with log printing, `dstate` setting, etc.) [#2450]
+
  - revised `nut.exe` (the NUT for Windows wrapper for all-in-one service)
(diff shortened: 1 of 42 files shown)


jimklimov:
Updated with target code base shortly after v2.8.3 release

jimklimov:
FIXME: Implement `dstate_addenum_dynamic()` - it is in header and comments already, but not in C code!

Issue #2454, opened 2024-05-23
nut-2.8.2 does not seem to honor DEADTIME


avg-I:
I have a single UPS and several computers powered off it.
One is a master that runs upsd and upsmon, others are slaves that run just upsmon.
Here is a snippet from upsmon.conf on one of the slaves:
```
MONITOR labups@neo.home.arpa 1  xxx xxx slave
POLLFREQ 15
POLLFREQALERT 5
DEADTIME 120
```

Recently, during a blackout, I had a glitch where the network interface on that slave went down for about 4 seconds.
Its upsmon started powering off the machine in about 5 seconds, which was not ideal in the situation.

Here are some logs:
```
May 23 05:55:50 super kernel: lagg0: link state changed to DOWN
May 23 05:55:54 super kernel: lagg0: link state changed to UP
May 23 05:55:55 super upsmon[1078]: Poll UPS [labups@neo.home.arpa] failed - Server disconnected
May 23 05:55:55 super upsmon[1078]: Communications with UPS labups@neo.home.arpa lost
May 23 05:55:55 super upsmon[1078]: UPS [labups@neo.home.arpa] was last known to be not fully online and currently is not communicating, assuming dead
May 23 05:55:55 super upsmon[1078]: Executing automatic power-fail shutdown
May 23 05:55:55 super upsmon[1078]: Auto logout and shutdown proceeding
```

I expected that upsmon would wait for `DEADTIME` before doing that.

What additional information should I provide?

jimklimov:
I suppose "link down/up" transitions broke the TCP session, so the `upsmon` client was forcefully disconnected from the data server while in a critical state, and behaved by design.

avg-I:
So, any communication problem between upsd and upsmon while on battery, and upsmon is supposed to immediately start powering off?

Then, what `DEADTIME` is for?
```
# DEADTIME - Interval to wait before declaring a stale ups "dead"
# 
# upsmon requires a UPS to provide status information every few seconds
# (see POLLFREQ and POLLFREQALERT) to keep things updated.  If the status
# fetch fails, the UPS is marked stale.  If it stays stale for more than
# DEADTIME seconds, the UPS is marked dead.
# 
# A dead UPS that was last known to be on battery is assumed to have gone
# to a low battery condition.  This may force a shutdown if it is providing
# a critical amount of power to your system.
```
Is that applicable only to a local (serial or USB connected) UPS?
Is there any control like that for network communication?

[...]

PR #2108, opened 2023-10-15
Revise `upsmon` reaction to UPS calibrations vs. critical state

jimklimov:
Addresses part of #2104

This PR revisits the changes from PR #2055 for issue #2044, which added handling of (presumed administrative) "OFF" and "BYPASS" states in `upsmon`, to report that while the power device remains manageable, its load is no longer fed (thus `upsmon` must decrease the counter of healthy power supplies).

Practice has shown that some devices begin their calibration rituals indistinguishably (at least in existing NUT drivers) from an "OFF" state, and only after a few seconds they switch to "OL+DISCHRG" (in the example case at #2104). This time gap sufficed for NUT to initiate system shutdown whenever the UPS started its regular calibration.

This PR introduces an `OFFDURATION` option to take into account the reported "OFF" state only if it has persisted long enough, and revises the triggers we use in `upsmon` to proclaim a device as being in critical state (hence to decrease the supplies counter, which can lead to initiating a shutdown).

PR #2462, opened 2024-06-05
clients/upsmon: restore handling of comm failure while on battery

avg-I:
Addresses #2454

Commit 2647f026f7f0 from #2108 updated `is_ups_critical()` such that a communication failure (`commstate == 0`) while on battery (`linestate == 0`) would be treated as immediately entering critical state.

There was no way to disable that behavior or add any grace period.  As such, that change made `DEADTIME` parameter ineffective as `lastpoll` being stuck should be equivalent to `commstate == 0` because any `lastpoll` update would set `commstate` to 1.

Also, that change in how a communication failure affects OB state was no explicitly communicated in the commit message or the pull request. They talked only about OFF, BYPASS and CAL states.

This change removes special handling of that condition from `is_ups_critical()` and reverts to the prior behavior where `recalc()` looks at `lastpoll` and `deadtime` to set LB flag after the configured communication timeout.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -51,6 +51,10 @@ https://github.com/networkupstools/nut/milestone/11
    * Addition of "NUT Simulated devices" support to `nut-scanner` in v2.8.2
      broke detection of (in-)ability to find and query "Old NUT" servers via
      `libupsclient.so` (internal flag got always enabled). [#2246]
+   * A fix for `upsmon` v2.8.1 setting of `OFFDURATION` [PR #2108, issue #2104,
+     revisiting PR #2055, issue #2044] was overly zealous and impacted also
+     the `OB` state in cases where communications to the data server were
+     severed and `DEADTIME` setting was not honored. [PR #2462, issue #2454]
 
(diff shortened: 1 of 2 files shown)


jimklimov:
Cheers @avg-I : after some more thought over the weekend, your fix makes sense. I've added the NEWS entry into your source branch, however I wonder if it is reasonable to continue logging about this situation as it can impact untimely hard power-offs of some systems and log messages can help with investigations?

I've posted that suggestion as a PR in your repo: https://github.com/avg-I/nut/pull/1

WDYT?

Also CCing @desertwitch as a reviewer of the original change :)

avg-I:
@jimklimov I think that we would get polling failure messages in that situation ("Poll UPS [%s] failed").
They are at even higher, ERROR level. 
The good thing about them is that they can be throttled using `POLLFAIL_LOG_THROTTLE_MAX`.

jimklimov:
Ok, that also makes sense, forgot about that while commuting.

desertwitch:
Makes sense, looking good - thanks!

Issue #1316, opened 2022-02-25
Upstreaming improvements from 42ity/nut fork


jimklimov:
A lot of work on NUT was done under the auspices of Eaton-backed 42ity project, as well as some earlier bundled releases that became part of its codebase. While much of the development was cross-posted for PRs, some was only pursued in the fork to perfect there first. 42ity took the decision to eventually move from the burden of maintaining a separate fork (thereby regularly solving merge conflicts) towards using the "upstream NUT" codebase directly, so I imported the "FTY" branch and identified the largest differing subjects to reconcile between the two:

* [ ]    DMF feature (not in first wave, a very big subject in itself): lua, libxml2, HAVE_XMLLINT, reference some mapping functions in "standard" binary driver ...
* [ ]    docs building (man-pages vs others; --with-doc=dist-auto / --with-doc=man=dist-auto)
* [ ]    nut-conf and installer scripts
* [ ]    drivers/snmp-ups.c - semi-static values, ambient info, maybe daisy-chain
* [x]    snmp-ups: sysContact, sysLocation => #320
* [x]    run-time debug settings via config files; separation of FG/BG runs from debug setting => #1293 
* [ ]    stc.c: str_concat()
* [ ]    nut-driver-enumerator-daemon
* [ ]    docs/man/nut-scanner.txt
* [ ]    docs/man/upsdrvsvcctl.txt support for --timeout-cmd / --timeout-args
* [ ]    docs/nut-names.txt - input.feed.color, input.feed.desc, input.realpower.nominal, outlet.group.n.input, many ambient.* data points
* [ ]    netxml-ups SSL support
* [ ]    apc-iem-mib.h
* [ ]    eaton-pdu-marlin-helpers.h/c
* [ ]    libserial-nutscan.la
* [ ]    struct init for info_lkp_t and snmp_info_t (varied number of fields)
* [ ]    drivers/eaton-ats16-nm2-mib.c - many different mappings
* [ ]    drivers/eaton-ats16-nmc-mib.c
* [ ]    drivers/eaton-pdu-marlin-mib.c
* [ ]    drivers/mge-mib.c
* [ ]    drivers/mge-xml.c
* [ ]    drivers/netxml-ups.c
* [ ]    drivers/powerware-mib.c
* [ ]    drivers/snmp-ups-helpers.c/h
* [ ]    check *.pc files (should use \@prefix\@, files in 42ity use /usr directly)
[...]

PR #2472, opened 2024-06-15
 FTY remerge from master as of 2024-06-15

jimklimov:
Bump the copy of 42ity fork contents to minimize difference with the master. Follows up from previous resync in PR #2275 (made before NUT v2.8.2 was released, including some features backported from the FTY branch), and related to #1316 generally.

Regarding parts that originated in 42ITy and were upstreamed earlier, the `nutconf` effort got largely updated in the upstream during the past few months.


jimklimov:
TODO (maybe in a follow-up PR if this one is basically OK for main NUT CI farm):
* Allow CI runners that react to `master` branch changes also cover the `FTY` branch.
* Probably need to solve https://github.com/networkupstools/nut/issues/2431#issuecomment-2169976264 for those builds to pass.

Issue #2529, opened 2024-07-11
CI: PyNUTClient src preparation sometimes fails on omnios build agent


jimklimov:
Does not seem to happen on every CI run, but tends to happen for scenarios like `slowbuild-run/MATRIX_TAG=gnu99-gnu++11-gcc-omnios` during a `distcheck`.

Build log complaints are:
````
...
Making all in python
Making all in module
  PYPI  Generate PyPI module source
SRCFILE="../../../../scripts/python/app/locale/ru/ru.po"; OUTFILE="../../../../scripts/python/app/locale/ru/LC_MESSAGES/NUT-Monitor.mo"; {  /usr/bin/ggrep -v -E '^.?POT-Creation-Date: ' < "${SRCFILE}" > "${OUTFILE}.tmpsrc" &&  /usr/bin/msgfmt -o "${OUTFILE}" "${OUTFILE}.tmpsrc" &&  rm -f "${OUTFILE}.tmpsrc" ;  }
sh[1]: ../../../../scripts/python/app/locale/ru/LC_MESSAGES/NUT-Monitor.mo.tmpsrc: cannot create [Permission denied]
*** Error code 1
make: Fatal error: Command failed for target `app/locale/ru/LC_MESSAGES/NUT-Monitor.mo'
Current working directory /home/abuild/jenkins-nut/workspace/nut_nut_PR-2528/nut-2.8.2.1/_build/sub/scripts/python
*** Error code 1
...
````

Gotta actually read into this, but probably the Sun `make` interprets some path variables differently than other `make`'s out there, and tries to write into read-only `dist`'ed source directory emulating a CD or other immutable media.

Curiously, this is not 100% reproducible and was not seen to happen e.g. on OpenIndiana so far. Maybe there's some parallel-build (or parallel-job) racing also?

jimklimov:
It would seem that the `*.po` files are "newer" in the unpacked `dist` area than the `*.mo` files, and so `make` decides to rebuild the `mo`'s. Probably fallout of git checkout and alphanumeric ordering, so the tracked `mo` files are instantiated first (or 50/50 random order?), and of not-building `NUT-Monitor desktop application` in the "root" scenario while auto-enabling it with `make distcheck-light` (so on the system where we can build it, it becomes active and the tarballed/unpacked files suffer the same timestamp mess).

[...]

PR #2530, opened 2024-07-11
scripts/python/Makefile.am: revise *.po => *.mo generation [#2529]

jimklimov:
Closes: #2529 
(hopefully)


jimklimov:
Seems BSD make dislikes the change.

Issue #2510, opened 2024-07-05
CI: Recipes about `ChangeLog` regularly fail in Linux+mingw builds


jimklimov:
Over the past months there have been several efforts and PRs to improve the NUT recipes (e.g. #2366, #2360), and in particular the generation of `ChangeLog` file (but only if it is older than Git metadata) and subsequent parsing into `ChangeLog.adoc` and HTML/PDF variants. On weaker CI systems the cycle can take 10 minutes, so a number of hoops are jumped through to avoid needless rebuilds of these files (and also aiding `nut-website` builds which generate its deliverables).

Currently this is handled by `Makefile.am` (to allow `make ChangeLog` in the root dir) and `docs/Makefile.am` (to handle PDF/HTML renditions and their prerequisites up the chain), and hoping for interaction between copies of `make` implementations that have a "job server" support to avoid parallel attempts to build the thing. There were earlier mishaps with several copies of a parser script writing into the same file, causing invalid content and markup; now the writes should go into temporary files and then be atomically renamed, to avoid at least that failure mode.

In the recent weeks, NUT CI farm occasionally fails the Linux+mingw builds (only them, and not always) indicating that the interim `ChangeLog.xml` (from `asciidoc`) did not pass structural `xmllint` checks. Either something misfires about not-writing to same file from several builds (job server issue?) or we indeed have some invalid characters in section titles (was a problem earlier) or whatever; but then it would fail every time?..

````
...
Using still-valid ChangeLog file generated earlier from same revision of Git source metadata in '.git'
a2x: ERROR: "xmllint" --nonet --noout --valid "/dev/shm/jenkins-nutci/nut_nut_PR-2480@2/scripts/Windows/nut_build/docs/tmp/html-single.ChangeLog.html.983992/ChangeLog.xml" returned non-zero exit status 4

mv: cannot stat './tmp/html-single.ChangeLog.html.983992/ChangeLog.html': No such file or directory
make[2]: *** [Makefile:1389: ChangeLog.html] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [Makefile:979: all-recursive] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [Makefile:802: all-recursive] Error 1
````

jimklimov:
Of 3 similar cross-building agents, overnight reproduced this issue on one with a build in `/dev/shm`. So maybe filesystem type (timestamp precision, locking, caching...) plays a role?

[...]

PR #2531, opened 2024-07-12
CI: Sanity-check generated `ChangeLog.html` with more attention to detail

jimklimov:
Outcome of part of investigation for issue #2510

Apparently we can end up with valid-looking but contentually wrong documents. Possibly `asciidoc` or `a2x` gives up generating the XML intermediate file (the prepared asciidoc text source is huge, several megabytes and thousands of sections) and we get just a beginning part of the change log, or we still have some clashes from parallel runs corrupting each other's files (should have been solved earlier in the year, but...)

Anyhow, we can try to catch and prevent delivery of generated pretty documents (starting with the simplest case of single HTML) which do not mention first, second and last section titles present in the `ChangeLog` (generated from git) exactly twice (table of contents and section itself).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ Makefile.in
 /ltmain.sh
 /missing
 /test-driver
+*-contentchecked
 *-spellchecked
 *-prepped
 *.usage-report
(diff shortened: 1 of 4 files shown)


jimklimov:
So far generation fails (per se, or per content) on one agent quite consistently, regardless of filesystem involved. Another that should be identical (just smaller and slower, and with an older kernel under same container image) passes well.

Trying a hack to retry ChangeLog HTML (single) generation once if it failed, checking its contents right after generation.

Issue #2522, opened 2024-07-09
Want to implement parallel `nut-scanner` scanning support on MacOS


jimklimov:
As noted in #2511, `sem_init()` is not implemented on MacOS; some other methods should be used. See e.g.
* https://stackoverflow.com/questions/1413785/sem-init-on-os-x
* https://stackoverflow.com/questions/641126/posix-semaphores-on-mac-os-x-sem-timedwait-alternative (sample fallback implem available)

PR #2539, opened 2024-07-17
Implement parallel `nut-scanner` scanning support on MacOS, fix NUT build on Solaris 8

jimklimov:
Closes: #2522

Detect dysfunctional (on MacOS) `sem_init()` (unnamed) and functional `sem_open()` (named) semaphore support.

Update `nut-scanner` to work with either technology.

Tested on MacOS and Linux to do successfully run parallel scans.

Not quite sure what should happen if `sem_init()`/`sem_open()` fails. This code drops the `max_threads_scantype` to 0, to avoid probing a `NULL` named semaphore pointer. Currently there are no similar provisions for the global limiting semaphore (to skip if null).


jimklimov:
Previous build stumbled on some of the issues solved by PR #2538 (and this one would solve some issues with the new MacOS agent that would preclude other PRs from getting built), so merged that and #2537 branches into this one.

Probably merging it later (after tests pass) would auto-close those too.

jimklimov:
While testing `SEMLIBS` definition fixes (with `-lrt` additionally needed for Solaris, and `-lpthread` explicitly for some of BSDs), found that Solaris 8 did not build some recent NUT changes (not part of CI farm). Fixed in recent commits here.

jimklimov:
The couple of failed scenarios are due to tinkering with the MacOS build agent. No `clang-cpp` there, and attempts to symlink a program name like that end up freezing.

Issue #2541, opened 2024-07-19
Want to fix `make distcheck*` (as used with CI builds) to actually inherit the `DISTCHECK_FLAGS`


jimklimov:
As found during work on #2536, while `ci_build.sh` and `Makefile.am` jump through hoops passing around `DISTCHECK_FLAGS` with contents of `CONFIG_OPTS` bash array prepared by `ci_build.sh` and tailored to the build agent's "idiosyncrazies" (such as special CFLAGS and pkg-config path settings for alternate build systems like Homebrew on MacOS, and which `configure` itself should likely not blindly impose), the `make distcheck(-something)` build in fact proceeds according to chosen configuration pre-set in the `Makefile.am`.

It was somewhat hard to see, with the distcheck'ed config logs hidden in the middle of a wall of text, but noticed thanks to that PR (SNMP required for some of DMF scenarios, and requires special options on MacOS) and reproduced with plain `make distcheck` oriented `ci_build.sh` runs not inheriting the settings either (just not failing due to not requiring that SNMP code is built).

This task is about revising the work already done, maybe there's just some shell/make-scripting magic to tweak and score.

jimklimov:
As the `automake`-generated scenario for `distcheck` target goes:
````
...
          && $(am__cd) $(distdir)/_build/sub \
          && ../../configure \
            $(AM_DISTCHECK_CONFIGURE_FLAGS) \
            $(DISTCHECK_CONFIGURE_FLAGS) \
            --srcdir=../.. --prefix="$$dc_install_base" \
          && $(MAKE) $(AM_MAKEFLAGS) \
          && $(MAKE) $(AM_MAKEFLAGS) $(AM_DISTCHECK_DVI_TARGET) \
          && $(MAKE) $(AM_MAKEFLAGS) check \
          && $(MAKE) $(AM_MAKEFLAGS) install \
          && $(MAKE) $(AM_MAKEFLAGS) installcheck \
          && $(MAKE) $(AM_MAKEFLAGS) uninstall \
          && $(MAKE) $(AM_MAKEFLAGS) distuninstallcheck_dir="$$dc_install_base" \
                distuninstallcheck \
...
````

[...]

PR #2542, opened 2024-07-19
Fix `make distcheck*` builds to use passed `DISTCHECK_FLAGS` to sub-`configure` script

jimklimov:
Closes: #2541

Also addresses a few more small issues noted on MacOS builds.


jimklimov:
Seems the last trick is disliked by Sun make...

Issue #2344, opened 2024-03-02
DMF: Complete the use of `HAVE_XMLLINT` to decide if we want to validate DMF XMLs


jimklimov:
Follow-up from #2275

Currently there's the definition of the flag and the comment, but I see no use of it in practice.

jimklimov:
The presence of the tool is part of `WITH_VALIDATE_DMF_SNMP` definition (part of `scripts/DMF/Makefile.am` logical structure). So this issue is about more clearly using the automake flag `HAVE_XMLLINT` instead of shell `test -n "${XMLLINT}"` checks.

PR #2544, opened 2024-07-22
`configure.ac`: use `HAVE_XMLLINT` (automake condition flag) directly…

jimklimov:
…to decide about DMF checks

Closes: #2344


jimklimov:
OOM fault on a build agent, otherwise good.

PR #2553, opened 2024-07-24
m4/nut_check_aspell.m4: avoid `exec -- 2>&1` trick, not all shells like it

jimklimov:
Just what it says on the tin.


jimklimov:
A rebased variant is merged in the above commit.

Issue #704, opened 2019-06-17
Lost communication with UPS flood syslog


tomiisp:
Hi,

I get this issue second time, nutups lost communication with UPS (via USB/Serial cable) and nut tools and syslog start eating all 4 cores (cpu quickly reach temperature 78C ), it produce huge log file (my poor SD card...) and it produce about **4500lines in log per second** !
Entries in log looks:
```
Jun 17 22:00:08 iotgwpc2 apcsmart[1285]: Warning: excessive comm failures, limiting error reporting

Jun 17 22:00:08 iotgwpc2 apcsmart[1285]: Communications with UPS lost: serial port write error: 1379(smartmode): Input/output error

Jun 17 22:00:08 iotgwpc2 apcsmart[1285]: message repeated 9 times: [ Communications with UPS lost: serial port write error: 1379(smartmode): Input/output error]
```

That USB/serial is only temporary solution, later UPS will be connected directly to onboard UART but this is insane amount of error messages and rate. 
Is this a bug or there is a option to limit this error messages ?

Orange PI PC2 - Armbian 4.19.38-sunxi64 #5.86 SMP
Network UPS Tools - UPS driver controller 2.7.4

/Tomi

fatbasstard:
Hitting the same behaviour now, any progress on this?

jimklimov:
I am not aware of anyone addressing this specifically, so probably fair to say it is a bug, and probably it is still present. Tested PRs for throttling the message emission (maybe slower backoff to retry connecting?) are welcome.

olicooper:
Just had the same happen to me on a Raspberry Pi. Filled my 250GB SSD which subsequently made the Home Assistant database get corrupted. No way of catching it that quickly since it happened while I was sleeping. I'm not happy about this at all.

Any solution or workaround to this? I've just disabled nut for now.

nnov4k:
Hello!
i have some problem, is any ideas, how to fix syslog flooding by apcsmart?

jimklimov:
Not that I'm aware of, nobody posted related PRs or investigation notes...

nnov4k:
baaaad :(
i made a script, which kills stupid driver than UPS deads... this fix a system stability, but problem remains...

ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="067b", ATTR{idProduct}=="2303", MODE="0660", GROUP="nut", RUN+="/bin/nut-switch-restart.sh"
[...]

PR #2564, opened 2024-07-29
Fix log flood from `apcsmart`

jimklimov:
Follows up from investigation at https://github.com/networkupstools/nut/issues/704#issuecomment-2254748943 and if that analysis was correct, stands a chance to alleviate or even fix that problem. Also revises memory-safety (avoiding unbound `strcpy()`) -- not sure it was a cause for this problem, but was not a right thing to leave in the code-base either.

@tomiisp @fatbasstard @olicooper @nnov4k @desertwitch : Better late than never: would you be able to try a custom build of NUT following https://github.com/networkupstools/nut/wiki/Building-NUT-for-in%E2%80%90place-upgrades-or-non%E2%80%90disruptive-tests to see if it handles the devices better? A couple of things I'd be interested in is if it any communications are broken by possibly truncated strings, and if the data-stale happens soon after a driver startup (NUT coding issue likely) or takes a while to appear (HW issue) - and newly, if the driver tries and succeeds to reconnect (would try every 60 loops).

For the git checkout, use this PR's source branch:
````
:; git clone https://github.com/jimklimov/nut -b issue-704 nut
:; cd nut
...
````

UPDATE:  One more thing of note: in `apc_read()` there are timeouts for the `select()` from the `upsfd`, different for varying `flags` bits. Maybe those are insufficient if the controller on the UPS is old and clocked out of spec, or the host (or its USB bus) is busy context-switching and traffic-sharing, and does not deliver the message in time?

If your tests with this fix still see the UPSes in peril, whether trying to recover like this or actually recovering, please check what happens just before that data-stale situation (is there a broken read, perhaps one timed out?) and if setting longer timeouts in `apc_read()` in your custom NUT build would help?


jimklimov:
NOTE: Disregard the aborted recent CI builds, mis-pushed branch was reverted.

jimklimov:
Converting to draft while this is being tested, so NUT CI won't rebuild it in vain against newer target branch as it evolves.

jimklimov:
Gentle bump. So many people complained about the issue, is anyone still interested in testing a prospective fix? :)

jimklimov:
The one failed scenario is a hiccup on build farm, PRed code seems still ok.

jimklimov:
Hm, got a random hiccup on a build agent. :\

Issue #2591, opened 2024-08-13
RFE: SHM and/or FUSE to access nut-client information


jimklimov:
In the 42ITy project there was a component to represent NUT and other readings as a virtual file system akin to `/dev/shm` for easier uniform access from other programs (and using entity names mapped to conform to tat project's standards).

This issue is about embracing a similar idea for NUT itself, to either have `upsd` represent an `shm` view for the local system (may require several OS-dependent implementations), and/or to make use of FUSE as an arguably more portable solution to wrap the abilities of `libnutclient` as a pure userland application (a mix of `upsc`, `upsrw`, `upscmd`) that might be able to use `upsmon.conf` (or a similar file) for privileged access to remote devices.
* consider R/W from `root`, maybe other user names?
* https://www.kernel.org/doc/html/latest/filesystems/fuse.html
* https://en.wikipedia.org/wiki/Filesystem_in_Userspace

Sounds like a fun project, design and PRs would be welcome.

CC FYI: @arnaudquette-eaton

PR #2592, opened 2024-08-13
Introduce a simple PoC for NUT readings seen in FUSE file system

jimklimov:
Take advantage of the simple option in #2591 RFE, with https://github.com/vi/execfuse ("shell FUSE binding" project) successfully used for a simple and inefficient wrapper around `upsc`.

Tested on Linux and MacOS


aquette:
Interesting @jimklimov 
But to me, the real disruptive next step is still smartnut, as detailed in the pieces of specs I've started, and as started decades ago with the HAL connector. Pure async / event based, direct connection to the consuming busses (dbus, mqtt for HA and others, ...), ...

jimklimov:
Yeah, this one is in no way fit for production, just was a fun thought experiment turned into a code experiment after I told people about FUSE during lunch same day :)

A client similar to `upsmon` (maybe an optional new facet of it?) that would cache the info efficiently etc. may be another matter, especially for people who currently have to parse `upsc` outputs themselves e.g. for good old MRTG graphing or `snmpagent` representation....

PR #2606, opened 2024-08-27
drivers/main.c: do not overwrite set allow_killpower flag with defaults

### LINKED ISSUE #2605: ups.conf: allow_killpower flag has no effect
I'm trying to get the `allow_killpower` flag of `ups.conf` to work with a user, but we can't get it active no matter what.
Despite all our tries, the `driver.flag.allow_killpower` attribute shows as `0` and the `driver.killpower` command has no effect.

```
Aug 27 19:40:21 BACKUP usbhid-ups[8090]: [D2:8090] send_to_one: sending SETINFO driver.flag.allow_killpower "0"
Aug 27 19:40:21 BACKUP usbhid-ups[8090]: [D2:8090] send_to_one: sending SETFLAGS driver.flag.allow_killpower RW NUMBER
```

The flag seems to get set here:
https://github.com/networkupstools/nut/blob/f31b1ebf76edd7bdb125ab5de91923c6d6da2a71/drivers/main.c#L923-L931

Is it possible this later line overwrites the previously set flag with what seems to be a default?
https://github.com/networkupstools/nut/blob/f31b1ebf76edd7bdb125ab5de91923c6d6da2a71/drivers/main.c#L2684-L2686

It seems to be done a bit differently for other parameters (checking against a previously set value), an example:
https://github.com/networkupstools/nut/blob/f31b1ebf76edd7bdb125ab5de91923c6d6da2a71/drivers/main.c#L2645-L2647

desertwitch:
only apply the defaults for `driver.flag.allow_killpower` when `allow_killpower` flag was not already set.
current state applies defaults regardless of this, overwriting a previously set `allow_killpower` flag with `0`.
fixes #2605 , tested and working with APC device using `usbhid-ups` on Slackware 15.0 w/ gcc 11.2.0.

### DIFF
--- a/drivers/main.c
+++ b/drivers/main.c
@@ -2681,7 +2681,9 @@ sockname_ownership_finished:
 			upslogx(LOG_WARNING, "Running as foreground process, not saving a PID file");
 	}
 
-	dstate_setinfo("driver.flag.allow_killpower", "0");
+	if (dstate_getinfo("driver.flag.allow_killpower") == NULL)
+		dstate_setinfo("driver.flag.allow_killpower", "0");
+
 	dstate_setflags("driver.flag.allow_killpower", ST_FLAG_RW | ST_FLAG_NUMBER);
 	dstate_addcmd("driver.killpower");
 
--- a/drivers/main.c
+++ b/drivers/main.c
@@ -2681,6 +2681,7 @@ sockname_ownership_finished:
 			upslogx(LOG_WARNING, "Running as foreground process, not saving a PID file");
 	}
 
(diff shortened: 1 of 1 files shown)


desertwitch:
> ❌ [Build nut 2.8.2.2094-master failed](https://ci.appveyor.com/project/nut-travis/nut/builds/50486603) (commit [a82ce2f91f](https://github.com/networkupstools/nut/commit/a82ce2f91f) by @desertwitch)

Seems a timeout was exceeded here, rather than something failing build-wise.

Issue #2396, opened 2024-04-13
Use `upsdebug_with_errno()` instead of a few cases calling `upsdebugx(..., errno, strerror())` explicitly


jimklimov:
We have a nice helper method and a handful of places that can be made more simple with this change.

Single-line hits are easy to find with `git grep`:
````
:;  git grep -E upsdebugx'.*(strerror|errno)' | grep -v _strerror
clients/upssched.c:                     upsdebugx(6, "read() from fd %d returned %" PRIiSIZE " (bytes): '%c'; errno=%d: %s",
clients/upssched.c:                                     upsdebugx(4, "read() from fd %d returned 0; errno=%d: %s",
drivers/blazer_ser.c:           upsdebugx(3, "send: %s", ret ? strerror(errno) : "timeout");
drivers/blazer_ser.c:           upsdebugx(3, "read: %s", ret ? strerror(errno) : "timeout");
drivers/main.c:                         upsdebugx(1, "Socket dialog with the other driver instance: %s", strerror(errno));
drivers/main.c:                 upsdebugx_with_errno(1, "Socket dialog with the other driver instance");
drivers/riello_usb.c:                   upsdebugx(3, "USB: Send_USB_Packet: send_usb_packet, err = %d %s ", err, strerror(errno));
drivers/riello_usb.c:                   upsdebugx(3, "USB: Send_USB_Packet: send_usb_packet, err = %d %s ", err, strerror(errno));
drivers/riello_usb.c:           upsdebugx(3, "USB: Get_USB_Packet: send_usb_packet, err = %d %s ", err, strerror(errno));
````

(Note we have quite a few helpers like `nut_usb_strerror`, `shut_strerror`, `modbus_strerror`... which do not easily fall into this cleanup's targets, but might benefit from similar helpers).

PR #2631, opened 2024-09-15
clients/, common/, drivers/: refactor debug/log strerror to helper funcs

desertwitch:
had some time to kill on public transportation and did some more detective work, refactors all found occurrences of non-specific `strerror` functions into their respective helper functions ``upslog_with_errno()`` and `upsdebug_with_errno()`, addresses #2396


desertwitch:
Whoops, missed that `%s`, interesting my system did not complain...

common/common.c
desertwitch, line 1: @jimklimov : On second glance these changes in `common.c` probably will not work, as they do not use `errno`. Is there a good way to work around that or just drop these changes from the PR?

desertwitch: After some more pondering came to the preliminary conclusion that it doesn't make much sense to use a `errno`-specific helper function for something that uses a different `errnum` to `strerror`, and is likely also outside of the scope of #2396, so I've decided to revert these changes for the time being. Feel free to comment back, if you/the CI are happy will happily squash & rebase to clean up the additional (unneeded) commits before any merges.

drivers/riello_usb.c
```diff
@@ -143,7 +143,7 @@ static int Send_USB_Packet(uint8_t *send_str, uint16_t numbytes)
 		err = usb_bulk_write(udev, 0x2, (usb_ctrl_charbuf) USB_buff_pom, 8, 1000);
 
 		if (err < 0) {
-			upsdebugx(3, "USB: Send_USB_Packet: send_usb_packet, err = %d %s ", err, strerror(errno));
```
jimklimov, line 146: I think the 3 hits in this file are not right: `err` is the return from `usb_bulk_write()` and probably returns the library's codes. While OS-based `errno` (hence `upsdebug_with_errno()` may add interesting context to troubleshooting, the acute problem would be diagnosed by `nut_usb_strerror(err)`. The original code was half-wrong too though :)

jimklimov (changes requested):

drivers/riello_usb.c
desertwitch: You're right we shouldn't drop the `err` here, was a bit over-eager on my part, suggestion:
`upsdebugx(3, "USB: Send_USB_Packet: send_usb_packet, err = %d %s ", err, nut_usb_strerror(err));`
or...
`upsdebug_with_errno(3, "USB: Send_USB_Packet: send_usb_packet, err = %d %s ", err,  nut_usb_strerror(err));`
for also including the OS-based `errno` although probably kind of irrelevant here?

(applied to the other occurrences as well)

jimklimov: I think the latter does make sense, to add context maybe (e.g. some generic error in lib and perms in OS)?..

Perhaps likewise in other USB code, if there still are cases...

desertwitch: Good thought, will amend accordingly and look for similar occurrences elsewhere while at it.

jimklimov (approved):

jimklimov:
@desertwitch : what's with the force-pushes, they seem to bring no new diff... rebases over current target? typos in commit comments?

Should I wait before merging? :)

desertwitch:
> @desertwitch : what's with the force-pushes, they seem to bring no new diff... rebases over current target? typos in commit comments?
> 
> Should I wait before merging? :)

Just squashing/cleaning up before merge, good to go now.

PR #2647, opened 2024-09-30
FTY branch remerge from master/dev as of 2024-09-30

jimklimov:
Minimal reconcilement of the two codebases since last sync in #2560


jimklimov:
Agent problem; macos worker fails timer ("OLOBOLOB") test in `make check-NIT` sometimes.

Issue #2616, opened 2024-09-03
Please document how to run with LIBUSB_DEBUG=x or enable libusb debug output with nut debug options


tormodvolden:
To investigate USB issues it is important to see the debug output from libusb. Either the env var LIBUSB_DEBUG must be set to e.g. 4, or the program can call libusb_set_debug() to set debug level per libusb context.

jimklimov:
As a quick shot with existing builds - does it suffice to export the envvar in USB-capable driver units/init-scripts (commonly via `nut.conf`)?

For later versions, I'd envision something like what's done in recent years for `nut_debug_level` via `DEBUG_MIN` settings and real-time protocol settings to dial the knob in-vivo.

"Per context" here maps to one process's usage of libusb (from init to close), right?

tormodvolden:
Yes, if there is a mechanism to set environment variables when executing the driver binaries, that will work.

Setting the env var on the command line when launching a program is technically easy but there are syntax differences between operating systems (or rather their shells), and when things e.g. run via systemd it becomes more difficult for many end-users.

Per-context refers to the libusb concept of context. You can have multiple libusb contexts in one process without them interfering with each other. They will for instance have their own list of discovered devices and their own set of callbacks for events, their own debug level and so forth. This makes sense e.g. for a larger program with plugins or modules that do USB but operate independently.

jimklimov:
Got it, thanks!

jimklimov:
For starters, linking docs to https://libusb.sourceforge.io/api-1.0/group__libusb__lib.html#ga2d6144203f0fc6d373677f6e2e89d2d2

masterwishx:
`exported LIBUSB_DEBUG = 4` is not working yet ?

jimklimov:
For environment variable setup through `nut.conf` file sourced for service startup (including drivers as a service - but not used directly as a program nor via `upsdrvctl`), either way - for a shell interpreter you would

````
export LIBUSB_DEBUG=4
````

Not `exported`, no spaces around `=` ...

masterwishx:
> For environment variable setup through `nut.conf` file sourced for service startup (including drivers as a service - but not used directly as a program nor via `upsdrvctl`), either way - for a shell interpreter you would
> 
> ```
> export LIBUSB_DEBUG=4
> ```
> 
> Not `exported`, no spaces around `=` ...

In nut plugin setup config all settings with spaces, but not a problem I can check both with and without spaces.
[...]

PR #2649, opened 2024-10-02
Support `LIBUSB_DEBUG=NUM` setting in `ups.conf`

jimklimov:
Addresses part of #2616

A quick hack for that issue, to `setenv()` the envvar during configuration parsing, so hopefully the library would see it:

````
:; ./drivers/usbhid-ups -a eco650 -x LIBUSB_DEBUG=qqq -d 1
Network UPS Tools 2.8.2.1082.2-1084-g0520091ac (development iteration after 2.8.2) - Generic HID driver 0.57
USB communication driver (libusb 1.0) 0.49
WARNING : Invalid LIBUSB_DEBUG value found in ups.conf for the driver, defaulting to 4
Enabling LIBUSB_DEBUG=4 from ups.conf
[timestamp] [threadID] facility level [function call] <message>
--------------------------------------------------------------------------------
[ 0.000011] [001a9956] libusb: debug [libusb_init] created default context
[ 0.000019] [001a9956] libusb: debug [libusb_init] libusb v1.0.24.11584
...
ups.timer.shutdown: -1
ups.timer.start: -1
ups.vendorid: 0463
[ 7.381093] [001aa8d0] libusb: debug [libusb_close]
[ 7.381098] [001aa8d0] libusb: debug [usbi_remove_event_source] remove fd 8
[ 7.381120] [001aa8d0] libusb: debug [libusb_exit]
[ 7.381123] [001aa8d0] libusb: debug [libusb_exit] destroying default context
[ 7.381126] [001aa8d0] libusb: debug [libusb_handle_events_timeout_completed] doing our own event handling
[ 7.381129] [001aa8d0] libusb: debug [handle_events] event sources modified, reallocating event data
[ 7.381134] [001aa8d0] libusb: debug [usbi_wait_for_events] poll() 2 fds with timeout in 0ms
[ 7.381137] [001aa8d0] libusb: debug [usbi_wait_for_events] poll() returned 0
[ 7.381140] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 6.1
[ 7.381143] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 5.1
[ 7.381146] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 4.1
[ 7.381149] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 3.7
[ 7.381153] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 3.6
[ 7.381157] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 3.1
[ 7.381160] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 2.1
[ 7.381165] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 1.4
[ 7.381167] [001aa8d0] libusb: debug [libusb_unref_device] destroy device 1.1
[ 7.381170] [001aa8d0] libusb: debug [usbi_remove_event_source] remove fd 7
[ 7.381177] [001aa8d0] libusb: debug [usbi_remove_event_source] remove fd 6
[ 7.381198] [001aa8d1] libusb: debug [linux_udev_event_thread_main] udev event thread exiting
````

A cleaner solution would be to call `libusb_set_debug()` specifically, when initializing a context for device communications, but needs some more careful coding :) so can be a follow-up sometime later.

FYI : CC @masterwishx @tormodvolden

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -191,6 +191,11 @@ https://github.com/networkupstools/nut/milestone/11
    of an UPS interface on a composite USB device or when looking at devices
    with non-default interface/endpoint/config numbers. [PR #2611]
 
+ - USB drivers should now accept a `LIBUSB_DEBUG=INTEGER` setting in `ups.conf`
+   (as well as an environment variable that can be generally set via `nut.conf`
+   or service unit methods or init script), to enable troubleshooting of LibUSB
+   itself. [issue #2616]
+
  - Introduced a new driver concept for interaction with OS-reported hardware
    monitoring readings. Currently instantiated as `hwmon_ina219` specifically
    made for Texas Instruments INA219 chip as exposed in the Linux "hwmon"
--- a/conf/ups.conf.sample
+++ b/conf/ups.conf.sample
@@ -82,8 +82,8 @@
 # Set maxretry to 3 by default, this should mitigate race with slow devices:
 maxretry = 3
 
-# These directives can be set outside and inside a driver definition, with
-# slightly different meanings per context:
+# These directives can be set outside and inside a driver definition,
+# sometimes with slightly different meanings per context:
 #
 # maxstartdelay: OPTIONAL.  This can be set as a global variable
 #                above your first UPS definition and it can also be
@@ -110,6 +110,13 @@ maxretry = 3
 #              background running mode. If both the global and driver level
 #              `debug_min` are set, the driver-level setting takes precedence.
 #              Command-line option `-D` can only increase this verbosity level.
+#
+# LIBUSB_DEBUG: OPTIONAL.  For run-time troubleshooting of USB-capable NUT
+#                drivers, you can specify the verbosity of LibUSB specific
+#                debugging as a numeric value such as `4` ("All messages
+#                are emitted").  Should not have any practical impact on
+#                other NUT drivers. For more details, see the library's
+#                documentation at e.g. https://libusb.sourceforge.io/api-1.0/
 #
 # user, group: OPTIONAL. Overrides the compiled-in (also global-section,
 #                when used in driver section) default unprivileged user/group
--- a/docs/man/nut_usb_addvars.txt
+++ b/docs/man/nut_usb_addvars.txt
@@ -125,18 +125,21 @@ As a rule of thumb for `usb_hid_desc_index` discovery, you can see larger
 `wDescriptorLength` values (roughly 600+ bytes) in reports of `lsusb` or
 similar tools.
 
-[NOTE]
-======
+*LIBUSB_DEBUG =* 'INTEGER'::
+
 Run-time troubleshooting of USB-capable NUT drivers can involve not only
 raising the common NUT debug verbosity (e.g. using the `DEBUG_MIN` setting
 in linkman:ups.conf[5] or protocol commands to change the `driver.debug`
 value), but may also benefit from LibUSB specific debugging.
-
-For the latter, currently you would have to export the environment variable
-`LIBUSB_DEBUG` before starting a NUT driver (may be set and "exported" via
-linkman:nut.conf[5]), to a numeric value such as `4` ("All messages are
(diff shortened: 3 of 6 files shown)


masterwishx:
checked , working fine for nut plugin in Unraid.
the question is libusbdebug should work only befor?

` usbhid-ups[29172]: Startup successful`

jimklimov:
Good question. The current implementation sets the environment variable for this process while reading configuration from `ups.conf` or CLI arguments.

*Maybe* it can also change the value during a reload signal/command processing (if it changed in `ups.conf`), that would be nice but I did not specifically code for that. I also did not add protocol commands to manipulate this at run-time like `debug` can be changed.

I also see now I've only tested it with a dump mode, which does not `fork()`. So if you don't see the libusb debug output in a normally running driver (which forks), *that* may be a bug of this PR and we need to have an equivalent of `export VARNAME` somewhere here.

jimklimov:
Per https://stackoverflow.com/a/17929641/4715872 discussion,

> Child processes created with fork, will inherit the current processes environment definitions, thus your changes and additions as well.

So there should be no need for a separate `export` equivalent - this is rather shells' construct to prepare the env when they fork off each process vs. whatever the shell interpreter uses itself.

masterwishx:
> Good question. The current implementation sets the environment variable for this process while reading configuration from `ups.conf` or CLI arguments.
> 
> _Maybe_ it can also change the value during a reload signal/command processing (if it changed in `ups.conf`), that would be nice but I did not specifically code for that. I also did not add protocol commands to manipulate this at run-time like `debug` can be changed. I would not object to PRs with such exercises, but not likely would have time to make them myself (inspiration can be taken from earlier PRs that added reload and socket-protocol modification support for `debug`/`debug_min` settings - explore the trail from approx #1914).
> 
> I also see now I've only tested it with a dump mode, which does not `fork()`. So if you don't see the libusb debug output in a normally running driver (which forks), _that_ may be a bug of this PR and we need to have an equivalent of `export VARNAME` somewhere here.

Using nut plugin for Unraid, i added `LIBUSB_DEBUG = 4` in ups.conf .

masterwishx:
asl ofound some libusb debug present after line i wrote above...

jimklimov:
So to be clear, you've added tge setting and it had effect all the lifetime of the driver (regular update loop and all)?

Then it works as it should (run-time modification improvements speculated above being separate goals) :)

masterwishx:
> So to be clear, you've added the setting and it had effect all the lifetime of the driver (regular update loop and all)?
> 
> Then it works as it should :) (run-time modification improvements speculated above being separate goals)

No not at all lifetime , when debugmin = default also soemtime when =6 : 

only befor `usbhid-ups[10044]: Startup successful`

but today i found it was also debug + libusb debug but then is gone from list . i can send debug file if you interested

masterwishx:
[nut debug log 9.10.24.txt](https://github.com/user-attachments/files/17311337/nut.debug.log.9.10.24.txt)

jimklimov:
Maybe I got it: when we fork, stderr is normally detached (this can be maybe tweaked when dev/testing), and usbdebugx posts to syslog. But libusb's own debug probably does not, or needs more settings if it can?

I can suggest avoiding forking, e.g. `-D` by default keeps NUT daemons not detaching from foreground...

tormodvolden:
Generally when you fork, the stderr fd is inherited. libusb logs to stderr by default, unless built with --enable-system-log.

jimklimov:
Added this feature to https://github.com/networkupstools/nut/wiki/Changing-NUT-daemon-debug-verbosity

tormodvolden:
Note that the libusb API link on the that wiki page is broken.

jimklimov:
Thanks, fixed! Wringing it through GitHub renderer took a while (double underscores treated as HTML `<em>` tags even if part of `link:`   markup otherwise (facepalm).

PR #2658, opened 2024-10-15
introduce further handling and notifiers for ALARM status

### LINKED ISSUE #2657: upsmon: ALARM status should be handled/handleable
This is more of an enhancement request, but I think it's also a bit of an issue considering alarms can have consequences.

In the current codebase various drivers utilize `ups.alarm` for raising device-/manufacturer-specific alarms. There is additional common code in `dstate.c`, evaluating if such `ups.alarm` variables are preset and, if so, set a status `ALARM`:
https://github.com/networkupstools/nut/blob/539d9709a5560cd1ea113f1d826e25f9c00ac674/drivers/dstate.c#L1696-L1700

Right now, in an unattended setting, specifically checking the UPS status and UPS variables for active alarms seems needed. This is because at present `upsmon` seems to have no specific code handling an `ALARM` status and hence it cannot be acted on with `upssched` or other means of notification processing. As the drivers seem to already be handling alarms one way or another, and this is unified by `dstate.c` with a status `ALARM` as common denominator, `uspmon` code seems to be the remaining piece of the puzzle to actively be able to handle and act on alarms and an associated `ALARM` status of the UPS.

We should introduce `upsmon` code with a flag `ST_ALARM` and associated `NOTIFY_ALARM` & `NOTIFY_NOALARM` for the user to be able to act on `ALARM`. Such code residing in `upsmon` would make sense to process an `ALARM` status (as set by `dstate.c`) as a common denominator for different ways of handling alarms within the UPS drivers themselves. UPS alarms can raise potentially grave issues of the UPS and users should be able to configure e.g. `upssched` tasks or notifiers accordingly:
```
NOTIFYFLAG ALARM EXEC
NOTIFYFLAG NOTALARM EXEC
```
`ALARM` being when the `ALARM` status is set on the UPS
`NOTALARM` being when the `ALARM` status is no longer set on the UPS

Ideally this could be accompanied by respective SYSLOG warning messages with the specific alarm message, both when set and later removed again by the UPS driver, although this might require more advanced tracking of the individual alarms and is possibly overkill as opposed to handling a simple `if (flag_isset(ups->status, ST_ALARM)`.
```
UPS %s is now in status ALARM (first-time) with message: %s.
UPS %s is no longer in status ALARM, all alarms have cleared.
```

desertwitch:
As outlined in #2657, `dstate.c` sets `ALARM` as a common denominator status when encountering `ups.alarm` variables reported through the different device-/manufacturer-specific implementations within the UPS drivers.  Previously a user had to either rely on audible alarms by the UPS or (in an unattended setting) periodically check the UPS status and variables manually to be informed of potentially severe UPS device alarm states/conditions.

This pull request introduces `ALARM` status handling in `upsmon` as the remaining piece of the puzzle to effectively act upon UPS device alarms as raised by the UPS drivers. This includes notifiers `ALARM` and `NOTALARM` for when the status is encountered and dissipates, respectively, allowing the user to now configure notifications and tasks e.g. through `upssched`. 

By default `upsmon` will treat any encountered `ALARM` status as (potentially) severe enough to warrant an UPS being promoted to critical in a no-communication situation. This can, however, be unwarranted due to there being no common standard for what constitutes an alarm. In fact mundane alarm states were previously observed with UPS devices in HE/ECO mode, which could lead to unwarranted shutdown situations on minor connection hiccups. For this reason a `0|1` setting `ALARMCRITICAL` was introduced into `upsmon.conf`, for such users to be able to prevent `upsmon` from treating the `ALARM` status as overly severe when it's not.

As for the default behaviour (of considering such an UPS as volatile) erring on the side of caution seemed to be reasonable, as e.g. an alarm for some failing circuitry and the affected UPS later dropping offline should be concern enough for the UPS to be considered immediately critical and the system pre-emptively being shutdown e.g. by reduction/depletion of `MINSUPPLIES`.

fixes #2657


desertwitch:
For further improvements down the road, perhaps we should also consider tracking `ups.alarm` variables and produce logs for newly registered alarms and later cleared again alarms respectively (although this would be more suitable in direction of `dstate.c` and/or the drivers, rather than within `upsmon`, but it would be nice to have a unified/common denominator solution in spite of different driver-side implementations... similar to as done with the `ALARM` status, so perhaps in `dstate.c`?). This would complement the `upsmon` side of handling things well, where the status is handled but not the individual alarm messages/variables. But this is, I think, outside the scope of this `upsmon`-specific PR.

jimklimov:
Thanks, I hope to have some time to look at it this week. So far it started very intensively.

desertwitch:
Thanks Jim, it's no stress at all, I've taken the liberty to broaden the scope here a bit where it made sense. 😉

PR #2637, opened 2024-09-23
EATON HID: add missing usages and paths

masterwishx:
Adding HE/ECO mode, Bypass On/Off for Eaton USB UPS Online Models + Adding Missing usages and paths

Continue work for #2636
Fixes #2485
Ref #2495 

Added: 

```
Variables: 

battery.charger.status: discharging
device.usb.version: 01.29
input.eco.switchable: normal
outlet.1.designator: AC OUTPUT
outlet.1.protect.status: protected
input.transfer.forced: enabled
input.eco.switchable: ECO
input.bypass.switch.on: off
etc ...

R/W variables:

input.eco.switchable
input.bypass.switch.on
etc ..

Commands: 

ecomode.disable - Take the UPS out of High Efficiency (aka ECO) mode
ecomode.enable - Put UPS in High Efficiency (aka ECO) mode
essmode.disable - Take the UPS out of Energy Saver System (aka ESS) mode
essmode.enable - Put UPS in Energy Saver System (aka ESS) mode
bypass.start - Put the UPS in bypass mode
bypass.stop - Take the UPS out of bypass mode

```

PR #2660, opened 2024-10-16
EATON HID: add missing usages and paths (addon) - ABM Statuses

masterwishx:
ABM Statuses + Adding Missing usages and paths

See also: https://www.eaton.com/content/dam/eaton/products/backup-power-ups-surge-it-power-distribution/backup-power-ups/power-xpert-9395/Eaton-ABM-White-Paper-AP162001EN.pdf

Continue work/addon for #2637

Added: 

```
Variables: 

battery.charger.status: discharging
battery.charger.type: ABM
etc ...
```

Added ABM statuses for Eaton Models with next paths:

```
UPS.BatterySystem.Charger.ChargerType
UPS.BatterySystem.Charger.Status
```


masterwishx:
@arnaudquette-eaton is last commits OK for adding status ? UPS.BatterySystem.Charger.Status

arnaudquette-eaton:
> @arnaudquette-eaton is last commits OK for adding status ? UPS.BatterySystem.Charger.Status

seems good, though
1) a bit redundant (or overlapping) with UPS.BatterySystem.Charger.Mode 
2) you may not limit to charging / discharging, and may expand words instead of using the ups.status "flag" mode

masterwishx:
> > @arnaudquette-eaton is last commits OK for adding status ? UPS.BatterySystem.Charger.Status
> 
> seems good, though
> 
> 1. a bit redundant (or overlapping) with UPS.BatterySystem.Charger.Mode
> 2. you may not limit to charging / discharging, and may expand words instead of using the ups.status "flag" mode

Thanks, if all fine from statuses perspective, I will try to combine them in one function with `if check`, as I think important not change the code logic if possible...

masterwishx:
> 1. a bit redundant (or overlapping) with UPS.BatterySystem.Charger.Mode

Please look at 2 last commits combined to one  functions, should look better now.

@jimklimov can we merge it also or to merge it in eco branch?

jimklimov:
> can we merge it also or to merge it in eco branch?

I think two PRs for two goals are okay. Does this one fully include the commits of the "eco branch"? If yes, merging this one will automatically close that PR as achieved in one shot.

masterwishx:
> > can we merge it also or to merge it in eco branch?
> 
> I think two PRs for two goals are okay. Does this one fully include the commits of the "eco branch"? If yes, merging this one will automatically close that PR as achieved in one shot.

Yes it is

masterwishx:
> > can we merge it also or to merge it in eco branch?
> 
> I think two PRs for two goals are okay. Does this one fully include the commits of the "eco branch"? If yes, merging this one will automatically close that PR as achieved in one shot.

but not include conversations ?

jimklimov:
Not sure I fully understand the question. Conversations will remain in github (if they are not part of Git committed code, e.g. new comments about logic), in respective issue/PR discussions. For easier history research later, I'll add a cross-link between these.

masterwishx:
> Member

Sorry , i meaned converation like here is 8 only  , and in #2637 was 110  and also comments of the code .
but its OK if you will add links to that pr

drivers/mge-hid.c
```diff
@@ -1571,14 +1540,14 @@ static hid_info_t mge_hid2nut[] =
...
 	/* Same as the one above, but for legacy units */
 	/* Refer to Note 1 (This point will need more clarification!)
 	{ "battery.charger.status", 0, 0, "UPS.BatterySystem.Charger.PresentStatus.Used", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_enabled_legacy_info }, */
 	/* This data is the actual ABM status information */
-	{ "battery.charger.mode", 0, 0, "UPS.BatterySystem.Charger.Mode", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_mode_info }, /* needs both ? from https://github.com/networkupstools/nut/pull/2637#discussion_r1772730590 */
-	{ "battery.charger.status", 0, 0, "UPS.BatterySystem.Charger.Status", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_status_info }, /* checked on Eaton 9E Model */
+	{ "battery.charger.mode", 0, 0, "UPS.BatterySystem.Charger.Mode", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_info }, /* needs both ? from https://github.com/networkupstools/nut/pull/2637#discussion_r1772730590 */
+	{ "battery.charger.status", 0, 0, "UPS.BatterySystem.Charger.Status", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_info }, /* checked on Eaton 9E Model */
```
jimklimov: @arnaudquette-eaton : I am offline today and feel a bit vague on how the mapping tables are processed.

From this commit, it seems we can set `battery.charger.status` from several sources above (once from the same `UPS.BatterySystem.Charger.Status` source, but with `eaton_abm_enabled_info` instead of `eaton_abm_info`)... Would this line even work (or would the loop finish at first hit... and what counts as a hit - seeing a reading from the source, or specifically a non-`NULL` return (so both referenced functions have a chance to run even this way)?

masterwishx:
@jimklimov please check if now all fine?

masterwishx:
Now all fine ....

masterwishx:
after compiled i have : 

`ups.status = OL DISCHRG ECO` instead of `OL ECO`

masterwishx:
> after compiled i have :
> 
> `ups.status = OL DISCHRG ECO` instead of `OL ECO`

@desertwitch can you say why i have now this :  
image

drivers/mge-hid.c
```diff
@@ -267,27 +275,55 @@ static const char *eaton_abm_status_fun(double value)
 
 	upsdebugx(2, "ABM numeric status: %i", (int)value);
 
-	switch ((long)value)
+	if (dstate_getinfo("battery.charger.type"))
```
desertwitch, line 278: I don't really understand this `if`-section. It seems here the decision is made for two different kinds of ABM handling, but based on the existence of a variable `battery.charger.type`? Why not `advanced_battery_monitoring == ABM_ENABLED_TYPE` that we already know? Does the existence of `battery.charge.type` imply a different kind of ABM is used and can we ensure that we're not sending other devices down the wrong road if basing this on some variable existence alone? Because this seems to say if that variable exists `2 = floating` and not `2 = discharging`.

drivers/mge-hid.c
```diff
@@ -300,28 +336,60 @@ static info_lkp_t eaton_abm_status_info[] = {
...
+
 /* Used to process ABM flags, for ups.status (CHRG/DISCHRG/RB) */
 static const char *eaton_abm_chrg_dischrg_fun(double value)
 {
 	/* Don't process if ABM is disabled */
 	if (advanced_battery_monitoring == ABM_DISABLED)
 		return NULL;
 
-	switch ((long)value)
+	if (dstate_getinfo("battery.charger.type"))
```
desertwitch, line 356: Same comment as above, why base this on variable existence and not `advanced_battery_monitoring == ABM_ENABLED_TYPE`? Would this not send a device down the wrong path if it does not expose the variable yet (perhaps temporarily not)?

drivers/mge-hid.c
```diff
@@ -227,7 +227,7 @@ static const char *eaton_abm_enabled_fun(double value)
 {
 	advanced_battery_monitoring = value;
 
-	if (advanced_battery_monitoring == ABM_ENABLED_TYPE)
+	if (dstate_getinfo("battery.charger.type"))
```
desertwitch, line 230: +1 as with my other reservations about the variable existence approach deciding on which ABM mode is in effect, if `ABM_ENABLED_TYPE` does not work for the comparison here perhaps it being `4` is not the correct value?

masterwishx:
> Took a look as requested, commented a few things that did not make sense to me (but I'm by no means an expert with the UPS drivers, so take it with a grain of salt). In general I think we need to take a cautious approach to not break functionality with a broader range of other devices, in favour of making things work on one series, so better to double and triple check here. 👀
> 
> > > after compiled i have :
> > > `ups.status = OL DISCHRG ECO` instead of `OL ECO`
> > 
> > 
> > @desertwitch can you say why i have now this : image
> 
> Might be caused by some of the ABM changes introduced, I could imagine a `battery.charger.status` wrongly being set to `discharging` might lead to a `DISCHRG` status further down the line also?

in general i compiled and all fine , only this stuff and wrong statuses becouse `if (advanced_battery_monitoring == ABM_ENABLED_TYPE)` not always working as needed , so changed back to:
`if (dstate_getinfo("battery.charger.type"))`  to separate two versions of ABM Statuses one is like was in original code : 
```
UPS.BatterySystem.Charger.ABMEnable
UPS.BatterySystem.Charger.Mode
```
and added for my 9E unit but dont have ABM to check , so  
```
UPS.BatterySystem.Charger.Status
UPS.BatterySystem.Charger.ChargerType
```

desertwitch:
> 
> ```
> UPS.BatterySystem.Charger.ABMEnable
> UPS.BatterySystem.Charger.Mode
> ```
> 
> and added for my 9E unit but dont have ABM to check , so
> 
> ```
> UPS.BatterySystem.Charger.Status
> UPS.BatterySystem.Charger.ChargerType
> ```

Ok, now I understand the reasoning for checking the variable existence (and assuming the ABM mode based on its existence) better. But, perhaps, not sure if this is possible (and logical), it would be a cleaner separation (and less prone to accidental error) for the different ABM paths to be separated in the HID lookup table also? And not for both HID paths to result in `eaton_abm_status_info` and `eaton_abm_chrg_dischrg_info`? Not sure what happens if multiple HID paths (assuming all are present in a device) get translated with the same `info_lkp_t` and function - last execution wins (previous one gets overwritten)?

drivers/mge-hid.c
```diff
@@ -1282,11 +1533,21 @@ static hid_info_t mge_hid2nut[] =
...
 	/* This data is the actual ABM status information */
-	{ "battery.charger.status", 0, 0, "UPS.BatterySystem.Charger.Mode", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_status_info },
+
+	/* NOTE: 9E Model and some other using ("UPS.BatterySystem.Charger.Status" and "UPS.BatterySystem.Charger.ChargerType")
+	 * instead of ("UPS.BatterySystem.Charger.ABMEnable" and "UPS.BatterySystem.Charger.Mode") */
+
+	{ "battery.charger.status", 0, 0, "UPS.BatterySystem.Charger.Mode", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_status_info }, /* needs both ? from https://github.com/networkupstools/nut/pull/2637#discussion_r1772730590 */
+	/* Same as above but for 9E Models that using "x.Status" instead and other units */
+	{ "battery.charger.status", 0, 0, "UPS.BatterySystem.Charger.Status", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_abm_status_info }, /* checked on Eaton 9E Model */
+    { "battery.charger.type", 0, 0, "UPS.BatterySystem.Charger.ChargerType", NULL, "%.0f", HU_FLAG_QUICK_POLL, eaton_charger_type_info },
```
desertwitch, line 1549: Reading some code comments, translation order seems important here, I think the checking for `battery.charger.type` existence will always fail at least in `eaton_abm_enabled_info` (translated on L1535) because it gets translated earlier than the actual variable that is being checked for (that is translated on L1549). Still think we should probably save the state somewhere internally as done in other functions as opposed to checking variable existence via `dstate`.

desertwitch, line 1549: Reading some code comments, translation order seems important here, I think the checking for `battery.charger.type` existence will always fail at least in `eaton_abm_enabled_info` (translated on L1535) because it gets translated earlier than the actual variable that is being checked for (that is translated on L1549). Still think we should probably save the state somewhere internally as done in other functions as opposed to checking variable existence via `dstate`.

jimklimov: +1 for caching in a (`static int`?) variable, on the grounds that tree walks in dstate lookups have a cost.

The variable might be e.g. negative until initialized, and then maybe read-only once we saw a definitive indication that the device has/hasn't ABM, etc.

masterwishx:
Now we good to go ...

masterwishx:
@jimklimov Compiled all working fine : 

```
Path: UPS.BatterySystem.Charger.ChargerType, Type: Feature, ReportID: 0x26, Offset: 0, Size: 8, Value: 5
hu_find_infoval: found Constant Charge (CC) (value: 5)

Path: UPS.BatterySystem.Charger.ChargerType, Type: Feature, ReportID: 0x26, Offset: 0, Size: 8, Value: 5
Set ABM variable to disabled, charger type status: 0
ABM is disabled
Lookup [5] failed for [battery.charger.status]

Path: UPS.BatterySystem.Charger.Status, Type: Feature, ReportID: 0x26, Offset: 8, Size: 8, Value: 2
Lookup [2] failed for [battery.charger.status]
```

@arnaudquette-eaton @jimklimov @desertwitch
The question if we should use this Table if ABM not availible ? 
https://github.com/networkupstools/nut/pull/2637#discussion_r1772730590

```
Unified charger info,bitmap with :
Status on bit 0-3:
0x0: Off
0x1: Charging
0x2: Floating
0x3: On, not charging
0x4: Discharging
0x5-0xF : Reserved
Mode on bit 4-6:
0: CC
1: ABM
2-7 : Reserved
Error on bit 7:
0: Good
1: In error
```

becouse i still have path : 
`Path: UPS.BatterySystem.Charger.Status, Type: Feature, ReportID: 0x26, Offset: 8, Size: 8, Value: 2` but not `ABM Mode`

for now i disabled for Constant charge as my case , so we have `Lookup [2] failed for [battery.charger.status]`


The only problem i have now and privies compiled :  

`ups.status = OL CHRG ECO`  @jimklimov  can be related to your commits ?

desertwitch:
> 
> > > after compiled i have :
> 
> > > `ups.status = OL DISCHRG ECO` instead of `OL ECO`
> 

Before it was `DISCHRG`, are you sure this is not still related to some ABM parsing (issue)? Will check in detail later this evening and test with my 5P for the other path that's not available on 9E.

masterwishx:
> > after compiled i have :
> > `ups.status = OL DISCHRG ECO` instead of `OL ECO`
> 
> Before it was `DISCHRG`, are you sure this is not still related to some ABM parsing (issue)? Will check in detail later this evening and test with my 5P for the other path that's not available on 9E.

i think i found :  

```
Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
process_boolean_info: chrg
```

L517 `snprintf(mge_scratch_buf, sizeof(mge_scratch_buf), "%s", "chrg");`
```
```

masterwishx:
Now when i set `advanced_battery_monitoring == ABM_DISABLED` in constant charge , befor in original code was : 
`advanced_battery_monitoring == ABM_UNKNOWN`

masterwishx:
> > after compiled i have :
> > `ups.status = OL DISCHRG ECO` instead of `OL ECO`
> 
> Before it was `DISCHRG`, are you sure this is not still related to some ABM parsing (issue)? Will check in detail later this evening and test with my 5P for the other path that's not available on 9E.

Will be cool , i really need to finish it as it much out of ECO i wanted implement here :(

masterwishx:
compiled , Strange but still have `OL CHRG ECO`

masterwishx:
in this case L437 :  
Path: UPS.BatterySystem.Charger.Status, Type: Feature, ReportID: 0x26, Offset: 8, Size: 8, Value: 2
` ABM CHRG/DISCHRG legacy string status (ups.status): chrg `

masterwishx:
> > > after compiled i have :
> > > `ups.status = OL DISCHRG ECO` instead of `OL ECO`
> > 
> > 
> > Before it was `DISCHRG`, are you sure this is not still related to some ABM parsing (issue)? Will check in detail later this evening and test with my 5P for the other path that's not available on 9E.
> 
> i think i found :
> 
> ```
> Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
> process_boolean_info: chrg
> ```
> 
> L517 `snprintf(mge_scratch_buf, sizeof(mge_scratch_buf), "%s", "chrg");`


@jimklimov @desertwitch 
not sure how to resolve this?
if we accept that we dont use ABM Table in Constant Charge then we have send charge by :

  
https://github.com/networkupstools/nut/blob/387444804fdccef0b8ab2993a84f8a4332cb6758/drivers/mge-hid.c#L517
```
L517 `snprintf(mge_scratch_buf, sizeof(mge_scratch_buf), "%s", "chrg");`
```
```
Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
process_boolean_info: chrg

```

but we have https://github.com/networkupstools/nut/blob/387444804fdccef0b8ab2993a84f8a4332cb6758/drivers/usbhid-ups.c#L2327

to check for 100% charge.battery so should be no CHRG ? but somehow its not working ?!?

masterwishx:
@jimklimov @arnaudquette-eaton 

!! Finally Finished !!

compiled , now its Fine `OL ECO` and logic as was in original code for Constant charge (ABM_UNKNOWN)

The only problem exists that i checked when `battery.charge=97` we still have `OL ECO` no CHRG , but also in original code checked from main (Unraid Plugin )`nut-2.8.2-x86_64-6master.ssl11.txz` !

jimklimov:
Bumped #2637 and #2660 against recent master changes (`ALARM` support landed into same spots almost everywhere).
NOTE: #2637 is still a strict subset of #2660.

So CI will spend some time re-checking them all...

desertwitch:
Not sure which PR to put this in, but as requested the paths pulled from my Eaton 5P850I during the first information rounds. This device seems to expose the "other type" of ABM (from what I gathered here) using `battery.charger.status` (resting/floating): https://gist.github.com/desertwitch/63a79170aad129807298380179357507.

masterwishx:
> Not sure which PR to put this in, but as requested the paths pulled from my Eaton 5P850I during the first information rounds. This device seems to expose the "other type" of ABM (from what I gathered here) using `battery.charger.status` (resting/floating): https://gist.github.com/desertwitch/63a79170aad129807298380179357507. In this case the device was `OL` and `resting`.

this pr continue /include old one

masterwishx:
> Not sure which PR to put this in, but as requested the paths pulled from my Eaton 5P850I during the first information rounds. This device seems to expose the "other type" of ABM (from what I gathered here) using `battery.charger.status` (resting/floating): https://gist.github.com/desertwitch/63a79170aad129807298380179357507. In this case the device was `OL` and `resting`.

Tried to avoid same paths like in your case ...

masterwishx:
> Not sure which PR to put this in, but as requested the paths pulled from my Eaton 5P850I during the first information rounds. This device seems to expose the "other type" of ABM (from what I gathered here) using `battery.charger.status` (resting/floating): https://gist.github.com/desertwitch/63a79170aad129807298380179357507. In this case the device was `OL` and `resting`.

I will test this pr soon, will be cool if you can test also to check that all working fine Becouse you have all paths for statuses

masterwishx:
Tested working Fine now when charging

masterwishx:
@jimklimov Tested all fine.
@desertwitch please test this pr if you will have time so we can merge it

masterwishx:
Now its really Finished :) 

@arnaudquette-eaton @jimklimov 

The only one thing is : 

After i made for `ABM_UNKNOWN` Like 9E UPS (in Constant Charge) to change status by ABM Status Table 
it seems now mostly `battery.charge = 84VDC (6x14V)` , before was `83VDC (6x13.8V)`

in original code was no CHRG, DSCHR statuses working by:

https://github.com/networkupstools/nut/blob/b2f14e6fcef026d66babc07bdb19d9554fa834b3/drivers/mge-hid.c#L552
https://github.com/networkupstools/nut/blob/b2f14e6fcef026d66babc07bdb19d9554fa834b3/drivers/mge-hid.c#L526

So is it right condition or we need to use `ABM_UNKNOWN` statuses same as when ABM_DISABLED ?
like: 
https://github.com/networkupstools/nut/blob/b2f14e6fcef026d66babc07bdb19d9554fa834b3/drivers/mge-hid.c#L512

image
image

drivers/usbhid-ups.c
```diff
@@ -2392,10 +2405,25 @@ static void ups_status_set(void)
...
+		(ups_status & STATUS(DISCHRG))) {
+		status_set("OB"); /* on battery */
+	}
+
+	/* Fix (OL+CHRG) for Eaton 9E Model */
+	if ((current_charge == onlinecharge_hovercharge) &&
+	   (ups_status & STATUS(ONLINE)) && (ups_status & STATUS(CHRG))) {
+		status_set("OL");
+	}
+
```
desertwitch, line 2426: Not sure about these recent additions, forcing `OB` on any status `!OL` + `DISCHRG` would set a device in e.g. `CAL DISCHRG` to `OB` as well and possibly raise notifications or even shutdowns (?). The second addition also, hard setting a `OL CHRG` to `OL` for all UPS using `usbhid-ups` (= multiple vendors and many devices) feels invasive to me and should be a user toggle if anything - in my opinion. Some users may want to know if their UPS is in fact (over-)charging despite being at 100%. I think we should obscure the actual UPS reported statuses as little as possible and really just when it's obvious it's a mistake or useless for all users, otherwise introduce configurational toggles for such things instead. Thoughts, @jimklimov ?

jimklimov: I agree; the combo of "Fix <XXX> for <model>" without a toggle or model detection spoke for itself :\

drivers/mge-hid.c
```diff
@@ -191,9 +192,28 @@ static char		mge_scratch_buf[20];
...
+
+/* Define if we have battery.charger.mode.staus by "x.ABMEnable and "x.Mode" */
+#define			ABM_CHARGER_NO_MODE -1
+#define			ABM_CHARGER_MODE 1
+
+/* Internal flag to process battery.charger.type.status for 9E Models and others */
+static int advanced_battery_status = ABM_CHARGER_NO_STATUS;
+
+/* Internal flag to process battery.charger.mode.status by "x.ABMEnable and "x.Mode" */
+static int advanced_battery_mode = ABM_CHARGER_NO_MODE;
```
desertwitch, line 216: I think this section complicates overall code readability a lot and the definitions are misleading into thinking these are device/HID return values (which may change in the future) when actually they are only internal states -1(unknown/uninitialized)/0(disabled)/1(enabled) that are better suited for recording with only an integer (in my opinion):

```
static int advanced_battery_status = -1;
static int advanced_battery_mode = -1;
```
Later we can check if the condition for `advanced_battery_status` is fulfilled and, if yes, set `advanced_battery_status = 1` and `advanced_battery_mode = 0` and vice-versa. Also, right now -1 is the same value for unknown and disabled, so better - I think - to treat -1 as unknown and 0 as disabled (as separate states).

drivers/mge-hid.c
```diff
@@ -191,9 +192,28 @@ static char		mge_scratch_buf[20];
...
 
 /* Internal flag to process battery status (CHRG/DISCHRG) and ABM */
 static int advanced_battery_monitoring = ABM_UNKNOWN;
```
desertwitch, line 219: I know this is not your code, but maybe also better to start here with `-1` and drop the `ABM_UNKNOWN` (= -1) definition for similar reason as in other comment (separation from HID/device-returned values).

drivers/mge-hid.c
```diff
@@ -300,28 +396,99 @@ static info_lkp_t eaton_abm_status_info[] = {
...
+			if (advanced_battery_monitoring == ABM_UNKNOWN)
+			{
+				snprintf(mge_scratch_buf, sizeof(mge_scratch_buf), "%s", "!dischrg");
+				break;
+			}
+			else
+			{
+				snprintf(mge_scratch_buf, sizeof(mge_scratch_buf), "%s", "dischrg");
+				break;
+			}
```
desertwitch, line 465: Not sure about these `if (advanced_battery_monitoring == ABM_UNKNOWN)`, why are we returning that we are not charging or discharging here? Would it not be better logically to either assume ABM is on (as ABM paths are present and ABM is not overall `ABM_DISABLED`) or exit the function prematurely (as is done for `ABM_DISABLED` earlier on L307/431)?

General charging and discharging is separately handled in `eaton_abm_check_chrg_fun` so - I think - probably we shouldn't clear flags based on `ABM_UNKNOWN` here and just let this function either still run or not run when ABM status is still `ABM_UNKNOWN`.

I personally think better would be to not let this function run when ABM is not specifically known to be enabled (yet), especially if we perhaps also don't know which ABM path it will take (yet), so I'd treat it the same as `ABM_DISABLED` and `return NULL` as on L307/431. At least from my current understanding...

drivers/mge-hid.c
desertwitch, line 219: I know this is not your code, but maybe also better to start here with `-1` and drop the `ABM_UNKNOWN` (= -1) definition for similar reason as in other comment (separation from HID/device-returned values).

drivers/mge-hid.c
desertwitch: Maybe also rename these variables to be more descriptive, e.g. `abm_batt_status_present` or `abm_batt_mode_present`, because right now their names also kind of imply these are modes/statuses values returned from the device itself.

drivers/mge-hid.c
```diff
@@ -300,28 +396,99 @@ static info_lkp_t eaton_abm_status_info[] = {
...
+	{ 5, "Constant Charge (CC)", NULL, NULL },
+	{ 0, NULL, NULL, NULL }
+};
+
 /* Used to process ABM flags, for ups.status (CHRG/DISCHRG/RB) */
 static const char *eaton_abm_chrg_dischrg_fun(double value)
 {
 	/* Don't process if ABM is disabled */
 	if (advanced_battery_monitoring == ABM_DISABLED)
 		return NULL;
```
desertwitch, line 432: Perhaps we should also handle here a transition from previous ABM enabled to ABM unknown/disabled?
Similar to as is done in the non-ABM functions `eaton_abm_check_chrg_fun`/`eaton_abm_check_dischrg_fun`:
https://github.com/networkupstools/nut/blob/4b1e9258b892cb0fa9b01c80bb12f147d04bf1ee/drivers/mge-hid.c#L449-L453
https://github.com/networkupstools/nut/blob/4b1e9258b892cb0fa9b01c80bb12f147d04bf1ee/drivers/mge-hid.c#L423-L427
Basically cleaning up any previously published ABM flags so the non-ABM mode can take over?

desertwitch: Nevermind, this seems not needed here as `eaton_abm_check_chrg_fun`/`eaton_abm_check_dischrg_fun` will always set non-ABM `chrg` and `dischrg` flags overriding no longer needed ABM-flags, in fact implementing such an addition would likely override such already correctly set flags due to function order.

drivers/mge-hid.c
```diff
@@ -300,28 +396,99 @@ static info_lkp_t eaton_abm_status_info[] = {
...
+		default:
+			/* Return NULL, not to get the value published! */
+			return NULL;
+		}
+	}
+	else
+	{
+		switch ((long)value)
+		{
+		case 1: /* charging status */
```
desertwitch, line 478: Should maybe also set `chrg` here (as with other ABM path)?

desertwitch, line 478: Should maybe also set `chrg` here (as with other ABM path)?

drivers/mge-hid.c
```diff
@@ -300,28 +396,99 @@ static info_lkp_t eaton_abm_status_info[] = {
...
+		case 3: /* floating status */
```
desertwitch, line 479: Should maybe also set `chrg` here for `case 1` (as done with other ABM path)?

masterwishx:
@jimklimov can we merge it?

jimklimov:
Sorry, I had a busy week with a release loiming at work, so did not yet read deeply into recent iterations of this PR.

Questions posted by @desertwitch do seem reasonable though.

masterwishx:
> Sorry, I had a busy week with a release loiming at work, so did not yet read deeply into recent iterations of this PR.
> 
> Questions posted by @desertwitch do seem reasonable though.

Please check when you will have more time, only make sense to me to rename `advanced_battery_status` to `abm_table_status` and `advanced_battery_mode` to `abm_table_mode`, the other I answered also befor above.
Let me know then.
@desertwitch  if thous names look better?

masterwishx:
@desertwitch Thanks for hint , should look better now ?

desertwitch:
I tried to quickly illustrate here my code review/comments: https://github.com/masterwishx/nut/pull/4
Includes treating `ABM_UNKNOWN` as `ABM_DISABLED` and what I considered improvements in readability and/or ease of logic. 
Haven't touched the `ECO` functions as I'm not 100% clear on the logic behind these, so this is mostly around ABM-related code.

masterwishx:
> I tried to quickly illustrate here my code review/comments: [masterwishx#4](https://github.com/masterwishx/nut/pull/4) Includes treating `ABM_UNKNOWN` as `ABM_DISABLED` and what I considered improvements in readability and/or ease of logic.
> 
> Haven't touched the `ECO` functions as I'm not 100% clear on the logic behind these, so this is mostly around ABM-related code. **Grain of salt:** Compiles OK but not (yet) tested in real life, built refactors around existing logic to show what I meant earlier. 😉

I'm sorry I didn't made sepate pr for status...
I saw your code and Thanks a lot for your effort but if you look closer it will break the fix for Eaton online ups with status but without ABM = Eaton 9E.

In addition to fix I leaved original code as it Becouse its already working for ups with mode  + ABM or for unknown without ABM.

And we don't know if there is ups with only status + ABM. 

In your case Eaton 5p that have status + mode + ABM + type paths.
But in status you have 19 that's not right value for code so we should use mode in your case!

In any case I compiled and tested package with lasted commit befor renaming today.
And using it for some days all working fine!

If you can check it will be cool, should not be issues in your case but good to know.

The last commit with renaming should not break old logic.

desertwitch:
> > I tried to quickly illustrate here my code review/comments: [masterwishx#4](https://github.com/masterwishx/nut/pull/4) Includes treating `ABM_UNKNOWN` as `ABM_DISABLED` and what I considered improvements in readability and/or ease of logic.
> > Haven't touched the `ECO` functions as I'm not 100% clear on the logic behind these, so this is mostly around ABM-related code. **Grain of salt:** Compiles OK but not (yet) tested in real life, built refactors around existing logic to show what I meant earlier. 😉
> 
> I'm sorry I didn't made sepate pr for status... I saw your code and Thanks a lot for your effort but if you look closer it will break the fix for Eaton online ups with status but without ABM = Eaton 9E.
> 
> In addition to fix I leaved original code as it Becouse its already working for ups with mode + ABM or for unknown without ABM.
> 
> And we don't know if there is ups with only status + ABM.
> 
> In your case Eaton 5p that have status + mode + ABM + type paths. But in status you have 19 that's not right value for code so we should use mode in your case!
> 
> In any case I compiled and tested package with lasted commit befor renaming today. And using it for some days all working fine!
> 
> If you can check it will be cool, should not be issues in your case but good to know.
> 
> The last commit with renaming should not break old logic.

Sorry, I'm not really sure what you mean, did you test it? What do the debug logs say?
Can you maybe provide a dump of all your HID paths too, so we can compare it a bit?

I preserved almost all of the old logic here, as far as I can see on another glance too.
For what it's worth I do have some implicit conversions going on, those might require explicit casting...

The only other difference though seems treating `ABM_UNKNOWN` as `ABM_DISABLED`. If in your case `x.ABMEnable` does not exist, and you also don't have `x.ChargerType=4` otherwise setting `ABM_ENABLED_TYPE` (_which, to be honest, I'm not sure is correct as I don't see that in the HID table... but I kept your logic there_), your UPS would be `ABM_UNKNOWN` and the ABM function would not run (correctly so).

If your UPS has a certain combination (e.g. "x.ChargerType=5" and "x.Status=2" without `x.ABMEnable`) that you know to be enabled ABM, we should make that specific combination become `ABM_ENABLED`, instead of generally letting an ABM function run with `ABM_UNKNOWN` also for any other devices where it might really be unknown (and possibly risk false pseudo-ABM values getting parsed, as opposed to just using the correct non-ABM status there).

As for my 5P, with `"x.ABMEnable=1 and "x.Mode=4","x.ChargerType=4" and "x.Status=19"`, both the current and also my code would correctly take `ABM_PATH_MODE` due to HID conversion function order (_MODE before _STATUS, no overwrite) and disregard the bogus `x.Status=19`, so this logic remains unchanged from the current code.

Anyway, it's your PR and your choice. But, full disclosure (and no disrespect intended), with the way things are right now, it took me a very long time to figure out the logic behind the current variable names, various ABM paths etc. and now having looked some more into it, I'm still not entirely sure we won't be breaking other devices with unknown ABM states, all in favour of fixing a specific device (which I've tried to safely address and improve with my suggestions). Let's see what @jimklimov says when he has more time at his disposal.

masterwishx:
> The only difference seems treating `ABM_UNKNOWN` as `ABM_DISABLED`. If in your case `x.ABMEnable` does not exist, and you also don't have `x.ChargerType=4` setting `ABM_ENABLED_TYPE` (_which, to be honest, I'm not sure is correct as I don't see that in the HID table... but I kept your logic there_), your UPS would be `ABM_UNKNOWN` and the ABM function would not run (correctly so).

I have x.ChargerType=4 also you have, but I don't have ABM `x.ABMEnable`.

And I have floating when battery.charge =100
After charge so this hid table working fine for me, that was the logic for fix 9E

masterwishx:
> If your UPS has a certain combination (e.g. "x.ChargerType=5" and "x.Status=2" without `x.ABMEnable`)

I have "x.ChargerType=4" it's constant charge (CC) if "x.Status=2" then it charging

desertwitch:
> > The only difference seems treating `ABM_UNKNOWN` as `ABM_DISABLED`. If in your case `x.ABMEnable` does not exist, and you also don't have `x.ChargerType=4` setting `ABM_ENABLED_TYPE` (_which, to be honest, I'm not sure is correct as I don't see that in the HID table... but I kept your logic there_), your UPS would be `ABM_UNKNOWN` and the ABM function would not run (correctly so).
> 
> I have x.ChargerType=4 also you have, but I don't have ABM `x.ABMEnable`.
> 
> And I have floating when battery.charge =100 After charge so this hid table working fine for me, that was the logic for fix 9E

> > If your UPS has a certain combination (e.g. "x.ChargerType=5" and "x.Status=2" without `x.ABMEnable`)
> 
> I have "x.ChargerType=4" it's constant charge (CC) if "x.Status=2" then it charging

Yes, and this logic is preserved in my code, as before:
https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L1626-L1627
https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L400-L416
https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L439-L449

masterwishx:
> Which leads me back to the question if you tested it with my code and it's not working?

Your code will not work for me if you make ABM_UNKNOWN with ABM_DISABLED to return.

I checked this befor when tryed same way, aslo original code was not working for me becose always (!chrg and   ! dschrg)
I will post logs later...

masterwishx:
> OK, so that clears up where the `ABM_ENABLED_TYPE` originates from, so we know that to be enabled ABM.

This only to show table of ABM battery type

masterwishx:
Also X.Status in you case =19, this wrong value we should use X.mode in your case.
And we can have ups that have only one of thouse. 
So that was the logic in my code:

To not change original code but to add fix for 9E

masterwishx:
this was in original code : 

```
Path: UPS.PowerSummary.PresentStatus.Discharging, Type: Feature, ReportID: 0x01, Offset: 24, Size: 8, Value: 0
Oct  4 14:54:17 DaRKNaS usbhid-ups[3017]: [D5:3017] process_boolean_info: !dischrg

Oct  4 14:54:17 DaRKNaS usbhid-ups[3017]: [D2:3017] Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
Oct  4 14:54:17 DaRKNaS usbhid-ups[3017]: [D5:3017] process_boolean_info: !chrg
```

this is my code that fine:  
```
Path: UPS.BatterySystem.Charger.ChargerType, Type: Feature, ReportID: 0x26, Offset: 0, Size: 8, Value: 5
Set ABM variable to unknown, charger type status: -1
ABM is disabled

Path: UPS.BatterySystem.Charger.Status, Type: Feature, ReportID: 0x26, Offset: 8, Size: 8, Value: 2
ABM numeric status: 2
ABM string status: floating
ABM is -1

Path: UPS.PowerSummary.PresentStatus.Discharging, Type: Feature, ReportID: 0x01, Offset: 24, Size: 8, Value: 0
process_boolean_info: !dischrg

Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
process_boolean_info: !chrg

Path: UPS.BatterySystem.Charger.Status, Type: Feature, ReportID: 0x26, Offset: 8, Size: 8, Value: 2
ABM CHRG/DISCHRG legacy string status (ups.status): !chrg
process_boolean_info: !chrg
```

masterwishx:
> OK, so that clears up where the `ABM_ENABLED_TYPE` originates from, so we know that to be enabled ABM.

i wrote you before inyour code that here you break the code logic :  
https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L434

masterwishx:
> > OK, so that clears up where the `ABM_ENABLED_TYPE` originates from, so we know that to be enabled ABM.
> 
> i wrote you before inyour code that here you break the code logic : https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L434

Also here for status not floating : 

https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L443


and all same  ...

desertwitch:
> Your code will not work for me if you make ABM_UNKNOWN with ABM_DISABLED to return.
> 

> i wrote you before inyour code that here you break the code logic : https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L434

Yes, and that is exactly my point, that - I think - it _should not work_ (still run an ABM function) in such a case where your UPS (**and any other UPS**) is `ABM_UNKNOWN`. But rather we should find out the exact HID combination that your UPS uses for enabled ABM and make _that_ `ABM_ENABLED`. My point is, just because still running the ABM function in an unknown ABM state works for your device, doesn't mean it won't break others and - I think - we should be careful here as this will affect many other devices also.

> Also X.Status in you case =19, this wrong value we should use X.mode in your case.

Yes, and X.mode is used by both your code and my code (see my comment above). 

> Also here for status not floating :
> 
> https://github.com/masterwishx/nut/blob/e4f263171a3112024e270beda4961b66fad1c708/drivers/mge-hid.c#L443
> 
> and all same ...

Exactly the same as in your code:
https://github.com/networkupstools/nut/blob/ebc0ff407c562b9d8473f70f67ebd65ed21467f7/drivers/mge-hid.c#L435-L437

Anyway, I'm (respectfully) signing off for the day (it is Sunday after all). No offense, I'm really not trying to blast your code and I have great respect for the effort you've put into this, just trying to explain my reservations. It's also up to Jim eventually, I'm also just a contributor (same as you) here and it's just my personal opinion on where safeguards and improvements could be done.

masterwishx:
> But rather we should find out the exact HID combination that your UPS uses for enabled ABM and make _that_ `ABM_ENABLED`. My point is, just because still running the ABM function in an unknown ABM state works for your device, doesn't mean it won't break others and - I think - we should be careful here as this will affect many other devices also.

I think you still not get the point of my code :)
My 9E model don't have ABM but have type and status tables.
So I checked and my code work well now on 9E model so I think should work for other UPS that have same tables and ABM_UNKNOWN , for all other the code same as original.

So the point of code just to make fix for 9E and other with same tables and leave original code as it already working for any other upses

Offcourse @jimklimov will decide what's better looks.
I really thanks to you for your help here and offcourse for continuing and updating amazing plugin for Unraid

masterwishx:
> Floating should result in `chrg` and not `!chrg` at value `2` of `UPS.BatterySystem.Charger.Status` according to the HID table. I do not understand why you are forcing `!chrg` here for any devices `ABM_UNKNOWN` (or even your own device, for that matter).

Becouse otherwise I will have OL CHRG every time also after battery.charge = 100, but when battery.charge = 100 change from 99, I got floating instead of charging in status table.

masterwishx:
> Last, I'm still not convinced about the extra ABM-enabling code of `ABM_ENABLED_TYPE`, which equates `UPS.BatterySystem.Charger.ChargerType` with value `4` to an enabled ABM. Where is this coming from? The code comments suggest my 5P, but it actually controls it through `UPS.BatterySystem.Charger.ABMEnable`? Your device seems to use `5`?

This part not really important Becouse if `4` ABM just can only print debug and flag will be owerwriten by ABM enabled to `1` if ABM exist.

My device use `5` constant charge (CC) 

Was needed to check if ABM enabled by other than ABM Enabled. 
But seems to enable/disable ABM only ABM enabled used. 
This was needed at start, and now I think if we don't use this flag, this function can be removed.

desertwitch:
> > Last, I'm still not convinced about the extra ABM-enabling code of `ABM_ENABLED_TYPE`, which equates `UPS.BatterySystem.Charger.ChargerType` with value `4` to an enabled ABM. Where is this coming from? The code comments suggest my 5P, but it actually controls it through `UPS.BatterySystem.Charger.ABMEnable`? Your device seems to use `5`?
> 
> This part not really important Becouse if `4` ABM just can only print debug and flag will be owerwriten by ABM enabled to `1` if ABM exist.
> 
> My device use `5` constant charge (CC)
> 
> Was needed to check if ABM enabled by other than ABM Enabled. But seems to enable/disable ABM only ABM enabled used. This was needed at start, and now I think if we don't use this flag, this function can be removed.

Ok, thanks for the clarification, so I think it's better if we remove this additional ABM-enabling code.
ABM seems to be enabled through `UPS.BatterySystem.Charger.ABMEnable` only and this should suffice?

>I think you still not get the point of my code :)
My 9E model don't have ABM but have type and status tables.

grafik
Your pull request description says: "Added ABM statuses for 9E", so I always assumed your UPS supports ABM and you are trying to make that work with the driver (which I factored into my code reviews and suggestions). But now I see that your UPS doesn't support ABM, which is also the reason you're getting `5` = `Constant Charge (CC)` for `UPS.BatterySystem.Charger.ChargerType`. So why are we modifying the existing ABM functions to "support" a device that does not have ABM? What values do you want to gain here and how can a/your UPS be "floating" if it's constant charge and does not have ABM?

I see the benefit in adding the additional ABM path for other ABM-capable devices, but if I'm not mistaken your device shouldn't use an ABM path/lookup table if it has no support for ABM...?

masterwishx:
> Ok, thanks for the clarification, so I think it's better if we remove this additional ABM-enabling code.
> ABM seems to be enabled through `UPS.BatterySystem.Charger.ABMEnable` only and this should suffice? Just because the charger type is ABM(-capable), it doesn't have to be enabled either probably?

Yes, but for other part it seems we are **going around** , maybe you will get it better if you will read from above comments of commits ?
https://github.com/networkupstools/nut/pull/2660#discussion_r1803749060


From this list you see ABM Eaton models , so mostly what left we have `5E = ABM_UNKNOWN` and `9E = ABM_UNKNOWN + x.Status and x.Type` tables .
We dont  know what of these use `x.Status` or `x.Mode` 

i will try again to explain :  For statuses i added missing path for x.Type x.Status etc but i have leaved the original code as is only added fix for 9E Models ! by added : 

```
if (x.status and not x.mode) 
  if unknown then change status of charge and discharge
```

So unknown will have same as original code also other


image

masterwishx:
> I see the benefit in adding the additional ABM path for other ABM-capable devices, but if I'm not mistaken your device shouldn't use an ABM path/lookup table if it has no support for ABM...? So in that sense my improvement suggestion of not letting an `ABM_UNKNOWN` device (like yours) proceed into an ABM function was kind of correct...?

Yes you right but as i found the x.status table is working fine when charging/floating/discharging however no ABM  in real  so it let me fix the charging states aslo to see real charging status by this table !

Maybe its like firmware cuted from 9SX that have ABM , i dont know but its working , but not regular chrg/dischrg statuses.
Sorry i didnt any other way to fix it ....
Also i didnt modified the original code that already works for other cases.

in any case again im sorry not started for this another pr than eco :(

masterwishx:
> correctly result in `OL CHRG` for your constant charge type charger

Always this state? Not shure it's correct, also have OB DSCHRG in this case.

masterwishx:
now with my code by x.status have that looks more right than always OL CHRG: 

image

desertwitch:
No need to apologize, I appreciate your point of view and effort. To me it just doesn't seem to matter what ABM paths are exposed by devices that are not ABM-capable. I think it's not an improvement to parse ABM paths just because they are there (due to common firmware, leftovers, ...), when the device itself doesn't support ABM. A non-ABM device reporting as "floating" seems to me more of a bug than an improvement. And yes, I'd expect a constant charge type charger to always display either `OL CHRG` (constant charge) or `OB DISCHRG` and this lines up with what the HID table says for constant mode too. So the non-ABM function already displays correctly your non-ABM paths (before this PR), in my opinion, and adding ABM into the mix makes no sense for your device - I think.

masterwishx:
> No need to apologize, I appreciate your point of view and effort. To me it just doesn't seem to matter what ABM paths are exposed by devices that are not ABM-capable. I think it's not an improvement to parse ABM paths just because they are there (due to common firmware, leftovers, ...), when the device itself doesn't support ABM. A non-ABM device reporting as "floating" seems to me more of a bug than an improvement. And yes, I'd expect a constant charge type charger to always display either `OL CHRG` (constant charge) or `OB DISCHRG` and this lines up with what the HID table says for constant mode too. So the non-ABM function already displays correctly your non-ABM paths (before this PR), in my opinion, and adding ABM into the mix makes no sense for your device - I think.

The original code was always send !chrg and !dischrg so i was not see any charge and discharge at all

desertwitch:
> > No need to apologize, I appreciate your point of view and effort. To me it just doesn't seem to matter what ABM paths are exposed by devices that are not ABM-capable. I think it's not an improvement to parse ABM paths just because they are there (due to common firmware, leftovers, ...), when the device itself doesn't support ABM. A non-ABM device reporting as "floating" seems to me more of a bug than an improvement. And yes, I'd expect a constant charge type charger to always display either `OL CHRG` (constant charge) or `OB DISCHRG` and this lines up with what the HID table says for constant mode too. So the non-ABM function already displays correctly your non-ABM paths (before this PR), in my opinion, and adding ABM into the mix makes no sense for your device - I think.
> 
> 
> 
> The original code was always send !chrg and !dischrg so i was not see any charge and discharge at all

You're right, my bad, because it incorrectly executed the ABM function when seeing your partial/unused ABM paths despite being `ABM_UNKNOWN` (even before your PR). This would be solved by treating `ABM_UNKNOWN` as `ABM_DISABLED`, as I suggested, and make it correctly use the non-ABM function for your device.

masterwishx:
> > > No need to apologize, I appreciate your point of view and effort. To me it just doesn't seem to matter what ABM paths are exposed by devices that are not ABM-capable. I think it's not an improvement to parse ABM paths just because they are there (due to common firmware, leftovers, ...), when the device itself doesn't support ABM. A non-ABM device reporting as "floating" seems to me more of a bug than an improvement. And yes, I'd expect a constant charge type charger to always display either `OL CHRG` (constant charge) or `OB DISCHRG` and this lines up with what the HID table says for constant mode too. So the non-ABM function already displays correctly your non-ABM paths (before this PR), in my opinion, and adding ABM into the mix makes no sense for your device - I think.
> > 
> > 
> > The original code was always send !chrg and !dischrg so i was not see any charge and discharge at all
> 
> You're right, my bad, because it incorrectly executed the ABM function when seeing your partial/unused ABM paths despite being `ABM_UNKNOWN` (even before your PR). This would be solved by treating `ABM_UNKNOWN` as `ABM_DISABLED`, as I suggested, and make it correctly use the non-ABM function for your device.

Yes but dont you think `UPS Status Timeline` in picture looks more right than always charge status ?
https://github.com/networkupstools/nut/pull/2660#issuecomment-2440140862

masterwishx:
Maybe @jimklimov and @arnaudquette-eaton can Reveal more about the right condition in this case?

desertwitch:
> 
> Yes but dont you think `UPS Status Timeline` in picture looks more right than always charge status ?
> 
> https://github.com/networkupstools/nut/pull/2660#issuecomment-2440140862

For a non-ABM device in constant charge mode it doesn't look correct, because my understanding of constant charge is that it's constantly applying charge (= always charging). This also aligns with the HID table and non-ABM paths exposed by your device, which also say it's charging:

```
Path: UPS.PowerSummary.PresentStatus.Discharging, Type: Feature, ReportID: 0x01, Offset: 24, Size: 8, Value: 0
Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
```

masterwishx:
> Let's see what the others say.

If you have time meanwhile, maybe you can check what values for our paths your ups will send in X.type=5 (cc) mode When ABM is disabled also to check statuses?

masterwishx:
> > Yes but dont you think `UPS Status Timeline` in picture looks more right than always charge status ?
> > [#2660 (comment)](https://github.com/networkupstools/nut/pull/2660#issuecomment-2440140862)
> 
> For a non-ABM device in constant charge mode it doesn't look correct to me, because my understanding of constant charge is that it's constantly applying charge (= always charging). This also aligns with the HID table and non-ABM paths exposed by your device, which also say it's charging (and are the only paths we should be looking at with a non-ABM-capable device _imo_):
> 
> ```
> Path: UPS.PowerSummary.PresentStatus.Discharging, Type: Feature, ReportID: 0x01, Offset: 24, Size: 8, Value: 0
> Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
> ```
> 
> Let's see what the others say... 🙂
> 
> Meanwhile I've updated my improvement recommendations: [masterwishx#4](https://github.com/masterwishx/nut/pull/4)

@jimklimov @arnaudquette-eaton Please confirm this is the right condition for Constant charge non ABM Eaton UPS , i merged this pr from @desertwitch as it looks he was right about this case !?

masterwishx:
> Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1

if you will have time , still interesting if your ups in ABM disabled mode will always send x.Charging =1 even in 100% charged ? 
that will  `OL CHRG`.
and when on battery x.Discharging =1 that will `OB DISCHRG` ...
Thanks

desertwitch:
> > Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
> 
> if you will have time , still interesting if your ups in ABM disabled mode will always send x.Charging =1 even in 100% charged ? that will `OL CHRG`. and when on battery x.Discharging =1 that will `OB DISCHRG` ... Thanks

5P850I - ABM disabled - `OL CHRG`:
```
Path: UPS.BatterySystem.Charger.ABMEnable, Type: Feature, ReportID: 0x29, Offset: 0, Size: 8, Value: 0
Path: UPS.BatterySystem.Charger.ChargerType, Type: Feature, ReportID: 0x26, Offset: 0, Size: 8, Value: 0
Path: UPS.PowerSummary.PresentStatus.Discharging, Type: Feature, ReportID: 0x01, Offset: 24, Size: 8, Value: 0
Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
```

5P850I - ABM enabled - `OL CHRG` (floating):
```
Path: UPS.BatterySystem.Charger.ABMEnable, Type: Feature, ReportID: 0x29, Offset: 0, Size: 8, Value: 1
Path: UPS.BatterySystem.Charger.ChargerType, Type: Feature, ReportID: 0x26, Offset: 0, Size: 8, Value: 4
Path: UPS.BatterySystem.Charger.Mode, Type: Feature, ReportID: 0x26, Offset: 8, Size: 8, Value: 3
Path: UPS.PowerSummary.PresentStatus.Discharging, Type: Feature, ReportID: 0x01, Offset: 24, Size: 8, Value: 0
Path: UPS.PowerSummary.PresentStatus.Charging, Type: Feature, ReportID: 0x01, Offset: 16, Size: 8, Value: 1
```

So we can see that in constant charge mode (with ABM disabled), we have `OL CHRG` (as I expected, always applying charge).

`UPS.BatterySystem.Charger.ChargerType` seems to change also, but I still think we should not trust it for ABM enabled/disabled deductions and let only `UPS.BatterySystem.Charger.ABMEnable` decide here. Because it didn't switch to `5` "Constant Charge (CC)", as we would've expected it to, but rather to `0` "None". So I think we could keep this as a textual representation of the charger type in the variables, but without any special significance for ABM enabled-ness (so I'd keep it as it is in the code right now).

masterwishx:
> and not let `UPS.BatterySystem.Charger.ChargerType` decide on anything

Thanks a lot, we will not let decide its just for the info of variable in table:
`battery.charger.type  Constant Charge (CC)`

From copilot: 

```
Charging Behavior:
The Eaton 9E UPS maintains constant charging even when the battery is at 100%.
This behavior is typical for online double-conversion UPS systems.
The UPS continuously charges the battery to ensure it’s ready for use during power outages.
Status Indicators:
When the UPS is operating normally (on utility power), you’ll see the status as “OL CHRG” (Online Charging).
When the UPS switches to battery power (during a power outage), you’ll see “OB DISCHRG” (On Battery Discharge).
```

masterwishx:
> Because it didn't stay either the same or switch to `5` "Constant Charge (CC)", as we maybe would've expected it to, but rather switched to `0` "None".

Yeah i see , but maybe in other Eaton models with ABM it switch to `CC` ?

masterwishx:
> 5P850I - ABM enabled - `OL CHRG` (floating):

battery.charge != 100 ?

desertwitch:
> > Because it didn't stay either the same or switch to `5` "Constant Charge (CC)", as we maybe would've expected it to, but rather switched to `0` "None".
> 
> Yeah i see , but maybe in other Eaton models with ABM it switch to `CC` ?

Maybe, we'll have to wait for more data on this (I think), so I'd keep it only as informational variable (for now).

masterwishx:
> battery.charge = 100 (floating is applying charge so `OL CHRG` is right here, resting would be `OL`)

Thanks got it

masterwishx:
> Maybe, we'll have to wait for more data on this (I think), so I'd keep it only as informational variable (for now).

we can leave as is for now or add something like: 

https://github.com/networkupstools/nut/blob/18e137597a00cde4839f90961d5e1a3bbb6e6f5c/drivers/mge-hid.c#L1629

`/* Published battery charger type (info variable that can be ABM, CC, ...) */ `

and: 
https://github.com/networkupstools/nut/blob/18e137597a00cde4839f90961d5e1a3bbb6e6f5c/drivers/mge-hid.c#L162C2-L162C136

` * UPS.BatterySystem.Charger.ChargerType      |                      4                      |            5 or 0             |       |      |`

but `0` is `none` in table of variable so not sure

desertwitch:
Maybe @jimklimov can help out here with the merge/rebase so that nothing ABM-related is lost. 🙂

masterwishx:
> Maybe @jimklimov can help out here with the merge/rebase so that nothing ABM-related is lost. 🙂

Fixed

masterwishx:
> Maybe @jimklimov can help out here with the merge/rebase so that nothing ABM-related is lost. 🙂

Please check now

desertwitch:
> > Maybe @jimklimov can help out here with the merge/rebase so that nothing ABM-related is lost. 🙂
> 
> Please check now

Looks good - thanks for fixing!

masterwishx:
@desertwitch not sure where the error ...

desertwitch:
> @desertwitch not sure where the error ...

I can't see any specific error in the logs, although I'm no expert navigating the CI, maybe something unrelated to this PR.

masterwishx:
> > @desertwitch not sure where the error ...
> 
> I can't see any specific error in the logs, although I'm no expert navigating the CI, maybe something unrelated to this PR.

Yes, sorry it's about CI so @jimklimov can help? I saw some new prs for CI

desertwitch:
Minor follow up to my code contributions: https://github.com/masterwishx/nut/pull/5 (if possible please merge into this PR)

masterwishx:
CC @jimklimov please check when you will have time if can be merged

Issue #2661, opened 2024-10-17
Added  ups.beeper.status  support for Masterpower MF-UPS650VA


yonesmit:
Hi,
Recently I start using a Masterpower MF-UPS650VA with nut. The device is detected correctly, it uses usbhid-ups driver and MGE HID 1.49 subdriver.
My problem was I couldn't control the beeper to silence it (ups.beeper.status). After some debugging I found that the feature was found in a different path than the ones that driver support (driver already support the feature in 2 different paths and it uses the first one it  founds).
Driver support ups.beeper.status in this 2 paths:
```
UPS.BatterySystem.Battery.AudibleAlarmControl
UPS.PowerSummary.AudibleAlarmControl
```
And my ups uses this path:

`UPS.AudibleAlarmControl`

So I added the 3 lines needed to support the beeper status feature in the path used by my ups and compiled the driver.

```
	{ "ups.beeper.status", 0 ,0, "UPS.AudibleAlarmControl", NULL, "%s", HU_FLAG_SEMI_STATIC, beeper_info },
...
	{ "beeper.disable", 0, 0, "UPS.AudibleAlarmControl", NULL, "1", HU_TYPE_CMD, NULL },
	{ "beeper.enable", 0, 0, "UPS.AudibleAlarmControl", NULL, "2", HU_TYPE_CMD, NULL },
```

Now the feature is correctly supported.

```
battery.charge: 100
battery.charge.low: 10
battery.runtime: 7200
battery.type: PbAC
device.mfr: KSTAR
device.model: UPS HID
device.serial: HID-C01
device.type: ups
driver.debug: 0
driver.flag.allow_killpower: 0
driver.name: usbhid-ups
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.port: auto
driver.parameter.product: UPS HID
driver.parameter.productid: 0001
driver.parameter.serial: HID-C01
driver.parameter.synchronous: auto
driver.parameter.vendor: KSTAR
driver.parameter.vendorid: 0463
driver.state: quiet
driver.version: 2.8.2.1081-1081-g78e0525d3
driver.version.data: MGE HID 1.49
driver.version.internal: 0.57
driver.version.usb: libusb-1.0.24 (API: 0x1000108)
input.frequency: 49.9
input.voltage: 230.5
outlet.1.status: on
output.voltage: 230.4
ups.beeper.status: disabled
ups.firmware:
ups.load: 8
ups.mfr: KSTAR
[...]

PR #2662, opened 2024-10-17
Update mge-hid.c: Added ups.beeper.status support for Masterpower MF-UPS650VA

yonesmit:
As explained here:
https://github.com/networkupstools/nut/issues/2661#issue-2593677212
This is a request to add suport for ups.beeper.status feature for Masterpower MF-UPS650VA
The only change is to find it in a different path from the already two paths supported.

Thanks
Best Regards,
yonesmit


jimklimov:
Notes for self:
* Update NEWS
* Update MGE subdriver version (now not here) - will be bumped by other PRs now in progress

PR #2680, opened 2024-11-15
Work247-ecomode_addon

masterwishx:
As add-on for https://github.com/networkupstools/nut/pull/2637 ECO/Bypass for Eaton Models
we can leave as is now only check range by r/w values or set also check range for eco/bypass by commands .

Also we can add ESS Mode staff (check range and set ups status) like in ECO mode
if it needed .

But only Eaton Models like `93PM G2` and `9395P` use ESS , not sure if they use NUT :)

### DIFF
--- a/drivers/mge-hid.c
+++ b/drivers/mge-hid.c
@@ -1755,12 +1755,12 @@ static hid_info_t mge_hid2nut[] =
 
 	/* Command to switch ECO Mode */
 	{ "ecomode.disable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "0", HU_TYPE_CMD, NULL },
-	{ "ecomode.enable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "1", HU_TYPE_CMD, NULL },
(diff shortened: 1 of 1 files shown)


masterwishx:
We can also make automatic ECO steps...
@jimklimov please confirm if we need all this

jimklimov:
Looks good, and I think more details is better too.

CCing @arnaudquette-eaton also.

masterwishx:
> Looks good, and I think more details is better too.
> 
> CCing @arnaudquette-eaton also.

What details you mean?
In new pr we can:

1. You think we can add ESS stuff also, hovewere
Only enterprise upses have it?

2. We can add also post limits not passed in debug when check range eco/bypass func.

4. We can add also maybe auto function for ECO that insert to bypass and then to eco.
Instead user manually make thouse steps.

Also for now can't leave eco mode by commands but maybe Becouse my ussue with timeout, not shure...

masterwishx:
> NOTE FIR SEKF: check that subdriver version is bumped since last release, but suffices once across several recent PRs.

It was bumped in eco pr

PR #2689, opened 2024-11-22
Updated Phoenix Contact modbus driver for newer models

RikyPlaza:
Updated Phoenix Contact modbus driver for newer models which are using different modbus addresses. Choice is made basing on the UPS firmware version and the model information have been changed in order to have a quick feedback over the new/old model detection.

### DIFF
--- a/drivers/phoenixcontact_modbus.c
+++ b/drivers/phoenixcontact_modbus.c
@@ -20,6 +20,34 @@
  *
  */
 
+#if (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_PUSH_POP) && ( (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT) || (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE) )
+# pragma GCC diagnostic push
+#endif
+#ifdef HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT
+# pragma GCC diagnostic ignored "-Wcovered-switch-default"
+#endif
+#ifdef HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE
+# pragma GCC diagnostic ignored "-Wunreachable-code"
+#endif
+/* Older CLANG (e.g. clang-3.4) seems to not support the GCC pragmas above */
+#ifdef __clang__
+# pragma clang diagnostic push
+# pragma clang diagnostic ignored "-Wunreachable-code"
+# pragma clang diagnostic ignored "-Wcovered-switch-default"
+#endif
+                /* All enum cases defined as of the time of coding
+                 * have been covered above. Handle later definitions,
+                 * memory corruptions and buggy inputs below...
+                 */
+                default:
+                        fatalx(EXIT_FAILURE, "no suitable definition found!");
+#ifdef __clang__
+# pragma clang diagnostic pop
+#endif
+#if (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_PUSH_POP) && ( (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT) || (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE) )
+# pragma GCC diagnostic pop
+#endif
+
 #include "main.h"
 #include <modbus.h>
 
--- a/drivers/phoenixcontact_modbus.c
+++ b/drivers/phoenixcontact_modbus.c
@@ -149,7 +149,6 @@ void upsdrv_updateinfo(void)
 		break;
 	default:
 		fatalx(EXIT_FAILURE, "Uknown UPS firmware version.");
-		break;
 	}	
 
 	status_init();
@@ -177,7 +176,6 @@ void upsdrv_updateinfo(void)
 		break;
 	default:
 		fatalx(EXIT_FAILURE, "Uknown UPS firmware version.");
-		break;
 	}
 
 	dstate_setinfo("output.voltage", "%d", (int) (tab_reg[0] / 1000));
@@ -192,7 +190,6 @@ void upsdrv_updateinfo(void)
 		break;
 	default:
 		fatalx(EXIT_FAILURE, "Uknown UPS firmware version.");
-		break;
 	}
 
 	dstate_setinfo("battery.charge", "%d", tab_reg[0]);
@@ -222,7 +219,6 @@ void upsdrv_updateinfo(void)
 		break;
 	default:
 		fatalx(EXIT_FAILURE, "Uknown UPS firmware version.");
-		break;
 	}
 
 	dstate_setinfo("battery.voltage", "%f", (double) (tab_reg[0]) / 1000.0);
@@ -305,7 +301,6 @@ void upsdrv_updateinfo(void)
 		break;
 	default:
 		fatalx(EXIT_FAILURE, "Uknown UPS firmware version.");
-		break;
 	}
 
 	if (errcount == 0) {
--- a/drivers/phoenixcontact_modbus.c
+++ b/drivers/phoenixcontact_modbus.c
@@ -135,7 +135,6 @@ void upsdrv_updateinfo(void)
 	switch (UPSModel)
 	{
 	case QUINT4_UPS:
-
 		mrir(modbus_ctx, 0x2000, 1, &actual_code_functions);
 
 		tab_reg[0] = CHECK_BIT(actual_code_functions, 2); /* Battery mode is the 2nd bit of the register 0x2000 */
@@ -146,10 +145,10 @@ void upsdrv_updateinfo(void)
 		tab_reg[1] = CHECK_BIT(actual_alarms1, 2); /* Battery discharged is the 2nd bit of the register 0x3001 */
 		break;
 	case QUINT_UPS:
-
 		mrir(modbus_ctx, 29697, 3, tab_reg); /* LB is actually called "shutdown event" on this ups */
 		break;
 	default:
+		fatalx(EXIT_FAILURE, "Uknown UPS firmware version.");
 		break;
 	}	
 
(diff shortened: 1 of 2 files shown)


drivers/phoenixcontact_modbus.c
```diff
@@ -151,17 +307,17 @@ void upsdrv_initups(void)
...
 		fatalx(EXIT_FAILURE, "Unable to create the libmodbus context");
-
+	
 	r = modbus_set_slave(modbus_ctx, MODBUS_SLAVE_ID);	/* slave ID */
 	if (r < 0) {
 		modbus_free(modbus_ctx);
 		fatalx(EXIT_FAILURE, "Invalid modbus slave ID %d",MODBUS_SLAVE_ID);
 	}
-
+
```
jimklimov, line 320: Here and several times above - needless added whitespace

drivers/phoenixcontact_modbus.c
```diff
@@ -77,49 +119,163 @@ void upsdrv_updateinfo(void)
...
-	if (CHECK_BIT(tab_reg[0], 12))
-		alarm_set("Low Battery (Voltage)");
-	if (CHECK_BIT(tab_reg[0], 13))
-		alarm_set("Low Battery (Charge)");
-	if (CHECK_BIT(tab_reg[0], 14))
-		alarm_set("Low Battery (Time)");
-	if (CHECK_BIT(tab_reg[0], 16))
-		alarm_set("Low Battery (Service)");
+
+	uint32_t alarms_word;
```
jimklimov, line 202: Here and above, variable definitions are added in the middle of scope. NUT coding style requires all definitions to be before code, so older systems can build current NUT too. Probably that is what CI failed about all over the place.

drivers/phoenixcontact_modbus.c
```diff
@@ -49,21 +54,58 @@ void upsdrv_initinfo(void)
...
+	uint16_t actual_alarms = 0;
+	uint16_t actual_alarms1 = 0;
+
+	switch (UPSModel)
+	{
+	case NEW_UPS:
+		
+		mrir(modbus_ctx, 0x2000, 1, &actual_code_functions);
+
+		tab_reg[0] = CHECK_BIT(actual_code_functions, 2); //battery mode is the 2nd bit of the register 0x2000
```
jimklimov, line 95: Please use block comment style `/* ... */`

drivers/phoenixcontact_modbus.c
```diff
@@ -49,21 +54,58 @@ void upsdrv_initinfo(void)
...
+		tab_reg[2] = CHECK_BIT(actual_code_functions, 5); //battery charging is the 5th bit of the register 0x2000
+				
+		mrir(modbus_ctx, 0x3001, 1, &actual_alarms1);
+
+		tab_reg[1] = CHECK_BIT(actual_alarms1, 2); //battery discharged is the 2nd bit of the register 0x3001
+		break;
+	case OLD_UPS:
+
+		mrir(modbus_ctx, 29697, 3, tab_reg); /* LB is actually called "shutdown event" on this ups */
```
jimklimov, line 104: Any idea if these numbers make more sense in hex? Are they perhaps bit-masks of some command combos?

drivers/phoenixcontact_modbus.c
```diff
@@ -24,17 +24,22 @@
...
 
 #define DRIVER_NAME	"NUT PhoenixContact Modbus driver"
-#define DRIVER_VERSION	"0.04"
+#define DRIVER_VERSION	"0.05"
 
 #define CHECK_BIT(var,pos) ((var) & (1<<(pos)))
 #define MODBUS_SLAVE_ID 192
 
+#define OLD_UPS 0
+#define NEW_UPS 1
```
jimklimov, line 33: I wonder how soon "old" vs. "new" would become obsolete? Maybe more symbolic names would be better (just a thought)?

drivers/phoenixcontact_modbus.c
```diff
@@ -49,21 +54,58 @@ void upsdrv_initinfo(void)
...
 	dstate_setinfo("device.mfr", "Phoenix Contact");
-	dstate_setinfo("device.model", "QUINT-UPS/24DC");
-
+	
 	/* upsh.instcmd = instcmd; */
 	/* upsh.setvar = setvar; */
+
+	uint16_t FWVersion;
+
+	mrir(modbus_ctx, 0x0004, 1, &FWVersion);
```
jimklimov, line 63: Makes sense to also populate `ups.firmware`

jimklimov (changes requested):
Thanks for a solid-looking enhancement. A `NEWS.adoc` entry would be good, beside addressing the specific comments I mentioned during a look at code.

drivers/phoenixcontact_modbus.c
RikyPlaza: You're right, i've put the models. To be honest there would be another model (TRIO) but i don't have the hardware to test the communication. As soon as it will arrive i can test the driver and see if changes are needed.

drivers/phoenixcontact_modbus.c
RikyPlaza: I agree, not sure if the variable should be defined somewhere else or i can just call the dstate_setinfo function.

drivers/phoenixcontact_modbus.c
RikyPlaza: Undersoot!

RikyPlaza:
@jimklimov Hello, regarding the `NEWS.adoc`, could you give me some references please? I've looked in other pull requests but i can't find any example. Thank you!

jimklimov:
> @jimklimov Hello, regarding the `NEWS.adoc`, could you give me some references please? I've looked in other pull requests but i can't find any example. Thank you!

Probably just add a paragraph (bullet point) similar to ones that are already there.

drivers/phoenixcontact_modbus.c
jimklimov: Well, the global variable makes sense if this info would be used elsewhere (e.g. instead of those `OLD_UPS/NEW_UPS` defines?)

drivers/phoenixcontact_modbus.c
jimklimov: Another options can be an `enum` to also constrain the values accepted at compile-time (and get warnings like that a `switch` does not look at all newly defined options).

drivers/phoenixcontact_modbus.c
jimklimov: No, got no particular preference here. If maintenance would be easier with decimal numbers - so be it. Maybe worth a short comment at start of the file though.

RikyPlaza (commented):
Files reviewed.

jimklimov:
FYI: If the *lack of default label* would also prove a problem with some other compilers, check other drivers for use of pragmas around that. Typically we end up with a monster like:

````
#if (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_PUSH_POP) && ( (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT) || (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE) )
# pragma GCC diagnostic push
#endif
#ifdef HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT
# pragma GCC diagnostic ignored "-Wcovered-switch-default"
#endif
#ifdef HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE
# pragma GCC diagnostic ignored "-Wunreachable-code"
#endif
/* Older CLANG (e.g. clang-3.4) seems to not support the GCC pragmas above */
#ifdef __clang__
# pragma clang diagnostic push
# pragma clang diagnostic ignored "-Wunreachable-code"
# pragma clang diagnostic ignored "-Wcovered-switch-default"
#endif
                /* All enum cases defined as of the time of coding
                 * have been covered above. Handle later definitions,
                 * memory corruptions and buggy inputs below...
                 */
                default:
                        fatalx(EXIT_FAILURE, "no suitable definition found!");
#ifdef __clang__
# pragma clang diagnostic pop
#endif
#if (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_PUSH_POP) && ( (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT) || (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE) )
# pragma GCC diagnostic pop
#endif
````

RikyPlaza:
> embered: a man page would be needed to describe the supported devices and the driver options (per `addvar()` in code) and maybe other nuances - known caveats, etc. Ideally this comes from a driver author, as they best know if there is anything special to know about the devices.

@jimklimov  Ok, as soon as the checks are ok i'll take a look to the man page. Should i modify something already present or should i make a new man page?

jimklimov:
New driver - new man page :)

As for checks, it's about 3-5hrs for all scenarios to run; overall time depends on how maby other PR iterations are queued...

RikyPlaza:
@jimklimov Hi, now i see that the there are errors regarding the missing default label in switch statements, but i had to remove them in commit f19c8e0c87ec54e1d9dca6f05a611af5bb8f2405 since the checks were also failing. I really don't understand what it the issue... 
In other drivers (blazer_usb.c, clone-outlet.c ecc) i see that the default label is preset.

jimklimov:
@RikyPlaza : did you try the construct with pragmas per https://github.com/networkupstools/nut/pull/2689#issuecomment-2499909828 ?

RikyPlaza:
@jimklimov Hello, if i consider the driver blazer_usb.c i don't see any pragma directive. I'm a little bit confused on how those drivers, without any pragma construct, pass the checks related to the default labels in switch statements...

RikyPlaza:
@jimklimov In any case now i try to insert that statemets and commit the driver. Let's see :)

jimklimov:
Different compilers and their different warning settings require different rules to be satisfied, and are sonetimes at odds:
* a switch should definitively handle all cases, so must have a default label - works well for general numeric variables;
* an enum's valid values are known at compile time, and each must be handled explicitly (even if as many labels preceding same code block), so...
* ...a default label is redundant in a switch that handles all enum values - head-on crash vs. the first rule here, hence the pragmas against warnings at this spot and the default label to be sure, as the least-worst solution.

Also note fall-through cases (lack of break/return/continue/... before next label) - these are considered a likely oversight by a programmer and in fact often are. Explicit `goto superfluous_label` to the next line defining this label solves the issue portably.

RikyPlaza:
@jimklimov no idea why the checks failed :)

RikyPlaza (commented):
No idea on the fails.

jimklimov:
CI setup issue, mis-added new workers :\

PR #2692, opened 2024-11-25
First commit of NHS drivers/nhs-nut.c

Challado:
First commit of driver NHS-NUT for NHS senoidal lines.
Signed-off-by: Lucas Willian Bocchi <lucas@lucas.inf.br>

MAINTAINER UPDATE: Started from mailing list submission at https://alioth-lists.debian.net/pipermail/nut-upsdev/2024-November/008046.html


jimklimov:
MAINTAINER NOTE: The initial driver code has some Linux-specific bits and can get a number of portability complaints from the NUT CI farm agents running on other platforms. This will be addressed later as the PR evolves to get merged.

drivers/nhs-nut.c
```diff
@@ -0,0 +1,2070 @@
...
+    float vin_high_crit = 0;
+    float calculated = 0;
+    float vpower = 0;
+    float pf = 0;
+    int bcharge = 0;
+    int min_input_power = 0;
+    
+    upsinfo ups;
+
+    upsdebugx(1,"Start updateinfo()");
```
jimklimov, line 1614: Generally (in many methods), for methodical tracing you can use `__func__` macro, e.g. `upsdebugx(4, "%s: starting...", __func__);` is a common sight in NUT code (maybe with parentheses - more often to include significant argument values in the log).

drivers/nhs-nut.c
```diff
@@ -0,0 +1,2070 @@
...
+    upsdebugx(1,"Start cleanup()");
+    if (serial_fd != -1) {
+        close(serial_fd);
+        serial_fd = -1;
+    }
+    upsdebugx(1,"End cleanup()");
+}
+
+void upsdrv_initups(void) {
+    upsdebugx(1,"Start initups()");
```
jimklimov, line 2040: `upsdrv_initups()` should take care of hardware communications init (maybe fail with `fatalx()` if `port` is not accessible).

The `upsdrv_initinfo()` method is called later to detect the device details (maybe fail with `fatalx()` if no compatible device was found) and set write-once values like device model etc. and populate the list of supported commands and/or settable variables, if any.

Please do check `docs/new-drivers.txt` for general driver code structure overview.

drivers/nhs-nut.c
```diff
@@ -0,0 +1,2070 @@
...
+    baudrate = DEFAULTBAUD;
+    upsdebugx(1,"Start initinfo()");
+    if (b)
+        baudrate = atoi(b);
+    if (device_path) {
+        if (strcasecmp(device_path,"auto") == 0)
+            strcpy(porta,DEFAULTPORT);
+        else
+            strcpy(porta,device_path);
+        serial_fd = openfd(porta,baudrate);
```
jimklimov, line 1573: `upsdrv_initups()` should take care of hardware communications init (maybe fail with `fatalx()` if `port` is not accessible).

The `upsdrv_initinfo()` method is called later to detect the device details (maybe fail with `fatalx()` if no compatible device was found) and set write-once values like device model etc. and populate the list of supported commands and/or settable variables, if any.

Please do check `docs/new-drivers.txt` for general driver code structure overview.

jimklimov (changes requested):
First of all, thanks for a solid-looking driver that yields a lot of info. As a stand-alone program, it looks mostly good (some spots with C memory handling are commented, e.g. with `upsdata` returns).

Most of the requested fixes (maybe I'll handle at least some of those notes myself) would be required to ensure portability of this driver to other platforms than yours, and to improve maintainability by using more of common NUT methods and approaches - probably including the serial-port communications.

Some general concerns are about which logical code landed into which standard methods for NUT driver structure, as documented at `docs/new-drivers.txt` - fixes there would be needed, but probably simple ones, moving blocks of code into suitably named methods from wrong current locations.

Also a `NEWS.adoc` entry to announce the driver, and `data/driver.list.in` HCL entry (or several) to suggest use of this driver for some NHS device name(s), would be useful. You have a lot of names and VA values mapped in a method - perhaps these names can become the first pass of HCL data enhancement.

Also needs a manual page to describe the supported devices and the driver options (per `addvar()` in code) and maybe other nuances - known caveats, etc. Ideally this comes from a driver author, as they best know if there is anything special to know about the devices.

Finally, as "naming is one of the hardest problems in IT" (citation needed), `nhs-nut` would need to be changed, I think. There was a guideline (alas, mostly overlooked) to name new drivers as `nutdrv_something` to help identify them in process listing, and there are several precedents for that. There is also one `nut-*ipmi*` set of files. But none with `nut` in the middle/end...
I think following the precedent of `bcmxcp`, `blazer`, `riello` or `sms` drivers suffuxed with `_ser` and/or `_usb` for different media, e.g. `drivers/sms_ser.c` - this one deserves to be `nhs_ser`.

jimklimov:
Hopefully got through compiler formalities to get the driver buildable with fair level of warnings with gcc and clang on Linux.

Some of the review comments I've raised before are now solved, others are still valid concerns (notably `upsdrv_initups()` and `return upsdata`, and conversion to more common NUT approaches to work with serial ports). Portability to non-Linux not immediately addressed, will look at it later.

Updated the docs as I suggested above, too, and renamed to `nhs_ser` to match `sms_ser` and similar drivers.

jimklimov:
@Challado : while looking at code more closely, I see we store many data points in structures with types like `unsigned int` or `bool`, and then interact with serial-port wire data as `uint8_t` -- potentially losing something in conversions and adding run-time load to do so. Why not use same data types in struct to see overflows at compile-time, and (optionally) perhaps even mark with `align(1)` to just send our struct on the wire as is?

You mentioned testing with a PC and an ARM, so portability issues with such approach (memory layout vs. packet) can be detected/addressed this way, right?

jimklimov:
Quiescing on my side for now due to dayjob starting, and to let CI collect more complaints...

Challado:
> @Challado : while looking at code more closely, I see we store many data points in structures with types like `unsigned int` or `bool`, and then interact with serial-port wire data as `uint8_t` -- potentially losing something in conversions and adding run-time load to do so. Why not use same data types in struct to see overflows at compile-time, and (optionally) perhaps even mark with `align(1)` to just send our struct on the wire as is?
> 
> You mentioned testing with a PC and an ARM, so portability issues with such approach (memory layout vs. packet) can be detected/addressed this way, right?

Jim
We made several test in read/output data in any forms and the best job that I can do it's use a uint8_t that match with type of byte returned from nobreak. Don't ask me why, but I have tryed several different approaches to match binary info with protocolo standarts but the best that I can do is that. Try to move or change it can cause more difficulties in match the binary data again and my afraid changing it is the lack of data.

drivers/nhs-nut.c
```diff
@@ -0,0 +1,2070 @@
...
+        return serial_fd;
+}
+
+char * strtolow(char* s) {
+  for(char *p=s; *p; p++) *p=tolower(*p);
+  return s;
+}
+
+upsinfo getupsinfo(unsigned int upscode) {
+    upsinfo data;
```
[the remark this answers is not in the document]
jimklimov: I missed that this is a struct, and these *can* be returned. May be sub-optimal (copying it) though, so `malloc()` and returning a pointer may be still better: https://stackoverflow.com/a/76458184/4715872

jimklimov:
@Challado : note, large changes of raw C text have landed as NUT code style was applied and some comments reformatted.
Logging verbosity also revised.

It would be great if you can confirm that this code base state still builds and works for your device, before I get into some more significant refactoring to share some of your code with common `serial.c/.h` and vice versa.

jimklimov:
@Challado : upon the closer inspection, I found that you parse hwinfo details (`ups.mfr` etc.) for every run of `upsdrv_updateinfo()`. Is that information expected to change often/ever? Usually that is seeded in `upsdrv_initinfo()` once and left for the duration of the program. (Some devices and drivers, e.g. SNMP, have a notion of semi-static values - like admin contacts or device location - that are re-fetched rarely).

jimklimov:
NOTE: I've also constrained builds of this driver to be just on Linux for now (hopefully including builds of the man page, and added respective wording to the `NEWS.adoc` and the man page itself) - hopefully `git revert c857014d62c7e0a4a5c4675d78294386e1be0bb0)` some time later would be all that is needed to restore its glory (in some other PR after merging this one, using common "contributor-owned" git workflow without rubbing elbows in someone else's repository/branch :) ).

I've checked that this arrangement does not preclude `(cd drivers && make nhs_ser)` attempts on those other systems, so experiments to extend the platform support should be easy to iterate.

jimklimov:
NOTE: CI hiccups about openbsd are due to build agent reconfiguration.

jimklimov:
Probably some clean-up remains e.g. to reshuffle code between methods, improve portability, etc. - but as a first code drop that passes CI this is "good enough" to not keep the PR indefinitely open.

drivers/nhs-nut.c
```diff
@@ -0,0 +1,2070 @@
...
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ */
+
+#include "config.h"
+#include "main.h"
+#include "common.h"
+#include "nut_stdint.h"
+#include "serial.h"
+#include <stdio.h>
+#include <linux/serial.h>
```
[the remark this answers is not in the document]
jimklimov: UPDATE: As found later, initial code relies on `struct serial_struct` and `TIOCSSERIAL` which are only defined there. So platform expansion would need finding equivalents for whatever those bits solved.

Issue #2401, opened 2024-04-15
strange data in logs, `file_report_buffer: expected 2 bytes, but got 512 `


fenugrec:
(same  device and system discussed in #2399 )

looking at logs, I thought it strange that a request would return 512 bytes :
```
   54.285255        [D1] Quick update...           
   54.285266        [D4] Entering libusb_get_report     
   54.287751        [D3] Report[get]: (3 bytes) => 16 00 00                                                                                                                                   
   54.287793        [D2] Path: UPS.BELKINControls.BELKINDelayBeforeStartup, Type: Feature, ReportID: 0x16, Offset: 0, Size: 16, Value: 0
   54.287813        [D4] Entering libusb_get_report     
   54.290724        [D3] Report[get]: (3 bytes) => 15 00 00                                          
.... trimmed
   54.302802        [D4] Entering libusb_get_report
   54.304747        [D3] Report[get]: (2 bytes) => 23 10
   54.304789        [D2] Path: UPS.BELKINStatus.BELKINBatteryStatus, Type: Feature, ReportID: 0x23, Offset: 0, Size: 8, Value: 16
   54.304804        [D3] Report[buf]: (2 bytes) => 23 10
   54.304816        [D2] Path: UPS.BELKINStatus.BELKINBatteryStatus, Type: Feature, ReportID: 0x23, Offset: 0, Size: 8, Value: 16
   56.260381        [D1] upsdrv_updateinfo...
   56.268762        [D2] file_report_buffer: expected 2 bytes, but got 512 instead
   56.268810        [D3] Report[err]: (512 bytes) => 34 64 00 00 00 00 00 00 d0 12 2d 0b 1b 5d 00 00
   56.268835        [D3]  4d 8a c5 09 1b 5d 00 00 30 b6 76 c7 fd 7f 00 00 70 13 2d 0b 1b 5d 00 00 10
   56.268855        [D3]  00 00 00 00 00 00 00 0a 00 00 00 00 00 00 00 53 39 c5 09 1b 5d 00 00 10 00
   56.268877        [D3]  00 00 30 00 00 00 20 b8 76 c7 fd 7f 00 00 40 b7 76 c7 fd 7f 00 00 b8 22 03
   56.268897        [D3]  00 00 00 00 00 75 70 64 61 74 65 69 6e 66 6f 00 c7 fd 7f 00 00 00 00 00 00
[...]

PR #2701, opened 2024-11-29
drivers/libusb1.c: fix wrong return value of nut_libusb_get_interrupt()

fenugrec:
cf github #2401 

When global var 'interrupt_size' is not set, a transfer size of SMALLBUF (512) is requested when calling libusb_interrupt_transfer(). Our function nut_libusb_get_interrupt() should return the actual transferred size. Otherwise, the caller will attempt to parse any uninitialized data in the buffer after the response.

In libusb0.c this was done correctly by using the return value of usb_interrupt_read(). With libusb1, the transferred size is written to tmpbufsize, and should therefore be used as the return val.

***************

can we trust libusb to not return a value outside the range of `usb_ctrl_charbufsize` ? I did not add a check for this and simply cast the int. [edit - as noted in the review comments, this condition is already checked... 4 lines above)

### DIFF
--- a/drivers/libusb1.c
+++ b/drivers/libusb1.c
@@ -1090,7 +1090,7 @@ static int nut_libusb_get_interrupt(
 		) {
 			return -1;
 		}
-		ret = (usb_ctrl_charbufsize)bufsize;
+		ret = (usb_ctrl_charbufsize)tmpbufsize;
 	}
 
 	return nut_libusb_strerror(ret, __func__);


drivers/libusb1.c
```diff
@@ -1090,7 +1090,7 @@ static int nut_libusb_get_interrupt(
 		) {
 			return -1;
 		}
-		ret = (usb_ctrl_charbufsize)bufsize;
+		ret = (usb_ctrl_charbufsize)tmpbufsize;
 	}
```
jimklimov, line 1094: For the question in PR ("can we trust libusb...") -- I think the if-clause just above range-checks the `tmpbufsize` to fit in `usb_ctrl_charbufsize` range, so any problems should be caught there.

fenugrec: oh yes you're right, the || ensures that condition is also evaluated, I missed that.

Issue #2702, opened 2024-11-29
Update NUT HCL with Ippon models supported by `usbhid-ups`


jimklimov:
https://alioth-lists.debian.net/pipermail/nut-upsdev/2024-November/008058.html

> Hi NUT:
>
> We kindly request adding the UPS products produced by Cyber Energy for Ippon customers to the NUT Hardware Compatibility List. All Ippon UPS products listed below feature Cyber Energy's USB ID 0483:A430 and are supported by the USBHID-UPS(8) driver. Please assist in updating the Manufacturer (Ippon) section of the NUT Hardware Compatibility List by adding the following information to the Model and Driver fields. Thank you!
>
> Ippon Model & Driver Information to be Added to the Network UPS Tools - Hardware compatibility list at https://networkupstools.org/stable-hcl.html
>
> Manufacturer: Ippon
> Driver: usbhid-ups<https://networkupstools.org/docs/man/usbhid-ups.html>
> Models:
> * Pacific 1000/2000/3000
> * Ampere 1000/1500/2000/3000 bl
> * Innova RTB 1000/1500/2000/3000
> * Innova TBE 1000/2000/3000
> * Innova TB 1000/2000/3000
> * Back Power Pro II 500/600/800
> * Back Power Pro II Euro 650/850
> * Kirpich 650/850/1050
> * Smart Power Pro II 1200/1600/2200
> * Smart Power Pro II Euro  1200/1600/2200

MAINTAINER NOTE: The latter two are already in the list since 2023 - commit ba976a54c2 from #701 (with different USB IDs though).

PR #2703, opened 2024-11-29
data/driver.list.in: add usbhid-ups supported Ippon models

jimklimov:
Closes: #2702


jimklimov:
Applied to the generated HCL nut-website pages:

hcl-ippon-20241201

PR #2704, opened 2024-12-01
drivers/mge-hid.c: comment regarding ChargerType

desertwitch:
follow-up for #2660 
as explained in https://github.com/networkupstools/nut/pull/2660/files#r1838070677

The current state suggests we're sure about the principle and values of `UPS.BatterySystem.Charger.ChargerType` and that ABM-enabledness deductions could (theoretically) also be made from this HID path in the future. I've amended that in this PR, so that the current state of investigations is reflected in a separate in-depth comment, as to avoid future development going down a possibly wrong path due to what were perhaps premature assumptions on our part. 😉

### DIFF
--- a/drivers/mge-hid.c
+++ b/drivers/mge-hid.c
@@ -159,7 +159,6 @@ static char		mge_scratch_buf[20];
  * Synthesis table
  * HID data                                   |             Charger in ABM mode             | Charger in Constant mode | Error | Good |
  * UPS.BatterySystem.Charger.ABMEnable        |                      1                      |            0             |       |      |
- * UPS.BatterySystem.Charger.ChargerType      |                      4                      |            5             |       |      |
  * UPS.PowerSummary.PresentStatus.ACPresent   |           On utility          | On battery  | On utility | On battery  |       |      |
  * Charger ABM mode                           | Charging | Floating | Resting | Discharging | Disabled   | Disabled    |       |      |
  * UPS.BatterySystem.Charger.Mode             |     1    |    3     |   4     |      2      |     6      |    6        |       |      |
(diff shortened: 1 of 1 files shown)


masterwishx:
We already confirmed that `UPS.BatterySystem.Charger.ChargerType` is only for the info and `UPS.BatterySystem.Charger.ABMEnable` to switch ABM (Enable in HID path point to it) or you still not sure ?

in your case switching to 0 maybe bug in firmware?

masterwishx:
> The additional comments are just to document our investigation for the future (as it may be hard to find inside the very long PR discussion).

Cool , Sorry . By mistake from comment i was thinking you are not  sure if it info variable :)

masterwishx:
> I removed the variable from the ABM table (in the comments) because it made no sense to be there when we're not using it for ABM enabled/disabled (and just as informational variable).

Other values like : 
`UPS.BatterySystem.Charger.Mode`  etc also represent only info values !?
So `UPS.BatterySystem.Charger.ChargerType` represent info of charger type also some extensions .
so not sure why to remove it  :(

desertwitch:
> > I removed the variable from the ABM table (in the comments) because it made no sense to be there when we're not using it for ABM enabled/disabled (and just as informational variable).
> 
> 
> 
> Other values like : 
> 
> `UPS.BatterySystem.Charger.Mode`  etc also represent only info values !?
> 
> So `UPS.BatterySystem.Charger.ChargerType` represent info of charger type also some extensions .
> 
> so not sure why to remove it  :(

The other variables all have a practical effect on either ABM enabled/disabled or the UPS status itself (charging/discharging). Putting it into this table with them, like it was, is misleading because one could then assume that it can also be used to determine ABM enabled/disabled, when in fact it is only a textual informational variable (for now).

masterwishx:
> But maybe in 10 years someone will read this table having no idea about our previous discussions and make wrong deductions from

OK then I hope your comment about will be enough, I just wanted not to loose from table  this var that it represent type of charger...

PR #2706, opened 2024-12-03
core ipc: Handle select returning EINTR

gdt:
In the IPC code, select is called, and the return value checked.  If it's -1, the program aborts.   However, it is legal under POSIX for system calls to be interrupted.   They can be restarted, or the caller can get -1 with `errno` `EINTR`.

This patch changes the code to continue the loop and thus re-enter select, on `EINTR`.

With it, tests get much further on NetBSD 10 amd64.


jimklimov (approved):
Thanks for the investigation and PRs!

PR #2707, opened 2024-12-03
tests: Do not expect signals to be messages

gdt:
POSIX does not require 1:1 delivery per kill() call.  Adjust the test to not require behavior that 1) POSIX does not require and 2) is only probabilistically observed to happen, even if "most of the time".

The test previously sent signals as rapidly as possible, observed to be on the order of 100 us via ktrace.   Sometimes, a signal sent twice was delivered once.  This is legal, and it is an error to rely on this.

This commit adds comments discussing signal semantics.  It adds a delay between some signals, to allow time for delivery.   We consider a system which does not deliver signals within 1 s to be deficient.


jimklimov (approved):
Thanks for the investigation and PRs!

PR #1245, opened 2021-12-24
Fix #439 Incorrect Output Voltage on CPS*EFPCLCD

flashydave:
This fixes the problem whereby the Output Voltage on Cyberpower CPS*EPFCLCD UPS's is masked.
The root cause is due to a legal but inappropriate Report Descriptor published as part of the HID.
Specifically it looks liike Report ID 17 was deleted (ie Report 16 is followed by Report 18) with the result that there was no attempt to reset the LogMin and LogMax values from those used by the HighVoltageTransfer Item for those used by the Output Voltage.  

The fix conditionally corrects the Report Descriptor structure by applying the same LogMin and LogMax values used by the Input Voltage Item after first checking the issue exists.  

The fix creates a generic framework for fixing Report Descriptors that can be used for different manufacturers by adding code to the appropriate subdriver rather than polluting the main code with UPS specific exceptions.

PR #2718, opened 2024-12-12
`usbhid-ups` generally: apply disable_fix_report_desc…; `cps-hid`: fix mismatched LogMax between input/output voltages (bad encoding)

jimklimov:
…interruptsize and interruptonly before comm_driver->open_dev() [#1512, #1575]

Follows up from PR #1245 that addressed #439 and a ton of related issues.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -234,6 +234,8 @@ https://github.com/networkupstools/nut/milestone/11
    * ...should now not log "insufficient permissions on everything" alone when
      some devices were accessible but just did not match -- clarify that case
      in the next line, when applicable. [PR #2699]
+   * ...should now track the fact of `assumed_LogMax` (typically when firmware
+     encoding logic is wrong, and `-1` is resolved by parser). [#1512, #1040]
 
  - Introduced a new driver concept for interaction with OS-reported hardware
    monitoring readings. Currently instantiated as `hwmon_ina219` specifically
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -82,6 +82,9 @@ https://github.com/networkupstools/nut/milestone/11
      its report descriptors were being discovered. Now this flag, as well as
      `interruptsize` and `interruptonly`, are considered before we first try
      to open the USB device handle. [#1575, #1512]
+   * In `cps_fix_report_desc()` we intended to fix-up input and output voltages
+     in certain cases against high voltage transfer, we only fixed-up one of
+     them. [#1245]
 
  - development iterations of NUT should now identify with not only the semantic
    version of a preceding release, but with git-derived information about the
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -203,6 +203,9 @@ https://github.com/networkupstools/nut/milestone/11
      by default: `powercom_sdcmd_byte_order_fallback`. [PR #2480]
    * `cps-hid` subdriver now supports more variables, as available on e.g.
      CP1350EPFCLCD model, including temperature. [PRs #2540, #2711]
+   * in `cps-hid` subdriver, `cps_fix_report_desc()` method should now handle
+     mismatched `LogMax` ranges for input and output voltages, whose USB Report
+     Descriptors are wrongly encoded by some firmware versions. [#1512]
    * USB parameters (per `usb_communication_subdriver_t`) are now set back to
      their default values during enumeration after probing each subdriver.
      Having an unrelated device connected with a VID:PID matching the
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -79,7 +79,9 @@ https://github.com/networkupstools/nut/milestone/11
      indefinite wait [#1922, #2392, #2686, #2670]
    * The `disable_fix_report_desc` option introduced for `usbhid-ups` driver
      since NUT v2.8.1 was not applied for early dialog with the device while
-     its report descriptors were being discovered. [#1575, #1512]
+     its report descriptors were being discovered. Now this flag, as well as
+     `interruptsize` and `interruptonly`, are considered before we first try
+     to open the USB device handle. [#1575, #1512]
 
  - development iterations of NUT should now identify with not only the semantic
    version of a preceding release, but with git-derived information about the
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -77,6 +77,9 @@ https://github.com/networkupstools/nut/milestone/11
      use of `struct timeval={-1,-1}` as a trigger to `select(..., NULL)`,
      as logged in one part of code and not handled in the other, for the
      indefinite wait [#1922, #2392, #2686, #2670]
+   * The `disable_fix_report_desc` option introduced for `usbhid-ups` driver
+     since NUT v2.8.1 was not applied for early dialog with the device while
+     its report descriptors were being discovered. [#1575, #1512]
 
  - development iterations of NUT should now identify with not only the semantic
    version of a preceding release, but with git-derived information about the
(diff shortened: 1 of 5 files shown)


jimklimov:
Awaiting end-user testing as detailed in https://github.com/networkupstools/nut/issues/1512#issuecomment-2541291060

AlexJakeGreen:
Result of `/lib64/nut/usbhid-ups -DDDDDD -d1 -a ups`:
[usbhid-ups_DDDDDD_d1_a_ups.txt.gz](https://github.com/user-attachments/files/18149734/usbhid-ups_DDDDDD_d1_a_ups.txt.gz)

Both `input` and `output` are wrong:
```
$ zgrep -E '(input|output).voltage:' usbhid-ups_DDDDDD_d1_a_ups.txt.gz 
input.voltage: 0.7
output.voltage: 0.7
```

I guess I collected observations properly ↓↓↓↓

Result of `/lib64/nut/usbhid-ups -DDDDDD -d1 -x disable_fix_report_desc -a ups`:
[usbhid-ups_DDDDDD_d1_a_ups_x_disable_fix_report_desc.txt.gz](https://github.com/user-attachments/files/18149735/usbhid-ups_DDDDDD_d1_a_ups_x_disable_fix_report_desc.txt.gz)
`input` is correct, `output` is wrong:
```
$ zgrep -E '(input|output).voltage:' usbhid-ups_DDDDDD_d1_a_ups_x_disable_fix_report_desc.txt.gz 
input.voltage: 225.4
output.voltage: 20.6
```


`ups.conf`:
[ups.conf.gz](https://github.com/user-attachments/files/18149737/ups.conf.gz)

jimklimov:
@AlexJakeGreen: thanks again for the logs, I caught the bug I think (got distracted and mixed up the languages, mindlessly used `^` as power-of operator when it was not). 

Can you please re-check with the descriptor fix-up?

Please note that I had to force-push the changes after an update from `master` branch, so you'd likely need to remove and recreate the build workspace, fetching the new state of the PR source branch.

Or if you're comfortable with git, try something like this to change the workspace files "in place":
````
:; git fetch --all && git reset origin/issue-1575 && git checkout -f
````
````

**Comment by nbriggs** [comment]
@jimklimov --
> Now the code should limit LogMax=32767 for the 16-bit signed value (CC @nbriggs - I hope this is the most correct guess when we imagine things in lieu of a good statement in device reports, e.g. 65535 would be wrong, right?)

If it's a signed value, yes, the range would be -32768 to 32767, but I haven't got the (mental) bandwidth at the moment to dig into this bug.

This all makes me wish the UPS code developers were able to read the USB HID spec correctly - or that there were a freely available small library of code they could all use to implement their side of the protocol.

I'm also left wondering if NUT needs a bigger hammer than fixing individual report descriptors, something like once it has identified the particular UPS it is talking to...  completely ignore the report framework provided by the unit and substitute a "correct" one (from an editable data file) against which the individual report values from the unit will decode to the proper values.

**Comment by jimklimov** [comment]
@nbriggs : thanks for the comment, and yes a loadable/editable logic might help. Eventually "DMF" merge should bring the transport for loadable mappings (initially for SNMP, with an eye for USB mappings later), but making something like it work for such nuances of USB reports would be a hassle of its own.

As for signedness, can you please refresh my memory - are USB values always signed, or can be unsigned too, and how can we learn that? Is it part of the report where we read `LogMax == -1` and decide to discard that chain of all-1 bits in favor of 255 or 65535 as the value? So in the particular case of making assumptions, do we know or not if the value was really meant to be signed or not?

**Comment by jimklimov** [comment]
Sounds great, thanks for checking!

```
   0.010763	HIDParse: LogMax is less than LogMin. Vendor HID report descriptor may be incorrect; interpreting LogMax -1 as 65535 in ReportID: 0x0f
   0.010795	HIDParse: LogMax is less than LogMin. Vendor HID report descriptor may be incorrect; interpreting LogMax -1 as 255 in ReportID: 0x12
   0.010811	HIDParse: LogMax is less than LogMin. Vendor HID report descriptor may be incorrect; interpreting LogMax -1 as 255 in ReportID: 0x13
   0.010971	Using subdriver: CyberPower HID 0.82
   0.010986	[D3] Attempting Report Descriptor fix for UPS: Vendor: 0764, Product: 0601
   0.011000	[D4] Original Report Descriptor: output 0x12 LogMin: 0 LogMax: 255 (assumed: yes) Size: 16
   0.011011	[D4] Original Report Descriptor: input 0x0f LogMin: 0 LogMax: 65535 (assumed: yes) Size: 16
   0.011021	[D3] Fixing Report Descriptor: set Input Voltage LogMax = 32767
   0.011031	[D3] Fixing Report Descriptor: set Output Voltage LogMax = 32767
   0.011040	[D2] Report Descriptor Fixed
   0.011050	[D1] 73 HID objects found
```

nbriggs:
This is what the spec says (page labelled 19 of hid1_11.pdf, physical page 29 of 97) --

5.8 Format of Multibyte Numeric Values
Multibyte numeric values in reports are represented in little-endian format, with the least significant byte at the lowest address.
The Logical Minimum and Logical Maximum values identify the range of values that will be found in a report. **If Logical Minimum and Logical Maximum are both positive values then a sign bit is unnecessary in the report field and the contents of a field can be assumed to be an unsigned value.** **Otherwise, all integer values are signed values represented in 2’s complement format.** Floating point values are not allowed.

nbriggs:
The number of bytes in the encoding of the LogMin and LogMax fields is only loosely tied to the size of the field that they are describing - but the implementers on the UPS side don't seem to quite get that.
It's all starting to come back to me...

If you're trying to describe a report field that is 16-bits and has (unsigned) values from 0..65535 then you *should* have a LogMin field containing value 0, and a LogMax field that contains value 65535.

SInce all numeric fields are interpreted as signed twos-complement values (except for that note above about the report **values**, not the values in the report description), to encode such a LogMax field you would have to express the **LogMax** field in a 4-byte encoding in the report description.   That's independent of the 2-byte **report value** that this LogMin and LogMax are describing.

Some coder at the UPS company took a shortcut, and set LogMin = 0, LogMax = -1 (because they used a 2-byte encoding with all bits set), and then NUT is left to decide what they actually intended.  _My_ interpretation of that is that they're trying to say 0..65535, because if they had meant 0..32767 they would have written that (0..7FFF which fits in the signed 2-byte repr.) but unless they're actually trying to represent voltages over 327 V deciding it's 32767 should be fine.

jimklimov:
Sounds plausible. I think I'll pick this up for a comment in LogMax fixup.

As for 327V... maybe 3-phase devices can claim 420V. I guess we'd see if there are any practical bug reports against 32767 (or their makers get specs right).

Timi7007:
> As for 327V... maybe 3-phase devices can claim 420V. I guess we'd see if there are any practical bug reports against 32767 (or their makers get specs right).

In the 60Hz part of the world it's a different story and I'm not an expert, but in our European 50Hz grid single phase is 230 V ± 10%/23 V so between 207V and 253V. Three-phase used to be 380V and 415V, they now met in the middle at 400V ± 10% setting the maximum to 440V.
Since the UPS should kick in even if the grid values are violated (e.g. overloaded generator) everything lower than these should also be functional. This coincidentally includes the 120V single-phase three-leg 60Hz US-power and it's three-phase potentials of 208V and 415V. Even the lowest international grid voltage of Japan's 100V.

That's just power theory though and I don't have a clue about USB drivers.

jimklimov:
Ok, so to err on the safe side, I'll assume 65535 for the 16-bit maximum (=>655V)?

Or... wait... with `Exponent = -1` we count in 0.1V not 0.01V - so a 32767 value is 3276,7V anyway? Did someone up in the thread try to confuse me? ;p

Timi7007:
I know of industrial setups with 690V three-phase power with 400V on each phase but haven't heard of UPSes for those and would assume anything that high would be over-voltage for everything used in IT.
I also don't know how NUT handles three-phase UPSes. Realistically measuring each phase and calculating total voltage from that would word just fine and may be somewhat more useful for diagnostics. So again no idea if 65535 would be required, I was just trying to say that 420V might not be unrealistic.

Timi7007:
My test was successful: https://github.com/networkupstools/nut/issues/1512#issuecomment-2548526211

PR #2736, opened 2024-12-23
Windows cross build documentation

RikyPlaza:
Updated documentation for cross compilation of nut project with windows target. 

The steps have been tested on Ubuntu 22.04 virtual machine (VirtualBox) but the binaries have not been tested yet.

### DIFF
--- a/drivers/cps-hid.c
+++ b/drivers/cps-hid.c
@@ -388,11 +388,16 @@ static int cps_fix_report_desc(HIDDevice_t *pDev, HIDDesc_t *pDesc_arg) {
 						"LogMin: %ld LogMax: %ld",
 						input_logmin, input_logmax);
 
-					pData->LogMin = CPS_VOLTAGE_LOGMIN;
-					pData->LogMax = CPS_VOLTAGE_LOGMAX;
-					upsdebugx(3, "Fixing Report Descriptor: "
-						"set Input Voltage LogMin = %d, LogMax = %d",
-						CPS_VOLTAGE_LOGMIN, CPS_VOLTAGE_LOGMAX);
+					/* TOTHINK: Should this be still about
+					 * the *HIGH* Voltage Transfer? Or LOW?
+					 */
+					if (hvt_logmin == input_logmin && hvt_logmax == input_logmax) {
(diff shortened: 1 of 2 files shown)


jimklimov:
Thanks, although I think a few typos for spellchecker to catch remain, some lines are too long, and I'm not sure I agree with libneon change.

jimklimov:
Actually, upon review most of the suggested fixes look questionable. For example:
* Repetitive `PKG_CONFIG_PATH` is not needed if you re-use the same terminal where you have already set the variables according to big note in the beginning.
* Explicit installation of `pkgconf` is also not needed if you followed `docs/config-prereqs.txt` and installed the needed toolkits.
* Not sure which changes VirtualBox Guest Additions would bring that prerequisites don't already deliver. In context of font library I suppose something from X11 about graphics core?.. Notably, as you build something for another architecture, the headers, libraries and tools made for the build host itself are not all directly impactful.
* Likewise, `PATH=${PREFIX}/bin:...` would include the location of Windows `*.exe` programs that your build host is not necessarily able to execute.
* Explicit `libtoolize` etc. are not needed as you call the `autogen.sh` just afterwards, which involves them.
* The CLI options added for `ci_build.sh` are not supported by the script.

Maybe something else... 

Still, this PR did highlight some parts that could be more understandable - so thanks! I'll cherry-pick some of the changes into a new PR and add some clarifications, but would rather close this one.

RikyPlaza:
@jimklimov Hi!
Regarding the re-definitions of the environment variables you're totally right, i inserted them since i was not fully dedicated to this topic and i had to work in spots... In any case i don't think it's dangerous to have them
Regarding the `PATH=${PREFIX}/bin:...` yes, but the libtool binaries are installed in that location and would not be reachable if not exported...
the explicit libtoolize has been inserted since the autogen.sh script is was not working on my system, maybe something related to the fact that i'm using a virtual machine

using the commands reported in the docs to cross build the software caused every time some errors, this is the only way i got it working.

Great for the cherry picks! Hope that at least something in this PR can be usefuk :)

Tomorrow i'll be back in the office, hope to have chanche to test the software on windows as soon as possible.

Regards

PR #2740, opened 2024-12-30
WIP: Ecoflow River3+ and Delta3+ support

ykuksenko:
Attempting to add EcoFlow support in response to https://github.com/networkupstools/nut/issues/2735. Specifically Delta 3 Plus and River 3 Plus models.

Working:
- Basic status works (charging, discharging, online, onbattery, ...).
- Charge level works.
- Discharge limit from app shows correctly. **Load turns off at battery.low with current mapping.**

Not working:
- Any sort of control does not seem to work for load or beeper/light. Or I do not understand how to do it.

Model differences:
- Delta 3 Plus (firmware V6.31.48.29):
  - Has a beeper; it does not trigger on power loss.
  - Seems to lose USB connection randomly, needing to replug and/or restart software. (Both upsc and usbhid-scripts saw this)
- River 3 Plus (firmware V1.32.75.41):
  - Has a light instead of beeper; it does trigger on power loss.
  - Did not see USB connection issues on River 3 Plus but did less testing with it.

Attribute Details/Questions:

1. Where should the `HU_FLAG_QUICK_POLL` flag be used? The boolean values all seem to be alarm and status related. Does this look okay?
2. Code comments show attempted commands to control beeper and load. Please correct if I missed something there.
3. `DesignCapacity` and `FullChargeCapacity` metrics seem to be percent. Are the mappings okay to use despite NUT docs wanting these in Ah?
4. `flow.[4].configactivepower` attribute is close to unit Wh ratings. It matches on Delta 3+ but is not correct for River 3+. What should Wh metrics be mapped to if anything? Maybe use to calculate/estimate `DesignCapacity`/`FullChargeCapacity`.
5. `RemainingCapacityLimit` attribute shows the percentage at which the load will be turned off. Is this okay to map to `battery.low`? It is definitely a useful thing to have, just not sure if with that mapping.
6. `RunTimeToEmpty` does not take `RemainingCapacityLimit` in to account. How should the user be informed of this? The unit is in minutes, does this need to be converted to seconds or can it be left as is?

---

## General points

  comments or text documents.

  case.

- [ ] ~~Did not extend obsoleted drivers with new hardware support features
  except the new `nutdrv_qx` which should cover them all).~~

- [ ] ~~For updated existing device drivers, bumped the `DRIVER_VERSION` macro
  or its equivalent.~~

  concept.

- [ ] ~~Updated `docs/acknowledgements.txt` (for vendor-backed device support)~~

  the changed codebase.
  *LGTM.COM* looks to be shutdown as of Dec 16 (2022?).*


kirashi3:
Thanks for getting the ball rolling on this! I only recently found out that [Ecoflow is planning to support its Power Manager](https://manuals.ecoflow.com/eu/product/power-manager) software for both the River 3 and Delta 3 series of portable batteries, though the Delta 3 seems delayed until "end of February."

I've got a Synology DS918+ currently powered by a Delta 3 Plus that I'd be happy to test with, if necessary. I'd just need to dig out a USB-C to USB-A cable (unless perhaps they're supporting UPS power management over SNMP?), and receive some instruction on modifying Synology's DSM via SSH, if that's what it takes.

Either way, I'ma subscribe to see where this leads, as it'd be super cool to program our NAS hardware to gracefully shutdown once an Ecoflow battery reaches whatever low-capacity threshold we specify.

jimklimov:
This PR source was assimilated into #2837, thanks for the extensive comments about investigation results!

kpfleming:
I can confirm that the Delta 3 (Delta 3 Max Plus in my case) does randomly lose the USB connection, and the only way to restore it is to power-cycle the Delta unit which is rather inconvenient when it is powering a rack full of IT equipment! I'll report this to EcoFlow, since at least one of their other products does not have this problem.

PR #2744, opened 2025-01-02
Fix linker warning for C++ tests

jimklimov:
Avoid linker warning with some C++ test programs, e.g.:
````
              CXXLD    cppnit
            ld: warning: file /export/home/abuild/nut/obj/nut-2.8.2.1775.28/_build/sub/clients/.libs/libnutclient.so:
              linked to ../clients/.libs/libnutclient.so: attempted multiple inclusion of file
````

NOTE: Two fixes are attempted here, but they misfire differently on different platforms/toolkits.
Maybe the current status-quo is the least-worst situation...


jimklimov:

PR #2756, opened 2025-01-06
Feature improvements for `upslog`

jimklimov:
Per NEWS:

* If you use the legacy CLI options for single-system logging (`-s <system>` and `-l <logfile>`) along with newer tuple(s) for multiple-system logging (repeatable `-m <system,logfile>`), previously the single-system options were overridden by the tuple(s); now they become part of the list.
* The `upsname` in the `system=upsname[@hostname[:port]]` parameter may be an asterisk `*` to query for devices currently served by the hostname.
* Same log file may safely be used in different logging tuples (it is then recommended to use `%UPSHOST%` in a custom formatting string).
* Fixed printing of `%UPSHOST%` when multiple systems are being logged.
* A `%t` for a TAB character can now be used in the formatting string.
* Added `-D` for debugging (and foregrounding by default), like with  other NUT daemons.

Also addresses warnings of new clang-19 because GH builders began using it.


jimklimov:
CI faults unrelated to the changes: CodeQL workers apparently got a clang update and a new wad of warnings.

Issue #1298, opened 2022-02-15
NUT `sudo make install` creates paths not chown'ed to run-time user accounts


jimklimov:
(Related to #1296 and #1297)

Filesystem objects created by plain `make install` remain owned by the user who made them (unprivileged in test installs to a scratch DESTDIR, root in "custom production" builds). NUT daemons are pre-configured to run as an unprivileged account (`nobody:nogroup`, `nut:nut`, etc.) and corresponding paths get that ownership (and `chmod` access constraints) in package metadata - but not in a plain local installation that bypasses packaging, so e.g. `/var/state/ups` owned by `root:root` (and in fact not created by `make install`) can not house state, pid (#1299) and pipe files for drivers and upsd; the /usr/local/ups/etc is not protected from the world reading it, etc.

Proposed solution: add a recipe like `make install-own` (depends on `install`, and separated to avoid conflicts for existing use-cases in dist, packaging, etc.) that would try to apply ownership+permissions according to `configure` script choices. Or revise if existing `install` recipe tries to do the right thing but for some reason fails to (without faulting the `make` call)?..

jimklimov:
PR above should solve this, even with service (re)start attempts for a number of platforms. More ideas popped up during that work, delayed to 2.8.4 as it is expected to deal with vanilla/reference packaging recipes - which crosses paths with this issue.

Issue #1209, opened 2021-11-28
CI/Packaging: Want integration with OpenSUSE Build Service for master branch


jimklimov:
Numerous projects, including components under zeromq that served as role models for NUT CI with cross-pollination over time, benefit from integration with https://build.opensuse.org/ and both having recent iterations available in package repositories made for multiple (Linux) distributions, and checking (github statuses and all) that the packages are buildable for those distributions.

I suppose this would need bringing NUT-provided "reference" packaging into shape (like `make package` already implemented for some OSes), possibly drawing inspiration from DEB and SPEC files for those target distributions as shipped by them already, or from 42ITy fork which developed a means to build on a private OBS instance: https://github.com/42ity/nut/tree/FTY/obs

Note that special recipe file naming may be involved to ship numerous SPEC files that would be used by OBS for this or that distribution.

I am not sure that NUT as a project should "insist" on replacing distro-specific packages: after all, tight and "correct" integration with the OS lifecycle is better done by domain experts, but providing a way for users and testers to easily install a standardly-built replacement without guessing configure parameters every time would be cool. And it might help co-ordinate the sprawl of NUT packages (names, content) across the board - letting each distro shine where it is best, and letting NUT seem the same everywhere :)

In a way this follows up from #869

jimklimov:
For further reference: OBS docs on CI/CD integration: https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.scm_ci_workflow_integration.html and a practical example from ZeroMQ community at https://github.com/zeromq/gsl/commit/3f113aca129f72a06ba2d9b63ccff52f6119c304

UPDATE: Also OBS blogs directly on this matter, like:
* https://openbuildservice.org/2021/05/31/scm-integration/
* https://openbuildservice.org/2021/06/14/scm-integration-improvements/

...and ultimately product documentation:
* https://openbuildservice.org/help/manuals/obs-user-guide/cha-obs-scm-ci-workflow-integration#sec.obs.obs_scm_ci_workflow_integration.setup.token_authentication.how_to_authenticate_scm_with_obs

jimklimov:
Finally got some progress about this, sufficient to iterate the old DSC and SPEC recipes into something useful for today's set of deliverable files. Thanks to @bluca @adrianS and others for hints. Some highlights:

[...]

PR #2761, opened 2025-01-10
Makefile.am, configure.ac, NEWS.adoc: implement `make install-as-root` [#1298]

jimklimov:
Addresses: #1298

More or less follows documentation for manual NUT installation, wiki suggestions and what some existing packaging scripts do. Should help in-place upgrades of various nature; maybe initial installations too (user/group should be pre-made)...

After doing the work and some thinking, I believe this is a step in direction of providing reference packaging recipes for NUT in the codebase (see #1209 etc.) and the code added to `Makefile.am` now is better pushed out into scripts like `scripts/Solaris/postinstall` eventually - with better guaranteed availability of tooling, use of correct commands, etc. especially as this gets routine ported to various distributions.

### DIFF
--- a/INSTALL.nut.adoc
+++ b/INSTALL.nut.adoc
@@ -306,11 +306,30 @@ Installation
 
 [NOTE]
 =====================================================================
+You should now gain privileges for installing software if necessary, e.g.:
 
-you should now gain privileges for installing software if necessary:
+	su -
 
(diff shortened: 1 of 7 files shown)


jimklimov:
CI issue with openbsd is a build agent hiccup

PR #2779, opened 2025-01-21
feat: add SNMP support for Baytech RPC3-NC PDUs

noahajac:
Some models of Baytech RPC3-NC PDUs do not have a value at ".1.3.6.1.4.1.4779.1.3.5.2.1.24.1", causing the sysOID match to fail. Added a 2nd MIB that looks at ".1.3.6.1.4.1.4779.1.3.5.5.1.3.2.1" instead.

This patch would fall under the "various RPC" entry in the hardware compatibility list, thus the list does not need updating.

I was unable to run `make distcheck`, it appears to require quite a few dependencies that are not on my system.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -745,6 +745,10 @@ https://github.com/networkupstools/nut/milestone/10
  - snmp-ups driver:
    * added support for Eaton EMP002 sensor for ATS16 NM2 sub-driver. [#2286]
    * mapping table updates for apc-mib sub-driver. [#2264]
+   * added support to monitor BayTech RPC3-NC PDUs, with `baytech-rpc3nc-mib`
+     serving same basic data points as were available in `baytech-mib.c`,
+     but checking for a different model OID subtree and different OIDs for
+     the device model information. [#2779]
 
  - usbhid-ups driver:
    * `arduino-hid` subdriver was enhanced from "initial bare bones" experimental
--- a/docs/nut.dict
+++ b/docs/nut.dict
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 3279 utf-8
+personal_ws-1.1 en 3281 utf-8
 AAC
 AAS
 ABI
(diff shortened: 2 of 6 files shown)


jimklimov:
Cheers, and thanks for the contribution. If the mappings are all the same, you can get away defining an additional `mib2nut_info_t` in the same old source, see how e.g. `ietf-mib.c` is used for Tripplite MIB as well (matched by its own `TRIPPLITE_SYSOID`).

As for the distcheck, we also have a `make distcheck-light` which auto-detects the available dependencies instead of requiring them,  and should not insist on building docs either (I think).

noahajac:
@jimklimov 
> Cheers, and thanks for the contribution. If the mappings are all the same, you can get away defining an additional `mib2nut_info_t` in the same old source, see how e.g. `ietf-mib.c` is used for Tripplite MIB as well (matched by its own `TRIPPLITE_SYSOID`).
> 
> As for the distcheck, we also have a `make distcheck-light` which auto-detects the available dependencies instead of requiring them, and should not insist on building docs either (I think).

Hi there. I initially tried to implement a 2nd `mib2nut_info_t` as you described, however `match_model_OID()` was failing. The OID for device.model/ups.model is different on the device I have and the one in the original MIB. I could change device.model/ups.model to a generic OID such as `.1.3.6.1.2.1.1.1.0`, which I assume should be present on any Baytech device, but wasn't sure how proper that would be. 

`make distcheck-light` gives me errors regarding libusb drivers. Installing the devel package for those then leads to missing neon lib errors. If it is strictly necessary, I can go through the deps and install them.

Thank you!

jimklimov:
Well, I guess passing distcheck on your side is not strictly required - it would try and might fail on CI side, after a few hundred CPU hours across different platforms, which is the authoritative answer for the ultimate merge anyway.

Passing (or especially failing) it locally in one use-case just speeds up that part.

Issue #2183, opened 2023-11-14
PyNUTClient needs `telnetlib` which is deprecated


jimklimov:
Raised in #2181 discussion - after Python-3.13 we will need another implementation for the layer of text-based interaction over TCP in Python.

aquette:
side notes: while playing with Home Assistant NUT addon, I'm thinking about improving a bit pyNUT... :D
scan, device dump esp., possibly the cmd cookies (not sure it supports...)
cheers @jimklimov ;)

Mausy5043:
Although I currently have no time to work on this (again), I'd like to point anyone who does to [this repo](https://github.com/Mausy5043/python-nut3). I've deprecated it a month or so ago. But the implementation of `pexpect` to replace `telnetlib` was completed in that repo. Hope this helps.

opoplawski:
We just started development on Fedora 41 which will ship with Python 3.13.  So it would be nice to see this addressed relatively soon.  The above comment seems promising.

Also FWIW:
According to https://docs.python.org/3.13/whatsnew/3.13.html:
PEP 594: Remove the telnetlib module, deprecated in Python 3.11: use the projects telnetlib3 or Exscript instead.
(Contributed by Victor Stinner in gh-104773.)

jimklimov:
Took some time to investigate telnetlib3, but it seems too great a paradigm change. With original PyNUT code, we open a client connection to the NUT server once, and in different methods post this or that request and parse replies. With telnetlib3+asyncio, it seems that each interaction must be a completely (pre-)defined I/O loop, so a separate connection, login, query, response, exit for each bit of work. Lots of avoidable extra server stress too, I guess.

jimklimov:
It seems that one recommended approach per https://discuss.python.org/t/pep-594-take-2-removing-dead-batteries-from-the-standard-library/13508/59 is to just copy `telnetlib.py` from a distro which shipped one, and use as part of the consuming project that is hard to update otherwise. At least, this can be our first step in the area (until someone comes up with a more modern port).

Although examples at one of the recommended replacements, https://github.com/knipknap/exscript/ / https://exscript.readthedocs.io/en/latest/ seem to be conceptually similar to what PyNUT worked like.

opoplawski:
Just a heads up that Python 3.13b2 has now landed in Fedora Rawhide so NUT is failing to build.

jimklimov:
[...]

PR #2792, opened 2025-01-31
Socketify pynut

cgarz:
As discussed in #2183 I have updated `PyNUT.py.in` to no longer require `nut_telnetlib.py`.

The update required only minimal changes to how the client sends and receives data from upsd.
Essentially, I just went through the file and switched all `self.__srv_handler` `write` to `send` and all `read_until` calls to a new function. The only other thing is fixing the timeout init argument.

Testing has been performed by simply adding:
```python
if __name__ == '__main__':
    nut = PyNUTClient(debug=True, login='upsuser', password='upspass')
    ups_name = nut.GetUPSNames()[0]

    GetUPSList =        nut.GetUPSList()
    GetUPSNames =       nut.GetUPSNames()
    GetUPSVars =        nut.GetUPSVars(ups=ups_name)
    CheckUPSAvailable = nut.CheckUPSAvailable(ups=ups_name)
    GetUPSCommands =    nut.GetUPSCommands(ups=ups_name)
    GetRWVars =         nut.GetRWVars(ups=ups_name)
    SetRWVar =          nut.SetRWVar(ups=ups_name, var='driver.debug', value='1')
    RunUPSCommand =     nut.RunUPSCommand(ups=ups_name, command='driver.reload')
    DeviceLogin =       nut.DeviceLogin(ups=ups_name)
    FSD =               nut.FSD(ups=ups_name)
    _help =             nut.help()
    ver =               nut.ver()
    ListClients =       nut.ListClients(ups=ups_name)
```
To the bottom of the script and running in pycharm debug mode to note the contents of all the variables and noting the logs/effect on the server.


jimklimov:
Sounds great, thanks!

For FSD, you can extend the testing setup in tests/NIT/nit.sh (called as `make check-NIT` in CI). There are dummy drivers and a dummy upsmon that just logs its actions so the script can review its log file to check the events it saw.

The stack runs on dedicated random ports, can use custom SHUTDOWNCMD, and not running as root, so it won't actually shut down the computer.

It does also call the python and C++ client tests. With FSD I think the Python client would be essentially another upsmon by role, so might need using that login? I'm at a conference now so can't think more deeply on that.

To develop tests, see NIT_CASE envvar and `make check-NIT-sandbox-devel &` (wait some 15 sec and follow printed comments to source the generated env file).

jimklimov:
I guess it's okay then.

jimklimov:
From CI failures, it seems the solution's syntax is for Python 3. Can you please rewrite it so the module would work on both new systems and those stuck with 2.7 or so?

Per https://stackoverflow.com/a/73333155/4715872 seems this would just need to dumb down the code a little and write two lines instead of one (assign and check condition separately).

````
  File "/dev/shm/jenkins-nutci/nut_nut_PR-2792@2/nut-2.8.2.2040.3/_build/sub/scripts/python/module/test_nutclient.py", line 6, in <module>
    import PyNUT
  File "/dev/shm/jenkins-nutci/nut_nut_PR-2792@2/nut-2.8.2.2040.3/_build/sub/scripts/python/module/PyNUT.py", line 122
    while (data_end_index := data.find(finished_reading_data)) == -1:
                          ^
SyntaxError: invalid syntax
````

jimklimov:
Thanks! The changelog proposal looks right. For NEWS.adoc, see one of the topmost section with (planned) changes between 2.8.2 and 2.8.3. There could be an entry about providing nut_telnetlib - if it was after 2.8.2 already, can be replaced :)

jimklimov:
Just love to see how far a bit of know-how can go! :)

jimklimov:
Might be, not online now to check more.

jimklimov:
In NEWS, PR/issue numbers can be comma-separated.

jimklimov:
I guess, any mentions of `git grep nut_telnetlib` in recipes, if the file should not be delivered (or git-tracked) anymore?

jimklimov:
Yes please, if you have a moment to perfect the PR :)

jimklimov:
I've updated the `configure.ac` to use the same logic block here as was in NUT v2.8.2, just replaced the check for `telnetlib` with a check for `socket` (so we are sure both the python interpreter is usable, and the module we actually need is available - and remains in core/distro for future NUT builds).

Local `./ci_build.sh` runs that did fail for me with your branch tip (`configure: error: Prerequisites for PyNUT not found, can't install as required for NUT-Monitor desktop application`) passed now, so I've pushed the change here.

jimklimov:
Thanks a lot, merging!

jimklimov:
Well, sadly or not, but development entails more than just a bit of great coding :)

jimklimov:
FWIW, current state of the module should be uploaded to https://test.pypi.org/project/pynutclient/2.8.2.2132/ now, so can be tested as a standalone module.

PR #2798, opened 2025-02-09
Q2 and Q6 protocols implementation

dvdesolve:
Deeply sorry for a long delay, but finally, here is basic implementation of Q2 and Q6 protocol variants. Because of that I've also narrowed down `INNOVART31` protocol to support just this specific model.

If I would be able to get Innova RT 3/3 model I'll make special variant for that too (just want to be on safe side and test things before public rollout), but as for now new Q2/Q6 proto should suffice for these models


jimklimov:
Great thanks! I've updated some files "cosmetically", but the PR looks reasonable and solid otherwise. Hope CI agrees soon-ish :)

dvdesolve:
I'll also try some extra tests tomorrow on our Innovas

jimklimov:
Thanks!

Issue #2708, opened 2024-12-03
Clean-up or legalise: revise `status_set()` values vs. the NUT standard dictionary


jimklimov:
The "Status data" section in `docs/new-drivers.txt` defines certain keywords that are "Possible values for status_set", stressing that "Anything else will not be recognized by the usual clients.  Coordinate with the nut-upsdev list before creating something new, since there will be duplication and ugliness otherwise."

In fact we do have a number of other values in code, currently:
````
:; git grep -w status_set | sed 's,^.*\(status_set.*\);.*$,\1,' | sort | uniq -c
      1 docs/new-drivers.txt:   status_set(val) -- add a status word (OB, OL, etc)
      1 docs/new-drivers.txt:Possible values for status_set:
      1 drivers/dstate.c:void status_set(const char *buf)
      1 drivers/huawei-ups2000.c: * be set via status_set(), those are listed in "status_name".
      1 drivers/huawei-ups2000.c: * for status_set(), for example, "OL" or "OB", the field
      1 drivers/usbhid-ups.c:/** Helper to both status_set("CAL") and track last_calibration_start timestamp */
      6 status_set( status_buf )
      1 status_set("ACFAIL")
      1 status_set("ALARM")
     31 status_set("BOOST")
      1 status_set("BOOST");    break
      1 status_set("BY")
     21 status_set("BYPASS")
     12 status_set("CAL")
     17 status_set("CHRG")
      2 status_set("COMMFAULT")
      1 status_set("DEPLETED")
     13 status_set("DISCHRG")
      1 status_set("ECO")
      5 status_set("FSD")
      3 status_set("HB")
      1 status_set("HB");   break
      1 status_set("HE")
      1 status_set("LB ")
     59 status_set("LB")
      1 status_set("LB");       break
     62 status_set("OB")
      1 status_set("OB");   break
     30 status_set("OFF")
      1 status_set("OFF");  break
      1 status_set("OL BYPASS")
     66 status_set("OL")
      1 status_set("OL");   break
     31 status_set("OVER")
      2 status_set("OVERHEAT")
     35 status_set("RB")
      1 status_set("SD")
      5 status_set("TEST")
      1 status_set("TIP")
     27 status_set("TRIM")
      1 status_set((u8 & 0x04) > 0 ? "OL" : "OB")
      1 status_set(alarms->status_value)
      1 status_set(const char *buf)
      1 status_set(ctx->arglist[counter])
[...]

PR #2565, opened 2024-07-29
Address spurious LB+RB log flood on APC BXnnnnMI devices

jimklimov:
Closes: #2347

It also adds some visibility around calibration status setting, extends "dstate" API with a `status_get()` method, and this helps avoid setting duplicate states (roughly like "OB LB OB") seen in some drivers earlier.

I hope this toggle allows to fix the problem in the field by optionally delaying spurious status propagation from the driver by `lbrb_log_delay_sec` at most, and if the device is otherwise "online" and is calibrating (unless `lbrb_log_delay_without_calibrating` flag was also set).

The fix goes to some lengths to try detecting the device model during init to default the setting to 3 sec for this line-up, otherwise defaults to 0 (immediate status propagation).

PR #2801, opened 2025-02-11
Fix handling of spaces in driver `status_set()` args

jimklimov:
Addresses part of issue #2708: cleans up some drivers to not use spaces, and for those where not easily avoidable (e.g. `snmp-ups` where strings come from mapping tables) tolerates them better by recursing into self for each found token (avoid duplicates).

Also fixes one bug with `status_get` checks added in #2565 (and maybe there's another, see below).

TODO: this is a relatively hot code path, some optimization may be useful:
* optimize recursion to not `strcmp` for space again when we know it is not there, maybe use a public wrapper method vs. implem with more args, or moving this logic into a new method for use-cases where we do not pass known-clean verbatim strings
* for `status_get` checks, avoid long repetitive string walks (keep an array of offsets to starts/ends of tokens in one string? keep all statuses as a dynamic array of strings and only combine into one at `status_commit` - also avoids `sprintfcat` over and over?)

TODO: Check if there's a bug about `status_get` stopping on first substring e.g. a contrived `BYPASS BY` may not report `BY` as having been set now. UPDATE: indeed, the bug was there. Inner and tail substrings were also impacted (`PA`, `PASS`).

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -479,8 +479,8 @@ relocated into new `shutdown.default` INSTCMD definitions. [#2670]
    * custom `distcheck-something` targets did not inherit `DISTCHECK_FLAGS`
      properly. [#2541]
    * added `status_get()` in NUT driver state API, to check if a status
-     token string had been set recently, and to avoid duplicate settings.
-     [PR #2565]
+     token string had been set recently, and to avoid duplicate settings;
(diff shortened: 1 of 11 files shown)


arnaudquette-eaton (approved):
LGTM

Issue #2807, opened 2025-02-17
Adding a dummy-ups script to examples?


sshambar:
I wrote a bash script to integrate my local Enphase's Envoy API (a simple JSON based local web API) with a dummy-ups driver's port file.  I've just published it (GPLv3) at https://github.com/sshambar/enphase-monitor

The script is simple to use, and provides information on "grid" online/offline status, battery SOC, and several other details (like "load" and "runtime") from the Enphase system.  I use it to integrate with current UPS + ATS (automatic transfer switch) to treat my house battery system as 2nd UPS. I didn't feel it merited writing a full driver as I'm not using it for FSD, outlet control or anything "interactive", I just didn't want my servers shutting down when my UPS failed (don't ask) and the house still had backup power.  Plus it's nice to have another way to monitor the Enphase system using the NUT cgi programs...

The script is probably something others could use as an basis for integrating with similar web APIs using the dummy-ups as a hook to the NUT system, so I thought perhaps it made sense to add as an "example" script someplace in the source?

The script has nice features such as:
 - retains existing/extra values in the port file
 - gracefully handles split-phase or 3-phase input/output values
 - handles no-comms with temporary rename of the port file (indicates "stale data")
 - support portfile value customization (things like battery voltage range, or battery.charge.low for LB trigger)
 - contains logic for login and token based API authentication, with automatic token renewal
 - has it's own config for authentication and API query timing etc
 - checks permissions on files that use secrets
 - is fully self-documented (leading comment in the script)
 - GPLv3 licensed
 - minimal requirements: bash, jq, base64 and curl
 - has details on setting up a systemd service to use with dummy-ups, and SELinux config
 - includes a "TEST" mode that loops through various states and randomly expires the token
 - ... and it's pretty well tested :)

I would have rather created a pull request against NUT, but I honestly don't know where I would add the file here... perhaps someone can suggest a directory if you think it might be useful.

In any case, anyone wanting to add Enphase local API support to NUT can use it as is.

PS. yes I know there's a "external" Enphase API, but it's request limits are severely restricted without paying a high fee...

jimklimov:
Sounds interesting, thanks!

Some new entry under the `scripts/` directory sounds right for this. You would need also a `README.adoc` to describe this script (e.g. with text above as a starting point), and a `Makefile.am` to `EXTRA_DIST` the new files and spellcheck the document. If it is about a couple of new files, maybe add the names to `scripts/Makefile.am` here and there.

[...]

PR #2813, opened 2025-02-23
scripts/external_apis: add example external protocol script

sshambar:
New scripts directory for external_apis, including one integrating the Enphase IQ Gateway web-API with NUT.

First draft of script addition discussed in issue #2807


jimklimov:
Per https://github.com/networkupstools/nut/issues/2807#issuecomment-2677090808 :

> Submitted a draft PR #2813 with a new directory scripts/external_apis and the enphase script/README etc in a subdirectory.
> 
> Looking for comments:
> 
> * if you'd like a different directory name, or would prefer a standalone directory directly below scripts/
> 
> * added a short README for the external_apis directory, and a "full" README in the enphase subdirectory
> 
> * didn't add the "full" README to spellcheck, and it has lots of "shell-command" like things that don't check well...
> 
> 
> I figured I'd checkin the current script as is, and if we add the new "stale file" features to dummy-ups, then I'll submit those additions in a later version as an option :)

Thanks, I'll brush it up a bit for better build integration (e.g. default paths from NUT's `configure` script can be built into your script if handled as a `.in` template). Likewise, path to installed script in the service unit should match where `make install` would place it (all if enabled somehow for installation at all, by e.g. a NUT `configure` option).

jimklimov:
@sshambar : most of NUT is licensed GPLv2+, WDYT about changing the wording in the PR'ed files to that (not v3+ to avoid surprises)?

PR #2825, opened 2025-02-26
CI: Fix Makefiles for parallel builds

jimklimov:
Standard automake `SUBDIRS` allow for nested projects with every directory being buildable alone as well, however the recursive builds are effectively sequential (iterating over the listed directories one by one). This is nicely deterministic for sequential builds, but not so nice for parallel builds with unbalanced workload in the directories (noted when `ChangeLog.(pdf|html)` takes minutes to build while the other cores are quiet).

A commonly recommended solution is to collapse everything to one Makefile which is not too feasible for us.

So this PR goes another way, to explicitly define the web of known dependencies between directories (or specifically the header files and our numerous helper, mock and client libraries that can be used in other directories too). This ends up with a highly parallelisable build routine, where docs and code can consume CPU simultaneously and less wall-clock time gets spent overall. However sequential builds, as well as a dozen of less time-critical recursive goals like `check` or `install` of built artifacts, are handled by exactly same code as before (whatever automake generates).

This PR also revises interaction between root, docs and docs/man Makefiles, to minimize duplicate work that might be done to build or check documentation - this is another notable time hog, so sad to do it e.g. every time we pass a `make check` (sometimes several times per build).

Hopefully NUT CI farm turnaround will be a bit faster now.

UPDATE: Hopefully the later commits also definitively resolve the mix of our desire to build a new set of `ChangeLog*` products when git metadata changes, but to not roll rebuilding them in ordinary cycles that visit these files several times (`make distcheck` etc.)


jimklimov:
Appveyor CI was finally appeased. Still, exceeded the 1hr timeout so the failure now was when compressing the build artifacts after successful builds/tests.

PR #2834, opened 2025-03-05
drivers/nutdrv_qx_gtec.c: multiple status letters can be reported at once

turekl:
A few small fixes of the GTEC driver: only letters for online/offline/bypass status are mutually exclusive, other states like low battery can be reported in addition to them. Also add alarm when battery is disconnected and set _replace battery_ status when battery test had failed.

Follow-up to: #2818


jimklimov:
Thanks, seems good as far as code change goes. I'll trust you on the subject matter :)

PR #2839, opened 2025-03-08
use novendor param as flag for don't read vendro name, need for dexp ups

crodar33:
As I found, in **hunnox_protocol** on step 3 and `asking_for = x0c` it tried to read some string (vendor data by protocol), but ups didn't provide this data and because empty value on next step it cause crash of ups logic and it are stops response.
As I found, it should be solved by use **novendor**, that was used in **blazer_usb**, I added use it in **nutdrv_qx**, but not sure if it will not broke other working setup. Please suggest better solution for it.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -213,6 +213,8 @@ Overall, NUT CI farm build times got 25%+ shorter (which is important as
      on Innova devices, but other models than RT 3/1. [#2798]
    * introduced a `gtec` subdriver and protocol, tested over USB with a
      Gtec ZP120N device. [#2818]
+   * fixed `hunnox_protocol()` to honour the optional `novendor` setting for
+     devices that are confused by such query, e.g. DEXP LCD EURO 1200VA. [#2839]
    * extended Voltronic protocol to support longer numbers as remaining
      `battery.runtime` value. [#2765]
 
--- a/UPGRADING.adoc
+++ b/UPGRADING.adoc
@@ -117,6 +117,11 @@ Changes from 2.8.2 to 2.8.3
   other devices using that MIB (negatively, if the older mappings were
   indeed correct for any practical cases, and were not a typo). [#2803]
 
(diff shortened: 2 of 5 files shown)


jimklimov:
Looks reasonable, thanks. In the example you posted voltage overrides with floating-point separated by comma - should be by a dot. I can fix the commit here, but double-check that your running configuration does act as you expect it to :)

crodar33:
hi, I have change settings locally, output of upsc dexp_nutdrv@localhost look like: 
```
upsc dexp_nutdrv@localhost
Init SSL without certificate database
battery.charge: 87
battery.voltage: 27.0
battery.voltage.high: 28.6
battery.voltage.low: 21.4
device.type: ups
driver.debug: 0
driver.flag.norating: enabled
driver.flag.noscanlangid: enabled
driver.flag.novendor: enabled
driver.name: nutdrv_qx
driver.parameter.default.battery.voltage.high: 28.6
driver.parameter.default.battery.voltage.low: 21.4
driver.parameter.langid_fix: 0x0409
driver.parameter.pollfreq: 30
driver.parameter.port: auto
driver.parameter.product: MEC0003
driver.parameter.productid: 0000
driver.parameter.protocol: hunnox
driver.parameter.subdriver: hunnox
driver.parameter.vendorid: 0001
driver.state: quiet
driver.version: 2.8.2.2324-2324-ged967d659
driver.version.data: Hunnox 0.02
driver.version.internal: 0.41
driver.version.usb: libusb-0.1 (or compat)
input.frequency: 49.8
input.voltage: 241.0
input.voltage.fault: 0.0
output.voltage: 241.0
ups.beeper.status: disabled
ups.delay.shutdown: 60
ups.delay.start: 0
ups.load: 23
ups.productid: 0000
ups.status: OL
ups.temperature: 29.0
ups.type: online
ups.vendorid: 0001
```

Issue #2829, opened 2025-03-02
CI: Add recipes to check completeness and reproducibility of `dist` archives


jimklimov:
Follow-up from #2825/#2828 and thoughts that came along: the `make dist` archives (further verified by `make distcheck`) identify source files to include into the archive by several means, including `EXTRA_DIST` spelled out in `Makefile.am`, template transformations in `configure.ac`, and numerous `*_SOURCE`, `*_HEADERS`, `*_DATA` etc. make variables used in the recipes.

In many cases, where optional builds are involved, we go the way of either declaring some files as sources for a program or library (or its certain optional feature), or explicitly `EXTRA_DIST`'ing them.

If we miss some mark during such manual maintenance, it is potentially possible that a "dist" archive created from a build configuration without some optional build products would not include the sources needed to create those products on a more capable system.

This issue proposes a couple of checks:
* `configure --without-everything && make distcheck(-light)` to produce a dist archive with the possibly incomplete set of source files, and then try to build everything possible on the current system (to make sure we have the sources for that)
* configure and create dist archives under various combinations of `--with-all`, `--with-dev`, `--with-docs` and specific features (may require dependencies to be installed on the agent doing this), unpack each into a separate directory, and verify that they have same contents.

This may be related to (solution automated along with) issue #1400

jimklimov:
PR #3158 (and some later refinement commits) makes tarballs of recent PR and branch builds available, and PR #3161 includes a way to build the dist archive from a dist archive, checking that the two contain exactly same sets and contents of files - so that they are indeed reproducible.

The other goals, to vary the `configure` options before making a dist archive, and to have a way to `configure --without-everything`, are not yet completed -- so not closing this issue.

PR #2842, opened 2025-03-09
docs/Makefile.am: fixed make dist, fix custom man sections

sshambar:
`make dist` was failing to build all man pages.  Changed how the *-prepped file dependencies are handled to fix.

Custom man sections were being implemented in configure, but automake runs before they are chosen, so the Makefile.in were missing the install logic.  Changed to use _DATA targets with directories that can be chosen at configure time

Also added index.txt to DIST target to fix distcheck

Fixed substitution bug in external_apis/Makefile.am


jimklimov:
Great catch, reasonable analysis and fix - thanks!

How did you stumble on this? On many systems, `make distcheck{,-light}` did pass - although this is a test to build it in same conditions as the original creation of the archive, so checking in another bare-bone system would have been more reliable.

I did have concerns about better checks of the archive validity, posted at #2829... I suppose your PR brings the solution a step closer :)

sshambar:
> How did you stumble on this? On many systems, make distcheck{,-light} did pass - although this is a test to 

I was trying to fix the Fedora RPM to build HEAD, and `make distclean; make dist` to make a source-tar always failed with
```cp: cannot stat './nut.conf.5': No such file or directory```
but running `dist check` again worked... traced the problem to the missing suffix rule. (I think distcheck worked because it builds html docs, which masks the man the problem).

Of course, then trying to build the RPM failed to install the man pages (because the install-man: rule was empty), and figured how the man custom-section change caused the automake issues... testing my fix exposed the missing index.txt dist file...

I found a couple other issues updating the RPM, but I figured fixing the build was best done first :)

**Comment by jimklimov** [comment]
I think the changes about `prepped` rules broke the build, at least for Sun make.

Your new ruleset seems factually wrong in any case (maybe *that* is what upsets the build):
```
+.txt.txt-prepped:
+       rm -f "$(abs_top_builddir)/docs/man/.prep-src-docs"*
+       $(MAKE) prep-src-docs
```

The `.prep-src-docs` is a single flag for dependency tracking (newer than all of its prerequisites) so you remove it every time you prep another text file, and then call `make prep-src-docs` which processes all PREP_SRC files (and makes their individual `*-prepped` contents). I suppose that new code has potential to overlap, deleting the flag and re-prepping all the files for each file.

Also, calling autotools make is more complicated than that (see other cases with AM_MAKEFLAGS, `+` for job server cooperation, etc.

Currently it seems to end up with no `*-prepped` files in distcheck work area, and fails to generate mans/htmls from them.

Also, the `.NOTINTERMEDIATE` keyword seems to be GNU Make specific, not an automake scripting trick, right?..

Was this bit critical for your fixes? I'll try to put back the previous code for these, to see if it helps.

**Comment by sshambar** [comment]
The problem I hit with .prep-src-docs as it was used is that it "satisfied" it's first target (nut.conf.5 in my case), as nut.conf.5 didn't have any explicit recipes... the correct solution would be to have the `.txt.txt-prepped` rule actually build individual files (so .prep-src-docs isn't needed at all), but you have some fancy business in the .prep-src-docs recipe that I didn't want to mess with...

Reverted... nut.conf.5 will still fail on `make distclean; make dist`, so the logic for the flag needs to be handled some other way (my recursive make is just what I've seen used for "meta-flags" that handle multiple targets as a batch... but yeah it's probably not universally portable :)

I guess .NOTINTERMEDIATE probably is a GNU thing, would .PRECIOUS be portable?

**Comment by jimklimov** [comment]
Actually nut.conf.{5,html} seems to have been a tip of the iceberg, just first in list. A `make distcheck -k`showed all (or a lot of) man pages to fail before I reverted that prep part of the fix :\

> fail on make distclean; make dist

Aren't you supposed to re-`configure` in between?

There's also a number of files generated by `autogen.sh` as templates for `configure` (so they are `EXTRA_DIST`'ed but are not in git, and I had a case with files generated from those templates going AWOL at `make clean`. Possibly the {,DIST,MAINTAINER,...}CLEANFILES lists need a revision if such scenarios fail (helped in that case)...

**Comment by jimklimov** [comment]
Seems the main trick in this regard is about `BUILT_SOURCES`. And having the tools to build man pages (when building from git sources), or having them originally (from release/dist tarball).

**Comment by jimklimov** [comment]
Ooh that is a wiggly wad of worms. "We want it built when we can, or used from tarball if available, or not built when asked not to, and cleaned then but not cleaned then" - and all this without the GNU make "if" conditionals :\

Hopefully I made the best pick of tricks in a lousy situation, waiting for CI to concur in at least the use-cases where it passed before this PR :)

If this iteration survives the automated testing, can you check how it behaves for your original use-case? I've tested with `./autogen.sh && ./configure && make clean && make distcheck(-light)` on systems with and without asciidoc, where `make distcheck-light` deliberately skips the build products it can not make (`--with-all=auto`), and now for `dist` part of the parent make it also fakes the pre-built man pages if missing (just creates a file with placeholder contents for each such missing name, so `make distdir` passes with new rules that require all rendered man pages to be there one way or another).

**Comment by sshambar** [comment]
I'll give it spin as soon as the CIs pass... looks like there's a few failures atm though...

**Comment by jimklimov** [comment]
Aw, schucks :\

Will check tomorrow I hope.

**Comment by jimklimov** [comment]
Also, did you see your original use-case fail with earlier 2.8.x releases (tarball and/or git tag), e.g. is it a recent regression or some edge case that was there for a long time?

**Comment by sshambar** [comment]
> Also, did you see your original use-case fail with earlier 2.8.x releases...

Well, I just pulled the v2.8.2 tag and it failed at `make dist` with the same "cannot stat './nut.conf.5':" error, so I'm not sure when it first happened :)

**Comment by jimklimov** [comment]
And... here goes... nothing (engine whirs, clucks, and dies down).

Well, let's see if different systems like this twist better.

Essentially, now there are a few more tricks applied for CI builds to pass their `make distcheck*` variants, with a dist archive generation now carefully requiring all man page files to be available (so faked for some CI circumstances, when `distcheck-ci` or `distcheck-light` are built, and man pages are neither available nor buildable in original workspace). This is a significant step towards #2829 at least in spirit.

I hope this would also fix your original issue to survive `./autogen.sh && ./configure && make distcheck` (note `-j` may be used, "MUST" not fail due to parallelism as such) as long as your workspace has the pre-built man pages (from tarball) or can produce them (from git, with asciidoc et al present and usable).

**Comment by sshambar** [comment]
BTW, since you appear deep in the weeds on fixing build issues, one thing I noticed from the rpm fixes I was working on is that the build is choosing the wrong python site-package directory... `nut_check_python.m4` (on my system):

`import site; print(site.getsitepackages().pop(0))` => '/usr/local/lib64/python3.13/site-packages'

But that's the `platlib` (arch-dependent), not the `purelib` directory... It might be better (at least on python3) to use:

`import sysconfig; print(sysconfig.get_path('purelib')` => '/usr/local/lib/python3.13/site-packages'

... or alternatively, offer a configure option to set the site-package directory?  (or perhaps in addition?)
Since it's unrelated to this PR, I could offer a new PR - but it's such a trivial change... :)

**Comment by sshambar** [comment]
Oh, and one other note (again, I can open an issue if you think it deserves it), rpm build on my patch release failed on the configure "PACKAGE_VERSION" generated by `make dist` with the error:

```
Invalid version (double separator '-'): 2.8.2.2324-2324-ged967d659: pkgconfig(libnutclient) = 2.8.2.2324-2324-ged967d659
```

I guess rpm wants to be able to strictly order upgrades using semantic versioning... Do you think it'd be better to add the "hash" as version "metadata"? (ref https://semver.org/#spec-item-10)

**Comment by jimklimov** [comment]
Hm, package versions in standard semver X.Y.Z form should be derived from the git tag (for a release snapshot build), X.Y.Z.W for later commits on master, and are more detailed like this for development branches since some PRs made for issue #1949

IIRC a `echo NUT_VERSION_FORCED=1.2.3 > VERSION_FORCED && ./configure && make` sort of trick should take care of this for intermediate builds. I am not sure I've documented this well in one spot, so searching for `VERSION_FORCED_SEMVER`, `VERSION_FORCED`, `NUT_DEFAULT_VERSION` in `NEWS.adoc`, `UPGRADING.adoc`, `docs/maintainer-guide.txt`, as well as `tools/gitlog2version.sh`, `configure.ac` and some makefiles for the practical logic about this, is probably a good start.

**Comment by jimklimov** [comment]
> BTW, since you appear deep in the weeds on fixing build issues, one thing I noticed from the rpm fixes I was working on is that the build is choosing the wrong python site-package directory... `nut_check_python.m4` (on my system):
> 
> `import site; print(site.getsitepackages().pop(0))` => '/usr/local/lib64/python3.13/site-packages'
> 
> But that's the `platlib` (arch-dependent), not the `purelib` directory... It might be better (at least on python3) to use:
> 
> `import sysconfig; print(sysconfig.get_path('purelib')` => '/usr/local/lib/python3.13/site-packages'
> 
> ... or alternatively, offer a configure option to set the site-package directory? (or perhaps in addition?) Since it's unrelated to this PR, I could offer a new PR - but it's such a trivial change... :)

I think a separate PR makes sense, for the (triplet of) options that people could specify overrides, and to try `purelib` first and only fall back to `platlib` or that zero'th element if for some reason previous attempts did not resolve.

**Comment by sshambar** [comment]
> I think a separate PR makes sense, for the (triplet of) options that people could specify overrides, and by default to try `purelib` first and only fall back to `platlib` or that zero'th element if for some reason previous attempts did not resolve.

I wasn't sure on you're preferred naming style for the 3 new args, but I submitted PR #2846 for the sysconfig fix.

I went with a version check on python rather than a fallback -- sysconfig ships with python3.2+, so it should "just work" those recent version (and fail if not, just like previous versions would for `import site`)

**Comment by sshambar** [comment]
> IIRC a `echo NUT_VERSION_FORCED=1.2.3 > VERSION_FORCED && ./configure && make` sort of trick should take care of this for intermediate builds. 

True, but I really like having the auto-incrementing "commits since tag" you get with the `git describe` logic you have... each new commit should trigger package updates without having to patch the configure line :)

> I am not sure I've documented this well in one spot,...

The most complete doc appears to be in `tools/gitlog2version.sh`, and the only place I could find that talked about the `git hash`

My comment was just identifying "ordering" vs "non-ordering" parts of the version string... something like adding
```
SUFFIX=${SUFFIX/-g/+g}
```
in gitlog2version.sh would put the "hash" in the non-ordered part of the "semver" world... just an suggestion :)

jimklimov:
Regarding semver - thanks for link to the standard, I did forget that bit about build metadata. Makes sense, more so to fix it before a formal release that adds this markup to public code.

sshambar:
Tried the latest build-dist, and it works for make dist and the rpm build... so things are looking good.

BTW, following #2829 I tried `configure --without-everything`; would be cool before `make dist`, but I guess you haven't added that option yet :)

jimklimov:
No, that option as such did not yet appear, just an idea for it :) That said, in practice it would likely be `--without-all` and maybe an experimental a combo of other `--without` options if any are needed (`docs`, `dev`, `cgi` are most often juggled along with custom `all` settings).

As for failing CI runs above, it seems some scenarios keep calling `make distcheck`, which now fails on systems without asciidoc (and no pre-built man pages). I think I've changed these all to call `make distcheck-ci` (which then routes to `distcheck` or `distcheck-fake-man`), but maybe the CI library built-ins somehow took precedence.

jimklimov:
Facepalm - my security bit me:
> `‘Jenkinsfile-dynamatrix’ has been modified in an untrusted revision`

It just ignored the urge to use `distcheck-ci` because it is not in a trusted pipeline definition change (not merged, not maintainer repo as PR source). Oh well, it was a rare chance to test that part :)

jimklimov:
Re-posted this chain of commits as a PR from my repo, hopefully *now* that will be picked up properly.

PR #2849, opened 2025-03-14
docs/Makefile.am: fixed make dist, fix custom man sections (repost)

jimklimov:
Same as #2842, re-posted from maintainer repo for CI to trust the recipe/pipeline changes involved.


jimklimov:
Now it is down to oddities of two non-GNU makes, will try to reproduce and fix on those machines.

PR #2697, opened 2024-11-26
Ecomode addon 3.2

masterwishx:
add-on for #2637,#2680,#2684 ECO/Bypass for Eaton Models.

[ Done ] We can add also maybe auto function for ECO mode, that put ups to Bypass and then to ECO .
Instead user manually make thouse steps.

PR #2850, opened 2025-03-16
Revise the stance on ECO/... modes in NUT standard, amend implementations

jimklimov:
Following the long discussion in #2708, we decided that vendor-defined states named by marketing buzzwords amounting to some economic/ecological mode of operation should not be exposed in `ups.status` with a sense of urgency.

Instead, a new `output.inverter.mode` was introduced (with similar semantics) which drivers may eventually learn to populate with a number of standard states (line-interactive, online, bypass(?)) if we are sure we know them to be in place, as well as tokens for `vendor:NAME:BUZZWORD` to report the additional information that may be or not be of interest to the end user.

CC @masterwishx @gdt @desertwitch @arnaudquette-eaton : comments and practical testing would be welcome - e.g. that the state report would correlate with device state changes, that it would be still reported by upsmon, and remain reported while active (like only get reported when changing state, then disappear; and not flip-flop etc.)

I think the main review question here is about the spec, if that change feels acceptable as the compromise between different wishes in that discussion, and can be published as part of next NUT release.

Implementation, as long as it does not crash anything, may be improved later (after a release - i.e. minor flaws are not a blocker, it is experimental and constrained in practice to a couple of drivers so far)...

Closes: #2697
(Originally includes commits from that PR, but may post them rebased eventually, so linking for GitHub automatics this way)


masterwishx:
I'm sorry but as posted in https://github.com/networkupstools/nut/issues/2719#issuecomment-2630311809
I cant test it Becouse I can't get out from eco mode to online by commands seems some bug in firmware...

jimklimov:
Well, at least the driver might pick up (or not) the currently active mode if reported by device. If some button switches the mode and we see notifications, that would be fun too...

masterwishx:
> Well, at least the driver might pick up (or not) the currently active mode if reported by device. If some button switches the mode and we see notifications, that would be fun too...

Yep the problem in my 9E no buttons to switch to /from eco mode.
Only to/from bypass buttons, but they also not working in current state. So I'm stuck in eco mode.
Only I hope I can disconnect batteries I think for enter to online back.

masterwishx:
> currently active mode if reported by device

Also was some issue when tryed to compile for unraid 7, so maybe @desertwitch can compile then I can check if current mode pickuped

gdt:
Why are there 43 commits?    I know opinions differ about PRs and other stuff, but I'm in the rebase camp, and I find unrebased hard to follow

gdt:
I guess semantics wise, I wonder if this is really about "inverter".    If it were just output.mode, that would seem better.   As I am beginning to just barely understand, some eco modes are about not only a true bypass of out from in, but shutting down some other electronics.  This is all about situations where  the inverter is not running anyway.      So maybe it's even "ups.mode".

The next thought I have, not sure it's helpful, is there is "what modes are set to operate as able" vs "what mode is it in now".   I think we are talking about "what is it in now" and not configuration.

So I land on `ups.mode` as where this should live but not vigorously.

jimklimov:
Naming might be addressed, the idea crossed my mind too especially as I went about looking for precedents and changing the handling of nutdrv_q voltronic (eco and eco_adv modes - the latter seems to be indeed about powering down electronics of the UPS).

masterwishx:
> Naming might be addressed, the idea crossed my mind too especially as I went about looking for precedents and changing the handling of nutdrv_q voltronic (eco and eco_adv modes - the latter seems to be indeed about powering down electronics of the UPS).

in `ESS mode` and i think maybe in `ECO Advanced` the inverter is not down but still ON and waiting to work , thats why we have about 2ms instead of 10ms when inverter is off in `ECO`

jimklimov:
> > Naming might be addressed, the idea crossed my mind too especially as I went about looking for precedents and changing the handling of nutdrv_q voltronic (eco and eco_adv modes - the latter seems to be indeed about powering down electronics of the UPS).
> 
> in `ESS mode` and i think maybe in `ECO Advanced` the inverter is not down but still ON and waiting to work , thats why we have about 2ms instead of 10ms when inverter is off in `ECO`

Well, that highlights why the discussion in #2708 and this PR happened - different devices and vendors have different implementations and meanings for similar sounding buzzwords and technologies, which may even change over time for the same vendor, but we in NUT have to represent somehow in a vendor-agnostic data scheme, so that (ideally) the end-users would not care which UPS brand and model they have to know the physical impact of some configuration change whatever it is named today.

masterwishx:
I agree, just maybe we need to mark when the inverter is working or not?

jimklimov:
> Why are there 43 commits? I know opinions differ about PRs and other stuff, but I'm in the rebase camp, and I find unrebased hard to follow

Looking at the list:
* about half are inherited from #2697 directly;
* half a dozen are merges from `upstream/master` to resolve conflicts as time went on (rebase might be nice about these but proved tedious to re-resolve those text mismatches);
* the remainder is the "payload" about:
  * changes of spec (may evolve further in this PR) following ideas from #2708;
  * refactoring of `dstate` to generalize what was happening in `status_set` and friends (might perhaps become a separate PR changeset)
  * introduction of `invmode*` methods and data (naming may evolve further in this PR);
  * changes of clients, drivers and tests for new concepts and methods (may evolve) and possibly for related issues found along the way (there were many PRs spawned by work on #2708, but maybe not everything not strictly on "ECO" topic went that way eventually).

I'll check what can get dumped overboard into agreeable PRs, to make the difficult and contentious PR easier to review :)

jimklimov:
> I agree, just maybe we need to mark when the inverter is working or waiting to work (like in standby) or off?

Maybe, at least that would be a finite amount of states/strings to report. If we know or have good reasons to assume its state.
And the vague extras with marketing buzzwords (what this PR named `output.inverter.mode` so far) renamed to something else?.. (Maybe even under `experimental.*` namespace for 2.8.3 release, as we still seem to have no good grasp about the syntax and name - just that it is something we have a way to know from some devices and might be interesting to represent to users).

jimklimov:
Offloaded some "noise" into #2852 - will revise this PR after merging that.

gdt:
> > I agree, just maybe we need to mark when the inverter is working or waiting to work (like in standby) or off?
> 
> Maybe, at least that would be a finite amount of states/strings to report. If we know or have good reasons to assume its state. And the vague extras with marketing buzzwords (what this PR named `output.inverter.mode` so far) renamed to something else?.. (Maybe even under `experimental.*` namespace for 2.8.3 release, as we still seem to have no good grasp about the syntax and name - just that it is something we have a way to know from some devices and might be interesting to represent to users).

`experimental` is a great idea.    Let's call it `experimental.mode` and just dump in the brand:buzzword or whatever.  That exposes what can be exposed, in a non-pleasing way, but then we can get some experience without having to answer the "you broke this variable" later.

As for inverter on or off, and other things, I think we can begin to define a unfied semantics, but I think that should start with documentation that has a definition and that maps state from as many implementations as possible, so that we are sure it's meaningful, clean, and captures what it should.   Then we can think about variables vs flags.  In general I lean to anything that's a detail in a variable, so that flags are as simple as possible.  It's easier to ignore a variable you don't know about than deal with unknown or split flags.

"Is the inverter operating" sounds like a great first candidate.   I don't think the possibility of coming up with a definition for that should deter us from having an experimental variable that simply has a vendor:buzzword state description.

gdt:
> I'll check what can get dumped overboard into agreeable PRs, to make the difficult and contentious PR easier to review :)

Thanks.  Definitely all the merges should at least be easy to rebase out, and I think it's good to have an expectation/practice that PRs are rebased, not only dropping that, but folding in any 'make big change', 'fixup to that big change' commit sequences.

jimklimov:
Rebased over recent master branch changes; replaced earlier proposal of `output.inverter.mode` with `ups.mode` (stricter defined, but currently not populated - probably up to community after the release, based on testing with specific devices and reading their specs; similarly for `output.inverter.latency`), and a more relaxed `experimental.ups.mode.buzzwords` which is handled like `invmode` was in earlier iterations of the PR, to pass vendor-inspired tokens as opaque strings and issue `upsmon` notifications (if enabled).

jimklimov:
@gdt @masterwishx @desertwitch : WDYT, is this updated take on the naming agreeable?

For now I still struggle to find a good way to convert actions into vendor-agnostic naming (e.g. if we want to avoid `ecomode.enable`, `essmode.enable` and such). It should be possible to provide a single optional "cmdparam" to an `INSTCMD`, tools like `upscmd` allow that too. So we could make some handling for `upsmode.enable vendor:eaton:ESS` etc., I suppose. Perhaps even have the device/driver expose valid values for the `cmdparam`. But I don't have time to experiment with that in the coming months, getting late on dayjob work :( and don't want this delaying the release even further. Takers welcome :)

I think the best I can offer right now is to rename the actions added since v2.8.2 tag as `experimental.*` as well so their move to a single `INSTCMD` later is not a breach of API (nor cause for clumsy aliasing) later on. Maybe it is lazily okay even to keep them as they are, eventually no driver would offer the action as supported and that would be the end of their short story in the field (in-tree at least).

jimklimov:
Looking at the git diff since v2.8.2, I think the added `input.transfer.*` and `input.bypass.*` criteria settings are okay, should be portable to similar concepts of different vendors/models.

The thoughts about INSTCMD above concerned these new entries that now seem vendor-specific but too implicit about that. Maybe not though ("go put my UPS into whatever definition of eco/green/efficient it knows"), but with Eaton devices in the original changeset supporting several modes, it would be hard to leave one standard operation anyway (and "ess" is a vendor-specific feature name already):
* ecomode.enable
* ecomode.disable
* ecomode.start.auto
* essmode.enable
* essmode.disable

@masterwishx : there are also things like `outlet.n.protect.status` and `input.eco.switchable` documented as having values `0-2` with an example value of `normal`. I think the logical mismatch can be fixed by defining them as opaque strings, and documenting the possibly supported values in the man page section for a particular sub-driver (e.g. "mge-hid might support X Y Z depending on device model").

masterwishx:
I will check the code later today I hope. 
About checking on hardware, when tryed last time compile master for current version of Unraid 7.0.1 I had some missing packages etc
So was unable to compile until @desertwitch compiled for his plugin, but will try it again maybe today (I'm not at home all day) if not then tomorrow.

gdt:
> So we could make some handling for `ups.mode.change vendor:eaton:ESS` etc., I suppose.

Given time limits, I think moving anything eco to full experimental label immediately is good, as given the lack of broadly-shared clear semantics, it's a bug for it to have gone in.   Medium term, using vendor tokens in a mode set command sounds good.  In the glorious future, when all vendors publish clear specifications, we can figure out the common ground, while meeting on neutral territory.  That's my hot take :-)

drivers/mge-hid.c
```diff
@@ -1135,6 +1151,51 @@ static info_lkp_t eaton_input_bypass_mode_off_info[] = {
...
+	upsdebugx(1, "%s: ECO Mode was enabled after switching to Bypass Mode", __func__);
+	return NULL;
+}
+
+/* High Efficiency (aka ECO) mode for auto start command */
+static info_lkp_t eaton_input_eco_mode_auto_on_info[] = {
+	{ 1, "dummy", eaton_input_eco_mode_auto_on_fun, NULL },
+	{ 0, NULL, NULL, NULL }
+};
+
```
masterwishx, line 1198: @jimklimov this part of #2697 in draft not sure how its got there ?

as wroted in pr seems we need only `input.eco.switchable` + `input.bypass.switch.on` to enable ECO mode as tested on 9SX
(in my 9E was also need `input.bypass.switch.off` to disable Bypass mode before switching to ECO mode but stucked in ECO mode etc...)

drivers/mge-hid.c
```diff
@@ -2072,10 +2133,12 @@ static hid_info_t mge_hid2nut[] =
...
-	{ "ecomode.disable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "0", HU_TYPE_CMD, NULL },
-	{ "ecomode.enable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "1", HU_TYPE_CMD, eaton_input_eco_mode_on_off_info },
-	{ "essmode.enable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "2", HU_TYPE_CMD, NULL },
-	{ "essmode.disable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "0", HU_TYPE_CMD, NULL },
+	{ "experimental.ecomode.disable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "0", HU_TYPE_CMD, NULL },
+	{ "experimental.ecomode.enable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "1", HU_TYPE_CMD, eaton_input_eco_mode_on_off_info },
+	{ "experimental.essmode.enable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "2", HU_TYPE_CMD, NULL },
+	{ "experimental.essmode.disable", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "0", HU_TYPE_CMD, NULL },
+	/* Command to switch ECO(HE) Mode with switch to Automatic Bypass Mode on before */
+	{ "experimental.ecomode.start.auto", 0, 0, "UPS.PowerConverter.Input.[5].Switchable", NULL, "1", HU_TYPE_CMD, eaton_input_eco_mode_auto_on_info },
```
masterwishx, line 2141: Also here from draft

drivers/mge-hid.c
```diff
@@ -3,6 +3,8 @@
  *  Copyright (C)
  *        2003 - 2015 Arnaud Quette <arnaud.quette@free.fr>
  *        2015 - 2024 Eaton / Arnaud Quette <ArnaudQuette@Eaton.com>
+ *        2020 - 2025 Jim Klimov <jimklimov+nut@gmail.com>
+ *        2024 - 2025 "DaRK AnGeL" <28630321+masterwishx@users.noreply.github.com>
```
masterwishx, line 7: @jimklimov 10x for this

masterwishx (approved):

masterwishx (approved):
LGTM

drivers/mge-hid.c
masterwishx: > draft not sure how its got there ?

Oops, sorry I forgot we talked about all kind prs 🤕

jimklimov:
@masterwishx : Removed mentions of `auto_on` and `experimental.ecomode.start.auto` - but they are still in git history, so if you later come back to this idea, the commit ca27e069987fc6de246c1c085f97f7838236e424 can be `git revert`'ed.

Got any other last thoughts, or can this be merged when CI agrees? :)

And thanks again for bearing with me (and others) over so many months and uncertainty :)

masterwishx:
> Did you have a chance to read/check that my changes to `vendor:X:Y` string tokens did not break your feature?

Yes it's looks fine.



> Removed mentions of `auto_on` and `experimental.ecomode.start.auto` - but they are still in git history, so if you later come back to this idea, the commit [ca27e06](https://github.com/networkupstools/nut/commit/ca27e069987fc6de246c1c085f97f7838236e424) can be `git revert`'ed.

Thanks, i beleave I can finish later this draft pr and add also
experimental.online.start.auto (back from eco mode), but can't test it Becouse issues posted before, maybe user that checked eco mode with 9sx can test it...

But all other changes seems all fine to me

jimklimov:
Great, thanks! Note that THIS PR does include all your commits from your draft #2697 (rebased over later `master` branch state though), and includes this latest reversible commit to drop one experiment. So if/when you do follow up, that would be another draft PR, hopefully based from the master branch of that time :)

masterwishx (changes requested):
This should be fine ,as tested on 9SX ,
Last function needs to 9E model like my to enter ECO mode but cant get to online back then !

PR #2852, opened 2025-03-17
Refactor `status_set()` et al and update a test, optimize `upsmon` a bit, retouch docs

jimklimov:
Assorted features and fixes inspired by #2708 discussion, to avoid cluttering the main PR's review scope.

Notably, the refactoring of `status_set()` allows other PRs to manage multiple-unique-token data similar to `ups.status` (as proposed for those "ECO" related changes), potentially not only in drivers.


gdt:
I only skimmed but this all seems sensible.

PR #2857, opened 2025-03-23
NUT-Monitor: improve portability

jimklimov:
Follow-up from PR #2845 

* Help some platforms find localization files better
* Help builds on Solaris/illumos find an ARCH-dependent build of python2 interpreter that works (the default one in PATH may point to a build dysfunctional for our purposes: only `64/_glib.so` is delivered by `pygobject-27` package, but the default interpreter may be the 32-bit binary)
* Add credit where due, in translation authors


jimklimov:
Something seems to upset Sun make (and/or userland tools) with these recipe changes...

Issue #2782, opened 2025-01-24
Portability: add a `configure` option to enforce RPATH/RUNPATH from learned library locations


jimklimov:
As was found in #2781, some distros that put additional packages into "uncommon" locations (e.g. not into `/usr/lib` searched by default), we can build and link NUT binaries following the `-L/some/path` instructions learned from command-line `./configure --with-something-libs` options or from pkg-config.

However to *run* such binaries, the run-time LD linker needs either its built-in defaults, system configuration (`crle` on Solaris/illumos, etc.) or the `LD_LIBRARY_PATH` (frowned upon) to point to additional locations, or for the binaries (programs and shared libs) to say what they prefer - via `RPATH` and `RUNPATH` attributes of those binaries (see `objdump`, `elfdump -d`, `dump -Lv` etc. for the platform).

Normally the same data that we found the libraries for linking with (`-L` options) should convey the `-R` values if needed for run-time. In practice, many `*.pc` files delivered by various projects do not bother to do so, and then when packaged into uncommon locations, the packaging recipes did not think of adding those instructions either. 

This issue tracks an idea to add an option to NUT `configure` script to rewrite the located `-L` options into additional `-R` (and/or `-Wl,-rpath,...` etc.) options for the linker, at least for "uncommon locations", and so to optionally compensate for poor quality of information provided by dependencies. Since #2674 we have a precedent in `m4/nut_check_libnss.m4` and `LIBSSL_LDFLAGS_RPATH` to build upon.

Note the NUT binaries do get the `RPATH` to NUT installation location built in, and typically expect the major-versioned symlink like `libupsclient.so.6` or `libnutscanner.so.2` to be present there. Investigating how autotools achieve that, and emulating it for third-party dependencies, may be another path to solution.

jimklimov:
A hot-fix for OmniOS specifically took the form of blanketly adjusting default `LDFLAGS` when `/opt/ooce/lib` is seen. This allows USB, CGI and similar artifacts to get built and to run, even if it is an overkill to have these RPATHs in each binary.

Proper and generic fix should get investigated later.

jimklimov:
Merged #2865 which took care of the opposite side of this issue, about NUT pkg-config metadata announcing rpath settings for its libraries when installed into non-system paths (as the compiler used to build NUT lists them, in a way the linker used supports the operation).

PR #2865, opened 2025-03-27
Optionally suggest rpath for NUT libraries for third parties

jimklimov:
Another side of the coin raised by issue #2782 : when we install NUT itself, especially its libraries into uncommon paths like `/usr/local/ups/lib`, third-party progs linked against them can not find them at run-time: the RPATH or RUNPATH has to be embedded (which is what this PR helps with), or LD_LIBRARY_PATH or equivalent frowned-upon tricks should be done.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -18,6 +18,10 @@ PLANNED: Release notes for NUT 2.8.4 - what's new since 2.8.3
 
 https://github.com/networkupstools/nut/milestone/9
 
+ - `lib/*.pc.in`: propagate `-R/PATH` to NUT library installation location
+   (by default not in system prefix) to help third-party clients link with
+   us automatically. [#2782]
+
  - (expected) Dynamic Mapping Files (DMF) feature supported, to allow
    the driver binaries to be built once and data mappings to be loaded
    and modernized on the fly (porting from 42ITy project)
--- a/configure.ac
+++ b/configure.ac
@@ -5067,6 +5067,7 @@ NUT_REPORT_SETTING_PATH([System exec-library path],
 
 dnl Check if we want to expose non-standard RPATH to third-party consumers
 dnl of libupsclient etc. Keep verbatim dollar-var here, expand in pkg-config!
+dnl NOTE: hardcode_libdir_flag_spec(_CXX) vars come from LT INIT above
 LDFLAGS_NUT_RPATH=''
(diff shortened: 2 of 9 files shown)


jimklimov:
Tested with wmNut build on TermUX-x11, with `--prefix=$HOME/nut-inst` :

```
~/nut-wmnut $ ldd ~/nut-inst/bin/wmnut
        libX11.so => /data/data/com.termux/files/usr/lib/libX11.so
        libXext.so => /data/data/com.termux/files/usr/lib/libXext.so
        libXpm.so => /data/data/com.termux/files/usr/lib/libXpm.so
        libupsclient.so.6 => /data/data/com.termux/files/home/nut-inst/lib/libupsclient.so.6
        libssl.so.3 => /data/data/com.termux/files/usr/lib/libssl.so.3
        libcrypto.so.3 => /data/data/com.termux/files/usr/lib/libcrypto.so.3
        libdl.so => /system/lib64/libdl.so
        libc.so => /system/lib64/libc.so
        libxcb.so => /data/data/com.termux/files/usr/lib/libxcb.so
        libandroid-support.so => /data/data/com.termux/files/usr/lib/libandroid-support.so
        ld-android.so => /system/lib64/ld-android.so
        libXau.so => /data/data/com.termux/files/usr/lib/libXau.so                       libXdmcp.so => /data/data/com.termux/files/usr/lib/libXdmcp.so

~/nut-wmnut $ readelf -d ~/nut-inst/bin/wmnut | grep PATH
  0x000000000000001d (RUNPATH)    Library runpath: [/data/data/com.termux/files/usr/lib:/data/data/com.termux/files/home/nut-inst/lib:/data/data/com.termux/files/home/nut-inst/lib]

~/nut-wmnut $ cat ~/nut-inst/lib/pkgconfig/libupsclient.pc
prefix=/data/data/com.termux/files/home/nut-inst
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
sysconfdir=/data/data/com.termux/files/home/nut-inst/etc
statepath=/var/state/ups
nutuser=nobody

Name: libupsclient
Description: UPS monitoring with Network UPS Tools
Version: 2.8.2.2524.8
Libs: -L${libdir} -Wl,-rpath -Wl,${libdir} -lupsclient
Cflags: -I${includedir}
Requires: openssl
```

PR #2870, opened 2025-03-31
Fix warnings found by NetBSD

jimklimov:
Also test added NUT CI farm worker with NetBSD

CC @gdt

### DIFF
--- a/docs/config-prereqs.txt
+++ b/docs/config-prereqs.txt
@@ -1181,9 +1181,9 @@ systems, can add dash-number suffixed symlinks to compiler tools (e.g.
 `gcc-7` beside the `gcc` installed by package) near the original
 binaries and into `/usr/lib/ccache`:
 ------
-:; ( cd /usr/bin && for TOOL in cpp gcc g++ ; do \
-        for VER in "-7" "-14" ; do \
-            ln -s "$TOOL" "$TOOL$VER" ; \
+:; ( cd /usr/pkg/bin && for TOOL in cpp gcc g++ ; do \
+        for VER in "7" "14" ; do \
+            ln -fs "../gcc$VER/bin/$TOOL" "${TOOL}-${VER}" ; \
         done ; \
      done )
 
--- a/docs/config-prereqs.txt
+++ b/docs/config-prereqs.txt
@@ -1088,6 +1088,19 @@ https://www.librebyte.net/en/cli-en/pkgin-a-netbsd-package-manager/
 (the latter also provides a nice cheat-sheet about `pkgin` operations).
 ======
 
+[NOTE]
+======
+On NetBSD 9.2 the `openpimi` and `net-snmp` packages complain that
+they require either OS ABI 9.0, or that `CHECK_OSABI=no` is set in
+a `pkg_install.conf`. Such file was not found in the test system,
+but can be created in `/etc/` and is honoured by `pkgin`:
+
+----
+:; grep CHECK_OSABI /etc/pkg_install.conf || \
+    ( echo CHECK_OSABI=no >> /etc/pkg_install.conf )
+----
+======
+
 ------
 :; pkgin install \
     git perl curl \
@@ -1151,9 +1164,8 @@ https://www.librebyte.net/en/cli-en/pkgin-a-netbsd-package-manager/
 
 [NOTE]
 ======
-(TBD) On NetBSD 9.2 this package complains that it requires
-OS ABI 9.0, or that `CHECK_OSABI=no` is set in `pkg_install.conf`.
-Such file was not found in the test system...
+There seems to be no FreeIPMI for NetBSD, but there's OpenIPMI
+(can help someone implementing actual support for it):
 
 ------
 :; pkgin install \
--- a/docs/config-prereqs.txt
+++ b/docs/config-prereqs.txt
@@ -1053,6 +1053,7 @@ Note that `PATH` for builds on NetBSD should include `local` and
 ------
 
 NOTE: You may want to reference `ccache` even before all that,
+e.g. in the system-wide `/etc/profile` or build-user's `~/.profile`,
 as detailed below:
 
 ------
@@ -1161,8 +1162,17 @@ binaries and into `/usr/lib/ccache`:
    )
 ------
 
-NOTE: For Jenkins agents, also need to `pkgin install openjdk17` (will be
-in `JAVA_HOME=/usr/pkg/java/openjdk17`).
+NOTE: For Jenkins agents, also need to `pkgin install openjdk21` (will be
+in `JAVA_HOME=/usr/pkg/java/openjdk21`). Note that the location may be not
+exposed in `PATH` by default, so you may want to edit your system-wide
+`/etc/profile` or build user's `~/.profile` with:
+
+------
+JAVA_HOME="/usr/pkg/java/openjdk21"
+export JAVA_HOME
+PATH="${JAVA_HOME}/bin:${PATH}"
+export PATH
+------
 
 OpenIndiana as of releases 2021.10 to 2024.04
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/config-prereqs.txt
+++ b/docs/config-prereqs.txt
@@ -1159,6 +1159,9 @@ Such file was not found in the test system...
 :; pkgin install \
     openipmi
 ------
+
+It still seems to not work for NUT's expectations of FreeIPMI
+(we have basic detection support for OpenIPMI but no code yet).
 ======
 
 Recommended: For compatibility with common setups on other operating
--- a/docs/config-prereqs.txt
+++ b/docs/config-prereqs.txt
@@ -1070,6 +1070,24 @@ resources when generating the `configure` script may be old and its directory
 creation mode is not safe for parallel-`make` installations. If this happens,
 you can work around by `make MKDIRPROG="mkdir -p" install -j 8` for example.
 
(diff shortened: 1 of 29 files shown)


gdt:
More or less read the code and looks good.

A nit is that the unigned char is necessary not because it might be an array but because POSIX says it's UB to pass a value that is other than EOF and those that are legal for unsigned char.   (It's implicit that implementations may optimize and are free to produce UB when the spec says UB.)   Not a big deal, but so many people fail to understand the spec/UB issue.

jimklimov:
Why is NetBSD's shell `sort` program so broken? :)

Usual alphanumeric sort: decent
```
-bash-5.1$ printf '5\n55\n8\n4\n1\nqwe\nabc\nabc3\nabc1\na2\n' | sort
1
4
5
55
8
a2
abc
abc1
abc3
qwe
```

With numeric argument: broken - pure numeric strings are sorted as numbers go (e.g. 55 after 8), but other strings just dumped in same order as they were, at start of output (numbers after letters ignored and not sorted either):
```
$ printf '5\n55\n8\n4\n1\nqwe\nabc\nabc3\nabc1\na2\n' | sort  -n
qwe
abc
abc3
abc1
a2
1
4
5
8
55
```

* Same on all other systems I saw so far:
````
$ printf '5\n55\n8\n4\n1\nqwe\nabc\nabc3\nabc1\na2\n' | sort  -n
a2
abc
abc1
abc3
qwe
1
4
5
8
55
````

**Comment by jimklimov** [comment]
I'd invoke principle of least surprise: output of a `sort` program should be reasonably sorted :)

I agree pedantically it is not incorrect, e.g. pure-numeric sort is not defined for non-numeric tokens, so any order of those including original or random is not-wrong. But it is also not a useful result. Like in those fun tales about mathematicians' replies.

Do you know OTOH a POSIX CLI option combo to sort alphanumerically and number-aware for punctuation/whitespace surrounded numbers with leading parts of the strings being equal, at the same time?

**Comment by jimklimov** [comment]
Valid points. And your example is intriguing. I figured expression of complex desires would go through key specs (how else), but did not realize you can specify the same key twice - always did them for different numbers...

"Live for a century, learn for a century!" indeed :)

**Comment by gdt** [comment]
I had never thought about specifying a key twice, but the spec clearly doesn't say you can't, and sure enough it worked.

**Comment by jimklimov** [comment]
One more issue cropped up, similar to #2782 seen in OmniOS builds: the pkgsrc libraries are located under `/usr/pkg/lib` which is not in the compilers'/linker's default search path. However their `pkg-config` metadata does not suggest `-R` options to link them, e.g.:
```
:;  cat /usr/pkg/lib/pkgconfig/libssl.pc
prefix=/usr/pkg
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: OpenSSL-libssl
Description: Secure Sockets Layer and cryptography libraries
Version: 3.3.2
Requires.private: libcrypto
Libs: -L${libdir} -lssl
Cflags: -I${includedir}
```

As a result, during NUT build the toolkits see them (thanks to `-L`) and link the binaries, but at run-time they fail to start with messages like:
```
upsd: Shared object "libssl.so.3" not found
```

In OmniOS this was accepted as a distro error, maybe was fixed since.

As a hot-fix, hacking `LDFLAGS="-R/usr/pkg/lib"` to NUT `configure` runs helps, but is not ideal.

**Comment by jimklimov** [comment]
Despite having the packages installed, NUT failed to make use of:
* libltdl => no nut-scanner : lacks a pkgconfig file so is not found in /usr/pkg/lib (works when libs/cflags options are added explicitly though)
* net-snmp : net-snmp/net-snmp-config.h: No such file or directory (which package can that be in? `find / -name '*snmp*.h'` indeed finds nothing here)

**Comment by gdt** [comment]
I agree that a pkg-config file with -L and not -R is a bug and should be patched in pkgsrc (and probably fixed upstream, but many upstreams take an "of course you are using Linux and installing in /usr" attitude).

As for things not having pkg-config, it is historically normal to add CPPFLAGS/LDFLAGS to a program build to tell it where dependencies are.   Then, various programs grew `foo-config` which you could call with `--cflags` etc and then those were abstracted into `pkg-config` which has a standard interface and data files.   But we have not yet arrived at "there is a rule that every program must have pkg-config and if not it's defective".

Overall, I would say that when building a program where dependencies are in some prefix, one has to pass CPPFLAGS/LDFLAGS, and it's a special case on most GNU/Linux that everything is in /usr.    When pkgsrc builds programs, it does pass CPPFLAGS/LDFLAGS.    This is almost certainly why the failure to have -R in openssl has not been noticed and fixed.

File: docs/config-prereqs.txt [hunk 1]
```diff
@@ -1181,9 +1181,9 @@ systems, can add dash-number suffixed symlinks to compiler tools (e.g.
 `gcc-7` beside the `gcc` installed by package) near the original
 binaries and into `/usr/lib/ccache`:
 ------
-:; ( cd /usr/bin && for TOOL in cpp gcc g++ ; do \
-        for VER in "-7" "-14" ; do \
-            ln -s "$TOOL" "$TOOL$VER" ; \
+:; ( cd /usr/pkg/bin && for TOOL in cpp gcc g++ ; do \
+        for VER in "7" "14" ; do \
+            ln -fs "../gcc$VER/bin/$TOOL" "${TOOL}-${VER}" ; \
```
gdt, line 1186: This suggests you are running into "nut can't just build with the system compiler", which is gcc7 on NetBSD 9, gcc10 on 10, and gcc12 on -current.

Or is this "intentionally do builds with different compilers to provoke different kinds of warnings, to check them all"?  In which case, nevermind, makes sense.

jimklimov: Yes, the idea is that https://github.com/networkupstools/jenkins-dynamatrix and https://github.com/networkupstools/nut/blob/master/Jenkinsfile-dynamatrix conspire to dynamically define a build matrix and cover different toolkits, with default scripts expecting patterns like `${COMPILER}-${COMPILERVER}` where versions are iterated. With known compiler versions, we can then automatically iterate a number of C standards known supported by the toolkit release.

By the silent default, it is up to autotools to find something (which is typically some `gcc` first seen via the `PATH`).

jimklimov:
Some hiccup in tests against Mozilla NSS - does not seem to honour the `-R` in linker flags so does not find `libssl3.so` during test case start.

jimklimov:
Will separate spellcheck revision as not netbsd-specific.

jimklimov:
So indeed, the linking calls go like this:

* `nut-scanner` (note just `-L/usr/pkg/lib/nss -L/usr/pkg/lib/nspr` as one mention of the pathS involved):
````
/bin/sh ../../libtool  --tag=CC   --mode=link gcc -I../../clients -I../../clients -I../../include -I../../include -isystem /usr/pkg/include  -m64 -I/usr/pkg/include -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Wno-system-headers -Wall -Wextra -Wsign-compare -pedantic -Werror   -m64 -R/usr/pkg/lib -o nut-scanner nut_scanner-nut-scanner.o libnutscan.la -lpthread

libtool: link: gcc -I../../clients -I../../clients -I../../include -I../../include -isystem /usr/pkg/include -m64 -I/usr/pkg/include -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Wno-system-headers -Wall -Wextra -Wsign-compare -pedantic -Werror -m64 -o .libs/nut-scanner nut_scanner-nut-scanner.o  ./.libs/libnutscan.so -L/usr/pkg/lib -L/usr/pkg/lib/nss -L/usr/pkg/lib/nspr /usr/pkg/lib/libltdl.so -lssl3 -lsmime3 -lnss3 -lnssutil3 -lplds4 -lplc4 -lnspr4 -lpthread -Wl,-rpath -Wl,/usr/local/ups/lib -Wl,-rpath -Wl,/usr/pkg/lib
````

* `upsc`:
````
/bin/sh ../libtool  --tag=CC   --mode=link gcc -I../include -I/usr/pkg/include/nss -I/usr/pkg/include/nspr  -isystem /usr/pkg/include  -m64 -I/usr/pkg/include -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Wno-system-headers -Wall -Wextra -Wsign-compare -pedantic -Werror  -m64 -R/usr/pkg/lib -o upsc upsc.o ../common/libcommonclient.la libupsclient.la  -Wl,-R/usr/pkg/lib/nss -L/usr/pkg/lib/nss -lssl3 -lsmime3 -lnss3 -lnssutil3 -L/usr/pkg/lib/nspr -Wl,-R/usr/pkg/lib/nspr -lplds4 -lplc4 -lnspr4  -lpthread

libtool: link: gcc -I../include -I/usr/pkg/include/nss -I/usr/pkg/include/nspr -isystem /usr/pkg/include -m64 -I/usr/pkg/include -Wno-reserved-identifier -Wno-unknown-warning-option -std=gnu99 -Wno-system-headers -Wall -Wextra -Wsign-compare -pedantic -Werror -m64 -o .libs/upsc upsc.o -Wl,-R/usr/pkg/lib/nss -Wl,-R/usr/pkg/lib/nspr  ../common/.libs/libcommonclient.a ./.libs/libupsclient.so -L/usr/pkg/lib/nss -L/usr/pkg/lib/nspr -lssl3 -lsmime3 -lnss3 -lnssutil3 -lplds4 -lplc4 -lnspr4 -lpthread -Wl,-rpath -Wl,/usr/local/ups/lib -Wl,-rpath -Wl,/usr/pkg/lib
````

Issue #1074, opened 2021-08-19
NUT and "Compensating for high input voltage" msg


djacobson:
I am trying to replicate this issue at the moment but, I was wondering if NUT would report an APC's "Compensating for high input voltage" and "No longer compensating for high input voltage" warning msgs as logged in APC PowerChute?

jimklimov:
Hi, sorry this was not noticed in time... Still, I believe these conditions should pop up as `TRIM` and `BOOST` device states, when they are active.

jimklimov:
Reporting is done by clients, most commonly `upsmon`. Indeed it does not have a message template (nor code and variables for memory) about beginning/ending the power compensations: `BOOST`, `TRIM` (aka `BUCK` in some docs but no drivers publish such a token in `ups.status`), `OVER`(load) -- so the issue remains valid.

In fact, feels related to recent work done to report entering/exiting `ECO` state, getting or clearing `ALARM`'s, etc.

jimklimov:
Posted a PR to this effect, should be one of the last items remaining in NUT v2.8.3 scope, to finally release it.

PR #2873, opened 2025-04-03
clients/upsmon.{c,h} et al: introduce {NOT,}{OVER,TRIM,BOOST} notifications

jimklimov:
Closes: #1074

FWIW, CC @desertwitch  :)


clients/upsmon.c
```diff
@@ -1278,6 +1280,32 @@ static int is_ups_critical(utype_t *ups)
...
+		}
+
+		if (flag_isset(ups->status, ST_BOOST)) {
+			upslogx(LOG_WARNING,
+				"UPS [%s] was last known to be "
+				"boosting incoming voltage "
+				"and currently is not communicating, assuming dead",
+				ups->sys);
+			return 1;
+		}
```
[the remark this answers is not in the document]
jimklimov: I did have similar design concerns, hence the "request" for a second opinion :) Thanks!

clients/upsmon.c
```diff
@@ -1270,21 +1272,32 @@ static int is_ups_critical(utype_t *ups)
...
+		if (flag_isset(ups->status, ST_OVER)) {
 			upslogx(LOG_WARNING,
-				"UPS [%s] was last known to be calibrating "
-				"and currently is not communicating, assuming dead",
+				"UPS [%s] was last known to be overloaded "
+				"and currently is not communicating, "
+				"but we make no assumptions if it is dead now",
 				ups->sys);
-			return 1;
 		}
```
desertwitch, line 1283: Hmm, I personally would have kept `OVER` for instant criticality on communications lost. Just not `TRIM`/`BOOST`. Because an overloaded UPS is - I think - a critical enough situation as it stands, and a possible fire hazard at that, so immediate shutdown seems warranted here (e.g. excess heat already melted the communication bus...)

desertwitch: On the other hand, a super short spike in power draw (mass drive spin-up comes to mind here) may as well temporarily overwhelm the UPS and its communication abilities to the point of causing such a short communication blip and taking down the system... Not sure, perhaps we should make this a configuration toggle with `OVER` + nocomms defaulting to critical (for playing it safe)?

jimklimov: Possibly, like with throttle variables for some other events/states.

Not sure I want to dive into this sort of change before the release though. As a newly reported situation we had not handled or reported before, this PR as is at least should not cause unexpected behaviors, I think - only more log noise when stars mis-align. Now just the more-frequent polling (like in an outage) remains, but I don't believe it would be a big problem per se.

desertwitch: Fair, otherwise LGTM.

Issue #2881, opened 2025-04-07
Forza FX-1500LCD (0665:5161) not supported by nutdrv_qx or blazer_usb on Windows/MSYS2 (unlike FX-2200LCD #2426)


andresg97:
## Summary

My Forza FX-1500LCD UPS, which uses the common USB ID 0665:5161, is not correctly detected or supported by either the `blazer_usb` or `nutdrv_qx` drivers in a recent NUT build on Windows using MSYS2. This is despite issue #2426 reporting success with the FX-2200LCD using the same VID/PID and the `nutdrv_qx` driver. The official Forza monitoring software *does* work with the UPS.

## Hardware Details

*   **UPS Model:** Forza FX-1500LCD
*   **USB Vendor ID (VID):** 0665
*   **USB Product ID (PID):** 5161
*   **Identification:** Appears in driver logs as "INNO TECH" "USB to Serial"

## Software Details

*   **Operating System:** Windows 11 
*   **NUT Installation:** Built from source using MSYS2 UCRT64 environment.
*   **NUT Version:** Git master branch, post-2.8.2 (specifically `2.8.2.2725-2725+g90c75da1b` from logs, built approx. April 2025)

## Steps Tried

1.  Used `blazer_usb` driver (suggested by older HCL) with `port = auto`. Failed ("No supported UPS detected").
2.  Used `blazer_usb` driver with explicit `vendorid=0665`, `productid=5161`. Failed ("No supported UPS detected").
3.  Tried forcing protocols (`megatec`, `mustek`, `megatec/old`, `zinto`) with `blazer_usb`. Failed.
4.  Tried forcing subdrivers (`cypress`, `phoenix`, `ippon`, etc.) with `blazer_usb`. Failed.
5.  Used `nutdrv_qx` driver with `port = auto`. Failed ("Device not supported!").
6.  Used `nutdrv_qx` driver with the configuration reported working for FX-2200LCD in issue #2426 (explicit VID/PID, product, vendor). **Failed ("Device not supported!")** - see logs below.
7.  Tried forcing protocols (`q1`, `megatec`, `voltronic`, `voltronic-qs`, etc.) with `nutdrv_qx`. Failed.
8.  Tried forcing subdrivers (`cypress`, `phoenixtec`, etc.) with `nutdrv_qx`. Failed.

## Expected Result

[...]

Issue #1479, opened 2022-07-13
Investigate C++ code and a few drivers that are currently effectively "neutered" by `#ifndef WIN32`


jimklimov:
To facilitate having a completely buildable (not necessarily completely functional) NUT codebase for Windows, some code was rudely commented away for the platform (builds only `#ifndef WIN32`) to not waste time for investigations at that point. It would be nice *now* to revisit it:

* [ ] clients/nutclient.cpp (C++11 may also be neutered in configure.ac, clients/Makefile.am and/or tests/Makefile.am)
* [ ] drivers/belkinunv.c - port locking (see TODO comments)
* [ ] drivers/clone-outlet.c - `last_connfail` support not build in WIN32, why?..

I think there were some more (I think there was a driver or tool wholly ifdef'ed out) but can't find easily - update the list while trawling the code for other tasks.

PR #2882, opened 2025-04-07
WIN32: add run-time warnings and in-code comment tags for known-incomplete platform code base with `NUT_WIN32_INCOMPLETE*`

jimklimov:
Follow-up from #1479 thanks to reminder from #2881: with this PR I revisited all mentions of `#if(n)def WIN32` to check context and comments of the code blocks to see if they are no-ops waiting for completion, and added either `NUT_WIN32_INCOMPLETE*()` macros for active message emission, or `NUT_WIN32_INCOMPLETE` in comments, either way to help more quickly locate and address those missing functionalities later.

### DIFF
--- a/clients/message.c
+++ b/clients/message.c
@@ -11,4 +11,4 @@ int main(int argc, char ** argv)
 
 	return 0;
 }
-#endif
+#endif	/* WIN32 */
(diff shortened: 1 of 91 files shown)


jimklimov:
Tested locally in different mingw systems, CI skipped.

PR #2893, opened 2025-04-10
Fix build on Debian 13

jimklimov:
Added a Debian 13 build agent (LXC container) to the NUT CI farm. The system has a few differences from Debian 12, notably mis-detection of systemd directories (new prereq package needed), and more complaints from its brew of clang-19 static analysis.

CC @axet


axet:
Thanks for digging it up. RIP my old debian install.

PR #2903, opened 2025-04-15
FTY remerge as of 20250415 with NUT tag v2.8.3-rc4 and 20250421 with tag v2.8.3-rc5

jimklimov:
Follow-up to #2875 to verify new RC still behaves nice


teoberi:
Why does NUT CI farm - Jenkins checks take so long a day and almost 9 hours so far in this case?

jimklimov:
It runs a matrix of several thousand builds with different dependencies and build settings, varied toolkits, compiler versions, distro releases, etc. Quite often just one combo acts up, so quality-wise across the big board this is worth it. 

The builds also compete for computers involved, so most of the time scenarios wait in a queue.

teoberi:
It took 2 days and a few dozen minutes to check each of the 3 Pull requests left open in the case of NUT 2.8.3.
It seems a bit long to me, but I hope the wait is over.

jimklimov:
Well... the architecture is open to people providing their hardware as workers for the NUT CI farm, although the original intent was to register unique workers (like when someone cares for NUT on AIX or HPUX, and has a machine, it could be made available as a swarm agent and dynamatrix cells to test for its toolkits would get generated). So far it is mostly VMs of various linux/bsd/illumos/... systems, some hosted on DigitalOcean donated resources. My home is also heated by one of such machines.

teoberi:
It works! It works!
The only thing that still bothers me is the -rc5 suffix!

jimklimov:
Note that many of the messages (like NUT version banners, SSL not-init, etc.) can now also be silenced with environment variables, commonly exported by an init-script, systemd unit, your shell profile (for CLI commands), etc. - see `nut.conf.sample` comments for an overview.

teoberi:
I know, you told me that and I already did it.

PR #2908, opened 2025-04-18
upsmon: introduce OVERCRITICAL for OVER+NOCOMM states

### LINKED ISSUE #2877: Add temporal limit and optional criticality for OVER(load)/TRIM/BOOST in upsmon
> Hmm, I personally would have kept OVER for instant criticality on communications lost. Just not TRIM/BOOST. Because an overloaded UPS is - I think - a critical enough situation as it stands, and a possible fire hazard at that, so immediate shutdown seems warranted here to reduce (over-)load (e.g. excess heat already messing with the communication bus...). I'm thinking poorly ventilated, dusty consumer UPS with daisy-chained power strips type of setups... 😉
> On the other hand, a super short spike in power draw (mass drive spin-up comes to mind here) may as well temporarily overwhelm the UPS and its communication abilities to the point of causing such a short communication blip and taking down the system... Not sure, perhaps we should make this a configuration toggle with `OVER` + nocomms defaulting to critical (for playing it safe)?

_Originally posted by @desertwitch in https://github.com/networkupstools/nut/pull/2873#discussion_r2028576305_

desertwitch:
closes #2877 

introduces `OVERCRITICAL` for upsmon.

```
*OVERCRITICAL* 'seconds'::

This setting handles how a UPS that is overloaded should be treated in
situations when it is not fully communicating. Because such a UPS may 
be in a potentially severe state, some users may want their systems 
to be shutdown either immediately or after a set timeout has elapsed. 
The OVERCRITICAL setting defines this timeout (in seconds), after which 
an overloaded UPS that is not communicating will be considered critical.
+
A negative value means an overloaded UPS will never be considered 
critical (at least not because of the overload itself). A value of zero 
means the UPS will instantly be considered critical when overloaded and 
not communicating. Built-in default value is -1 (seconds).
+
NOTE: This setting only affects the behavior when a UPS is both 
overloaded and not communicating. When the UPS is communicating 
normally, an overload condition may raise notifications but won't 
trigger a system shutdown on its own.
```

We may later extend this to `TRIM` and `BOOST` but to me that makes no sense at present. Boosting/trimming input voltage is to me not a potentially severe state but rather within expected operation of any UPS, so nothing FSD-worthy even in NOCOMM.

As for the default value, I think something not `-1` would be safer/preferable, but that'd change the present release behavior?

### DIFF
--- a/INSTALL.nut.adoc
+++ b/INSTALL.nut.adoc
@@ -569,6 +569,19 @@ binaries that the package would put into an obscure location like
 would refer to new locations specified by the current build, so such old
 binaries would just consume disk space but not run.
 
+It is recommended to first try installing into a prototyping area, so you
+can review which files get delivered and perhaps which locations you may
+have to tune for a more tightly tailored replacement of an older (packaged)
+installation, in case whole swathes of files that you expect to be present
+in the current system (libraries, drivers, docs) are not getting installed
+by the new build into same path names:
+
+----
+:; rm -rf /tmp/nut ; make DESTDIR=/tmp/nut install -j 8
+:; (cd /tmp/nut && find . | sort) | while read N ; \
+   do [ -e "/$N" ] || echo "=== MISSING: /$N" >&2 ; done
+----
+
 Replacing any NUT deployment
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/INSTALL.nut.adoc
+++ b/INSTALL.nut.adoc
@@ -206,7 +206,7 @@ To build NUT from a Git checkout you may need some additional tools
 (referenced just a bit below) and run `./autogen.sh` to generate the
 needed files. For common developer iterations, porting to new platforms,
 or in-place testing, running the `./ci_build.sh` script can be helpful.
-The "<<Installing_inplace,Building NUT for in‐place upgrades or non‐disruptive
+The "<<Installing_inplace,Building NUT for in-place upgrades or non-disruptive
 tests>>" section details some more hints about such workflow, including some
 `systemd` integration.
 
@@ -271,8 +271,8 @@ do this, you will have problems later on when you try to start upsd.
 Build and install
 ~~~~~~~~~~~~~~~~~
 
-NOTE: See also <<Installing_inplace,Building NUT for in‐place upgrades
-or non‐disruptive tests>>.
+NOTE: See also <<Installing_inplace,Building NUT for in-place upgrades
+or non-disruptive tests>>.
 
 [[Configuration]]
 Configuration
@@ -429,7 +429,7 @@ You are now ready to configure NUT, and start testing and using it.
 You can jump directly to the <<Configuration_notes,NUT configuration>>.
 
 [[Installing_inplace]]
-Building NUT for in‐place upgrades or non‐disruptive tests
+Building NUT for in-place upgrades or non-disruptive tests
 ----------------------------------------------------------
 
 NOTE: The NUT GitHub Wiki article at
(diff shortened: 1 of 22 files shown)


desertwitch:
Actually this makes me wonder if we should make this more broad to cover not only non-communication situations but also generally anytime a overloaded UPS is not fully online (could be `OB`/NOCOMM...)?

Maybe as part of a more generic `OVERTIMEOUT` instead of `OVERCRITICAL`, or both? I could well see users wanting an overloaded UPS to FSD a system immediately (probably not) or after a short timeout (more likely) when drawing (excessive) power from battery to protect their UPS and battery health... this style of setting could be flexible enough here to cover all bases?

jimklimov:
I think it suffices to leave this for the no-comm situation. *Probably* the UPS is protected from protracted overload by hardware in some way, whether by popping or melting a fuse, or turning off with no warning to the load (like an EPO contact might do for fire emergencies), and in software we can't make reliable guesses that would be good for everyone.

Now that OVER is a notifiable event, `upssched` can take over the field work for end users who want to time-limit this situation for their particular UPS.

I might envision that the situation may happen in redundant setups with N+M PSU in the servers with each fed from a different UPS, where the individual line power draw is normally shared but might jump up when one of the UPSes is off for long maintenance, or quickly restarting, etc. - which is something end-users might know and we do not.

desertwitch:
> I think it suffices to leave this for the no-comm situation. *Probably* the UPS is protected from protracted overload by hardware in some way, whether by popping or melting a fuse, or turning off with no warning to the load (like an EPO contact might do for fire emergencies), and in software we can't make reliable guesses that would be good for everyone.
> 
> 
> 
> Now that OVER is a notifiable event, `upssched` can take over the field work for end users who want to time-limit this situation for their particular UPS.
> 
> 
> 
> I might envision that the situation may happen in redundant setups with N+M PSU in the servers with each fed from a different UPS, where the individual line power draw is normally shared but might jump up when one of the UPSes is off for long maintenance, or quickly restarting, etc. - which is something end-users might know and we do not.

Agreed, sounds good. 👍

jimklimov (approved):
Simple but useful, complete with docs - LGTM :)

Thinking of it, probably there are trivial augeas and C++ lines to add for the new keyword to become known to config parsers.

jimklimov (commented):
Thanks for the quick update!

Looking at the naming that popped up in the context, the nearby `ALARMCRITICAL` is a flag (0/1) while other similar items accounted in seconds (when not negative) were a something-duration. WDYT about renaming this new feature into `OVERDURATION` for consistency, would it make better sense?

desertwitch:
> Thanks for the quick update!
> 
> Looking at the naming that popped up in the context, the nearby `ALARMCRITICAL` is a flag (0/1) while other similar items accounted in seconds (when not negative) were a something-duration. WDYT about renaming this new feature into `OVERDURATION` for consistency, would it make better sense?

Yep, that actually makes better sense - I'll amend that, do you want me to add `ALARMCRITICAL` to the augeas/cpp files as part of this PR while we're at it or better as a separate follow-up PR?

jimklimov:
Yes, please - better now to save on CI cycles :)

jimklimov:
Hm, that recent force-push seems to include commits from other in-flight PRs :\

desertwitch:
Hmm, seems a rebase got messed up somewhere - will reopen with a clean branch.

> Hm, that recent force-push seems to include commits from other in-flight PRs :\

Yep, something must've gone wrong during the rebasing - will reopen with a clean branch.

jimklimov:
It also helps to prepare a new branch locally, then `git reset` the PR'ed one to the result.

desertwitch:
> It also helps to prepare a new branch locally, then `git reset` the PR'ed one to the result. Or use the interactive mode `git rebase -i upstream/master` and remove the lines for unneeded commits.

Yeah, sorry about that, I totally should've just done that - Friday evening... 😮‍💨 
To be continued in #2909

PR #2909, opened 2025-04-18
upsmon: introduce OVERDURATION for OVER+NOCOMM states

### LINKED ISSUE #2877: Add temporal limit and optional criticality for OVER(load)/TRIM/BOOST in upsmon

desertwitch:
continuation of #2908
closes #2877 

- introduces `OVERDURATION` for upsmon
- adds `ALARMCRITICAL` to augeas/cpp files

```
*OVERDURATION* 'seconds'::

This setting handles how a UPS that is overloaded should be treated in
situations when it is not fully communicating. Because such a UPS may 
be in a potentially severe state, some users may want their systems 
to be shutdown either immediately or after a set timeout has elapsed. 
The OVERDURATION setting defines this timeout (in seconds), after which 
an overloaded UPS that is not communicating will be considered critical.
+
A negative value means an overloaded UPS will never be considered 
critical (at least not because of the overload itself). A value of zero 
means the UPS will instantly be considered critical when overloaded and 
not communicating. Built-in default value is -1 (seconds).
+
NOTE: This setting only affects the behavior when a UPS is both 
overloaded and not communicating. When the UPS is communicating 
normally, an overload condition may raise notifications but won't 
trigger a system shutdown on its own.
```

We may later extend this to `TRIM` and `BOOST` but to me that makes no sense at present. Boosting/trimming input voltage is to me not a potentially severe state but rather within expected operation of any UPS, so nothing FSD-worthy even in NOCOMM.

As for the default value, I think something not `-1` would be safer/preferable, but that'd change the present release behavior?


desertwitch:
@jimklimov : Apologies for the messed up rebase, not sure what went wrong there.
In retrospect should've reset the branch (seen your message too late), not my day... 😉

jimklimov:
Good point, thanks!

PR #2911, opened 2025-04-21
Fix NUT SEMVER support for concept of release-candidates

jimklimov:
If current commit is git-tagged for a pre-release (or e.g. `NUT_VERSION_FORCED` value in tarball builds does similar), and conveys that this source is an `X.Y.Z-(rc|alpha|beta)N` then treat its SEMVER as `X.Y.Z` (and track pre-release status, suffix, and tag info) even though the precise detailed version may be `X.Y.(Z-1)` based.

Example:
````
:; ./server/upsd -V
Network UPS Tools upsd 2.8.2.2878.3-2881+g45029249f+v2.8.3+rc6 (pre-release iteration of 2.8.3)
````

Including into v2.8.3 release trail before the release window closes, as it is the NUT release which introduced the NUT SEMVER definition and tooling.


jimklimov:
CI hiccups unrelated to PR change.

New deb13 worker has different libgpiod than another (updated one from deb12, recently another from deb11) causing a previously unseen codepath build?..

PR #2919, opened 2025-04-26
drivers/cps-hid.c: add report descriptor fix for ConfigActivePower

### LINKED ISSUE #2917: Cyber Power EC850LCD reports wrong output.realpower.nominal
Hello, 

From the spec of the device and APCUd it's 510W.

However, nut is reporting 450W.

It's connected through USB. And I'm using nut 2.8.3 on unraid as a plugin.

desertwitch:
fix #2917 

Findings in the linked issue suggest that at least one model in the `cps_fix_report_desc` model range suffers also from a broken descriptor for reporting the UPS nominal power (`ConfigActivePower`). The value seemingly gets correctly sent by the UPS, but is then clipped by a too low `LogMax` of that affected descriptor:  [nut-issue-2917.log](https://github.com/user-attachments/files/19923182/nut-issue-2917.log)

```
Apr 26 08:43:04 Tower rc.nut:    0.035598#011[D3:181672] Report[get]: (3 bytes) => 18 fe 01
Apr 26 08:43:04 Tower rc.nut:    0.035600#011[D5:181672] PhyMax = 0, PhyMin = 0, LogMax = 450, LogMin = -1
Apr 26 08:43:04 Tower rc.nut:    0.035601#011[D5:181672] Unit = 0000d121, UnitExp = 7
Apr 26 08:43:04 Tower rc.nut:    0.035602#011[D5:181672] Exponent = 0
Apr 26 08:43:04 Tower rc.nut:    0.035603#011[D5:181672] hid_lookup_path: 00840004 -> UPS
Apr 26 08:43:04 Tower rc.nut:    0.035604#011[D5:181672] hid_lookup_path: 0084001c -> Output
Apr 26 08:43:04 Tower rc.nut:    0.035604#011[D5:181672] hid_lookup_path: 00840044 -> ConfigActivePower
Apr 26 08:43:04 Tower rc.nut:    0.035606#011[D1:181672] Path: UPS.Output.ConfigActivePower, Type: Feature, ReportID: 0x18, Offset: 0, Size: 16, Value: 450
```

We can see above that 510W (correct) gets sent (`Report[buf]: (3 bytes) => 18 fe 01`) but is clipped by `LogMax = 450` (wrong).

This PR introduces a simple enough fix as part of `cps_fix_report_desc`, operating under the assumption that other models with broken descriptors may also require this descriptor fixed and that the value sent by the UPS is in principal the correct value.

Hard-setting the `LogMax` to a reasonable ceiling value of `2048` (W/VA) as an early investigation fix should cover most UPS of the affected ranges and have little negative side affects as long as correct values are in principal sent by the UPS. This can later be expanded to more advanced fiddling, should further investigations result in incorrect values getting sent for either.

### DIFF
--- a/drivers/cps-hid.c
+++ b/drivers/cps-hid.c
@@ -669,25 +669,29 @@ static int cps_fix_report_desc(HIDDevice_t *pDev, HIDDesc_t *pDesc_arg) {
 
 	/* Fix for nominal power reporting getting clipped by LogMax. */
 	if ((pData=FindObject_with_ID_Node(pDesc_arg, 24 /* 0x18 */, USAGE_POW_CONFIG_ACTIVE_POWER))) {
+		long power_logmax = pData->LogMax;
+
 		upsdebugx(4, "Original Report Descriptor: ConfigActivePower "
 			"LogMin: %ld LogMax: %ld",
-			pData->LogMin, pData->LogMax);
+			pData->LogMin, power_logmax);
 
-		/* Set a generous maximum that won't restrict UPS reporting.
-		 *
-		 * Current findings suggest that the values sent by the UPS are
-		 * accurate, but then get clipped by too strict LogMax thresholds:
-		 * https://github.com/networkupstools/nut/issues/2917#issuecomment-2832243477
-		 *
-		 * This could later be expanded to more advanced fiddling, where necessary.
-		 */
-		pData->LogMax = CPS_WATTAGE_LOGMAX;
+		if (power_logmax < CPS_WATTAGE_LOGMAX) {
+			/* Set a generous maximum that won't restrict UPS reporting.
+			*
+			* Current findings suggest that the values sent by the UPS are
+			* accurate, but then get clipped by too strict LogMax thresholds:
+			* https://github.com/networkupstools/nut/issues/2917#issuecomment-2832243477
+			*
+			* This could later be expanded to more advanced fiddling, where necessary.
+			*/
+			pData->LogMax = CPS_WATTAGE_LOGMAX;
 
-		upsdebugx(3, "Fixing Report Descriptor: "
-			"set ConfigActivePower LogMax = %ld",
-			pData->LogMax);
+			upsdebugx(3, "Fixing Report Descriptor: "
+				"set ConfigActivePower LogMax = %ld",
+				pData->LogMax);
 
-		retval = 1;
+			retval = 1;
+		}
 	}
 
 	if (!retval) {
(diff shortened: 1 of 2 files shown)


drivers/cps-hid.c
```diff
@@ -666,6 +667,29 @@ static int cps_fix_report_desc(HIDDevice_t *pDev, HIDDesc_t *pDesc_arg) {
...
+
+		/* Set a generous maximum that won't restrict UPS reporting.
+		 *
+		 * Current findings suggest that the values sent by the UPS are
+		 * accurate, but then get clipped by too strict LogMax thresholds:
+		 * https://github.com/networkupstools/nut/issues/2917#issuecomment-2832243477
+		 *
+		 * This could later be expanded to more advanced fiddling, where necessary.
+		 */
+		pData->LogMax = CPS_WATTAGE_LOGMAX;
```
jimklimov, line 684: I think it should not be blanket-applied to every case, but if e.g. the original LogMax is smaller (so we do not truncate a 10KW UPS to 2KW either)? And perhaps if the value (like 510 in the practical case) is even smaller, if known at this point?

jimklimov (changes requested):
Commented about a change I think makes sense, ready to negotiate though :D 

Also a `NEWS.adoc` entry would be nice, and a bump of subdriver version.

drivers/cps-hid.c
desertwitch: Heh, was already at it while you commented, with the limited information we have perhaps just checking if smaller than reasonable is enough for now (until we learn more)? Let me know, will then add to NEWS and bump subdriver version.

desertwitch: I thought about going the value route too, but I wasn't sure if we can access the value at this point as other descriptor fixes seem not to interact with the value within the function. A value-based fix would definitely also make sense, if we can realise it, e.g. check if value > logMax && logMax < reasonable ceiling => set logMax to reasonable ceiling.

jimklimov: Yep, the change looks good. Checking for current value being under the reported LogMax is a cherry on top, and a bit of no-op (upon second thought) - we would anyway trim (or not) using the larger threshold value than originally reported.

desertwitch: Thanks for double-checking! I'll also rename the constant to something less ambigous, I think... will get to the remaining changes hopefully later tonight.

jimklimov (approved):

jimklimov:
CI error seems to have been a build agent hiccup, although one I have not seen before.

Not relevant for changes in this PR.

PR #2931, opened 2025-04-30
drivers/dstate: improve ALARM status safeguards, documentation and handling

### LINKED ISSUE #2928: dstate/dummy-ups: ALARM status does not clear after dirty entrance into state
Stumbled upon this with `dummy-ups` and the 2.8.3 release version:

Setting `ups.status` to `OL ALARM` correctly registers the alarm state. (Note: I did not expose `ups.alarm` hence as `N/A`)
`Apr 29 14:44:59 TowerOLD upsmon[3362741]: UPS ups@127.0.0.1: one or more active alarms: [[N/A]]` 
Setting `ups.status` back to `OL` does not clear the alarm state, `upsc` (still) reports `ups.status: ALARM OL`
Setting `ups.status` to `OL TRIM` in an attempt to clear the alarm state, `upsc` now reports `ups.status: ALARM OL TRIM`
`Apr 29 14:48:49 TowerOLD upsmon[3362741]: UPS ups@127.0.0.1: trimming incoming voltage` (nothing else)
Setting `ups.status` back to `OL` in another attempt to clear both states,  `upsc` now reports `ups.status: ALARM OL`
`Apr 29 14:52:34 TowerOLD upsmon[3362741]: UPS ups@127.0.0.1: no longer trimming incoming voltage` (nothing else)

I'm not sure if this is limited to `dummy-ups`, I haven't had time to investigate further so far.

desertwitch:
fix #2928

work in progress

### DIFF
--- a/docs/new-drivers.txt
+++ b/docs/new-drivers.txt
@@ -218,6 +218,10 @@ formatting string to those expectations.  In this case, you would use
 	char *fmt = "Mega-Zapper %d";
 	dstate_setinfo_dynamic("ups.model", fmt, "%d", rating);
 
+Please note that `ups.alarm` should no longer be manually set, but rather
+the appropriate alarm functions should be used instead. For more details,
+see below in the `UPS alarms` section.
+
 Setting flags
 ~~~~~~~~~~~~~
 
@@ -261,9 +265,10 @@ Possible values for status_set:
 	BOOST   -- UPS is boosting incoming voltage
 	FSD     -- Forced Shutdown (restricted use, see the note below)
 
-Internally, an `ALARM` value would be added (typically as first in the list)
-if the `ups.alarm` is currently not empty. For more details, see below in
-`alarm_set()` description.
+`ALARM` should no longer be raised through the UPS status.
+An `ALARM` value will be added internally (typically as first in the list)
+if an alarm was set and committed through the appropriate alarm functions.
+For more details, see below in the `UPS alarms` section.
 
 [NOTE]
 ======
--- a/drivers/dstate.c
+++ b/drivers/dstate.c
@@ -1801,6 +1801,7 @@ static int status_set_callback(char *tgt, size_t tgtsize, const char *token)
 		 */
 		upsdebugx(2, "%s: (almost) ignoring ALARM set as a status", __func__);
 		if (!alarm_status && !alarm_active && strlen(alarm_buf) == 0) {
+			upsdebugx(2, "%s: Assume sloppy driver coding that ignored alarm methods and did not set an alarm message: set an injected N/A value", __func__);
 			alarm_init();	/* no-op currently, but better be proper about it */
 			alarm_set("[N/A]");
 		}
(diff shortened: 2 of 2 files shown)


jimklimov:
CI complains about `FAIL driver_methods_utest`

jimklimov (approved):
Looks good, thanks

desertwitch:
CI will fail this again, but I think we need to talk in which direction to go here before further work.
Right now there are some open questions regarding both the safeguard logic and the test cases (which fail).

**The safeguard logic assumes that a driver is using `status_init()` but not `alarm_init()` (and its related functions).**
This is generally logic enough, although not sure how realistic that someone uses modern `status_init()` & not `alarm_init()`?

In any case, we can separate this into legacy logic and modern logic quite well (see last commit) - so that's so far so good...
**BUT** -- Perhaps the safeguard logic should cover a rudimentary driver only using `dstate_setinfo` but neither functions?
I think that scenario is more realistic than one modern function being used, but not the other?

**The test logic (and resulting failures) makes no sense to me, it assumes odd combinations of both modern/legacy code.**
This makes it hard to separate into what's legacy code, to activate the safeguards, and what's modern/current standard code.

The segmentation fault results from the safeguard thinking it's a cleared alarm state, nulling the buffer and then the test function subsequently querying a no longer existing `ups.alarm` (being already deleted) and dereferencing the received null pointer.

https://github.com/networkupstools/nut/blob/dfe6062959c30431de4c4e716655ac42f4cdadee/tests/driver_methods_utest.c#L142-L147

https://github.com/networkupstools/nut/blob/dfe6062959c30431de4c4e716655ac42f4cdadee/tests/driver_methods_utest.c#L158-L198

The mixture in these test cases makes no sense, either it should be legacy enough `status_set("ALARM")` but without any `alarm_` functions or it should be the full batch of `alarm_` functions. A mixture of both, I think, is a really unrealistic scenario.

I see three credible scenarios:

- A legacy driver just setting alarm via `ups.status` without either `status_` or `alarm_` functions.
- A legacy driver setting alarm via `ups.status` with `status_` functions but no `alarm_` functions.
- A modern driver setting alarm via and `alarm_` functions only, in which case it should persist `status_` changes.

So we see that legacy drivers could couple alarm states with the `ups.status` and should safeguard that accordingly, while also keeping the alarm state internally coupled to the status (as done in my last commit). In contrast, a modern driver should be able to call `status_` functions (including `status_init()`) without resetting appropriately published (and still active) `alarm_` states.

desertwitch:
P.S. Maybe the most clean solution would be to fix up the - on first glance +/- 3 remaining drivers - and completely remove this fallback logic in favour of appropriately using both the `status_` and `alarm_` functions (separately). Dragging along such legacy logic adds unnecessary complexity and coupling of alarm states to the UPS status... something the `alarm_` functions were probably intended to decouple in the first place? I mean we can theoretically still show a manually set `ALARM` state in the status for visibility, but just not make any assumptions and not trigger any further modern functions and injections?

jimklimov:
Thinking (slowly and in the background though, got some work to catch up with)...
It may well be that I did not see the forest behind one tree I was chopping at here (for #2708 related work).

desertwitch:
> Thinking (slowly and in the background though, got some work to catch up with)... It may well be that I did not see the forest behind one tree I was chopping at here (for #2708 related work).

No worries, just wanted to drain my thoughts to "paper" (sorry it ended up in a wall of text), mostly so I don't forget. I'll keep this as is for now, but will get started on a second alternative PR on how it would look without legacy code and simply handling textual alarms vs. modern alarms in `upsmon` instead (not needing to inject variables that don't exist in `dstate`). For later comparison. 😎

desertwitch:
superseded by #2934

PR #2934, opened 2025-04-30
dstate: clean separation of UPS-statused based ALARM and modern alarm_*-driven ALARMs

desertwitch:
follow up to #2931, fixes #2928 
implements the ideas of https://github.com/networkupstools/nut/pull/2931#issuecomment-2841705269

this PR amends the various involved parts to clean separation of legacy UPS-status coupled `ALARM`
and modern `alarm_`-driven `ALARM`, which are decoupled from UPS status aside from the token itself.

it removes making assumptions on legacy UPS-status coupled `ALARM` and removes workarounds trying
to port any such legacy `ALARM` statuses into the UPS-status decoupled `alarm_` schema, because this would
mean later having to couple them back to UPS-status again to be able to exit these alarm states. (within `dstate`)

this PR allows for both to co-exist peacefully with the only limitation being that a `ups.alarm` does not get
injected when it is not otherwise published directly and manually by the (legacy) driver. `ups.alarm` published
manually by the (legacy) UPS driver will still raise appropriate (seperate) notifications, be tracked by `upsmon`
and function normally as if they were raised by the modern `alarm_` functions (from the `upsmon` perspective).

the `upsmon` does not care where in `dstate` an `ALARM` token was set, nor if done so by legacy or modern method.

the tests now cover both legacy, mixed and modern scenarios respectively, as well as some more curious edge-cases.
further work is in progress to modernize the few remaining legacy drivers to use the `status_` and `alarm_` functions.
this ongoing work will later be published in separate PRs.

PR #2936, opened 2025-05-03
drivers/, dummy-ups: modernize status/alarm handling, ALARM instruction for dummy modes

### LINKED ISSUE #2937: clone.c, clone-outlet.c: UPS status handling needs modernization
follows up on work in #2934 , #2936 
issue should serve as a reminder for outstanding driver refactor work

`clone.c` and `clone-outlet.c` are the remaining drivers that directly set `ups.status`
instead of using contemporary best-practice `status_*` _dstate_ functions, we should update these as well.

desertwitch:
follow up to #2934 
fixes #2937 
implements ideas of #2928, #2931 

modernizes remaining 4 drivers to use the contemporary best-practice status/alarm handling:
```
 - `clone`, `clone-outlet`, `nhs_ser`, `nutdrv_qx_ablerex` driver updates:
   * Refactored to follow modern handling of status and alarm conditions,
     aligning with current driver design practices. This includes fixing
     copy-paste related issues in alarm reporting and removing some alarm
     messages that should instead be reflected as status flags.
```

introduces an `ALARM` instruction for dummy modes in `dummy-ups` to allow simulating modern alarms:
```
 - `dummy-ups` driver updates:
   * A new instruction `ALARM` was added for the `Dummy Mode` operation
     of the driver, enabling simulation of UPS alarm states more closely
     in line with modern, real-world UPS driver implementations. This
     follows the updated principle of keeping alarm states decoupled from
     the `ups.status` variable, with alarms now raised via common alarm
     functions rather than direct manipulation.
```
```
Another, more recently introduced instruction is `ALARM`, which allows
the simulation of UPS alarms in much the same way they would be raised
by real driver implementations. Modern drivers decouple alarm states
from the `ups.status` variable, and raising `ALARM` by setting it as a
status token is discouraged in favor of using modern, common functions
for raising alarms within the driver code.

The `ALARM` instruction is intended to simulate this behavior as closely
as possible. The value following an `ALARM` instruction is treated as
the alarm message and is eventually published as the `ups.alarm`
variable, with the `ALARM` token also set in `ups.status` by internal
driver mechanisms, rather than by directly manipulating the variable.
Multiple `ALARM` instructions will have their messages combined into
the `ups.alarm` variable, just as would happen with real driver logic.

Conversely, any `ALARM` instruction on its own will clear active alarm
states. See below for an example of setting and resetting alarms:

	ALARM [UPS too warm to charge]
	ALARM [UPS circuit is overheating]
	TIMER 5
	ALARM
	TIMER 5
	ALARM [UPS too cold to charge]
```


desertwitch:
Had some more time and figured this would also fit into this PR (and save on CI cycles), fixes #2937.
This PR should now conclude the effort to modernize all drivers to use `status_*` and `alarm_*` functions.

jimklimov (approved):
Looks reasonable, but there's a lot of change in clone driver that I couldn't quite wrap my head around. Did you have some chance to test that it still works (no worse than before)? :)

desertwitch:
> Looks reasonable, but there's a lot of change in clone driver that I couldn't quite wrap my head around. Did you have some chance to test that it still works (no worse than before)? :)

Yes, all thinkable scenarios were tested by me and the driver works as before. It may look like many changes but it's mostly moving things around to reduce redundant expensive checks via dstate to locations where we can also check it from the data we have. Happy to answer any questions.

I think the only behavioral change in `clone` really was also setting `OFF` when an outlet is dead to make this consistent with `clone-outlet` and obvious to the user that the outlet is not giving power.

As far `clone-outlet`, the driver did just set `FSD` or `OFF` and return before, but that basically exposed nothing else from the UPS statuses and may have left the user wondering what other condition the UPS was in. I figured for completeness there'd be no harm giving them a full picture of the UPS statuses while also raising those two even as possible combinations of each other as they are not mutually exclusive.

drivers/clone.c
```diff
@@ -49,8 +50,13 @@ static struct {
 		long	start;
 		long	shutdown;
 	} timer;
+	struct {
+		char	*off;
+		char	*on;
+		char	*status;
```
desertwitch, line 56: Stores the user set arguments for re-use, instead of `getval`ing them all over

drivers/clone.c
```diff
@@ -136,11 +142,21 @@ static int parse_args(size_t numargs, char **arg)
...
 			snprintf(ups.status, sizeof(ups.status), "%s", arg[2]);
 
-			online = strstr(ups.status, "OL") ? 1 : 0;
+			/* Status is published later in upsdrv_updateinfo() */
+			return 1;
+		}
+
+		if (ups.load.status && !strcasecmp(arg[1], ups.load.status)) {
+			if (!strcasecmp(arg[2], "off") || !strcasecmp(arg[2], "no")) {
+				outlet = 0;
```
desertwitch, line 151: Sets internal state from data we have, for later updating from directly in `upsdrv_updateinfo()`, instead of then needing to `dstate_getinfo` it again otherwhere to the same effect.

drivers/clone.c
```diff
@@ -478,23 +498,11 @@ static int sstate_dead(int maxage)
 
 static int instcmd(const char *cmdname, const char *extra)
 {
-	const char	*val;
-
-	val = dstate_getinfo(getval("load.status"));
-	if (val) {
-		if (!strcasecmp(val, "off") || !strcasecmp(val, "no")) {
-			outlet = 0;
-		}
```
desertwitch, line 487: Moved to `parse_args` where we have this information, if the outlet is off should not be set inside an instant command handling function, but where the data actually comes in and is parsed. This also obsoletes the need for the redundant dstate check which is expensive.

drivers/clone.c
```diff
@@ -562,6 +571,10 @@ void upsdrv_initinfo(void)
 		battery.runtime.low = strtod(val, NULL);
 	}
 
+	ups.load.off = getval("load.off");
+	ups.load.on = getval("load.on");
+	ups.load.status = getval("load.status");
```
desertwitch, line 576: Record the user given arguments into the internal state struct, reduces needs for constant `getval`ing them within update loops.

drivers/clone.c
```diff
@@ -598,22 +611,54 @@ void upsdrv_updateinfo(void)
...
+
+	online = status_get("OL");
+
+	if (!outlet) {
+		/* Outlet was declared OFF (by upstream or assumption) */
+		status_set("OFF");
+		if (ups.load.status && !reported_off) { /* first entrance */
+			reported_off = 1;
+			upslogx(LOG_WARNING, "%s: upstream reports outlet off (setting OFF)",
+				__func__);
```
desertwitch, line 625: Inform user that the outlet is **reported** off and set `OFF` to make it consistent with `clone-outlet`, the user can see on a first glance the outlet is not giving power.

drivers/clone.c
```diff
@@ -598,22 +611,54 @@ void upsdrv_updateinfo(void)
...
 				char	buf[SMALLBUF];
-				snprintf(buf, sizeof(buf), "INSTCMD %s\n", val);
+				upslogx(LOG_WARNING, "%s: sending '%s' to upstream (load.off)",
+					__func__, ups.load.off);
+				snprintf(buf, sizeof(buf), "INSTCMD %s\n", ups.load.off);
 				sstate_sendline(buf);
 			}
+
+			if (!ups.load.status) {
+				/* Better than nothing if no load.status argument was given,
```
desertwitch, line 653: If no argument was given where to check the outlet's actual status, we assume it to have successfully turned off after sending `ups.load.off` for lack of a variable for checking. If an argument was given, no assumptions are made and the actual status of the outlet is checked. 

Same as before, but it was just set here and overridden later in an unexpected place `instcmd`, this way should be cleaner and more traceable with debug messages.

drivers/clone-outlet.c
```diff
@@ -491,20 +493,20 @@ void upsdrv_updateinfo(void)
...
 		upsdebugx(2, "FSD flag set (%s: -1 < [%ld] <= %ld)", prefix.timer.shutdown, outlet.timer.shutdown, outlet.delay.shutdown);
-		dstate_setinfo("ups.status", "FSD %s", ups.status);
-		return;
+		status_set("FSD");
 	}
 
-	upsdebugx(3, "%s: power state not critical", getval("prefix"));
-	dstate_setinfo("ups.status", "%s", ups.status);
+	status_set(ups.status); /* FIXME: Split token words? */
+	status_commit();
```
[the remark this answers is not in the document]
desertwitch: Intended because the user should see all actual statuses and `OFF`/`FSD` should not be mutually exclusive, I did drop the debug message - sorry - but I think it made no sense when a UPS may well be `OB LB` or `OVER` even if not `OFF` or `FSD` yet, which would also be somewhat critical.

I did also pave it over for ease of coding around it, other than the argument above, truth be told, since all statuses (also critical ones) are reported summarily now.

jimklimov:
Super, thanks for the analyses!

Issue #2769, opened 2025-01-16
PowerMaster / PowerGuide (legacy?) serial Protocol support


3v1n0:
Hi, I've got an [Atlantis Land OnePower A03-S1200](https://www.atlantis-land.com/download/SCHEDATECNICA/A03-S1200.pdf) UPS and it doesn't seem to be recognized by the classic serial protocols so far (nor I see anything similar in the source code of the drivers I've quickly looked at).

The device shows up as a serial converter only on the USB side `Bus 001 Device 002: ID 1a86:7523 QinHeng Electronics CH340 serial converter` so I don't think there's any HID support (although I can't fully confirm it since in the NAS I'm running it, I've a limited number of kenel modules and I'm currently remote, so I can't plug it elsewhere :)).

However, thanks to [this read](https://infeeeee.github.io/sysadmin/njoy-ups.html) I ended up to try the [powerguide-1.0.0](https://www.powermonitor.software/#PowerGuideSoftware) software (easier to strace and debug compared to the PowerMaster java blobs) which gave me a some simple protocol information...

```
WRITE: "\r"
READ: "#-1\r"
WRITE: "K19CuRVvhy5ZyMgqZgH\r"
READ: "#0\r"
WRITE: "\r"
READ: "#-1\r"
WRITE: "X19\r"
READ: "#C930D6DF88DB1F30\r"
WRITE: "\r"
READ: "#-1\r"
WRITE: "B\r"
READ: "#I235.0O000.0L000B100V27.0F50.2H00.0R060S\200\204\320\200\300\200\r"
WRITE: "\r"
READ: "#-1\r"
WRITE: "B\r"
READ: "#I237.0O000.0L000B100V27.0F50.2H00.0R060S\200\204\320\200\300\200\r"
WRITE: "X41\r"
READ: "#1200,PE02022.002,000000000000,000000000000000000\r"
WRITE: "X72\r"
READ: "#1200,0720,230,40,70,05.20\r"
WRITE: "X27\r"
READ: "#230,290,162,10,300\r"
WRITE: "X15\r"
READ: "#10\r"
WRITE: "X87\r"
READ: "#\307\227\360\360\r"
WRITE: "P19\r"
READ: "#-1\r"
WRITE: "K601\r"
READ: "#0\r"
WRITE: "K411\r"
READ: "#-1\r"
WRITE: "\r"
READ: "#-1\r"
WRITE: "B\r"
[...]

PR #2940, opened 2025-05-05
drivers/nutdrv_hashx: Add new driver for PowerManager+ serial supported devices

3v1n0:
From the main commit:

```
Add an initial driver for the "#*" protocol support that is used by many
UPS out there that rely on the proprietary software suites PowerMaster+,
PowerMaster, PowerGuide and so on...

The naming of the driver has been chosen since all the UPS replies start
with an "#", while there's not a definitive common pattern for the queries
or commands.

Apparently the same protocol is used by various UPS devices, both online
and line-interactive (given that many producers advertise to use the
said programs), and only in serial mode (AFAIK). Although this driver has
solely been tested with the Atlantis Land S1200 line-interactive UPS.

So there may be missing features (such as the bypass mode) or lack of
information. However most of the decoding has been done and the driver
is enough to update the UPS status for monitoring and shutdown purposes.

As for now, the status updates and immediate commands features have been
implemented.
```

Few further things:
 - Mentioned more devices to the supported devices list (as per what their website state)
 - I would like to make the `nut-scanner` to look for these devices, but it would imply duplicating lots of `scan_eaton_serial.c` code or move it to a common place, so maybe that's another PR? As we don't have a generic place for serial scanning AFAIK.

## General points

  comments or text documents.

  case.

  concept.

- [x] Updated `data/driver.list.in` if applicable (new tested device info)

  the changed codebase.

Closes: #2769

### DIFF
--- a/drivers/nutdrv_hashx.c
+++ b/drivers/nutdrv_hashx.c
@@ -510,7 +510,7 @@ void upsdrv_updateinfo(void)
 	upsdebug_hex(6, "Poll: status bytes", status_bytes, status_bytes_size);
 
 	if (!status_bytes || status_bytes_size < 5) {
-		upslogx(LOG_ERR, "Status bytes are not enough: %" PRIiSIZE,
+		upslogx(LOG_ERR, "Status bytes are not enough: %" PRIuSIZE,
 		        status_bytes_size);
 		dstate_dataok();
 		return;
--- a/drivers/nutdrv_hashx.c
+++ b/drivers/nutdrv_hashx.c
@@ -125,7 +125,8 @@ static struct hashx_cmd_t {
 {
 	{ "beeper.enable", COMMAND_SET_BEEPER_ENABLED":1" },
 	{ "beeper.disable", COMMAND_SET_BEEPER_ENABLED":0" },
-	/* FIXME: We need a test.beeper command too. */
+	{ "experimental.test.beeper.start", COMMAND_START_BEEPER_TEST },
+	{ "experimental.test.beeper.stop", COMMAND_STOP_TEST },
 	{ "test.battery.start.deep", COMMAND_START_BATTERY_RUNTIME_TEST },
 	{ "test.battery.start.quick", COMMAND_START_BATTERY_TEST },
 	{ "test.battery.stop", COMMAND_STOP_TEST },
--- a/drivers/nutdrv_hashx.c
+++ b/drivers/nutdrv_hashx.c
@@ -1,0 +1,599 @@
+/* nutdrv_hashx.c - Driver for serial UPS units with #* protocols
+ *
+ * Copyright (C)
+ *   2025 Marco Trevisan (Treviño) <mail@3v1n0.net>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ *
+ */
+
+#include "common.h"
+#include "config.h"
+#include "dstate.h"
+#include "main.h"
+#include "nut_stdint.h"
+#include "powerpanel.h"
+#include "serial.h"
+#include "str.h"
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+#define ENDCHAR		'\r'
+#define IGNCHARS	""
+
+#define DRIVER_NAME	"Generic #* Serial driver"
+#define DRIVER_VERSION	"0.01"
+
+#define SESSION_ID	"OoNUTisAMAZINGoO"
+#define SESSION_HASH	"74279F35A48F5F13"
+
+/* These commands have been tested on an Atlantis Land A03-S1200 */
+#define COMMAND_GET_SESSION_HASH		"X19" /*-> #16-hex-values */
+#define COMMAND_SET_SESSION_ID			"K19" /* :value -> Status value */
+#define COMMAND_SET_BEEPER_ENABLED		"K60" /* :0|1 -> Status value */
+#define COMMAND_SET_FREQUENCY_MODE		"K39" /* :0|1 -> Status value */
+#define COMMAND_GET_BYPASS_CONDITION_MANDATORY	"X87" /* -> #\xc7\x97\xf0\xf0 */
+#define COMMAND_SET_BYPASS_CONDITION_MANDATORY	"K87" /* :0|1 -> Status value */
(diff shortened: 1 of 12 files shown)


jimklimov:
Thanks, looks like a very solid first entry with all the bells and whistles. CI however remains upset on some platforms, with various reports from different versions of different compilers, but one that pops up in several - and confuses me really - is:
```
nutdrv_hashx.c:624:2: error: code will never be executed [-Werror,-Wunreachable-code]
        upslogx(LOG_NOTICE, "instcmd: unknown command [%s] [%s]", cmd_name, extra);
```

Looking at https://github.com/networkupstools/nut/pull/2940/files#diff-2c7611b130b6008d3222e6c8227300aefd78788c142d6abf25fb7208ff7e25b3R605-R626 I think the general logic allows for code to never reach a `return` statement inside the loop, if indeed the `cmd_name` argument matches no entry in `hashx_cmd[]` (so it always `continue`'s until the end of loop).

But too many compilers a decade apart (including recent clang-20) think this would not happen :\

Please take another look at the loop criteria, maybe they see something we missed upon a quick glance?

Also, some toolkits complain that:
```
nutdrv_hashx.c:514:11: error: format specifies type 'ssize_t' (aka 'long') but the argument has type 'size_t' (aka 'unsigned long') [-Werror,-Wformat]
  513 |                 upslogx(LOG_ERR, "Status bytes are not enough: %" PRIiSIZE,
      |                                                                ~~~~~~~~~~~
  514 |                         status_bytes_size);
      |                         ^~~~~~~~~~~~~~~~~
```
...Probably should use `PRIuSIZE` here (for unsigned `size_t`).

jimklimov:
Gotta review closer what the GCC scenarios are upset about, possibly something with `(dist)check` (logs they point to are inconclusive, a CI's own backlogged bug).

Curiously the scenarios with mingw-on-linux and FreeBSD builds somehow fail with `ChangeLog.html` generation; possibly your essay in a commit comment was harder than what the simple parser script and/or Makefile shell snippet used as a sanity checker know how to handle (I think diacritics in non-ASCII names should not be a problem anymore, but that is also a possibility) :-\

jimklimov:
FWIW, the issues I've seen and commented on above can be reproduced locally with `./autogen.sh && ./configure --with-all=auto --enable-Werror --with-docs CC=clang CXX=clang++ && make -ks -j` to try developing fixes :)

I'll give it a shot for the simpler ones...

jimklimov:
Aha, for ChangeLog checker - it must be about the parentheses in user name (`grep` thinks they are special regex tokens, probably).

jimklimov:
Ultimately, for the loops around `hash_cmd[]` it for whatever reason disliked the maths with `sizeof`:
```
for (i = 0; i < sizeof (hashx_cmd) / sizeof (*hashx_cmd); ++i) {
```

A dumber and simpler approach used all over NUT, with explicit sentinel values to mark the end of array for iteration, was a simple fix that just worked, it seems. Neither gcc nor clang complained for me locally, will see what the larger NUT CI farm thinks about this...

3v1n0:
> Ultimately, for the loops around `hash_cmd[]` it for whatever reason disliked the maths with `sizeof`:
> 
> ```
> for (i = 0; i < sizeof (hashx_cmd) / sizeof (*hashx_cmd); ++i) {
> ```
> 
> A dumber and simpler approach used all over NUT, with explicit sentinel values to mark the end of array for iteration, was a simple fix that just worked, it seems. Neither gcc nor clang complained for me locally, will see what the larger NUT CI farm thinks about this...

Oh actually it was right to complain.. I missed the for check in the 2nd loop :joy:

3v1n0:
@jimklimov thanks for the fixups, what about this aspect:

> I would like to make the nut-scanner to look for these devices, but it would imply duplicating lots of scan_eaton_serial.c code or move it to a common place, so maybe that's another PR? As we don't have a generic place for serial scanning AFAIK.

jimklimov:
I think `nut-scanner` serial scanning fix is a matter for another PR, possibly an issue for that is already backlogged... for years. At least in the backs of many minds :)

Not really sure what a good approach might be, other than starting all drivers to see if they accept the device as known - with a matrix of baud ratea and other serial nuances, and hoping the UPS controller would not freeze or brick due to spam it does not understand :\

jimklimov:
Seems to be about `ChangeLog.html-contentchecked` and native SED implementations on *BSD and illumos (so my fix to escape parentheses does not work there). Will take a look, probably not before tomorrow.

drivers/nutdrv_hashx.c
```diff
@@ -0,0 +1,662 @@
...
+	}
+
+	if (status_bytes[0] & 0x20) {
+		upsdebugx(6, "Poll, status: Battery is critically low");
+		if ((status_bytes[0] & 0x08)) {
+			alarm_set("TEST MODE: Battery is critically low!");
+		} else {
+			alarm_set("Battery is critically low, UPS is about to turn off!");
+		}
+	}
```
3v1n0, line 588: @jimklimov Regarding this logic, the UPS seems to provide two kinds of state: normal, low battery, critically low battery... Now, I'm unsure if the mapping is done properly here for NUT, given that
 - High battery I assume is fine when the battery has no warnings, right?
 - Low battery bit is set when also the signaled percent is normally 0, but from that state in my tests the UPS can run quite a lot more
 - Then the critically low bit is set and after some more minutes IIRC PowerMaster+ was turning the UPS off.

Now... I'm unsure if the mapping is fine here, given that AFAIK there's no way to notify the critically low level here, so I decided to be conservative and set the `LB` state when we receive the low-battery bit.

Maybe I can provide a driver parameter to define whether the LB should be set only when in critical state?

It's a bit unclear to me what are the nut levels: HB -> <nothing> -> LB or HB -> LB?

jimklimov: This sort of meanings clean-up was actually discussed in recent months at #2708 and related tickets (and mailing list, IIRC). Possibly also #2215.

I think nobody came up with what "HB" reliably stands for, as very few drivers actually used it and (IIRC/OTOH) with different implied situations for "well-charged" or "dangerously over-charged"... this might have been raised in the "hover-charging" concept discussion after all, like when laptops charge to only 90% or so to prolong battery hardware life - so getting to 100% is "unhealthy" and something to report about.

There are few devices that have by themselves any thresholds to differentiate "just low" vs. "critically low" battery charges, but for example `drivers/netxml-ups.c` (for Eaton enterprise UPSes that can post alerts to their network clients) explicitly goes to `status_set("FSD");  /* shutdown imminent */` when the device-specific criteria are met (e.g. status bit raised).

Other than that, there are criteria in `upsmon` about treating a situation as critical or not, which is a large decision logic block riddled with tunables (which users can for behaviors of their devices while the client itself should be device-agnostic and decide only based on the readings it sees and config tunables it might have). Generally speaking, this is about a device being "OB + LB" and not in a known "CAL"(ibration) state. But also about being in a non-nominal state (trimming, boosting, simply on-battery, etc.) and losing communications for a bit too long.

Probably the best way at the moment is for your driver to report "LB" either when the device says so the first time, or with one or another situation based on a user tunable (effectively shutting down earlier or later in an outage), so `upsmon` would see it as "OB + LB" and trigger "FSD" internally. Also do `status_set("FSD");` anyway when you know it is critically low (and not charging).
* I guess the "critical LB" may mean the state that is not healthy for battery chemistry, possibly a fire/poison hazard, so not something you want to reach in practice and do want to quickly get out of.

Typically, NUT users who want to manage the time after an outage before shutdowns are triggered either use UPSes that have a manageable setting for "LB" report, or set up `upsmon`+`upssched` to script the timer-based activity. 

Maybe due to this, no support evolved for many layers of discharge. After all, either you do begin a shutdown, or you don't, so one trigger-reading is enough...

PR #2945, opened 2025-05-08
Add a way to `man nut` (new man section support, new page text)

jimklimov:
An overview page that could be installed for end users to ask for `man nut` was sorely missing.


docs/man/nut.txt
```diff
@@ -0,0 +1,196 @@
...
+
+For each monitored power distribution device, a NUT driver instance should
+be running on a machine that has a communication link to that device.
+
+NOTE: There is currently no support for "multiplexing" information or commands
+for devices with multiple-media/multiple-driver access, e.g. to gain redundant
+connection or a better data collection than any one driver provides. For some
+devices it may be possible to run several drivers independently (probably for
+your monitoring system to cherry-pick data points from one driver or another),
+while others might only enable any one link exclusively.
```
sebastiankuttnig, line 91: Sorry to hijack, but I read up a bit while working on my PR. Coincidentally I’ve got a basic failover setup running with a custom NUT driver that I wrote a while back. It receives a `upsc`-style data dump from a local HTTPS endpoint, using `neon`, and presents that as a read-only monitoring UPS.

The useful part for me is that the external program behind the API can swap out or manipulate the UPS data on the fly during a failover, and that driver doesn’t need to know. I also use it to inject some additional "UPS" variables for my environment sensors, such as temperature and humidity. If this seems like something the community might find useful, I could clean up the driver a bit and showcase it in a separate PR.

I guess it could also be handy for development work or for people who want to build failover setups, bridge drivers into NUT or otherwise manipulate NUT driver output without C, as long as their "API" returns a classic upsc-style output.

PR #2957, opened 2025-05-15
Instcmd setvar brushup

jimklimov:
Follows up from #2955 to consistently log instant command and variable setting, so drivers differ less for end-user experience.

Previously all drivers tended to do this differently, which made troubleshooting harder. Now they would all follow the same logic and message structure, at least in broad strokes, and specifically would try to `upslog*()` => syslog any operations that can impact power delivery to the useful load. So if a system crashes with a self-inflicted power outage, there's a chance it would have already recorded why.

While here, also identified some cases where `upslog*()` methods' first argument was a number, not a syslog macro name. Probably some time ago someone could not choose between `upslogx()` and `upsdebugx()`?..

NOTE: Code of these methods is peppered with TODO/FIXME comments regarding cleaner use of return codes (INVALID/UNKNOWN/FAILED/...) where apparently some reply types in headers and dstate (and data server side) appeared later than these methods. This was largely ignored and may be a focus of another PR eventually; some more such comments were added where the current responses seemed imperfect, however this was on a best-effort basis when something caught my eye, so probably some similar cases were missed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -157,11 +157,7 @@ EXTRA_DIST = LICENSE-GPL2 LICENSE-GPL3 LICENSE-DCO MAINTAINERS
 # Since the renaming of documentation to `*.adoc` extension to help IDE
 # and GitHub UIs to render the source files in a pretty fashion, we need
 # to list them:
-EXTRA_DIST += INSTALL.nut.adoc UPGRADING.adoc TODO.adoc NEWS.adoc README.adoc
-
-# As long as we do not tarball the script (or should we?), do not deliver
-# its doc either:
-# EXTRA_DIST += ci_build.adoc
+EXTRA_DIST += INSTALL.nut.adoc UPGRADING.adoc TODO.adoc NEWS.adoc README.adoc ci_build.adoc
 
 # Tarballs created by `make dist` include the `configure.ac` and `m4/*` sources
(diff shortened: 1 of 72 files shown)


desertwitch:
Maybe also worth adding to the developer documents in docs/ to have a common ground on this written down somewhere.

jimklimov:
I gave this a bit of thought, and maybe better hand off the docs part (perhaps to you, if volunteering) so it would also serve as a review with separate eyeballs  :)

IIRC the `sddefault` approach added for 2.8.3 could be also un(der)-documented...

desertwitch:
Sure, I hope I'll have time at the weekend to look through it and come up with something.

jimklimov:
Superb, thanks!

desertwitch:
I'm afraid I'm a bit lost here still, so from my understanding this standardises the debug and error levels to use common severities within instcmd and setvar? So developers should use these instead of randomly picking a debug severity of say 6 when it should better be 3? The `_STARTING` functions in turn emit a standardised log that handling was started within the driver, so it is visible straight away that the command made it to the driver handling part?

jimklimov:
I guess one could say that. Being macros, these values can be re-defined during a build (perhaps passed in CFLAGS), but even then they would be consistent so when troubleshooting different drivers, we would see these entries at same verbosity levels everywhere.

PR #2959, opened 2025-05-15
Help users address bad state files

jimklimov:
As recently noted in the mailing list discussion, it is possible to shoot oneself in the foot by pre-creating the state files (e.g. local unix socket) just because driver logs mention that one is absent. This PR should help the users fix their foot back :)

````
:; mkdir /tmp/dummy-ups-test
:; NUT_STATEPATH=/tmp ./drivers/dummy-ups -s test -x port= -DDDDDD
...
   1.001350     [D5] send_to_all: SETINFO driver.state "init.quiet"

Fatal error: unable to create listener socket

bind /tmp/dummy-ups-test failed: Address already in use

Things to try:

 - ps -ef | grep 'dummy-ups'    (Linux, GNU userland)
 - ps -xawwu | grep 'dummy-ups' (BSD, Solaris, embedded)
   To check if another copy of the driver is running; if not:

 - ls -la /tmp/dummy-ups-test
   To check if a (non-socket) filesystem object already exists there

 - rm -rf /tmp/dummy-ups-test
   To remove any offending files (a new driver instance creates its own)

   1.001437     Exiting.
...
````

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -74,6 +74,8 @@ https://github.com/networkupstools/nut/milestone/9
      potential for buffer overflow. [#2915]
 
  - common driver code:
+   * Update reports of failed socket file creation, to help troubleshooting
+     some error cases in the field.
    * Removed workarounds trying to migrate legacy driver raised `ALARM`
      status tokens into modern `alarm_*` function logic. Rather, we keep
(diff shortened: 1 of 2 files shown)


jimklimov:
NUT CI farm errors: seems to be a network glitch during build, retrying.

PR #2963, opened 2025-05-20
common.c: VA copies with buffer upsizing, upsdrvquery.c: str handling fixes

### LINKED ISSUE #2948: SIGSEGV in drivers/upsdrvquery.c: upsdrvquery_read_timeout()
Hello!

I hit a segmentation fault in `drivers/upsdrvquery.c` during driver development today, when in debug level 6.
Note that the displayed `errno` is related to a previous unrelated failure - simulating a dead driver socket for failover.

```
   0.004037     [D5] upsdrvquery_read_timeout: tv={sec=3, usec=000000}
WARNING: vupslog: vsnprintf needed more than 1024 bytes: 1108 (2 => No such file or directory), extending to 2132

Program received signal SIGSEGV, Segmentation fault.
0x0000155555472b59 in __strlen_avx2 () from /lib64/libc.so.6
(gdb) bt
#0  0x0000155555472b59 in __strlen_avx2 () from /lib64/libc.so.6
#1  0x000015555535da40 in __printf_buffer () from /lib64/libc.so.6
#2  0x000015555538342d in vsnprintf () from /lib64/libc.so.6
#3  0x00000000004114d1 in vupslog ()
#4  0x0000000000411985 in s_upsdebugx ()
#5  0x000000000040dddb in upsdrvquery_read_timeout ()
#6  0x0000000000405b73 in upsdrv_updateinfo ()
#7  0x0000000000404847 in main ()
(gdb) 
```

https://github.com/networkupstools/nut/blob/093e44b1c8db556f356263951f906db3551d5f3e/drivers/upsdrvquery.c#L366-L371

I don't have time to look into this more at the moment, thought I'd log it not to forget.

Possibly related to the SIGSEGV in the other PR #2947.

sebastiankuttnig:
I took a preliminary shot at this, feel free to change or move around as required.

If upsizing log buffers we will most probably need copying of variadic arguments, as outlined here:
https://github.com/networkupstools/nut/issues/2948#issuecomment-2869130856

Some systems have `va_copy`, others `__va_copy`, others nothing - this should account for all eventualities.
In case absolutely no variadic argument copying is supported, truncated logs are shown and buffers not upsized.

The PR also addresses the needed fixes in `drivers/upsdrvquery.c` to account for non-null-terminated output.

fixes #2948


sebastiankuttnig:
> ❌ [Build nut 2.8.3.3180-master failed](https://ci.appveyor.com/project/nut-travis/nut/builds/52086098) (commit [bd2d526a1f](https://github.com/networkupstools/nut/commit/bd2d526a1f) by @sebastiankuttnig)

> Build execution time has reached the maximum allowed time for your plan (60 minutes).

Seems unrelated, but would definitely appreciate a sanity-check on this and particularly if other parts here and elsewhere may make re-use of variadic arguments in a way that would mangle them as we've seen before (and also need fixing with `va_copy`).

drivers/upsdrvquery.c
```diff
@@ -364,8 +364,10 @@ ssize_t upsdrvquery_read_timeout(udq_pipe_conn_t *conn, struct timeval tv) {
 #endif  /* WIN32 */
 
 	upsdebugx(ret > 0 ? 5 : 6,
-		"%s: received %" PRIiMAX " bytes from driver socket: %s",
-		__func__, (intmax_t)ret, (ret > 0 ? conn->buf : "<null>"));
+		"%s: received %" PRIiMAX " bytes from driver socket: %.*s",
+		__func__, (intmax_t)ret, (int)(ret > INT_MAX ? INT_MAX : ret),
```
jimklimov, line 368: According to code just above, we can end up with `ret = -1` for example. I think these Elvis-style arguments should take that into account (for `.*` and for `s` formats), to behave sanely if `ret <= 0`?

drivers/upsdrvquery.c
```diff
@@ -486,7 +488,7 @@ ssize_t upsdrvquery_prepare(udq_pipe_conn_t *conn, struct timeval tv) {
 		goto socket_error;
 	if (upsdrvquery_read_timeout(conn, tv) < 1)
 		goto socket_error;
-	if (strcmp(conn->buf, "ON")) {
+	if (strcmp(conn->buf, "ON")) { -- WARNING: conn->buf is NOT null terminated --
```
jimklimov, line 491: Was this supposed to be a comment? ;)
I see it is inside a comment block, but probably a `//` is safer in this context, even if we discourage it in buildable code. Just in case it gets enabled, many compilers (or their chosen standards) would not be utterly confused.

include/common.h
```diff
@@ -546,6 +546,13 @@ int vsnprintfcat(char *dst, size_t size, const char *fmt, va_list ap);
 int snprintfcat(char *dst, size_t size, const char *fmt, ...)
 	__attribute__ ((__format__ (__printf__, 3, 4)));
 
+/* Define a missing va_copy using __va_copy, if available: */
+#ifndef va_copy
```
jimklimov, line 550: Do we have a guarantee that `va_copy` is a macro wherever defined (e.g. not a method)? Maybe better check for `#ifndef HAVE_VA_COPY` more portably? We define one in `configure.ac` anyway :)

jimklimov (changes requested):
Mostly looks reasonable, but in a few points either I misunderstood, or the code should be improved. Thanks!

configure.ac
```diff
@@ -1433,6 +1433,64 @@ AS_IF([test x"${ac_cv_func_strdup}" = xyes],
     [AC_MSG_WARN([Required C library routine strdup not found; try adding -D_POSIX_C_SOURCE=200112L])]
     )
 
+AC_CACHE_CHECK([for va_copy(dest,src)],
+  [ac_cv_have_va_copy],
+  [
+    save_CFLAGS="$CFLAGS"
+    CFLAGS="$CFLAGS -Werror=implicit-function-declaration"
```
jimklimov, line 1440: I think these options are compiler-specific and should be limited to `AS_IF([test "x${CLANGCC}" = "xyes" || test "x${GCC}" = "xyes"], ...` as some other parts of the script already do. Also below in the next check.

common/common.c
```diff
@@ -3206,9 +3206,22 @@ static void vupslog(int priority, const char *fmt, va_list va, int use_strerror)
...
+		ret = vsnprintf(buf, bufsize, fmt, va_snprintf);
+		va_end(va_snprintf);
+#else
+		/* Without va_copy(), we cannot safely retry vsnprintf()
+		 * and will accept truncation if the buffer is too small */
 		ret = vsnprintf(buf, bufsize, fmt, va);
+#endif
 
 		if ((ret < 0) || ((uintmax_t)ret >= (uintmax_t)bufsize)) {
+#ifdef HAVE_VA_COPY_VARIANT
```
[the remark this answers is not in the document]
sebastiankuttnig: Because if the variadic arguments do not get copied, but rather instantly consumed from the original set on the first try (due to lack of `va_copy()`), the do-while would keep running and retrying upsized buffers on the original set of (already consumed) arguments which would result in the mangling we've seen resulting in the issue #2948. Not building this block will make it fall through to the truncation case instead and not retry with the consumed, non-copied set of VAs.

include/common.h
sebastiankuttnig: I've researched this in depth before and everywhere I looked this was only ever a macro.

The rationale behind `va_copy` and not HAVE_* for the definition itself was:

- If the environment already provides a va_copy macro, we should trust it (to know best) and not override it.
- If it doesn't, we can define it with a va_copy variant that fits our expecations, we wouldn't override.
- If we rely on HAVE_VA_COPY alone, we risk defining va_copy even when the environment already provides a valid one, possibly leading to conflicts or subtle bugs (if our test expectation of va_copy diverges from system's implementation).
- Conversely, the HAVE_VA_COPY, HAVE___VA_COPY, HAVE_VA_COPY_VARIANT logic can be used by the calling code to ensure that a test compliant va_copy must exist and behave according to tested expectations, not just any va_copy (hence not using `#ifdef va_copy` in the calling code places but HAVE_*)

Of course this was just my personal thought process, so happy to revisit this (or have it patched over 😉 )

drivers/upsdrvquery.c
sebastiankuttnig: Good point, this should indeed be guarded against `-1`.

configure.ac
sebastiankuttnig: That's a good point, how do you usually handle such test cases requiring to fail when they need to, while maximising portability but also not missing out testing for needed bits on some platforms?

configure.ac
```diff
@@ -1433,6 +1433,64 @@ AS_IF([test x"${ac_cv_func_strdup}" = xyes],
...
+
+    AC_COMPILE_IFELSE(
```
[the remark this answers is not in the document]
sebastiankuttnig: Thanks for the pointer. I agree that AX_RUN_OR_LINK_IFELSE might be a stronger check, especially when it might compile but not actually link or behave properly. For va_copy, since it seemed to always be a macro and typically either work or be missing outright, I leaned toward the lighter AC_COMPILE_IFELSE here - happy to revisit/have it patched over as well here.

sebastiankuttnig:
Thanks for the detailed feedback, happy to collaborate since this is a crucial piece of the codebase and I'm not so well versed with the broader ecosystem yet, wanted to kick off a PR to get things rolling - hope my thought process so far made at least some sense!

common/common.c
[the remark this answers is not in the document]
sebastiankuttnig: Fwiw, the comment two lines above was supposed to address this, but can move it down right above the excluded block for better visibility?

drivers/upsdrvquery.c
sebastiankuttnig: Fixed up to a `FIXME:` (ha) which should also be more grep-able.

configure.ac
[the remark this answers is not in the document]
sebastiankuttnig: Fixed up to `AX_RUN_OR_LINK_IFELSE`, should also be more portable and not depend on `CFLAGS`.

common/common.c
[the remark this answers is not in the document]
sebastiankuttnig: Fixed up, added a more detailed comment to the block in question - logic remains for now.

include/common.h
sebastiankuttnig: I've changed this to your suggestion for now so we can see how this behaves in CI, and you're definitely more experted at portability concerns than me from looking at this project (and the CI... hats off...)

jimklimov (approved):
Super, thanks!

PR #2969, opened 2025-05-23
upsdrvquery.{c,h}: add new one-shot functions for more versatile API use

sebastiankuttnig:
### Overview 
```
 - `upsdrvquery` API updates:
   * Added `upsdrvquery_oneshot_conn()` for issuing one-shot queries using an
     existing `udq_pipe_conn_t *` connection. The caller manages the
     connection's lifecycle, and the function includes a best-effort call to
     restore broadcast mode after the query to return the connection as it was.

   * Added `upsdrvquery_oneshot_sockfn()` for initiating one-shot queries using
     a socket filename. Shares internal logic with the existing
     `upsdrvquery_oneshot()`, which uses a UPS and driver name, respectively.

   * Introduced `upsdrvquery_restore_broadcast()` to explicitly restore
     broadcast mode (`BROADCAST 1`) on a connection, helping return it to a
     consistent and talkative state.

   * Revised connection ownership handling: internal functions like
     `upsdrvquery_prepare()` and `upsdrvquery_request()` no longer close
     connections they do not own. Responsibility for cleanup is now delegated to
     the caller to avoid unintended side effects and better align with expected
     usage patterns.
```

as initially outlined in and hopefully useful to #2962


drivers/upsdrvquery.c
```diff
@@ -635,10 +635,31 @@ ssize_t upsdrvquery_oneshot(
...
+	const char *query,
+	char *buf, const size_t bufsz,
+	struct timeval *ptv
+) {
+	udq_pipe_conn_t	*conn = upsdrvquery_connect(sockfn);
+
+	return upsdrvquery_oneshot_conn(conn, query, buf, bufsz, ptv);
+}
+
+ssize_t upsdrvquery_oneshot_conn(
```
sebastiankuttnig, line 655: Hm, on second thought... perhaps we should either make it absolutely clear here that the connection is being freed by this function (through a comment), or not free here (but inside the other two functions which are more beginning-to-end-type implementations), or not expose this helper function at all.

If I were to call this function, I would probably not expect it to free a connection that I create and own (in my calling code), so maybe it'd make more sense to free only in functions which also establish the connection (`upsdrvquery_oneshot` & `upsdrvquery_oneshot_sockfn`) but not inside `upsdrvquery_oneshot_conn` that takes an (external) connection it doesn't "own".

@jimklimov : wdyt - any perspectives from the maintainability side of things?

jimklimov: Yeah, that would be... surprising. I agree that whoever allocates the connection, owns it and should free it - so not the code shared by everyone of those then.

jimklimov:
How does that look in your testing? Any planned changes or can it be merged when CI is content enough? :)

sebastiankuttnig:
> How does that look in your testing? Any planned changes or can it be merged when CI is content enough? :)

All good on my side, can be merged as far as I'm concerned, if it passes CI. 👍

PR #2977, opened 2025-06-03
Fix man page section assignments for end-user interactive clients

jimklimov:
There was some mess in existing pages; bolt down the man page section for these programs: `upsc`, `upscmd`, `upsrw` and `NUT-Monitor`. Being (semi-)interactive end-user clients, they belong in section 1 rather than 8 (system daemons).
* This is a sort-of breaking change for existing packaging recipes; less so for saved URLs on rendered nut-website (they do not include numbers, just the `.html` extension).

UPDATE: Due to the questionable nature of the man page section change, this PR was split: #2978 takes care of general docs improvements (including the fix of man sections from `1` to `8` where they got mis-referenced in existing docs vs. existing layout), and this PR was rebased on top of that to propose the re-organization of the layout itself. Also, should this go further, moving tools like `nut-scanner` and `nutconf` to section 1? If we do end up breaking everyone's packaging recipes, better limit the pain to one NUT release.

Raised a discussion on mailing list: https://alioth-lists.debian.net/pipermail/nut-upsdev/2025-June/008151.html

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -201,6 +201,10 @@ https://github.com/networkupstools/nut/milestone/9
    * Option to `configure --enable-docs-man-for-progs-built-only` was added,
      to differentiate NUT builds that deliver man pages for only built programs
      (legacy default) or for all of them (as needed for docs sites). [#2976]
+   * Several man page sources revised to tack the correct section code (e.g.
+     `upsc`, `upscmd`, `upsrw` and `NUT-Monitor` are clients that end-users
+     can interact with (section 1), while `upsmon` and `upssched` are rather
+     system services (section 8). [#2977]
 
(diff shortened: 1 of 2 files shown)


jimklimov:
Due to the questionable nature of the man page section change, this PR was split: #2978 takes care of general docs improvements (including the fix of man sections from `1` to `8` where they got mis-referenced in existing docs vs. existing layout), and this PR was rebased on top of that to propose the re-organization of the layout itself.

Raised a discussion on mailing list: https://alioth-lists.debian.net/pipermail/nut-upsdev/2025-June/008151.html

jimklimov:
Rebased over recent state of PR #2978 as it got merged.

jimklimov:
For now suspending this idea, maybe even until after a NUT v2.8.4 release. May lead to path re-assignments for some programs as well.

PR #2982, opened 2025-06-10
drivers/apc_modbus.c: Fix interframe delay

marcan:
The delay is supposed to be from *end* of a frame to *start* of the next. Also fix the type used for timestamp calculations (useconds_t is not big enough and overflows).

Fixes one aspect of issue #2609

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -157,6 +157,10 @@ https://github.com/networkupstools/nut/milestone/9
    * Added APC BVKxxxM2 to list of devices where `lbrb_log_delay_sec=N` may be
      necessary to address spurious LOWBATT and REPLACEBATT events. [#2942]
 
+ - `apc_modbus` driver updates:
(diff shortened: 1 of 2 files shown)


jimklimov:
Looks meaningful, thanks! Is there a particular reason to use signed `int64_t` and not e.g. `uint64_t` here (or would there be any benefit from changing this, e.g. the 2x time range possible with loss of signedness)?

marcan:
Timestamps can be negative if the system time is before 1970 :) (unlikely, but legal)

jimklimov:
So now we get roughly 2^43 full seconds before/after epoch (rounding 1000000 to 10^20), right? Should sufffice :)

jimklimov:
@marcan : just in case - have you tested this change over the past days? Does the driver behave better (or not-worse) than before it?

marcan:
I have, this change completely eliminated all register read errors for me over the past 24+ hours.

PR #2986, opened 2025-06-13
phoenixcontact_modbus: Add support for shutdown variables, mode selector, PC reset timing…

haulfi:
…and battery switch threshold

- Introduced new variables to the phoenixcontact_modbus driver especially for QUINTS 4 USB:
  - shutdown variables
  - mode selector 9 = PC_MODE
  - PC reset delay after main power recovers
  - automatic switch to battery mode if main power is below or above a defined threshold.
  
  Signed-off-by: Uli <ulfathasangaaryev@gmail.com>

## General points

  comments or text documents.

- [ ] Please star NUT on GitHub, this helps with sponsorships! ;)

  case.

  or its equivalent.

  concept.

  the changed codebase.


drivers/phoenixcontact_modbus.c
```diff
@@ -21,727 +21,864 @@
...
+ #include "nut_stdint.h"
+ #include <stdbool.h>
+ 
+ 
+ #if !(defined NUT_MODBUS_LINKTYPE_STR)
+ # define NUT_MODBUS_LINKTYPE_STR	"unknown"
+ #endif
+ 
+ #define DRIVER_NAME	"NUT PhoenixContact Modbus driver (libmodbus link type: " NUT_MODBUS_LINKTYPE_STR ")"
+ #define DRIVER_VERSION	"0.1"
```
jimklimov: `0.08` => `0.09` (and two digits, so `0.1` is wrong, `0.10` is better).

drivers/phoenixcontact_modbus.c
```diff
@@ -21,727 +21,864 @@
...
+ #define TRIO_2G_10A_UPS_PARTNUMBER 2907161
+ #define TRIO_2G_10A_UPS_DESCRIPTION "TRIO-UPS-2G/1AC/24DC/10"
+ 
+ #define TRIO_2G_20A_UPS_PARTNUMBER 1105556
+ #define TRIO_2G_20A_UPS_DESCRIPTION "TRIO-UPS-2G/1AC/24DC/20"
+ 
+ 
+ /* --- Shutdown & Mode Configuration --- */
+ 
+ // Macro for fetching a uint16_t config value with fallback
```
jimklimov: Here and below, please use block comments `/* ... */` even for single line texts

drivers/phoenixcontact_modbus.c
```diff
@@ -21,727 +21,864 @@
...
+ {
+	 addvar(VAR_VALUE, "delay.shutdown", "Delay before initiating PC shutdown (in seconds)");
+	 addvar(VAR_VALUE, "delay.pc_shutdown", "Time allowed for PC to shutdown (in seconds)");
+	 addvar(VAR_VALUE, "delay.pc_reset", "Duration of output off before reboot (in seconds)");
+	 addvar(VAR_VALUE, "delay.mains_return", "Delay before switching back to mains (in seconds)");
+	 addvar(VAR_VALUE, "mode.selector", "UPS operating mode (e.g. 9 = PC-Mode)");
+	 addvar(VAR_VALUE, "bat.warning_soh", "Battery warning SOH threshold (%)");
+	 addvar(VAR_VALUE, "bat.voltage_low", "Threshold [V] to switch to battery mode");
+	 addvar(VAR_VALUE, "mains.voltage_high", "Threshold [V] to return to mains mode");
+	 addvar(VAR_VALUE, "mains.return_delay", "Time [s] above threshold before switching back to mains");
```
jimklimov: Can you please add these explanations also to the man page?

drivers/phoenixcontact_modbus.c
```diff
@@ -21,727 +21,864 @@
...
+	 The difference is caused from the way they are reported in the datasheet,
+	 keeping them in the same format as the datasheet make more simple the maintenence
+	 of the driver avoiding conversions while coding.
+ */
+ 
+ /* driver description structure */
+ upsdrv_info_t upsdrv_info = {
+	 DRIVER_NAME,
+	 DRIVER_VERSION,
+	 "Ulfat Hasangarayev <ulfathasangarayev@gmail.com>\n",
```
jimklimov: Do not remove original authors, several can be listed in metadata (see other drivers about examples).

drivers/phoenixcontact_modbus.c
```diff
@@ -21,727 +21,864 @@
...
+ 
+ static const delay_param_t delay_params[] = {
+	 { "ups.delay.shutdown",     REG_PC_SHUTDOWN_DELAY},
+	 { "ups.delay.pc_shutdown",  REG_PC_SHUTDOWN_TIME},
+	 { "ups.delay.pc_reset",     REG_PC_RESET_TIME},
+	 { "ups.delay.mains_return", REG_MAINS_RETURN_DELAY},
+	 { "ups.mode.selector",      REG_MODE_SELECTOR_SWITCH},
+	 { "ups.bat.warning_soh",    REG_WARNING_SOH_THRESHOLD},
+	 { "ups.bat.voltage_low",    REG_VOLTAGE_BELOW_BATTERY},
+	 { "ups.mains.voltage_high", REG_VOLTAGE_ABOVE_MAINS }
```
jimklimov: I believe these strings are also driver configuration parameters (per `addvar()` below)?

If any of these concepts are already described by standard NUT variable/command names listed in docs/nut-names.txt, it would be beneficial to add corresponding code so `upsrw`/`upscmd` can be used to set them at run-time, same way as for any other supported UPS with such features. But this may be improved later.

drivers/phoenixcontact_modbus.c
```diff
@@ -21,727 +21,864 @@
...
+ 
+ int write_uint32_reg_bit(modbus_t *ctx, int reg, int bit_index, bool bit_value)
+ {
+	 uint16_t regs[2];
+	 if (modbus_read_registers(ctx, reg, 2, regs) != 2) {
+		 upslogx(LOG_ERR, "Failed to read 32-bit register 0x%04X: %s", reg, modbus_strerror(errno));
+		 return -1;
+	 }
+ 
+	 uint32_t val = ((uint32_t)regs[0] << 16) | regs[1];
```
jimklimov: Declare all vars on top of scope please.

drivers/phoenixcontact_modbus.c
```diff
@@ -21,727 +21,864 @@
...
+	 errcount = 0;
+ 
+	 upsdebugx(2, "upsdrv_updateinfo");
+	 
+ 
+	 switch (UPSModel)
+	 {
+	 case QUINT4_UPS:
+		 
+		 for (size_t i = 0; i < sizeof(delay_params) / sizeof(delay_params[0]); i++) {
```
jimklimov: Declare all vars on top of scope please (and not inside `for` loops).

haulfi:
Thanks @jimklimov for your feedback.

I updated the code, and please dont hesitate to add more feedback :)

jimklimov:
@haulfi : Now that the driver code looks good for CI, I wonder if these added driver-configuration names (for concepts that overlap with existing standard NUT variable or instant command names in `docs/nut-names.txt`) can be reconciled? Just to have this done on main branch before merging the PR, to minimize the surprise for end-users about their meaning ("is X conceptually same as Y or...?")

And as noted in earlier review comments, I'm also wondering if for at least some of these it would make sense to back them as ordinary NUT variables (rather than internal C numeric variables), which can be set at run-time by `upsrw` and other clients (so the initial value via `ups.conf` configuration would be like `override.battery.voltage.low=11.5` etc.?) Note there are `input.transfer.(trim|boost).(low|high)` - one of these is likely equivalent to what you named `ups.mains.voltage_high`, for example.

jimklimov:
Note: AppVeyor CI build itself was OK, just overflowed their time allocation so could not finish some job book-keeping.

drivers/phoenixcontact_modbus.c
```diff
@@ -790,14 +794,13 @@
...
-	 addvar(VAR_VALUE, "delay.pc_reset", "Duration of output off before reboot (in seconds)");
-	 addvar(VAR_VALUE, "delay.mains_return", "Delay before switching back to mains (in seconds)");
+	 addvar(VAR_VALUE, "timer.shutdown", "Time allowed for PC to shutdown (in seconds)");
+	 addvar(VAR_VALUE, "timer.start", "Duration of output off before reboot (in seconds)");
+	 addvar(VAR_VALUE, "delay.start", "Delay before switching back to mains (in seconds)");
 	 addvar(VAR_VALUE, "mode.selector", "UPS operating mode (e.g. 9 = PC-Mode)");
-	 addvar(VAR_VALUE, "bat.warning_soh", "Battery warning SOH threshold (%)");
-	 addvar(VAR_VALUE, "bat.voltage_low", "Threshold [V] to switch to battery mode");
-	 addvar(VAR_VALUE, "mains.voltage_high", "Threshold [V] to return to mains mode");
-	 addvar(VAR_VALUE, "mains.return_delay", "Time [s] above threshold before switching back to mains");
```
jimklimov: @haulfi : Looking over changes in your later (fix) commits, which added descriptions in man page for the older config-variable names, and changed the names to fit NUT standard settings (so not fitting the proposed man page changes).

It seems the original PR had two names for "return delay" aspects: `mains.return_delay` apparently linked to `mains.voltage_high` (so I understood it as "feed the load from battery in BUCK mode or complete transformer mode if input voltage is low but not necessarily zero, and going back to feeding the load from mains when voltage returns and stays high for longer than X seconds"), and `delay.mains_return` which seemed less specific (and I understood as "if the UPS turned off, and power returned, wait X sec to make sure it really returned and is stable, and turn on the UPS"), which upon later review both mapped to `REG_MAINS_RETURN_DELAY` and in the fix you dropped one of those names.

Just wanted to clarify if the device really supports only one of those meanings, and then which one - "power off, wait, power on" or "feed from battery, wait, feed from mains"? They are quite different behaviors (about PC going down or not), so it is rather important to clarify :)

jimklimov: @haulfi : gentle bump, let me know if you would have time to answer the questions above in the coming days, please?

haulfi: @jimklimov So sorry for the late response. Had busy time with work. The return delay which link to REG_MAINS_RETURN_DELAY  is a timer for switching the PC  automatically on after the main power recovered. The main voltage must be within the threshold for a 10sec, to intiate automatic pc restart.

jimklimov:
@haulfi : thanks for the comments, I've retouched the PR to use standard NUT variables (both about consistent names and dstate as storage), so they should be configurable in `ups.conf` as e.g.  `default.ups.timer.shutdown = 30`.

Can you please check if this updated code base state actually works with your device, so it can be safely merged?

jimklimov:
One last note (reflected in latest docs commits): currently any change in the dstate-backed values above is (should be) still only processed during driver init via `default.*` settings - they are not handled by a `setvar()`.

I'll repost the points below as issues for future work:
* Probably someone with an actual device should check and code which of these are meaningful and safe to apply on the fly (with `upsrw` or similar NUT clients).
* It would also be helpful to actually read back these values from the device (whichever ones it serves... gotta be some?) and initialize these variables to them, not the device to hard-coded defaults or configured values every time the driver starts, right?

PR #3019, opened 2025-07-15
Extend gitlog2changelog to report start/end commits and allow to tweak them; allow to disable generation of changelog formats or completely

jimklimov:
A massive part of developer iteration time when docs are enabled is repetitive re-generation of the `ChangeLog` document (export from git to text, by default since `v2.6.0`) and especially generation of HTML and PDF variants from that.

This PR allows to disable changelog completely, or reduce its size (e.g. since most-recent release till HEAD) when we do want docs recipes challenged but not wasting a lot of time on that.

For (pre-)release builds, it should by default enable the same scope as before.


jimklimov:
NOTE: This spectacularly misfired when trying to release v2.8.4, because many CI agents do not have doc tools but the `configure.ac` change required to build changelog for (pre-)release tags if the option remained at auto.

Issue #3063, opened 2025-08-20
upsdrvctl.exe uses incorrect path separator on Windows


twoeater:
Description
I'm trying to run upsdrvctl.exe on a Windows system. However, the application seems to be using a forward slash (/) as a path separator instead of a backslash (\\), which is the standard for Windows.

This leads to a "No such file or directory" error because the full path C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.4.443-master\mingw64\sbin/usbhid-ups.exe is invalid on Windows.

Steps to Reproduce:

1.  Run the following command:
```shell
PS C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.4.443-master\mingw64\sbin> .\upsdrvctl.exe start
```

2. Observe the error message:
```shell
Network UPS Tools upsdrvctl.exe - UPS driver controller 2.8.4 release
Can't start C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.4.443-master\mingw64\sbin/usbhid-ups.exe: No such file or directory [지정된 파일을 찾을 수 없습니다. ]
```

Additional Context
I attempted to use the -r option to specify the path, but this did not resolve the issue. The application still seems to construct the path internally using the incorrect separator.

Example with -r option:
```shell
PS C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.4.443-master\mingw64\sbin> .\upsdrvctl.exe -r "C:\test\" start
```

The output remains the same, showing the invalid path with the forward slash:
```shell
Network UPS Tools upsdrvctl.exe - UPS driver controller 2.8.4 release
Can't start C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.4.443-master\mingw64\sbin/usbhid-ups.exe: No such file or directory [지정된 파일을 찾을 수 없습니다. ]
```

jimklimov:
Windows for a couple decades supports both slashes; not sure indeed about mixing them though. Probably a platform-specific "pathsep" macro or similar trick is needed. Can you experiment and PR a solution?

jimklimov:
Sorry just saw PR #3064.

Also, which Windows version did this happen on?

twoeater:
I wrote in PR.
It is be my mistake. It is not path separator issued.
Problem is usbhid-ups.exe are located in "bin", but ctl is default ref in "sbin".
but drivepath can modified  ups.conf.
Sorry.

twoeater:
```shell
[...]

PR #3065, opened 2025-08-21
nut.exe are call upsdrvctl.exe in bin.

twoeater:
but upsdrvctl.exe located in sbin.
So, change bin to sbin.

## General points

  comments or text documents.

- [ ] Please star NUT on GitHub, this helps with sponsorships! ;)

  case.

  or its equivalent.

  concept.

  the changed codebase.

### DIFF
--- a/scripts/Windows/wininit.c
+++ b/scripts/Windows/wininit.c
@@ -148,7 +148,7 @@ static DWORD run_drivers(void)
 	char command[NUT_PATH_MAX];
 	char *path;
 
-	path = getfullpath(PATH_BIN);
+	path = getfullpath(PATH_SBIN);
 	if (nut_debug_level < 1) {
 		snprintf(command, sizeof(command), "%s\\upsdrvctl.exe start", path);
 	} else {


twoeater:
[https://github.com/networkupstools/nut/issues/3063#issuecomment-3208741379](https://github.com/networkupstools/nut/issues/3063#issuecomment-3208741379)
Sorry to bother you.
I can't Set up a build environment. I don't have mingw experience.
Please approval test build. then i trying to test.

twoeater:
```shell
PS C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.4.443-master\mingw64\bin> .\nut.exe -N
Running in non-service mode

EventLog : Starting

EventLog : Starting drivers

EventLog : Starting upsd

Network UPS Tools upsdrvctl.exe - UPS driver controller 2.8.4.8.3-11+g53a17cf87 (development iteration after 2.8.4)
Network UPS Tools upsd 2.8.4.8.3-11+g53a17cf87 (development iteration after 2.8.4)
listening on 0.0.0.0 port 3493
Found 1 UPS defined in ups.conf
Network UPS Tools 2.8.4.8.3-11+g53a17cf87 (development iteration after 2.8.4) - Generic HID driver 0.67
USB communication driver (libusb 1.0) 0.50
Warning: report descriptor too short (expected 745, got 33)
Please check your Windows Device Manager: perhaps the UPS was recognized by default OS
driver such as HID UPS Battery (hidbatt.sys, hidusb.sys or similar). It could have been
"restored" by Windows Update. You can try https://zadig.akeo.ie/ to handle it with
either WinUSB, libusb0.sys or libusbK.sys.
HIDParse: LogMax is less than LogMin. Vendor HID report descriptor may be incorrect; interpreting LogMax -1 as 255 in ReportID: 0x00
HIDParse: LogMax is less than LogMin. Vendor HID report descriptor may be incorrect; interpreting LogMax -1 as 255 in ReportID: 0x00
Using subdriver: APC HID 0.100
Defaulting lbrb_log_delay_sec=3 for American Power Conversion model Back-UPS BX1600MI; consider also setting the lbrb_log_delay_without_calibrating and/or onlinedischarge_calibration and/or onlinedischarge_log_throttle_sec flag(s) in your configuration
Listening on named pipe \\.\pipe\usbhid-ups-apc-bx1600
EventLog : Starting upsmon

EventLog : upsd - listening on 0.0.0.0 port 3493

Network UPS Tools upsmon 2.8.4.8.3-11+g53a17cf87 (development iteration after 2.8.4)
UPS: apc-bx1600@192.168.0.8 (primary) (power value 1)
Using power down flag file C:\killpower
Warning: no custom notification command defined, just so you know
EventLog : upsmon - Startup successful

EventLog : upsmon - upsnotify: failed to notify about state NOTIFY_STATE_READY_WITH_PID: no notification tech defined, will not spam more about it

EventLog : upsd - User monuser@192.168.0.8 logged into UPS [apc-bx1600]
```

It's working.
Not need to driverpath in ups.conf.
just CLI clean run.

jimklimov:
Odd, i think `upsdrvctl(.exe)` should be in same dir as drivers - is it also in Windows builds? (drvpath which may be binpath per configuration of that build).
* UPDATE: no it is not, details in later posts below

NUT for Windows env setup (semi-native with MSYS2 or cross with Linux+mingw) is documented in docs/config-prereqs.txt and scripts/Windows/README.adoc - you may have to pre-build some prerequisites on this or that platform. Also see appveyor.yml which sets it up from scratch for CI builds.

jimklimov:
Also, I'm away from computers till September, so can only look at this in detail in a couple of weeks. Just in case, hoping someone else might chime in sooner...

drivers/upsdrvctl.c
```diff
@@ -1652,7 +1652,7 @@ int main(int argc, char **argv)
 #ifndef WIN32
 	driverpath = xstrdup(DRVPATH);	/* set default */
 #else	/* WIN32 */
-	driverpath = getfullpath(NULL); /* Relative path in WIN32 */
+	driverpath = getfullpath(PATH_BIN); /* Relative path in WIN32 */
```
jimklimov, line 1655: Nice, the method (in `common.c`) actually has a possible NULL deref problem if there is no backslash. And also looks only for that one, ignoring forward slashes (that recent Windows do accept... at least if not mixing two styles in same string - not sure yet what happens then).
* UPDATE: Checked locally in `cmd` with `dir "c:/temp\tools"` - that worked okay in Win11 (as long as quotes were used, otherwise forward slash was assumed to start an invalid CLI argument)

This problem *probably* should never happen since we parse the output of WIN32 API method `GetModuleFileName()` but still that bit does not seem clean now that I looked at it.

jimklimov, line 1655: I think the whole PR is flawed here by assuming the drivers would be in `PATH_BIN`. This is a coincidence due to `configure` script arguments involved (or defaulted) in a particular build; the pedantically correct one would be the `DRVPATH` (or a relative path to that, starting from location of `upsdrvctl.exe`).

I believe here the root-cause problem is rather derived from assuming in this code spot that the tool is in same location as drivers (NULL relative path, compared to program module location), despite this `drivers/Makefile.am` piece:
```
# always build upsdrvctl
sbin_PROGRAMS = upsdrvctl
```
(because drivers are often hidden in some `/lib/nut` or under `$libexec`, and the tool to generally manage them should be in a more visible location).

The puzzle to solve here is in fact the relative path to drivers, I think, because the absolute paths built into the code may be flawed (in the default build config for Win32 in `scripts/Windows/build-mingw-nut.sh` script, the `--prefix=/` and a shifted `make install DESTDIR=$INSTALL_DIR` is used).

scripts/Windows/wininit.c
```diff
@@ -148,7 +148,7 @@ static DWORD run_drivers(void)
 	char command[NUT_PATH_MAX];
 	char *path;
 
-	path = getfullpath(PATH_BIN);
+	path = getfullpath(PATH_SBIN);
```
jimklimov, line 151: Here the change to `PATH_SBIN` is correct (matching the install location of `upsdrvctl(.exe)` per `drivers/Makefile.am`).

jimklimov: Hm, it seems the relative path to another directory should be added with slash and dot-dot like `"/../lib"` (in one example); but "thanks" to `--prefix=/` in those builds, I see paths actually almost ready to use with this approach - e.g.
```
$ grep PATH scripts/Windows/nut_build_x86_64-w64-mingw32/include/config.h
#define CONFIG_FLAGS "--host=x86_64-w64-mingw32 --build=x86_64-linux-gnu --prefix=/ --enable-keep_nut_report_feature PKG_CONFIG_PATH=/usr/x86_64-w64-mingw32/lib/pkgconfig --with-all=auto --with-doc='man=auto html-single=auto html-chunked=skip pdf=skip' --without-systemdsystemunitdir --with-pynut=app --with-augeas-lenses-dir=/augeas-lenses --enable-Werror"

#define ALTPIDPATH "/var/state/ups"
#define CGIPATH "//cgi-bin"
#define CONFPATH "//etc"
#define DRVPATH "//bin"
#define HTMLPATH "//html"
#define PIDPATH "/run"
#define STATEPATH "/var/state/ups"
```

...but lacking the `/../` part - so now I am surprised it did find anything (whether binaries or configs).

Needs some more debugging I guess.

Also, the `getfullpath()` is used inconsistently I think, as it currently uses `xstrdup()` to return what it finds and needs the caller to `free()` the returned string, while methods that return PID, config, etc. locations generally return pointers to someone else's strings that must not be freed (env, built-in defaults). So I guess these calls can leak when repetitively called on Win32 builds.

jimklimov:
Supplanted by PR #3224

jimklimov:
PR #3224 with hopefully a fix for this situation was merged to master. Testing that the result works is still desirable :)

Issue #2289, opened 2024-01-28
apc_modbus "buffer overflow detected" - usb enabled libmodbus


cbrherms:
Running in to a similar (if not the same), issue as mentioned in https://github.com/networkupstools/nut/pull/2063#issuecomment-1905299561

This is with a 2020 SMT2200RM12U, connected via USB and attempting to use apc_modbus.
Have been able to build the modified libmodbus fine (and attempted to use apc-test-client but showed a similar error), but upon building nut with this modified lib for the driver, copying driver to the correct place, and installing over the apt package of nut as described in wiki/comments on the issue thread, the driver won't load.

Output when trying to start the driver directly:

```
cbrherms@nutpi:~$ sudo /lib/nut/apc_modbus -a apc-2200 -u nut -DDDDDD -d1
   0.000001     [D5] send_to_all: SETINFO driver.state "init.starting"
Network UPS Tools - NUT APC Modbus driver 0.01 (2.8.1-403-g746cb8611)
   0.000478     [D5] send_to_all: SETINFO driver.version.usb "libusb-1.0.25 (API: 0x1000109)"
   0.000527     [D1] Using USB implementation: libusb-1.0.25 (API: 0x1000109)
   0.000716     [D5] do_upsconf_args: confupsname=(null), var=pollinterval, val=1
   0.000755     [D5] do_upsconf_args: call do_global_args()
   0.000792     [D3] do_global_args: var='pollinterval' val='1'
   0.000829     [D6] testval_reloadable: var=pollinterval, oldval=2, newval=1, reloadable=1, reload_flag=0
   0.000891     [D1] testval_reloadable: setting 'pollinterval' exists and differs: new value '1' vs. '2'
   0.000934     [D6] testval_reloadable: verdict for (re)loading var=pollinterval value: 1
   0.000989     [D5] do_upsconf_args: confupsname=(null), var=maxretry, val=3
   0.001034     [D5] do_upsconf_args: call do_global_args()
   0.001086     [D3] do_global_args: var='maxretry' val='3'
   0.001129     [D5] do_upsconf_args: confupsname=apc-2200, var=driver, val=apc_modbus
   0.001175     [D5] do_upsconf_args: call main_arg()
   0.001216     [D3] main_arg: var='driver' val='apc_modbus'
   0.001252     [D5] do_upsconf_args: not a main_arg()
   0.001295     [D5] do_upsconf_args: this is a 'driver' setting, may we proceed?
[...]

PR #3069, opened 2025-09-02
apc_modbus: Fix invalid memory access for USB vendor/product/serial

marcan:
As requested in #2289. Only compile tested, but hopefully it should be a fairly obvious fix.

### DIFF
--- a/drivers/apc_modbus.c
+++ b/drivers/apc_modbus.c
@@ -79,7 +79,7 @@ static const HIDNode_t modbus_rtu_usb_usage_tx = 0xff8600fd;
 /* Variables */
 static modbus_t *modbus_ctx = NULL;
 #if defined NUT_MODBUS_HAS_USB
-static USBDevice_t usbdevice;
+static USBDevice_t usbdevice = {0};
 static USBDeviceMatcher_t *reopen_matcher = NULL;
 static USBDeviceMatcher_t *regex_matcher = NULL;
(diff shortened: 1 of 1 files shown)


drivers/apc_modbus.c
```diff
@@ -79,7 +79,7 @@ static const HIDNode_t modbus_rtu_usb_usage_tx = 0xff8600fd;
 /* Variables */
 static modbus_t *modbus_ctx = NULL;
 #if defined NUT_MODBUS_HAS_USB
-static USBDevice_t usbdevice;
+static USBDevice_t usbdevice = {0};
```
jimklimov, line 82: Thanks! Worried a bit about this line - so hoping that all systems tested by CI would agree this is a valid syntax, otherwise would need to pre-init via `memset`. Some build agents are kept to intentionally use older toolkits to ensure this code base can build wherever, as much as we can.

marcan: As far as I know this is nothing new and valid syntax going back to at least C89. I also don't think it's actually necessary, since "uninitialized" static objects are implicitly zero initialized (bss section), but I just wanted to play it safe.

jimklimov: Right, I guess `{0}` shortcut should be okay; it was the named structure field initializers that had "issues" at high warnings/pedanticity levels.

jimklimov:
CI is somehow laggy this week, gotta check what's with the machine after the current builds complete...

PR #3071, opened 2025-09-02
docs/Makefile.am: Fix installation of the pdf files

bigon:
Little typo, big consequences.


jimklimov:
Maintainer Note: the functionality was (incorrectly) introduced after v2.8.2 release, so only now should it actually work - but nothing previously working was regressed due to this.

Issue #3082, opened 2025-09-08
usbhid-ups: let us know which data points from report descriptors were ignored by mapping


jimklimov:
During research for #3061 and #3080, and as a thought rolling lazily in the back of my mind with numerous earlier investigations, the `usbhid-ups` driver on one hand collects a Report Descriptor from the device and parses it into a list of `pDesc` structures, and on another it may use that info while walking data mappings in a particular `drivers/*-hid.c` table of `hid_info_t[]`. Both data walks (what we saw from the device, and what we tried to use in the subdriver - and possibly lacked a mapping for) are reported early in driver startup.

This ticket is about taking the next step useful for troubleshooting/debugging and for driver evolution (to add new mappings over time): report what we saw from the device but did not use in code at run-time.

PR #3095, opened 2025-09-18
Doc links

jimklimov:
* Refer to nut-website rendered pages from asciidoc markup comments (to help git source readers)
* Similar for `usbhid-ups` if too few mappings are used/defined (follow-up from #3082)
* [ ] TODO: Same for `snmp-ups` (also if "auto" matched "ietf" or "tripplite" - no vendor extensions)
* [ ] TODO: Same for `nutdrv_qx`

Also fix a dist tarballing bug reported on mailing list, thanks @gdt


gdt:
I looked at this roughly and it seems like improvements.  Thanks for putting up with being nut maintainer!

PR #3114, opened 2025-10-04
scripts: Avoid `tail -n +/-NUM` syntax as less portable than `tail +/-NUM`

jimklimov:
Addresses part of issue #3099

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -1354,7 +1354,11 @@ consider_cleanup_shortcut() {
     if [ -s "${CI_BUILDDIR}"/configure ] ; then
         # FIXME: Consider CONFIG_SHELL, maybe from script shebang,
         #  here - like autogen.sh does
-        if sh -n "${CI_BUILDDIR}"/configure 2>/dev/null ; then
+        USE_CONFIG_SHELL=sh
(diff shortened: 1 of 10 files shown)


jimklimov:
Odd, tests did fail (shellcheck mismatch for NDE) on the older Linux distros. So something did behave differently :\

GH status says
* slowbuild-run/MATRIX_TAG="linux-ubuntu-trusty-make-shellcheck"
* slowbuild-run/MATRIX_TAG="linux-centos-7-make-shellcheck"

But I only see one in the build log though (linux-centos-7-make-shellcheck):

* https://ci.networkupstools.org/job/nut/job/nut/job/PR-3114/3/pipeline-overview/?selected-node=2921

will investigate sometime...

jimklimov:
Indeed, at least CentOS 7's `tail` can do `tail -n +3` but cannot `tail +3` because no such file `"+3"` :\
And classic Solaris `tail` has no `-n` :\

jimklimov:
CI in run 5: errors on openbsd worker not related to scripting changes (got some FS troubles).
Re-running as build 6 (same commit so old notifications that are not getting overwritten - remain).

Issue #1765, opened 2023-01-04
Refactor modbus drivers: use NUT common code for ModBus over USB link


jimklimov:
In spirit of #1763 consider providing driver configuration options to help match the devices.

Note that currently the point may be moot, with several independent drivers (phoenixcontact_modbus generic_modbus huawei-ups2000 socomec_jbus adelsystem_cbi) not even mentioning "usb" in their sources, so it is possibly offloaded to `libmodbus` build installed into the end-user OS. This may however be a consideration of we proceed with #50 to unify them as subdrivers under a single binary.

PR #3122, opened 2025-10-12
Make `apcupsd-ups` driver IPv6-capable

jimklimov:
Follows up from #3118, #1209 efforts.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -50,6 +50,10 @@ https://github.com/networkupstools/nut/milestone/12
      framework introduced a `upsdrv_tweak_prognames()` hook method to let
      them manipulate the array of `prognames[]` aliases. [PR #3101]
 
+ - `apcupsd-ups` driver updates:
+   * Abandoned use of obsolete `gethostbyname()` in favour of `getaddrinfo()`.
+     Extended to be IPv6-capable along the way. [#1209]
+
  - `nutdrv_qx` driver updates:
    * Define an internal `QX_FLAG_MAPPING_HANDLED` to check if the subdriver
      code (mapping table) and the data found from device walk sit together
--- a/docs/man/apcupsd-ups.txt
+++ b/docs/man/apcupsd-ups.txt
@@ -41,6 +41,14 @@ For instance:
 		port = localhost
 		desc = "apcupsd client"
 
+For IPv6 addresses, use square brackets for the hostname part (so colons in
+it are not mistaken to be the port, especially if using the default one):
+
+	[apcupsd]
+		driver = apcupsd-ups
+		port = "[::1]:3551"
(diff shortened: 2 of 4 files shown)


jimklimov:
Also revised how our other drivers for networked `port` values parse their `device_path` - if anyone needs similar fixes? TLDR: a couple of modbus drivers can benefit from attention about IPv6 ability:
* `netxml-ups.c` uses `ne_uri_parse(device_path, &uri)` to parse the provided URL into schema, host, port, etc.
* `snmp-ups.c` uses `g_snmp_sess.peername = xstrdup(hostname);` and later `snmp_open(&g_snmp_sess);` which parses the `peername` into transport, host, port... -- see https://net-snmp.sourceforge.io/dev/agent/structsnmp__session.html and https://linux.die.net/man/3/snmp_sess_session
* `nut-ipmipsu.c` currently only uses local device access, it seems; there are comments regarding syntax for remote IPMI (`id?@host`), but it is not parsed that way.
  * In `nut-scanner` there is `ipmi_ctx_open_outofband` scanning however. I did not quite find docs, but in sources https://fossies.org/linux/freeipmi/libfreeipmi/api/ipmi-api.c it seems the internal `_setup_hostname()` used in different `ctx open` methods does parse the "hostname" string into `hostname_ptr` and `port_ptr`.
* NUT client drivers (dummy, clone, failover) should be using NUT common library and parse through it for access to remote NUT servers (not verified though).
* `powerman-psu.c` uses `pm_connect()` which per https://linux.die.net/man/3/libpowerman takes care of host-port parsing
* modbus TCP RTU support may be spread around `MODBUS_DRIVERLIST = phoenixcontact_modbus generic_modbus huawei-ups2000 socomec_jbus adelsystem_cbi apc_modbus` which currently do not share code much (issues like #1765 exist). Half of them only work with serial RTU (not sure if this is a problem... can e.g. serial-tcp bridges be used for remote monitoring of serial-only UPSes?); of the rest, one seems IPv6 capable and two are not:
  * `apc_modbus`: uses `_apc_modbus_parse_host_port()` and handles bracketed IPv6 if present, and port
  * `huawei-ups2000`: FIXME or not? Only calls `modbus_new_rtu()` assuming a serial port
  * `phoenixcontact_modbus`: FIXME or not? Only calls `modbus_new_rtu()` assuming a serial port
  * `socomec_jbus`: FIXME or not? Only calls `modbus_new_rtu()` assuming a serial port
  * `adelsystem_cbi`: FIXME: in `modbus_new()` wrapper uses a naive approach to parsing a port (`strchr` for first `:`) so is not IPv6 capable
  * `generic_modbus`: FIXME: in `modbus_new()` wrapper uses a naive approach to parsing a port (`strchr` for first `:`) so is not IPv6 capable

Issue #1962, opened 2023-06-08
Riello: No shutdown.return after fsd


Dave2526:
Hello,
I try to setup my Riello NPW800with nut. It´s working except one "Issue":
When the upssched foreces the shutdown of my server, The UPS didn´t shutdown completly and also didn´t restart after the Power is back. The scrpit seem to set set the shutdown.return command but it isn´t working. Could there be a interrupt with the ForcedShutdown Command?
If I set the shutdown resturn by "upscmd -u admin -p XXXXXXX UPS shutdown.return" it shut down after my selected 40 Seconds.

I Think it´s a little Thing but I didn´t find it...

Here is the System Journal:
````
Jun 08 17:27:01 server systemd[1]: Stopped target Timer Units.
Jun 08 17:27:01 server systemd[1]: Stopped target Sound Card.
Jun 08 17:27:01 server systemd[1]: Stopped target Preparation for Remote File Systems.
Jun 08 17:27:01 server systemd[1]: Stopped target Host and Network Name Lookups.
Jun 08 17:27:01 server systemd[1]: Stopped target Graphical Interface.
Jun 08 17:27:01 server systemd[1]: Stopped target Cloud-init target.
Jun 08 17:27:01 server systemd[1]: Removed slice Slice /system/modprobe.
Jun 08 17:27:01 server sshd[878740]: pam_unix(sshd:session): session closed for user mainuser
Jun 08 17:27:01 server systemd[1]: Stopping Session 3 of User mainuser...
Jun 08 17:27:01 server systemd[1]: unattended-upgrades.service: Deactivated successfully.
Jun 08 17:27:01 server systemd-logind[665]: System is powering down.
Jun 08 17:27:01 server systemd[1]: nut-monitor.service: Deactivated successfully.
Jun 08 17:27:01 server systemd-logind[665]: Creating /run/nologin, blocking further logins...
Jun 08 17:26:58 server upsscript[892787]: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Jun 08 17:26:58 server upsscript[892786]: UPS Shutdown+return forced; Sytem Shutdown forced, Goodbye!
Jun 08 17:26:58 server upsscript[892785]: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #3124, opened 2025-10-12
Drop `riello_{ser,usb}` fencing of shutdown-related commands based on power status

jimklimov:
Closes: #1962

Investigation detailed in that issue.

NOTE: A hypothesis there about not all drivers now initializing their `instcmd` support when started for killpower activity may be or not be moot, but was almost not the case here. Additional inputs (reading the device status... or not) did impact the check for availability of those commands in an unusually conditional run-time code.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -698,9 +698,13 @@ spellcheck spellcheck-interactive:
 		SUBDIR_MAKE_VERBOSE=0 ; \
 	  fi ; \
 	  export SUBDIR_MAKE_VERBOSE ; \
-	  $(MAKE) $(AM_MAKEFLAGS) SPELLCHECK_TGT='$@' SUBDIR_MAKE_VERBOSE="$${SUBDIR_MAKE_VERBOSE}" -k -s $(SPELLCHECK_DIRS) && exit ; \
+	  ( $(MAKE) $(AM_MAKEFLAGS) SPELLCHECK_TGT='$@' SUBDIR_MAKE_VERBOSE="$${SUBDIR_MAKE_VERBOSE}" -k -s $(SPELLCHECK_DIRS) && exit ; \
 	  echo "WARNING: FAILED fanned-out attempt in $@, retrying with NUT_MAKE_SKIP_FANOUT" >&2 ; \
-	  $(MAKE) $(AM_MAKEFLAGS) NUT_MAKE_SKIP_FANOUT=true SPELLCHECK_TGT='$@' -k -s $(SPELLCHECK_DIRS)
+	  $(MAKE) $(AM_MAKEFLAGS) NUT_MAKE_SKIP_FANOUT=true SPELLCHECK_TGT='$@' -k -s $(SPELLCHECK_DIRS) ) || exit ; \
+	  if [ x"$@" = xspellcheck-interactive ] ; then \
+		echo "SUCCESS: $@: follow up with spellcheck-quick to revise and update timestamps"; \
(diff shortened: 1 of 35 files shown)


jimklimov:
@useful64 : can you please check if this behaves better for you - i.e. serves both `load.*` and `shutdown.return` both when started as a daemon and just standalone for killpower?

Would it be possible to check that these commands work always, regardless of online/onbatt status? (According to protocol docs, it may take 600 sec for the UPS to actually shut down after it was commanded)

useful64:
> @useful64 : can you please check if this behaves better for you - i.e. serves both `load.*` and `shutdown.return` both when started as a daemon and just standalone for killpower?
> 
> Would it be possible to check that these commands work always, regardless of online/onbatt status? (According to protocol docs, it may take 600 sec for the UPS to actually shut down after it was commanded)

Hi, I can't check it until next week. I'll report the results next weekend.

jimklimov:
@useful64 : code-wise that seemed okay, so merged now - you can test from the master branch. Still hoping to learn if the change goes well or not :)

useful64:
Hi @jimklimov, I tested from the master branch and **now it worked** without starting the driver before.

Here are the logs:
```
$ sudo upsdrvctl shutdown riello
Network UPS Tools upsdrvctl - UPS driver controller 2.8.4.437-437+g9b4f848d1 (development iteration after 2.8.4)
Network UPS Tools 2.8.4.437-437+g9b4f848d1 (development iteration after 2.8.4) - Riello USB driver 0.17
Warning: This is an experimental driver.
Some features may not function correctly.

Checking if an already running driver instance can handle the shutdown command for us
Can't open /var/lib/nut/riello_usb-riello: No such file or directory
Initiating UPS [riello] shutdown
Processing command 'shutdown.stop' while ups.status is 'OB TRIM'
riello_instcmd: attempting an operation [shutdown.stop] [(null)] which may impact power state on [riello]
Processing command 'shutdown.return' while ups.status is 'OB TRIM'
riello_instcmd: attempting a power state change operation [shutdown.return] [(null)] on [riello]
Shutting down
UPS [riello]: shutdown request was successful with 'shutdown.default'
upsnotify: notify about state NOTIFY_STATE_STOPPING with libsystemd: was requested, but not running as a service unit now, will not spam more about it
WARNING: upsdrvctl was called directly on a system with systemd support.
    Please consider using 'upsdrvsvcctl' instead, to avoid conflicts with
    nut-driver service instances prepared by 'nut-driver-enumerator'!
```
And the journal logs:
```
oct 24 17:51:18 a riello_usb[82021]: Initiating UPS [riello] shutdown
oct 24 17:51:19 a riello_usb[82021]: Processing command 'shutdown.stop' while ups.status is 'OB TRIM'
oct 24 17:51:19 a riello_usb[82021]: riello_instcmd: attempting an operation [shutdown.stop] [(null)] which may impact power state on [riello]
oct 24 17:51:19 a riello_usb[82021]: Processing command 'shutdown.return' while ups.status is 'OB TRIM'
oct 24 17:51:19 a riello_usb[82021]: riello_instcmd: attempting a power state change operation [shutdown.return] [(null)] on [riello]
oct 24 17:51:20 a riello_usb[82021]: Shutting down
oct 24 17:51:20 a riello_usb[82021]: UPS [riello]: shutdown request was successful with 'shutdown.default'
oct 24 17:51:20 a riello_usb[82021]: upsnotify: notify about state NOTIFY_STATE_STOPPING with libsystemd: was requested, but not running as a service unit now, will not spam more about it
```

Thanks!

jimklimov:
Great, thanks for confirmation!

PR #3125, opened 2025-10-12
Revise `upsdrv_initups()` vs `upsdrv_initinfo()`, be sure to reference instcmd and setvar handlers

jimklimov:
Raised by discussion of #1962

Changes to make driver code closer to what `docs/new-drivers.txt` requires (`initups` is for preparing the comms line, `initinfo` is for actually using it to detect presence of an UPS and its characteristics). They should not significantly change execution order in normal operations, but makes the code structure better (more like expected) for maintainability and review.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -698,9 +698,13 @@ spellcheck spellcheck-interactive:
 		SUBDIR_MAKE_VERBOSE=0 ; \
 	  fi ; \
 	  export SUBDIR_MAKE_VERBOSE ; \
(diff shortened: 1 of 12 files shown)


jimklimov:
Checked with `usbhid-ups`, no surprises for normal start-up.

Arguably it is still beside standard driver requirements, with `callback: calling HIDDumpTree(); in case of problems with device data please note that a wrong subdriver could have been chosen above; consider testing others with an explicit driver option` being in `upsdrv_initups()` to eventually confirm that the picked subdriver with HID mapping is good enough against the report descriptor, and its data is not bogus during interpretation (else try some other size, offset, subdriver...).

Then it goes to `upsdrv_initinfo: Performing an initial UPS data walk with subdriver MGE HID 1.57...` and stuff like `analyze_mapping_usage()` now happens there.

Issue #3137, opened 2025-10-19
Ippon TAE 1000 support


dvdesolve:
I've bought brand new TAE 1000 UPS. Despite being equipped with both RS232 and USB slots seems like internally it uses CH340 converter for USB frontend effectively providing two serial ports:
```
Bus 003 Device 002: ID 1a86:7523 QinHeng Electronics CH340 serial converter
```

So I tried only USB connection but think that the same will be true for RS232 connect (I can try later if you wish). As for now it's working with Q1 protocol ruled by `nutdrv_qx` driver. udev rules should be properly set to allow access for `nut` user. Personally  I have the following lines in my `/etc/udev/rules.d/52-nut.rules`:
```
# Ippon Back Basic series
SUBSYSTEM=="usb", ATTRS{idVendor}=="0665", ATTRS{idProduct}=="5161", MODE="0660", GROUP="nut"
# Ippon Innova TAE series
SUBSYSTEM=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="7523", MODE="0660", GROUP="nut"
```

First-order approximation for `ups.conf`:
```
[tae1000]
    driver      = nutdrv_qx
    protocol    = q1
    desc        = "TAE 1000 UPS"

    port        = /dev/ttyUSB0
```
gave me the following result:
```
battery.voltage: 2.24
device.type: ups
driver.debug: 0
driver.flag.allow_killpower: 0
driver.name: nutdrv_qx
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.port: /dev/ttyUSB0
driver.parameter.protocol: q1
driver.parameter.synchronous: auto
driver.state: quiet
driver.version: 2.8.4
driver.version.data: Q1 0.08
driver.version.internal: 0.45
driver.version.usb: libusb-1.0.29 (API: 0x0100010B)
input.frequency: 49.9
input.voltage: 218.8
input.voltage.fault: 0.0
output.voltage: 229.7
ups.beeper.status: disabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.load: 0
ups.status: OL
ups.temperature: 51.0
ups.type: online
```

As it can be seen there is no vendor or rating information at all. Battery voltage readings are incorrect. I can confirm that this UPS model support all proper Megatec Qx commands but with some specialties for `I` command:
```
#222222222222222222222222222R1.00.48
```
[...]

PR #3139, opened 2025-10-20
Protocol for Ippon Innova TAE UPSes

dvdesolve:
Simple PR to add specific protocol for Ippon Innova TAE (also see #3137). Nominal ratings, manufacturer and firmware version are reported by this variant of protocol

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -229,6 +229,10 @@ several `FSD` notifications into one executed action. [PR #3097]
  - The `configure` script should now try harder to report specifically
    the "purelib" location as `PYTHON*_SITE_PACKAGES`. [#1209]
 
+ - Revised use of NUT macros for USB Vendor IDs listed in different drivers,
+   so the names would be better exposed in generated `udev.rules` and similar
+   files. [PR #3139]
+
 
 Release notes for NUT 2.8.4 - what's new since 2.8.3
 ----------------------------------------------------
--- a/drivers/apc-hid.c
+++ b/drivers/apc-hid.c
@@ -35,7 +35,7 @@
 #define APC_HID_VERSION "APC HID 0.101"
 
 /* APC */
-#define APC_VENDORID 0x051d
+#define APC_VENDORID	0x051d
 
 /* Tweaks */
 static char * tweak_max_report[] = {
--- a/drivers/apc_modbus.c
+++ b/drivers/apc_modbus.c
@@ -49,7 +49,7 @@
 #if defined NUT_MODBUS_HAS_USB
 
 /* APC */
-#define APC_VENDORID 0x051D
+#define APC_VENDORID	0x051d
 
 /* USB IDs device table */
 static usb_device_id_t apc_modbus_usb_device_table[] = {
--- a/drivers/arduino-hid.c
+++ b/drivers/arduino-hid.c
@@ -41,7 +41,7 @@
 
 /* Arduino */
 #define ARDUINO_VENDORID	0x2341
-#define ARDUINO_VENDORID2	0x2A03
+#define ARDUINO_VENDORID2	0x2a03
 
 /* USB IDs device table */
 static usb_device_id_t arduino_usb_device_table[] = {
--- a/drivers/bcmxcp_usb.c
+++ b/drivers/bcmxcp_usb.c
@@ -26,13 +26,13 @@ upsdrv_info_t comm_upsdrv_info = {
 #define MAX_TRY_OPENUSB 32
 
 /* Powerware */
-#define POWERWARE 0x0592
+#define POWERWARE_VENDORID	0x0592
 
 /* Phoenixtec Power Co., Ltd */
-#define PHOENIXTEC 0x06da
+#define PHOENIXTEC_VENDORID	0x06da
 
 /* Hewlett Packard */
-#define HP_VENDORID 0x03f0
+#define HP_VENDORID	0x03f0
 
 static USBDevice_t curDevice;
 
@@ -81,10 +81,10 @@ static void *phoenixtec_ups(USBDevice_t *device) {
 /* USB IDs device table */
 static usb_device_id_t pw_usb_device_table[] = {
 	/* various models */
-	{ USB_DEVICE(POWERWARE, 0x0002), &powerware_ups },
+	{ USB_DEVICE(POWERWARE_VENDORID, 0x0002), &powerware_ups },
 
 	/* various models */
-	{ USB_DEVICE(PHOENIXTEC, 0x0002), &phoenixtec_ups },
+	{ USB_DEVICE(PHOENIXTEC_VENDORID, 0x0002), &phoenixtec_ups },
 
 	/* T500 */
 	{ USB_DEVICE(HP_VENDORID, 0x1f01), &phoenixtec_ups },
--- a/drivers/blazer_usb.c
+++ b/drivers/blazer_usb.c
@@ -49,6 +49,39 @@ upsdrv_info_t upsdrv_info = {
 	{ NULL }
 };
 
+/* Unregistered vendor 0x0001 (commonly identified as Fry's Electronics) */
+#define NONAME0001_VENDORID	0x0001
+
+/* Unregistered vendor 0xFFFF */
+#define NONAMEFFFF_VENDORID	0xffff
+
+/* ST Microelectronics */
+#define STMICRO_VENDORID	0x0483
+
+/* Sysgration Ltd. */
+#define SYSGRATION_VENDORID	0x05b8
+
+/* Cypress Semiconductor */
+#define CYPRESS_VENDORID	0x0665
+
+/* Phoenixtec Power Co., Ltd */
+#define PHOENIXTEC_VENDORID	0x06da
+
+/* Lakeview Research */
+#define LAKEVIEW_VENDORID	0x0925
+
+/* Unitek UPS Systems */
+#define UNITEK_VENDORID	0x0f03
+
+/* GE */
+#define GE_VENDORID	0x14f0
+
+/* QinHeng Electronics */
+#define QINHENG_VENDORID	0x1a86
+
+/* Legrand */
+#define LEGRAND_VENDORID	0x1cb0
+
 #ifndef TESTING
 
 static usb_communication_subdriver_t *usb = &usb_subdriver;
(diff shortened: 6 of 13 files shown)


jimklimov:
> https://github.com/networkupstools/nut/issues/3137#issuecomment-3423243292
> With proposed PR I'm getting (...) nominal values as well as manufacturer and firmware version info reported now. However, battery pack detection together with battery charge estimation still not working without specifying both `battery_voltage_reports_one_pack` and `override.battery.packs = 12`

I think you can cover that in e.g. `innovatae_initups()` - see some other drivers that do complex work here. The rough idea is that if e.g. battery voltage is in 1.6-2.4V range or so, assume `battery_voltage_reports_one_pack` and `battery.packs` is at least one multiple of 6 (to get N*12V nominal for PbAc). Not sure if there's a better way to guess the number more precisely. Maybe some other Qx queries can shed the light?..

Also, do you have to explicitly put `protocol=innovatae` into `ups.conf` now? I think a "claim" method could somehow automatically cover this and differentiate from other devices, e.g. check for bogus "mfr" and/or numbers there and something good here? It might mean that this protocol should get mentioned before potent competitors in `subdriver_list[]` of `nutdrv_qx.c`, probably.

For the manufacturer you essentially serve a static string (`Ippon`). I think a processing method can be used for an `I` query, so if that returns something bogus (empty, repetitions of characters) - return a predefined value, else return what is served. Maybe this fallback logic can be added to the common parser of `item_t[]` tables with a new `QX_FLAG_<something>`. Anyhow, a separate PR can address that if needed (also for `nutdrv_qx_masterguard.c` that seems to use similar `QX_FLAG_ABSENT` defaulting for something other than a `ups.delay.*` value).

Finally, if you would make changes to the PR per above ideas, can you please also add a `NEWS.adoc` entry? :)

dvdesolve:
Just tried with full autodetection - it detects Megatec 0.09 protocol variant:
```
окт 21 19:44:13 desolve-desktop nut-driver@tae1000[5335]: Battery runtime will not be calculated (runtimecal not set)
окт 21 19:44:13 desolve-desktop nut-driver@tae1000[5335]: Autodetected 12 as number of battery packs [24/2.24]
окт 21 19:44:13 desolve-desktop nut-driver@tae1000[5335]: Using 'guesstimation' (low: 20.800000, high: 26.000000)!
окт 21 19:44:13 desolve-desktop nut-driver@tae1000[5335]: No values for battery high/low voltages
окт 21 19:44:12 desolve-desktop nut-driver@tae1000[5335]: Using protocol: Megatec 0.09
окт 21 19:44:09 desolve-desktop nut-driver@tae1000[5335]: USB communication driver (libusb 1.0) 0.50
```
as a result the following data is available:
```
battery.charge: 0
battery.runtime: 22222
battery.voltage: 2.25
battery.voltage.high: 26.00
battery.voltage.low: 20.80
battery.voltage.nominal: 24.0
device.mfr: 2222222
device.model: 22222
device.serial: 22222222
device.type: ups
driver.debug: 0
driver.name: nutdrv_qx
driver.parameter.pollfreq: 30
driver.parameter.port: /dev/ttyUSB0
driver.state: quiet
driver.version: 2.8.4.1-0+g975e2fdc9
driver.version.data: Megatec 0.09
driver.version.internal: 0.47
driver.version.usb: libusb-1.0.29 (API: 0x0100010B)
input.current.nominal: 4.0
input.frequency: 50.0
input.frequency.nominal: 50
input.voltage: 226.0
input.voltage.fault: 0.0
input.voltage.nominal: 230
output.voltage: 229.7
ups.beeper.status: disabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.firmware: R1.00.48
ups.load: 0
ups.serial: 22222222
ups.status: OL
ups.temperature: 47.0
ups.type: online
```
Battery voltage is still incorrectly displayed so as battery charge (though battery packs are detected in a right way). Battery runtime seems to be irrelevant because of incorrect interpretation of `I` command reply (in `nutdrv_qx_megatec.c`)

**Addition**: even with `battery_voltage_reports_one_pack` option and full autodetection battery voltage and charge are still incorrectly calculated

**Addition 2**: the same is true for `innovatae` protocol - regardless of the `battery_voltage_reports_one_pack` setting I have the following data dump:
```
battery.charge: 0
battery.voltage: 2.21
battery.voltage.high: 26.00
battery.voltage.low: 20.80
battery.voltage.nominal: 24.0
device.mfr: Ippon
device.type: ups
driver.debug: 0
driver.name: nutdrv_qx
driver.parameter.pollfreq: 30
driver.parameter.port: /dev/ttyUSB0
driver.parameter.protocol: innovatae
driver.state: quiet
driver.version: 2.8.4.1-0+g975e2fdc9
driver.version.data: INNOVATAE 0.01
driver.version.internal: 0.47
driver.version.usb: libusb-1.0.29 (API: 0x0100010B)
input.current.nominal: 4.0
input.frequency: 50.0
input.frequency.nominal: 50
input.voltage: 226.0
input.voltage.fault: 0.0
input.voltage.nominal: 230
output.voltage: 229.9
ups.beeper.status: disabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.firmware: R1.00.48
ups.load: 0
ups.status: OL
ups.temperature: 49.0
ups.type: online
```
Number of battery packs is correct:
```
окт 21 20:19:41 desolve-desktop nut-driver@tae1000[7925]: Battery runtime will not be calculated (runtimecal not set)
окт 21 20:19:41 desolve-desktop nut-driver@tae1000[7925]: Autodetected 12 as number of battery packs [24/2.25]
окт 21 20:19:41 desolve-desktop nut-driver@tae1000[7925]: Using 'guesstimation' (low: 20.800000, high: 26.000000)!
окт 21 20:19:41 desolve-desktop nut-driver@tae1000[7925]: No values for battery high/low voltages
окт 21 20:19:41 desolve-desktop nut-driver@tae1000[7925]: Using protocol: INNOVATAE 0.01
окт 21 20:19:40 desolve-desktop nut-driver@tae1000[7925]: USB communication driver (libusb 1.0) 0.50
```
Only when I specify both parameters (`battery_voltage_reports_one_pack` and  `battery.packs`) I'm able to get right readings and estimations:
```
battery.charge: 100
battery.packs: 12
battery.voltage: 27.00
battery.voltage.high: 26.00
battery.voltage.low: 20.80
battery.voltage.nominal: 24.0
device.mfr: Ippon
device.type: ups
driver.debug: 0
driver.flag.battery_voltage_reports_one_pack: enabled
driver.name: nutdrv_qx
driver.parameter.override.battery.packs: 12
driver.parameter.pollfreq: 30
driver.parameter.port: /dev/ttyUSB0
driver.parameter.protocol: innovatae
driver.state: updateinfo
driver.version: 2.8.4.1-0+g975e2fdc9
driver.version.data: INNOVATAE 0.01
driver.version.internal: 0.47
driver.version.usb: libusb-1.0.29 (API: 0x0100010B)
input.current.nominal: 4.0
input.frequency: 50.0
input.frequency.nominal: 50
input.voltage: 227.3
input.voltage.fault: 0.0
input.voltage.nominal: 230
output.voltage: 229.9
ups.beeper.status: disabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.firmware: R1.00.48
ups.load: 0
ups.status: OL
ups.temperature: 49.0
ups.type: online
```
```
battery.charge: 100
battery.packs: 12
battery.runtime: 22222
battery.voltage: 27.00
battery.voltage.high: 26.00
battery.voltage.low: 20.80
battery.voltage.nominal: 24.0
device.mfr: 2222222
device.model: 22222
device.serial: 22222222
device.type: ups
driver.debug: 0
driver.flag.battery_voltage_reports_one_pack: enabled
driver.name: nutdrv_qx
driver.parameter.override.battery.packs: 12
driver.parameter.pollfreq: 30
driver.parameter.port: /dev/ttyUSB0
driver.state: quiet
driver.version: 2.8.4.1-0+g975e2fdc9
driver.version.data: Megatec 0.09
driver.version.internal: 0.47
driver.version.usb: libusb-1.0.29 (API: 0x0100010B)
input.current.nominal: 4.0
input.frequency: 50.0
input.frequency.nominal: 50
input.voltage: 226.0
input.voltage.fault: 0.0
input.voltage.nominal: 230
output.voltage: 230.7
ups.beeper.status: disabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.firmware: R1.00.48
ups.load: 0
ups.serial: 22222222
ups.status: OL
ups.temperature: 49.0
ups.type: online
```

My suggestion is that battery pack autodetection is a bit wrong - if corresponding flags and values are already supplied by the user the rest of math works fine. If autodetection kicks in during `initinfo` then we lose some multiplications on the way

dvdesolve:
> For the manufacturer you essentially serve a static string (Ippon). I think a processing method can be used for an I query, so if that returns something bogus (empty, repetitions of characters) - return a predefined value, else return what is served. Maybe this fallback logic can be added to the common parser of item_t[] tables with a new QX_FLAG_something. Anyhow, a separate PR can address that if needed (also for nutdrv_qx_masterguard.c that seems to use similar QX_FLAG_ABSENT defaulting for something other than a ups.delay.* value).

Regarding static manufacturer string - you're absolutely right! At first I've searched for hardcoded `ups.mfr` and `device.mfr` params and found exactly Megatec variant so the whole line is borrowed from that source file. Also it seems like TAE (and TBE) series is/are brand new and may report that bogus info as `I` reply. Unfortunately, I have no other UPSes of these series at hand and can't test this assumption. However, because we're making separate protocol variant for TAE series we can, in principle, leave `mfr` params hardcoded for now - if later we would find that not all specimens of this series report in the same way we could fix this behavior.

The same idea applies for your 2nd paragraph as well:
> Also, do you have to explicitly put protocol=innovatae into ups.conf now? I think a "claim" method could somehow automatically cover this and differentiate from other devices, e.g. check for bogus "mfr" and/or numbers there and something good here? It might mean that this protocol should get mentioned before potent competitors in subdriver_list[] of nutdrv_qx.c, probably.

It's a pity that I only have one sample of this new series. May be soon I will be able to get some other samples and test them too. As for now this special version for TAE seems to additionally report rating info as well as firmware version in comparison with basic Q1 variant - for example, my Back Basic series doesn't provide neither nominal values nor FW version. Moreover, Back Basic 2200 doesn't tell me temperature at all while Back Basic 850 does

dvdesolve:
Also I've found that reply to `I` actually contains two extra spaces just before carriage return symbol (thanks, hex dump!). It makes the whole picture complete because in `megatec` protocol sources I've found two reply examples which confirms my initial assumption about FW version in `I` reply string

dvdesolve:
Oh, I deeply sorry about forgetting to add news entry! I'm used to you're doing that part :)
BTW what data dumps should I provide?

jimklimov:
Hopefully now the USB device (when `port="auto"`) should be also automatically detected as supported, e.g. `udev.rules` should now assign the relevant permissions.

I am still not sure if it would pick the correct `protocol=...` for it, or if there is more magic needed to ensure this. Some other recent `innova*` subdrivers also lack such magic, so maybe the correct handler code for those devices can only be assigned explicitly?

dvdesolve:
> I am still not sure if it would pick the correct `protocol=...` for it, or if there is more magic needed to ensure this. Some other recent `innova*` subdrivers also lack such magic, so maybe the correct handler code for those devices can only be assigned explicitly?

Unfortunately, simple UPSes like Back Basic series by Ippon (and some others too, ofc) lack model-specific queries/answers so we can't differentiate between them confidently. One possible way (quite brittle, btw) is to evaluate combinations of available queries and bogus answers like `222222...` in TAE series. But anyway it looks much simpler to choose specific protocol from NUT-provided list. Possible drawback is that we're duplicating a lot of code with only small additions on top of it which can later lead to harder maintenance in case something goes wrong.

It would be great to implement some kind of inheritance for protocol development. A lot of protocol variants use basic requests such as Q1, Q2, Q6 and so on (in case of nutdrv_qx driver) and only adds/removes some minor parts on top of it. So basic inheritance would allow the use of already established "root" protocol variants with simple edits based upon them.

Thank you for fixing this PR and merging!

jimklimov:
As a side note, for this particular driver family, the "inheritance" is largely implemented, thanks to `blazer-common` methods referenced from different subdrivers' mapping tables.

PR #3141, opened 2025-10-21
nutdrv_qx_masterguard: Fix masterguard_add_slaveaddr return value and makes the deep battery test command works

atanas-vladimirov:
The masterguard_add_slaveaddr function, defined as
a preprocess_command handler, must return the length of the command
string it has processed. It was incorrectly returning 0.

This caused the main qx_process function to see a 0-byte command
length, resulting in a send: timeout (0) debug log and a failure
to send the command to the UPS.

This commit fixes the function to return the correct command length
(which was already stored in the local variable l), resolving the bug
for all commands that use the ,XX slave address placeholder.

The test.battery.start.deep (TUD) command requires the slave address
as a parameter to function.

Following the fix to masterguard_add_slaveaddr this command can 
now be properly enabled.

Tested on Masterguard A1000 and A2000 units.

### DIFF
--- a/docs/nut.dict
+++ b/docs/nut.dict
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 3564 utf-8
+personal_ws-1.1 en 3566 utf-8
 AAC
 AAS
 ABI
@@ -1319,6 +1319,7 @@ TSR
 TST
 TT
 TTT
+TUD
 TXF
 TXG
 TXV
@@ -3087,6 +3088,7 @@ ske
 skel
 sl
 slackpkg
+slaveaddr
 slaveid
 slavesync
 slibtool


jimklimov:
Just... wow. Thanks! :)

Worth a `NEWS.adoc` entry and a subdriver version bump.

atanas-vladimirov:
Hey Jim,

Thanks :)

>Worth a NEWS.adoc entry and a subdriver version bump.

Let me know what should I do and I can try :)

jimklimov:
Update this line: `#define MASTERGUARD_VERSION "Masterguard 0.03"` and describe the problem and solution in `NEWS.adoc`. There is already an entry under `nutdrv_qx driver updates` for your earlier PR, maybe add another near it or a few new words into it.

atanas-vladimirov:
>Update this line: #define MASTERGUARD_VERSION "Masterguard 0.03" ..

Should I bump it to "Masterguard 0.04" or something else? 

Also, now I see that a few of the checks failed, not sure why and if I have to do something ....

jimklimov:
Yeah, `0.04` seems right. CI had an out of `/tmp` space situation, will try to clean/bump that worker.

atanas-vladimirov:
Thanks, Jim,

I have pushed a new commit, let me know if something else is needed.

jimklimov:
Thanks! We'll see if the spell checkers on different CI platforms agree with `TUD` etc., but otherwise looks good, thanks!

atanas-vladimirov:
Yep, it seems the spell checker is not happy - let me know if I have to do anything to fix it :)

jimklimov:
Dictionary updated...

PR #3145, opened 2025-10-24
Some fixes after recent PR merges

jimklimov:
Notably python-default changes.

Also future-proof CircleCI with MacOS.


jimklimov:
PyPI fixed as of https://github.com/networkupstools/nut/actions/runs/18811168989/job/53672534508 and resulting https://test.pypi.org/project/pynutclient/2.8.4.657/ upload

PR #3156, opened 2025-10-31
usbhid-ups: consider FULLYCHARGED and NOTFULLYCHARGED separately

### LINKED ISSUE #3146: eaton 9e3000 not getting info
recently the sd card on my rasberry pi died, and now i i'm configuring again the nat server

i have 2 ups 
- apc 1400
- eaton 9e3000

the apc ups don't give me any problem

with the older configuration with the command 
nut-scanner -U

i get from both UPS all the info

but now with new version of nut the eaton ups
don't give me this info: serial, product, vendor

note: from dmsg i can see the serial number from both the UPS
so i think this problem is releted to nut

and now i see a new strange comment from both ups
 ###NOTMATCHED-YET###bcdDevice

i copied the old config file but now i get the error
for the eaton ups "file not found"

any idea?

--------------------
PC: rasberry pi 4 
OS: rasberry pi OS 64bit (debian trixie)
nut version: 2.8.1

jimklimov:
…because many subdrivers report neither; fall back to checking `battery.charge=?100` to set `CHRG` flag (avoid `OL CHRG` when UPS seems charged)

Closes: #3146


jimklimov:
Disregard the OBS failures, test-packaging recipes need to be completed by another issue.

jimklimov:
Maybe makes sense to track a previous `battery.charge` reading whenever it changes, to only report `CHRG` if the value grows (e.g. wall power/genset is weak, and despite having an input, the UPS slowly loses charge overall?), and maybe track "when" to evaluate the speed (runtimecal style)? There was however a ticket to make it a common feature for all drivers; maybe fits better into that solution then...

biergaizi:
> avoid OL CHRG when UPS seems charged

I do not believe this is technically the correct thing to do. For example, acid-acid batteries are electrically always being charged in a process known as float-charging, regardless of how much energy the batteries have, so `OL CHRG` are reported by most drivers, as they accurately reflect the physical charger status as reported by the UPS firmware. Depending on the implementation detail of the BMS, other battery types may also have the charger activated even when the batteries are at "100%" during maintenance tasks such as "equalization charging".

So my interpretation of the "CHRG" flag is that the flag should reflect the "is charging" register status reported by the UPS, for many UPS models the charger is always on. This is intended and not a bug. Hiding the flag when the battery is at 100% is a bug, because the flag no longer reflects the register status. I'd rather say: `CHRG` may or may not appear at 100% battery, it's *driver implementation-specified* and *should not* be be relied upon by any user.

```c
int	misleading_CHRG =    (current_battery_charge_value < previous_battery_charge_value && reports_CHRG);
```

I'd say it's `honest_CHRG`.

jimklimov:
This is an interesting take on the problem, thanks.

Looking for more context, in `docs/nut-names.txt` there is such a line:
```
 `battery.charger.status` replaces the historic flags `CHRG` and `DISCHRG`
```

while in `docs/new-drivers.txt` I see:
```
   CHRG    -- The battery is charging
   DISCHRG -- The battery is discharging (inverter is providing load power)
- the `CHRG` and `DISCHRG` flags are being replaced with
  `battery.charger.status`.  See the linkdoc:user-manual[NUT command and
  variable naming scheme,nut-names] for more information.
```

So at least there is a reason to reflect the real state ("charger is working, even if it always does") at least in the dedicated variable. I am not sure whether this happens currently, especially with this (possibly misleaded) effort to hide the CHRG flag.

Issue #1400, opened 2022-04-26
CI: Generate downloadable tarballs as part of NUT CI farm builds


jimklimov:
Previously this was done by buildbot, but it is planned for deprecation.

jimklimov:
See https://javahelps.com/manage-github-artifact-storage-quota for hints about making a github action for this to keep a few recent builds afloat and stay within quota.

Maybe this would benefit from a pre-defined Docker container to have all the tools and prerequisites installed once.

jimklimov:
This may be useful also for #1209.

jimklimov:
PR #3158 achieved the generation and publication of tarballs; not sure (due to GHA security nuances) if it can or not inform PR discussion readers about the result, or remove older tarball builds associated with the PR branch to keep storage available. Will see in later runs...

jimklimov:
With PRs merged and amended between NUT v2.8.4 and v2.8.5 releases, a "GHA-01" job in commit checks list should carry an archive with "dist" tarball as well as another with built docs for all formats so we can check asciidoc markup vs. rendering.

PR #3158, opened 2025-11-06
CI: Generate downloadable tarballs as part of NUT CI farm builds; make default OBS recipes buildable

jimklimov:
Closes: #1400

Likely helps move #1209 and #2829 forward eventually.

Probably should be merged to `master` so the workflow actually runs on GHA (due to safer `pull_request_target` usage, allegedly needed for edits of PR comments; with `pull_request` the code did run here for iterations when the YAML was not broken :) ).

### DIFF
--- a/.github/workflows/01-make-dist.yml
+++ b/.github/workflows/01-make-dist.yml
@@ -4,7 +4,7 @@
 # See also:
 #   https://github.com/actions/upload-artifact
 #   https://docs.github.com/en/actions/reference/workflows-and-actions/variables
-name: "GHA-01: Make dist and docs tarballs"
+name: "GHA-01: Make dist and docs tarballs, see workflow page for links"
 
 on:
   push:


jimklimov:
After some battle with the machine, I found out in GHA debug logs that despite `permissions` peppered around the YAML, I get:
```

GITHUB_TOKEN Permissions
  Actions: read
  Contents: read
  Issues: read
  Metadata: read
  PullRequests: read
  RepositoryProjects: read
  SecurityEvents: read
```
because per https://stackoverflow.com/a/78444521/4715872
> as it turned out from comments, when PRs come from a public fork, GITHUB_TOKEN defaults to read-only permissions; the workaround here is to use `pull_request_target` instead of `pull_request.` This can make you vulnerable to security issues, as you potentially merge unreviewed third-party code.

This makes sense, and my experiments to use a dedicated `MAKE_DIST_TOKEN` in the context of PRs from forks also failed, so the latest commits convert the workflow to use `pull_request_target` again (presumably this makes PRs use the workflow definition from the target branch, which was absent until this merge), and will see in later PRs how this goes.

PR #3161, opened 2025-11-07
Introduce `make distcheck-completeness` to verify that our dist archives exactly reproduce themselves

jimklimov:
Addresses much of issue #2829, at least in part of reproducibility of our tarballs.


jimklimov:
Another quick merge because workflow scripts must be in target (master) branch.

PR #3169, opened 2025-11-08
CI: finishing touches for make dist workflow and docs

jimklimov:
Checking results of #3167 for #1400


jimklimov:
Yay, master-branch commits now see comments after merge:

image

jimklimov:
Also seen in PR like above:
image

Issue #3138, opened 2025-10-19
[Feature request] Show units in upsc report


Ricc68:
Looking at the upsc report it stands out to me that it shows numbers without units: in general numbers without units are meaningless because they lack key information to understand them.
It would be a significant improvement in my view if the data emitted by upsc is shown with their units.

jimklimov:
Protocol units for standard values are listed in `doc/nut-names.txt`, and to an extent should be reported via protocol (STRING, NUMBER, RANGE etc.) although I think various authors were relaxed about the latter (so most things are opaque strings by default) and there's a backlogged issue about rectifying that. This one is unlikely to be addressed before the basics like that are reliably covered (essentially it is about reporting a secondary type like NUMBER&&SECONDS). Emission of that might be under an optional key, if there isn't one already.

jimklimov:
As a bit of a follow-up and refresher of my memory, `data/cmdvartab` file in NUT sources does contain "layman descriptions" of various standard data point and command names, often including units of measurement in that text. Since many values are proportions, a single word for such unit as "percent" would not cut it anyway without the context "percentage of what (load, charge, capacity, time/age etc.)?"

This file is handled (and data eventually served) by `upsd` via `server/desc.c` logic like `desc_get_var()`, called from `server/netget.c` methods like `get_cmddesc()` or `get_desc()` (for VARs), near the `get_type()` for the aspect noted in earlier post, responding to protocol queries like `DESC`. If an entry is missing in that file (e.g. `unmapped.*` or `experimental.*` variable is encountered), it would not be served. Likewise if the whole file is not installed on the server side.

I see client code `upsrw.c` and `upsset.c` (CGI) calling that, so you can e.g. `upsrw -l myups` to see the descriptions (at least of settable values). On a side note, oddly, the Python module and GUI client don't deem to mention it though.

[...]

PR #2549, opened 2024-07-23
Propagate `IMMUTABLE` flag on socket protocol (driver=>`upsd`); use it in `netset`; sanity-check if (defaulted) `NUMBER` values are not in fact `STRING`s

jimklimov:
Addresses various sides of issue #266 and should help hunt down cases where we do not provide a `NUMBER`/`STRING` flag for the readings.

Currently this PR does not enforce the `STRING:N` type but logs messages to that effect to help with further troubleshooting. One problem was about picking the `aux` value (e.g. allowed string length) at a point far away from code that should have defined the variable type in the first place. Another concern is about propagating the guessed `STRING` type "upwards", so that `netset` can set it safely. Although if it lacks any type flag in the first place, most of those checks are (before and still) bypassed and any value should be settable.

Possibly the better place for this sanity-checking logic of newly posted data vs. flags of a state entry is in core `state.c` code, to transplant pieces of both `netset.c::set_var()` and `netget.c::get_type()` to restrict type usage that should conform to declared flags.

Another aspect is bringing `IMMUTABLE` flag presence into consideration when trying to set values over NUT protocol (it is not exposed in `GET TYPE` of the networked protocol to avoid changing the RFC).

Screenshot with a driver that has `override.battery.charge=10` in `ups.conf`:
````
:; (echo 'GET TYPE dummy battery.charge'; sleep 1; echo 'GET TYPE dummy driver.version'; sleep 1) | ( telnet localhost 3493 ; sleep 1 )
Trying 127.0.0.1...
Connected to localhost.localdomain.
Escape character is '^]'.
TYPE dummy battery.charge NUMBER
TYPE dummy driver.version NUMBER
Connection closed by foreign host.

### Server side
Jul 23 02:19:08 pve nut-server[3327143]: 1448.475949        [D2] Connect from 127.0.0.1
Jul 23 02:19:08 pve nut-server[3327143]: 1448.475982        [D6] Entering check_command: GET
Jul 23 02:19:08 pve nut-server[3327143]: 1448.475984        [D6] check_command: Calling command handler for GET
Jul 23 02:19:08 pve nut-server[3327143]: 1448.475988        [D3] get_type: UPS[dummy] variable battery.charge is IMMUTABLE
Jul 23 02:19:08 pve nut-server[3327143]: 1448.475991        [D3] get_type: assuming UPS[dummy] variable battery.charge is a NUMBER
[...]

Issue #2524, opened 2024-07-09
Add new templates options


mikewhippsHarvest:
I'd like an option to select a template at runtime when running UPSSTATS.CGI. 

Something like a query string element 
    `upsstats.cgi?template=mytemplate.html`, 
or possibly even 
    `upsstats.cgi?template=/path/to/mytemplate.html`

It would be good to allow the use of `json` or `xml` as an extension instead of `html`, as in
    `upsstats.cgi?template=mytemplate.json`, 
 
The specific use case is to emit data in JSON format to import into PRTG for monitoring while leaving the friendlier HTML version alone for those wanting to query NUT directly in a browser.

jimklimov:
This is a really interesting point, would you endeavour to make a PR?

For the time being, especially regarding json, check the third-party projects list, IIRC some emitters were there.

But having some out of the box was backlogged for a while...

mikewhippsHarvest:
By PR do you mean Pull Request? My skill level won't run to that, I'm afraid :( I last wrote C (not even C++) about 1984

I have a working JSON template set up as upsstats-single.html. The Content-type: header is wrong, but PRTG doesn't seem to care.

We're still evaluating so this would be nice in the future. I imagine others would find it useful, though.

neg-dean:
I'm keen to understand the real timeline to this feature being added to the NUT package. Being able to retrieve the data shown on either the table or graph view of the cgi interface as JSON would be a major advantage allowing us to make a PRTG sensor to bring the data & monitoring into PRTG

jimklimov:
Timeline? Rather, milestones: Somebody writes the code, posts a PR, it passes CI and works in fact and does not add regressions and errors / predictable flaws, it gets merged and eventually published as part of a release...

jimklimov:
Some goals of this issue (JSON emission from CGI and `upsc`) were addressed in PRs below:
* #3171 
* #3178

PR #3171, opened 2025-11-09
upsstats: Add JSON output mode via ?json parameter

atanas-vladimirov:
When the ?json (or &json) parameter is added to the URL, upsstats.cgi will now bypass all HTML template processing.

Instead, it returns a "Content-Type: application/json" response containing the full data for the requested UPS(es), including all variables, raw status, and parsed status.

This provides a modern API for external monitoring tools without affecting the existing HTML template functionality.

Here is an example output:

```
{"upses": [
  {
    "host": "attic@localhost",
    "desc": "Masterguard A1000",
    "status_raw": "OL",
    "status_parsed": ["ONLINE"],
    "vars": {
      "ambient.0.temperature": "20.9",
      "ambient.0.temperature.high": "40",
      "ambient.0.temperature.low": "10",
      "battery.charge": "100",
      "battery.packs": "0",
      "battery.runtime": "8622",
      "battery.type": "PbAc",
      "battery.voltage": "40.86",
      "battery.voltage.high": "44.85",
      "battery.voltage.low": "35.88",
      "battery.voltage.nominal": "41.40",
      "device.mfr": "Masterguard",
      "device.model": "A1000",
      "device.part": "6A131",
      "device.serial": "5099",
      "device.type": "ups",
      "device.uptime": "171420",
      "driver.debug": "0",
      "driver.flag.allow_killpower": "0",
      "driver.name": "nutdrv_qx",
      "driver.parameter.pollfreq": "30",
      "driver.parameter.pollinterval": "2",
      "driver.parameter.port": "/dev/cuaU0",
      "driver.parameter.protocol": "masterguard",
      "driver.parameter.synchronous": "auto",
      "driver.state": "quiet",
      "driver.version": "2.8.4",
      "driver.version.data": "Masterguard 0.03",
      "driver.version.internal": "0.45",
      "driver.version.usb": "libusb-1.0.29 (API: 0x0100010B)",
      "input.frequency": "49.9",
      "input.frequency.high": "53.00",
      "input.frequency.low": "47.00",
      "input.frequency.nominal": "50.00",
      "input.voltage": "226.9",
      "input.voltage.high.critical": "276",
      "input.voltage.low.critical": "160",
      "input.voltage.nominal": "230",
      "load.high": "140",
      "output.current": "0.31",
      "output.current.nominal": "4.37",
      "output.voltage": "229.3",
      "output.voltage.nominal": "230",
      "ups.beeper.status": "disabled",
      "ups.delay.shutdown": "60",
      "ups.delay.start": "180",
      "ups.firmware": "10.06",
      "ups.firmware.aux": "03.09",
      "ups.id": "00",
      "ups.load": "7",
      "ups.power": "70",
      "ups.power.nominal": "1000",
      "ups.status": "OL",
      "ups.temperature": "20.9",
      "ups.test.result": "OK",
      "ups.type": "online"
    }
  }
]}
```

### DIFF
--- a/clients/upsstats.c
+++ b/clients/upsstats.c
@@ -1084,7 +1084,7 @@ static void json_escape(const char *in) {
 			case '\r': printf("\\r"); break;
 			case '\t': printf("\\t"); break;
 			default:
-				if (*in >= 0 && (unsigned char)*in < 32) {
+				if ((unsigned char)*in < 32) {
 					/* Print control characters as unicode */
 					printf("\\u%04x", (unsigned int)*in);
 				} else {


jimklimov:
@atanas-vladimirov : Cool, thanks!

I've long wondered if it would make sense to offer two JSON format queries for this, with flat sort of JSON for NUT data points (passing dot-separated names as such), or splitting those names by dots into an object, e.g.:
```
  "battery: {
      "charge": "100",
      "packs": "0",
      "runtime": "8622",
      "type": "PbAc",
      "voltage": {
          "": "40.86",
          "high": "44.85",
          "low": "35.88",
          "nominal": "41.40",
      }
  },
...
```
but now seeing it written out, I understood I have little idea if this split brings any value and/or how to handle items like `battery.voltage` that both have a value themselves and can be seen as a root of a collection (can a zero-length string be used as a key?) -- so overall do not know if this makes sense at all after all :)

In updated man page, I am not sure there's a double-asterisk in asciidoc markup.

In NEWS.adoc, you can add links to this PR and to the issues (#2524?) that it addresses.

Per CI logs, there are a few C code issues:

```
upsstats.c:385:71: error: variable 'newport' may be uninitialized when used here [-Werror,-Wconditional-uninitialized]
  385 |                 if (currups && hostname && (!strcmp(newhost, hostname)) && (port == newport)) {
      |                                                                                     ^~~~~~~
upsstats.c:363:18: note: initialize the variable 'newport' to silence this warning
  363 |         uint16_t        newport;
      |                                ^
      |                                 = 0
upsstats.c:608:24: error: unused variable 'sp' [-Werror,-Wunused-variable]
  608 |         char    stat[SMALLBUF], *sp, *ptr, *last = NULL;
      |                                  ^~
upsstats.c:1175:10: error: mixing declarations and code is incompatible with standards before C99 [-Werror,-Wdeclaration-after-statement]
 1175 |                         char *ptr, *last = NULL;
      |                               ^
3 errors generated.
```

There was also a lingering idea about adding a JSON output mode to `upsc` (or making a new similar client for that). Maybe these new methods (`json_escape`, perhaps even `display_json`) should not be confined as `static` code in `upsstats`, but become a new `common/` source and `include/` header (and a `common/Makefile.am` recipe for intermediate "la" library) that can be linked into both clients eventually.

Are both methods your original code, or is there some attribution that makes sense ("inspired by this URL on Stack Overflow", etc.?)

jimklimov:
Just for cross-linking, it is also worth noting some other issues here (later iterations of this code might grow into these areas):
* #2549 - eventually JSON entries for known `NUMERIC` values (int/double/...) can be reported as such, without quotes; 
* #3138 - somehow passing a JSON schema for data points (as a separate static file, or a generated nearby object?) so consumers can know about some numbers being seconds, others Amps or Volts, etc.

atanas-vladimirov:
Hey Jim,

Thanks for your comment.

> In updated man page, I am not sure there's a double-asterisk in asciidoc markup.

I've build the docs and the html and man pages looks good to me:

image

and the man:

image

Of course, I can change that  if you wish :)

>In NEWS.adoc, you can add links to this PR and to the issues (https://github.com/networkupstools/nut/issues/2524?) that it addresses.

Done.

>Per CI logs, there are a few C code issues:

Thanks. I tried to fix that but it looks that the jobs failed again, not sure if the same errors were hit. By the way, I'm building on OpenBSD and I do not see those issues there, not sure why :).

>Are both methods your original code, or is there some attribution that makes sense ("inspired by this URL on Stack Overflow", etc.?)

To be honest, I used an AI to help me with this particular code :)

About your other thoughts - re json splitting names by dots, I don't think it will "look" good for the end, i.e. it will be more confusing. Of course, the idea of the JSON is not to be read by a user, but a "machine" ....

and

>There was also a lingering idea about adding a JSON output mode to upsc (or making a new similar client for that). Maybe these new methods (json_escape, perhaps even display_json) should not be confined as static code in upsstats, but become a new common/ source and include/ header (and a common/Makefile.am recipe for intermediate "la" library) that can be linked into both clients eventually.

This sounds like a bigger task for me and not sure that I can handle it at the moment, but who knows what will happen in the near future :)

atanas-vladimirov:
I believe I found the last build error and also fixed the double-asterisk in the doc :)

jimklimov:
The spellcheck error (wanted to replace `upses` with `UPSes`) highlighted a flaw with the proposed JSON structure: top-level object should better be named `devices` as NUT also deals with ATS/STS, ePDU, genset, solar-controller etc. machinery, as long as they all have similar concepts and known protocols. You may want to bring up the `device.type` into the higher level as you do for `status_raw`, though I am not sure such duplication (and later possible confusion) is worth it, compared to knowing a `"devices".[X]."vars"."device.type"` should be there and hold a string...

atanas-vladimirov:
Hi,

I have replaced the `upses` with `devices` and left the `device.type` unchanged (for now :).

jimklimov:
Took a little swing at generalizing the code to print out JSON, so it can be used by more clients or even NUT server (given a special query), but indeed it is not trivial and worth a separate PR (that I started but not quickly sure how to finish, if at all). Posted issue #3172 to track the idea at least, and https://github.com/jimklimov/nut/tree/issue-3172 with (incomplete currently) PoC of structural code changes. Although looking at it, and how `upsc.c` handles this as a loop of data-point queries too, maybe the change should only be for clients (not "common" code), maybe even part of `libnutclient` as such?..

Issue #3172, opened 2025-11-10
Optionally show JSON as `upsc` output


jimklimov:
Related to #2524 and #3171, this is a long lingering thought that apparently was not recorded in the issue tracker: add an option to `upsc` (or an alternate client?) to display device(s') information as a JSON structure. Depending on tools that `upsc` output is piped into, JSON nowadays might be friendlier than a key-value list of the (g)olden days.

Loosely also related to #1306 (might as well "dump data" in JSON optionally... maybe add that as a NUT protocol command/option?) and #3138 (add a sub-object with schema for the data points shown).

Part of this quest is converting JSON methods originally added in #3171 into share-able common code that different clients (maybe even server) can use to emit consistently structured documents.

PR #3178, opened 2025-11-11
upsc: Add -j option for JSON output

atanas-vladimirov:
This commit introduces a new `-j` command-line flag to `upsc` to provide machine-readable JSON output for all operating modes.

To support this and share logic with `upsstats`, the JSON string escaping logic has been refactored into a new, reusable module:

- A new helper function `json_print_esc()` now lives in `common/strjson.c`
- `common/Makefile.am` builds this as a new `libcommonstrjson.la` library.
- `clients/Makefile.am` links both `upsc` and `upsstats` against this new library and adds the `common/` include path.
- `upsstats.c` is updated to use the shared `json_print_esc()` function.

Closes: #3172

### DIFF
--- a/clients/upsclient.c
+++ b/clients/upsclient.c
@@ -1769,9 +1769,14 @@ int upscli_splitaddr(const char *buf, char **hostname, uint16_t *port)
 	}
 
 	/* Check that "long" port fits in an "uint16_t" so is in IP range
-	 * (under 65536) */
+	 * (under 65536).
+	 * FIXME: If it is a non-numeric string, try to resolve via
+	 *  "services" naming database, with a C equivalent of:
+	 *  :;  getent services ssh
+	 *      ssh                   22/tcp
+	 */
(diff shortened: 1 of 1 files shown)


jimklimov:
Updated with error-handling, so a `fatalx()` as used before would not leave `stdout` with empty (or worse, half-generated) JSON documents. Here `stderr` from `fatalx()` is still reported, callers can redirect to hide one or another, e.g.:

```
:; ./clients/upsc -j zzz@localhost:zzz
upscli_splitaddr: no port specified after ':' separator
{"error": "invalid UPS definition.\nRequired format: upsname[@hostname[:port]]"}
Error: invalid UPS definition.
Required format: upsname[@hostname[:port]]

:; echo $?
1
```

There may be errors (essentially due to incompatible server on the other side) when a wrong amount of response elements come back, and we `fatalx()` in the middle of a list or object. I tried my best guesswork for adapting the output of `{"error": "some text"}` into these cases and closing the document, but this was not tested to produce any sort of valid JSON.

PR #3180, opened 2025-11-12
upsstats*.html.sample: Enhance model detection in upsstats HTML templates

atanas-vladimirov:
Some devices lack `ups.model`, causing the CGI to incorrectly report "Not supported".

Tested on Masterguard A1000.

### DIFF
--- a/UPGRADING.adoc
+++ b/UPGRADING.adoc
@@ -45,6 +45,11 @@ Changes from 2.8.4 to 2.8.5
   directories (and recipe variable names). This may impact some packaging
   recipes which do not rely on `make install` alone. [#3049]
 
+- Some fixes were applied to HTML templates for NUT CGI clients. It can be
+  useful for NUT deployments being upgraded to compare the files they have
+  installed (and possibly customized) with the `*.html.sample` files delivered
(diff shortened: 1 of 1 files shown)


conf/upsstats-single.html.sample
```diff
@@ -24,7 +24,15 @@
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
 @REFRESH@
-<title>@HOSTDESC@ : @VAR ups.model@ on @HOST@</title>
+<title>
+@HOSTDESC@ :
```
jimklimov, line 28: Did you check if any rendering aspects change in title or table cells that have now added the sort of whitespace around text in their contents? They were like "closing angle bracket"-"evaluated string" and now there's an end of line in-between...

atanas-vladimirov: Yes, I have double-checked how it looks after the change(s) - you can see it here - https://hodor.bsdbg.net/cgi-bin/nut/upsstats.cgi

jimklimov: Thanks for the link, looks good :)

PR #3181, opened 2025-11-12
nutdrv_qx_masterguard: Adds a clear.fault.record instant command

atanas-vladimirov:
Also adds two experimental variables:

- battery.ageing.factor ("Battery ageing factor in promille) and
- battery.calibration.factor (Battery calibration factor in percent).

Tested on A1000 and A2000 units

### DIFF
--- a/drivers/nutdrv_qx_masterguard.c
+++ b/drivers/nutdrv_qx_masterguard.c
@@ -26,7 +26,7 @@
 #include <stddef.h>
 #include "nut_stdint.h"
 
-#define MASTERGUARD_VERSION "Masterguard 0.04"
+#define MASTERGUARD_VERSION "Masterguard 0.05"
 
 /* series (for un-SKIP) */
 static char masterguard_my_series = '?';


jimklimov:
I see you've caught on to the idea of spinning numerous feature branches simultaneously ;)
This was a very bright week so far :)

atanas-vladimirov:
Hey Jim,

Yes, but I still make mistakes and try to learn :)))
By the way I have some doubts about this experimental.* variables thing and meaning. For example, I wonder if those two battery variables should be experimental or not? 
What I saw during my tests is that they are not shown together with all other variables, but are logged by the driver in the logs. Any thoughts on that will be appreciated :)

Edit:

Here is how it looks:

```
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.input_fault_voltage: 225.5
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.battery.ageing.factor: 999
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.battery.calibration.factor: 100
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.fault_1: 0: none
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.fault_2: 0: none
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.fault_3: 0: none
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.fault_4: 0: none
Nov 13 09:43:12 hodor nutdrv_qx[87544]: experimental.fault_5: 0: none
Nov 13 09:43:42 hodor nutdrv_qx[87544]: experimental.input_fault_voltage: 224.5
Nov 13 09:43:42 hodor nutdrv_qx[87544]: experimental.battery.ageing.factor: 999
Nov 13 09:43:42 hodor nutdrv_qx[87544]: experimental.battery.calibration.factor: 100
Nov 13 09:43:43 hodor nutdrv_qx[87544]: experimental.fault_1: 0: none
Nov 13 09:43:43 hodor nutdrv_qx[87544]: experimental.fault_2: 0: none
Nov 13 09:43:43 hodor nutdrv_qx[87544]: experimental.fault_3: 0: none
Nov 13 09:43:43 hodor nutdrv_qx[87544]: experimental.fault_4: 0: none
Nov 13 09:43:43 hodor nutdrv_qx[87544]: experimental.fault_5: 0: none
```

Another question (not related to this PR) is about experimental.input_fault_voltage - I still can't find its meaning and what the UPS is reporting as value is the same as `input.voltage`. So, I wonder if any changes are needed there.
The same questions applies for the fault_X records .....

Because now the drives is filling the logs with unuseful records ....

jimklimov:
Primarily, the `experimental.*` namespace allows to move forward quickly. Changes to `nut-names.txt` should happen via some community discussion and concensus, to make sure the concepts we reflect there make sense for different vendors and devices. For example, much of last/this year revolved around the concept of "ECO" which means absolutely different things to different makers and users, ultimately a greenwashing buzzword with no real single meaning - not a "term" that can be useful in technical or scientific sense.

jimklimov:
I did not quickly find where those logs are coming from, there does not seem to be any special logging handling for `experimental` in `nutdrv_qx_masterguard.c`, `nutdrv_qx.c`, `main.c` or even `dstate.c` I think.

The use of `addvar()` also looked fishy in this subdriver, but apparently happens in many others so could have been a copy-paste with initial submission(s) - this is a way to add command-line options to a driver program. Maybe something in nutdrv-qx unified processing of those somehow, but I do not quickly see any C variables related to those strings, nor value-getters in code loops (which would also be incorrect if they were/are there - tree walks and string-number parsers too expensive to do regularly, especially over same inputs for same outputs).

jimklimov:
I'll be fairly busy today, so probably this is something you would have to track down yourself. Maybe a debugger would help to step through the code (I had good experience with NetBeans IDE for that, some with VSCode - see notes in NUT docs).

atanas-vladimirov:
What I found so far is that the logging comes from `nutdrv_qx` and because of the `QX_FLAG_NONUT` flag:
For example:
```
		/* Loop on all existing values */
		for (envalue = item->info_rw; envalue != NULL && strlen(envalue->value) > 0; envalue++) {

			if (envalue->preprocess
			&&  envalue->preprocess(envalue->value, sizeof(envalue->value))
			) {
				continue;
			}

			/* This item is not available yet in NUT, so publish these data in the logs */
			if (item->qxflags & QX_FLAG_NONUT) {

				snprintfcat(buf, sizeof(buf), " %s", envalue->value);

			/* This item is available in NUT, add its enum to the variable */
			} else {

				dstate_addenum(item->info_type, "%s", envalue->value);

			}

		}

		if (item->qxflags & QX_FLAG_NONUT)
			upslogx(LOG_INFO, "%s, settable values:%s",
				item->info_type,
				strlen(buf) > 0 ? buf : " none");

	}
```

And the question that I can't answer is - should those experimental* variables have to stay the way they are now, or not?
I mean all from the list:

```
experimental.input_fault_voltage: 225.5
experimental.battery.ageing.factor: 999
experimental.battery.calibration.factor: 100
experimental.fault_1: 0: none
experimental.fault_2: 0: none
experimental.fault_3: 0: none
experimental.fault_4: 0: none
experimental.fault_5: 0: none
```

One thing is for sure - they are valid for Masterguard UPSes.

jimklimov:
CC @clepple @aquette @efuss : cheers guys, have you got any insights for questions above, please? :)

efuss:
> Another question (not related to this PR) is about `experimental.input_fault_voltage` - I still can't find its meaning and what the UPS is reporting as value is the same as `input.voltage`.

According to the protocol specification (which I'm presumably not allowed to quote here) it's used to catch a short glitch in input voltage that causes the UPS to switch to battery. The doc also says it will be reset to `input.voltage` once having been read out.

efuss:
> CC @clepple @aquette @efuss : cheers guys, have you got any insights for questions above, please? :)

Not more than what I said in my [original PR](https://github.com/networkupstools/nut/pull/975).
It's the way the UPS logs faults (not utility fail, but things like `inverter mode output short-circuit fault`, `inverter fault` or `battery fault`). I didn't (and still don't) know a better way of reporting these messages.

atanas-vladimirov:
@efuss thanks for your explanation. Now  I wonder:

-  experimental.input_fault_voltage - based on what you said we can just remove/hide/delete this value and don't care about it. The UPS firmware will take care and use it the way it is supposed to be. The end user will be notified by other ways if the input voltage is bad - or I'm wrong?
-  we can leave the experimental.fault_X variables to be logged in the syslog (the way they are now) or maybe improve it by logging them only they have registered a fault. What do you think?
- and how about the battery.ageing.factor and battery.calibration.factor - we can report them together with the other variables instead to the syslog? Alos, we can set them with QX_FLAG_SEMI_STATIC so they are not pulled so frequently, but only on instant command run or setting a variable.

Once again, thank you!

efuss:
> -  experimental.input_fault_voltage - based on what you said we can just remove/hide/delete this value and don't care about it. The UPS firmware will take care and use it the way it is supposed to be. The end user will be notified by other ways if the input voltage is bad - or I'm wrong?

No. According to what the doc says, it's used to catch input voltage glitches that otherwise may go unnoticed becaus they are shorter that the poll interval.

> -  we can leave the experimental.fault_X variables to be logged in the syslog (the way they are now) or maybe improve it by logging them only they have registered a fault. What do you think?

The problem is that the UPSen always report the last five entries (with time stamps). So even if the last fault is years ago, the entries are still non-empty.

> - and how about the battery.ageing.factor and battery.calibration.factor - we can report them together with the other variables instead to the syslog? Alos, we can set them with QX_FLAG_SEMI_STATIC so they are not pulled so frequently, but only on instant command run or setting a variable.

I didnt't yet look at these yet. Maybe next week.

> Once again, thank you!

You're welcome. Glad to see someone else is using the driver.

atanas-vladimirov:
> No. According to what the doc says, it's used to catch input voltage glitches that otherwise may go unnoticed becaus they are shorter that the poll interval.

Okay, I understand the statement about the voltage glitches and that they can occur, but how logging this variable to the syslog will help the end user? Or you mean that if I (as a user) can check the logs for particular incident and based on the value of `input_fault_voltage` might found why my UPS switched on batteries at that moment? 
I'm missing the whole picture...  

> The problem is that the UPSen always report the last five entries (with time stamps). So even if the last fault is years ago, the entries are still non-empty.

With this PR we will have the instant command `clear.fault.record` so I imagine the following workflow:

- the UPS works and suddenly an internal fault is triggered;
- the UPS starts to report `experimental.fault_X` in the logs and the user can notice it;
- the user clears the fault records and the `experimental.fault_X` stopes to be logged until the next time and event occurs. 

This way we won't filling the logs with "empty" fault records.

> I didnt't yet look at these yet. Maybe next week.

Okay, great! Have a lovely weekend!

> You're welcome. Glad to see someone else is using the driver.

I'm glad that you made this driver! The Masterguard devices are really good even at that age and I like them very much :)

efuss:
> Okay, I understand the statement about the voltage glitches and that they can occur, but how logging this variable to the syslog will help the end user?

The fact that it gets syslog-ed is just because it's an experimental variable name. If it wasn't, you would just look it up in your favourite monitoring software (like collectd/InfluxDB/GGrafana).

> This way we won't filling the logs with "empty" fault records.

Yes, that would work, but I would regard it as a cludge.

I'm not a nut expert, I just wrote two masterguard drivers because I needed them; do we have or need a more general ups fault logging facility? Are Masterguard the only ones that report such failures?
I could imagine a general nut facility where the driver could notify an upper layer of these fault records (in this case, maybe including time stamps relative to some ups internal epoch) and the upper layer would manage the rest.

atanas-vladimirov:
Thanks!

I'm also new to NUT and do not understand many things (yet :))). Also, I do not have answers to the questions you asked.
So, Jim (@jimklimov) any input on how we can move forward (and what we can do) will be greatly appreciated :)

jimklimov:
Well, I think there are some UPSes that can report some faults. NUT has a similar facility with alarms (but they are mostly for extreme issues that need attention, not "one of five alerts about something during the past year" as seems to be the case with these ones here). I am not sure there is a common cross-vendor concept around this, but it may be worth asking on the mailing lists and waiting a few weeks/months for some consensus.

As for filling the logs, those `upslogx(LOG_<LEVEL>, ...)` about the experimental values can be replaced by `upsdebugx(<2-3>, ...)` and only see them in sufficiently verbose runs. Or reduce the syslog facility associated with the message to something that is not usually recorded.

@efuss : maybe you in fact can shine some light at the use of `addvar()` there (hm, I just found they are not mentioned under `docs/*`) - usually this method allows NUT drivers to accept such and such strings as command-line or configuration-file options (flags or values), but they become useful as long as there is a  `getval()` somewhere to parse them (typically during init and saved into a C number, boolean or string variable to avoid subsequent getval's). In `drivers/nutdrv_qx_masterguard.c` I only see one `getval("slave_address")` as part of `masterguard_claim()` method, so the rest of the dozen or so strings seem to have no effect if used on command line? Unless I missed something centralized about `experimental.*` mapping handling?

efuss:
> @efuss : maybe you in fact can shine some light at the use of `addvar()` there

Sorry, no. It's five years since I wrote the driver and even digging through our internal SVN repository doesn't help (the `addvar()` lines are present in the first commit).

Maybe (I thought that) these are necessary for using custom variables?

The only thing I remember is that some of them are really special (and set in `masterguard_claim()`):

- `series` is needed to find out which commands are avalilable
- `slave_address` needs to be filled into several commands
- `output_voltages` is needed to know which values `output.voltage.nominal` may be set to
- `fault_n`, which are log entries.

jimklimov:
@atanas-vladimirov : I am too often caught up in scope creep, such as "what to do about experimentals and their logging" here :)

I think your PR as originally proposed is complete, adding the command and experimental values in a way similar to precedent, so can be merged. The rest of issues can be pursued and evaluated on their merits separately (such as figuring out if `addvar` is or is not an overkill in this and other subdrivers).

@efuss : thanks for your insights, and original contribution.

jimklimov:
> I could imagine a general nut facility where the driver could notify an upper layer of these fault records (in this case, maybe including time stamps relative to some ups internal epoch) and the upper layer would manage the rest.

Maybe a holistic `device.faults.NUM.{desc,timestamp,...?}` among NUT common variables could be it? And maybe some `device.faults.{count,slots,newest,since-last}` etc. to help make sense of them in one look?

atanas-vladimirov:
@jimklimov thanks!

jimklimov:
It seems that the `QX_FLAG_NONUT` should not be set for values we do now have a place for (e.g. `experimental.*` namespace) so there is little to warn about, at least via `upslogx` (maybe quieter `upsdebugx` as noted above?)

Issue #3194, opened 2025-11-21
nut-driver-enumerator.sh: Systemd "nut-driver@.service" script fails when NUT_DRIVER_ENUMERATOR_CONF is set


avolkov-1221:
**Description:**

If "NUT_DRIVER_ENUMERATOR_CONF" variable in **`nut-driver-enumerator.sh`** is not empty, an informational message is printed about sourcing the config file. However the systemd **`nut-driver@.service`** script expects only the device ID output and fails when extra messages appear.

**Steps to Reproduce:**

- Set NUT_DRIVER_ENUMERATOR_CONF.
- Start the systemd nut-driver.target.

**Expected:**
The **nut-driver.target** should start without issues.

**Actual:**
**`nut-driver@.service`**  fails due to the extra informational message from **`nut-driver-enumerator.sh`**.

Suggested Fix:
Suppress or redirect informational messages when the **`nut-driver-enumerator.sh`**  script runs under systemd.

jimklimov:
Good point, but I think a better fix would be to direct the message to `stderr`?

avolkov-1221:
> Good point, but I think a better fix would be to direct the message to `stderr`?

Then it might as well go into stddbg, if such a thing existed :).

If we keep this message, we’d also have to redirect the rest of the messages there as well; otherwise some would go to stdout while others end up in stderr. In other words, that would mean we’d need to implement a logging system.

IMHO, for now it’s better to remove it, and if someone really needs it, it can always be printed directly from the sourced configuration file itself.

jimklimov:
I think most other messages are on stderr already, and this is consistent with other tools; I am in favor of having the debugs available (even if indeed hidable in upsdebugx style).

PR #3195, opened 2025-11-21
scripts/upsdrvsvcctl/nut-driver-enumerator.sh.in: remove "Sourcing conf file.." message [#3194]

avolkov-1221:
If "NUT_DRIVER_ENUMERATOR_CONF" variable in `nut-driver-enumerator.sh` is not empty, an informational message is printed about sourcing the config file. However the systemd `nut-driver@.service` script expects only the device ID output and fails when extra messages appear.


jimklimov:
Supplanted by #3197

PR #2476, opened 2024-06-15
common/str.c, include/str.h: introduce `str_concat()`

jimklimov:
Address a small part of #1316
Closes: #2343

Imported from FTY/DMF branch, aggregated from commits:
* d5237b61f1112d6dd01f299b2f836c0974798e5f
* 27cbd4d3e2cbac93c6e47a772f4e91f84749b9c6

Currently has no consumers in upstreamed code, though.

PR #3200, opened 2025-11-25
FTY remerge from master as of 20251125

jimklimov:
Catch up with development in upstream NUT v2.8.5 (to be) cycle.


jimklimov:
Now the remaining difference between FTY and master branches seems to be all about the DMF feature (and maybe a few miniscule differences like #2476).

PR #3206, opened 2025-11-30
Meanwell NTU-Series experimental driver

jhite1234:
This pull request adds a new NUT driver named meanwell_ntu which provides support for the Mean Well NTU series hybrid inverter and UPS units. Supported models include the NTU-1200, NTU-1700, NTU-2200, and NTU-3200, along with all voltage variants such as -112, -124, -148, -212, -224, and -248. Communication is handled over RS-232 or USB-serial at 9600 baud with 8N1 configuration and carriage-return terminated commands. Development and testing were performed using a Mean Well NTU-1200-124.

The driver implements support for the NTU Q and I commands. The Q command provides real-time device data including battery voltage, battery charge percentage, UPS temperature, output voltage, input voltage when utility is present, output frequency in battery mode, and real load percentage derived from the PPP field. It also reads a 19-bit status field. The I command provides device metadata including manufacturer, model, serial number, firmware revision, and battery voltage thresholds.

The NTU status bits are exported as info variables under the experimental namespace, following the conventions documented in nut-names.txt. Examples include experimental.inverter.mode, experimental.bypass.mode, experimental.utility.present, experimental.overtemp, experimental.inverter.fault, and others. These bits are also used to populate standard NUT status indicators such as OB, OL, BYPASS, LB, OVER, CHRG, and DISCHRG.

The driver follows the naming rules for ups.mode described in nut-names.txt. Since the NTU is a line-interactive design with the ability to enter bypass mode, ups.mode is set to bypass when the bypass flag is active and is set to line-interactive in all other cases. Whether the unit is currently using the inverter or utility power is indicated through ups.status rather than ups.mode.

Additional behavioral handling is included. Input voltage is only reported when the utility.present bit is set so that stale values are not displayed. When in bypass mode, some NTU units report zero for output voltage, so the driver mirrors the input voltage to output voltage to ensure accurate reporting. Load percentage is taken from the PPP field, with fallback to the coarse digital load step field if needed.

The driver supports two instant commands, load.off and load.on, which map to the device’s shutdown and startup control commands.

The following files were created or updated: drivers/meanwell_ntu.c, drivers/Makefile.am, docs/man/meanwell_ntu.txt, and docs/man/Makefile.am.

The driver has been tested on real hardware across battery operation, utility operation, bypass mode, and transitions between these states. All expected status indicators, voltages, temperatures, and load values were confirmed. Mode transitions were verified through repeated cycling of the AC input to the inverter.

All new info variables that do not correspond to standardized names have been placed under the experimental namespace. The documentation builds cleanly. The spellcheck target did not run on my system because aspell was not installed, but the documentation files themselves appear correct.

This driver is ready for review. I am willing to make adjustments to naming, documentation, standardization of fields, additional device listings, or protocol detail if requested by reviewers.

MAINTAINER UPDATE: Data sheet about numerous models, and lots of tech details from hardware mounting to protocol, are at https://www.meanwell.com/Upload/PDF/NTS,NTU-E.pdf - I'll add a copy to NUT Protocols library.

### DIFF
--- a/.github/workflows/01-make-dist.yml
+++ b/.github/workflows/01-make-dist.yml
@@ -50,6 +50,15 @@ jobs:
       with:
         fetch-depth: 0
         fetch-tags: true
+        # NOTE: pull_request_target protects us by using the workflow definition
+        # from target branch (so trusted operations like artifact management can
+        # be used), but it also uses the code from that branch too, so fix it back.
+        # Maybe use "github.event.pull_request.merge_commit_sha" instead?.. though
+        # it may be null if the PR is not mergeable or is still checked for that,
+        # or may return the commit for PREVIOUS pushed iteration's merge to target.
+        # See more ideas in https://github.com/actions/checkout/issues/518 :
+        ref: "${{ github.event.pull_request.number > 0 && github.event.pull_request.head.sha || github.ref_name }}"
+        persist-credentials: false
 
     # https://github.com/marketplace/actions/substitute-string
     # See also examples in https://github.com/dhimmel/dump-actions-context/


jhite1234:
The Jenkins CI failure appears to be a Groovy syntax error (“expecting ')', found '}'” at line 457) in the Jenkins shared library, not related to this PR. I did not modify any Jenkins or Groovy files. Please let me know if there is anything you’d like me to adjust on the NUT side.

jimklimov:
Oh, thanks

jimklimov:
Hello, and thanks for the contribution. I wanted to clarify a couple of things: 
* is this work done/contributed on behalf of the vendor (e.g. should `acknowledgements.txt` be updated?)
* the protocol examples look a lot like Megatec Q<x> although the main query is without a dialect number like `Q1`, `Q2` or `Q6` used in most cases. Can you please check if nutdrv_qx does not by chance already support the device somehow, and if your tailored support would not benefit from being its subdriver (with mappings and all) instead of a new driver? Not to diminish your work, just quickly checking if we do need a new entity?..

jimklimov:
There are also some asciidoc nuances for the man page:
* better use `linkman` macro for NUT programs and config files
* there's a list format with double colons (see other pages to be consistent) commonly used for configurable options and here could be for examples of reported data points

jhite1234:
I'm not with the vendor. Just an end user trying to get my DIY UPS setup working with NUT.

I tried/checked several drivers for compatibility but nothing seemed to quite fit. The nutdrv_qx/Megateq Q1 Q2 structure doesn't quite fit the Meanwell Q response. The I and C commands aren't part of it at all. I might be mistaken, but I think shoehorning it into that driver wouldn't be the right move. I'm not a developer or anything though so I don't really know if that's right. 

I will see about making corrections to the man page formatting. 

If there's anything else I can to do make things be more in line with NUT let me know.

jimklimov:
> I tried/checked several drivers for compatibility but nothing seemed to quite fit. The nutdrv_qx/Megatec Q1 Q2 structure doesn't quite fit the Meanwell Q response. The I and C commands aren't part of it at all. I might be mistaken, but I think shoehorning it into that driver wouldn't be the right move.

Okay, thanks for checking. Then a separate driver it is :)

jhite1234:
As far as I can tell, the CI errors aren't stemming from the new driver. Unless something I did broke something else.  I think I'm done tinkering with this version for now. If there's anything else I need to do just let me know.

jimklimov:
Hm, on CI that's an odd one:
```
FAILED man-source sanity check for: /tmp/jenkins-swarm/jenkins-nutci/nut_nut_PR-3206@2/docs/man/meanwell_ntu.txt
/tmp/jenkins-swarm/jenkins-nutci/nut_nut_PR-3206@2/docs/man/meanwell_ntu.txt:	data
```

The recipe uses `file` to check that it is looking at "ASCII text" sources or something similar (and not at dangling symlinks, generated HTML, etc.) and I'm not sure what happens here - but it is constrained to, and repetitive on, illumos/Solarish tooling. First idea was that it could be about UTF8 BOM chars, etc. but those seem to not be there.

This is reproducible (e.g. on OpenIndiana) in command-line too:
```
$ file docs/man/m*.txt
docs/man/macosx-ups.txt:        ascii text
docs/man/masterguard.txt:       English text
docs/man/meanwell_ntu.txt:      data              <<< THIS
docs/man/metasys.txt:   ascii text
docs/man/mge-shut.txt:  English text
docs/man/mge-utalk.txt: ascii text
docs/man/microdowell.txt:       English text
docs/man/microsol-apc.txt:      ascii text
```

Will check deeper...

jimklimov:
Ach, close enough: UTF quote chars slipped in, and the file has no UTF BOM, so these multi-byte sequences outside ASCII range make it a non-text for the tool, see `vendor's` in the hexdump:
```
...
000001C8 55 50 53 20 x 75 6E 69 74 x 73 20 74 68 x 61 74 20 69  UPS units that i
000001D8 6D 70 6C 65 x 6D 65 6E 74 x 20 74 68 65 x 20 76 65 6E  mplement the ven
000001E8 64 6F 72 E2 x 80 99 73 20 x 41 53 43 49 x 49 20 73 65  dor▒..s ASCII se
000001F8 72 69 61 6C x 20 70 72 6F x 74 6F 63 6F x 6C 2E 0A 0A  rial protocol...
...
```

jimklimov:
I'll take care of the formatting etc. now, gimme a minute :)

jhite1234:
Yeah, I would not have found that. Thank you.

Issue #3055, opened 2025-08-08
CI: Revise code that deals with NUT versions to compare them numerically component-by-component


jimklimov:
> ...scripts expecting more or less alphanumeric comparisons (so single-digit version components) do not help either... 

 _Originally posted by @jimklimov in [#3053](https://github.com/networkupstools/nut/issues/3053#issuecomment-3167641819)_

Cover shell, perl, python, sed/grep/awk, POSIX make, maybe more. Check also nut-website.

This is more or less a required prerequisite for conversion to "MAJOR.MINOR" releases aligned with standard SEMVER usage, with ".PATCH" part left for hot-fixes over existing releases (so eventually moving away from "release is a snapshot of master, stale and obsolete after the next commit lands" stance).

Note that NUT SEMVER deals with up to 5 dot-separated numeric components (and then some suffix parts), see https://networkupstools.org/docs/developer-guide.chunked/versioning.html for more details.

jimklimov:
With commits listed above (PR #3213, also https://github.com/networkupstools/nut-website/commit/a99c10dc630e7d0cdb3490a4772697024c648b33), the shell scripting part can now expand leading semver components into fixed-width zero-padded numbers, so they can be sorted and compared alphanumerically. 

For the nut-website part, this can already be used for historic website building (e.g. optional rendering of files depends on comparison of requested version with a version known to have first delivered such files, so whether the one we are building now is "same or newer" than that).

jimklimov:
TODO:
* [x] Use in rendering of the NUT DDL, see e.g. https://networkupstools.org/ddl/APC/Smart-UPS_1000.html where multiple reports are grouped by NUT versions, newer to older
* [ ] find where else NUT version sorting can be used, to cover those places; revise release processes, CI scripts, etc.

jimklimov:
nut-website DDL generator revised with the new helpers logic

jimklimov:
Further investigation:
* https://github.com/networkupstools/nut/blob/master/m4/ax_compare_version.m4 includes the well-known Autoconf Archive macro `AX_COMPARE_VERSION(A, OP, B, …)` for comparing version strings. Its documented behavior is exactly about comparing version strings across `eq/ne/le/ge/lt/gt` and also `eqN/neN` prefix comparisons. => https://www.gnu.org/software/autoconf-archive/ax_compare_version.html
[...]

PR #3213, opened 2025-12-11
Fix and update `gitlog2versions.sh` and `docs/nut-versioning.adoc`; add tests for the script

jimklimov:
Addresses part of issue #3055 to help ensure multi-digit NUT versions do not cause surprises.

Sort of follows up from #3099 to more portably use `SED` as well.

Utilized on nut-website by commit https://github.com/networkupstools/nut-website/commit/a99c10dc630e7d0cdb3490a4772697024c648b33

### DIFF
--- a/tools/gitlog2version.sh
+++ b/tools/gitlog2version.sh
@@ -473,7 +473,9 @@ filter_extra_width() {
                 ".") echo "." ;;
                 0*|1*|2*|3*|4*|5*|6*|7*|8*|9*)
                     if $NUMERIC && [ x = x"`echo \"$LINE\" | sed 's,[0-9],,g'`" ] ; then
-                        printf '%0.*d' "${NUT_VERSION_EXTRA_WIDTH}" "${LINE}"
+                        # NOTE: Not all shells have printf '%0.*d' (variable width)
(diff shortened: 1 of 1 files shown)


jimklimov:
For reference, toolkits on Solaris 8 are too old to run both the test script itself, and `gitlog2version.sh`, and the NDE script. A bit can be enhanced by using third-party `bash` there, although not all test cases pass even then.

These updates for `gitlog2version.sh` work for `bash` running both the test suite and the script, but with `bash` running the suite and system default `ksh` running `gitlog2version.sh` the expanded numbers fail (`printf` functionality lacking).

Side note, the NDE script (not really of interest to the platform fails at least because `${#VANAME}` extensively used to count the characters there (e.g. emptiness of a string without tossing kilobytes around the `test` program/builtin) is not supported by that shell. File reading tricks like `UPSCONF_DATA=$(upslist_normalizeFile_filter < $UPSCONF)` are also not supported.

Another side note, zero-padded numbers should not be used in shell math or with `expr`, because they are assumed to be octal. So this new format is purely for alphanumeric comparisons of versions for different releases/iterations as strings.

Issue #3234, opened 2025-12-24
nut-scanner IP address reporting is bogus


jimklimov:
Scanned a VM network recently, and reports saw some IP addresses repeated and others empty. Is something overlapped in memory with pthreads in use? Is it poking all the correct IPs and only reporting them wrong (e.g. some buffer reuse in `inet_ntop()` or somesuch)?..

Reproduced with a native NUT for Windows build, and with a Linux build running in WSL:

```
:; LD_LIBRARY_PATH=`pwd`/clients/.libs:`pwd`/tools/nut-scanner/.libs:$LD_LIBRARY_PATH ./tools/nut-scanner/nut-scanner -O -m 10.10.10.0/24
upscli_tryconnect: Connection to host timed out: '10.10.10.12'
upscli_tryconnect: Connection to host timed out: '10.10.10.12'
upscli_tryconnect: Connection to host timed out: '10.10.10.3'
upscli_tryconnect: Connection to host timed out: '10.10.10.17'
upscli_tryconnect: Connection to host timed out: '10.10.10.20'
upscli_tryconnect: Connection to host timed out: '10.10.10.22'
upscli_tryconnect: Connection to host timed out: '10.10.10.0'
upscli_tryconnect: Connection to host timed out: '10.10.10.7'
upscli_tryconnect: Connection to host timed out: '10.10.10.8'
upscli_tryconnect: Connection to host timed out: '10.10.10.19'
upscli_tryconnect: Connection to host timed out: '10.10.10.23'
upscli_tryconnect: Connection to host timed out: '10.10.10.24'
upscli_tryconnect: Connection to host timed out: '10.10.10.32'
upscli_tryconnect: Connection to host timed out: '10.10.10.40'
upscli_tryconnect: Connection to host timed out: ''
upscli_tryconnect: Connection to host timed out: '10.10.10.18'
upscli_tryconnect: Connection to host timed out: '10.10.10.29'
upscli_tryconnect: Connection to host timed out: '10.10.10.30'
upscli_tryconnect: Connection to host timed out: '10.10.10.11'
...
```

jimklimov:
The problem was indeed "cosmetic", in conversions to strings. Just in case it pops up worse, fixed in PR #3235

Issue #3207, opened 2025-12-01
upsdrvctl.exe not starting


aarnold-sd68:
I'm not sure if I've done something wrong but when using the package built by appveyor on it tried to find upsdrvctl.exe in the bin directory.  But the files are in the ../sbin directory.  this can be fixed by copying the files over, but I'm not sure that the correct way no am I sure what this might break.  It seems to work, for what I need. Is there something I'm missing?

jimklimov:
Sorry, seems like our bug, slipped in investigation. Does description for PR #3065 look like the same thing (I think that particular solution is wrong though, that's why it slipped, but the issue may be still valid)?

aarnold-sd68:
That does sound like the same problem yes.  Is my workaround going to cause any issues, at least for now?

jimklimov:
I don't think it should, just not as pretty to use as everything working out of the box.

jimklimov:
PR #3224 with hopefully a fix for this situation was merged to master. Testing that the result works is still desirable :)

jimklimov:
Hm, got stuck for me showing help (in `confpath()` getter), checking....

jimklimov:
@aarnold-sd68  @twoeater : BTW, how are we doing at this issue, can it be closed with current changes in master (recent PRs merged)?

aarnold-sd68:
As far as all my testing has shown this issue(s) have been solved.  The stuff in sbin seems to work when left in sbin :D.

twoeater:
@jimklimov 

In the 2.8.4 Windows release, the driver path still references the sbin directory.
but the driver files are located in bin.

ups.conf
```
[myups]
   driver = genericups
   port = 192.168.0.6
   desc = "Something descriptive"
```

<img width="1121" height="615" alt="Image" src="https://github.com/user-attachments/assets/b35100f0-3279-49d7-9c93-cc62556a5c47" />

```shell
C:\Users\Me\Downloads\NUT-for-Windows-x86_64-RELEASE-2.8.4\NUT-for-Windows-x86_64-SNAPSHOT-2.8.4.443-master\mingw64\sbin>upsdrvctl.exe start
Network UPS Tools upsdrvctl.exe - UPS driver controller 2.8.4 release
[...]

Issue #3219, opened 2025-12-16
Windows and NUT CGI Web Pages


aarnold-sd68:
I'm taking a look at the web interface for Nut, Nut-cgi I believe it's called, but I'm not having any luck :(.  Is it even possible to get it working on windows?  I can get the main page working, Network UPS Tools with the statistics and settings links.  But clicking the links gets me a page not found error.  If I run https://serverip/cgi-bin/upsstats.cgi.exe I get this error :Error: can't open template file (upsstats.html) upsstats: Can't open 
\\?\C:\mingw64\cgi-bin\..\etc/upsstats.html: Invalid argument.  Which I'm assuming has something to do with it :).  The file does exist in etc directory but that is a weird way of displaying the path.  If I type in the path to the upstats.html file I do get the page all be it with no data, which I believe makes sense as it needs to come from the upstats.cgi.exe.    If I add ?json to the end I do get some data.  

oh and I also uncommented the "I_HAVE_SECURED_MY_CGI_DIRECTORY" in upsset.conf

jimklimov:
yeah, the path looks weird; there probably should be a separator between `cgi-bin/..` at least

Maybe there's a need to tweak code and provide a PR to address this... I am not aware of anyone running this yet, and the Windows port has certain known incomplete bits, so there may well be unknown ones too.

BTW it would be helpful for a repro case to post how you managed to get this far :)

aarnold-sd68:
In IIS I set the root to install for default website to c:\mingw64\.  This then gives me access to to cgi-bin and etc and all the other directories.  I had to enable cgi-exe as this is disabled by default in iis.  I updated the hosts.conf,upsset.conf and renamed upsstats.html.sample and upsstates-single.html.sample to not include the sample.  My URL is https://servername/cgi-bin/upsstats.cgi.exe?json which gives me some data, but doesn't give me the pretty template view.  if I go to https://servername/html I get the start page with statistics and settings but the links don't work when you click them :(.  https://servername/cgi-bin/upsstats.cgi.exe gives me the error : Error: can't open template file (upsstats.html) upsstats: Can't open \\?\c:\mingw64\cgi-bin\..\etc/upsstats.html: Invalid argument.  Loading : https://servername/cgi-bin/upsset.cgi.exe gives me this error Error: Can't open upsset.conf to verify security settings.. I've enabled Basic and windows auth and disabled anonymous, this was done to secure the server a little bit, might do more later.  

[...]

PR #3235, opened 2025-12-24
Follow-up fixes for NUT for Windows launcher, and parallel nut-scanner logging

jimklimov:
Closes: #3234
Follow-up to: #3207 et al


jimklimov:
Largely commented on in https://github.com/networkupstools/nut/issues/3219#issuecomment-3693471641 - lots of future work seen from experimental results, but existing issues/showstoppers are resolved.

Issue #3116, opened 2025-10-09
CyberPower CP1500AVRLCD3 doesn't reliably work unless NUT_DEBUG_LEVEL set


kenlasko:
As is apparently quite common with CyberPower UPSs, I have had a hell of a time trying to get my CyberPower UPS to reliably stay connected to NUT in a homelab Kubernetes environment.  No matter what I did, the UPS would go offline within a few minutes with repeated `DATA STALE` errors.  I tried various iterations of modifying:
- maxretry
- pollfreq
- pollinterval
- pollonly = "enabled"
- MAXAGE
- DEADTIME

Nothing seemed to work. I then tried to set debug logging through the envvar `NUT_DEBUG_LEVEL=5`.  As soon as I set that, UPS communication worked flawlessly for the next 24h. I've since dropped the debug level to 1 and it still works. If I remove the `NUT_DEBUG_LEVEL` env var, the UPS reverts to its previous behaviour by going offline within a few minutes.

Curious as to why the debug setting seems to be the magic trick to get my CyberPower UPS working?

My environment is the following:
- Kubernetes 1.34.1 running on Talos Linux 1.11.2
- Using `instantlinux/nut-upsd:2.8.3-r2` container
- UPS connected to Beelink mini-PC along with another Belkin UPS that has always worked fine with this implementation.

ups.conf
```
maxretry = 10
[belkin]
    driver = usbhid-ups
    port = auto
    vendorid = 050D
    productid = 0900
    desc = "Belkin UPS"
    pollinterval = 5

[cyberpower]
    driver = usbhid-ups
    port = auto
    vendorid = 0764
    productid = 0601
    desc="CyberPower CP1500AVRLCD3"
    pollfreq = 30
    pollinterval = 30
    pollonly = "enabled"
```

upsd.conf
```
LISTEN 0.0.0.0 3493
MAXAGE 30
```

upsmon.conf
```
MONITOR belkin@localhost 1 monitor upsmon master
MONITOR cyberpower@localhost 1 monitor upsmon master
SHUTDOWNCMD "echo 'System shutdown not configured!'"
DEADTIME 30
```

jimklimov:
There was a bug (though years before 2.8.3) hit after code refactoring, where a debug message printing call was also evaluating the variable somehow (incremented it?)

So after debug messaging code was fixed to do nothing if the verbosity was not high enough, that useful code got skipped.

I really wonder if anything similar might survive unnoticed this long?..

[...]

PR #3247, opened 2026-01-07
usbhid-ups: fix cyberpower issues resolved when debug logging is enabled

com6056:
Fixes https://github.com/networkupstools/nut/issues/3116 which is caused by a slight difference in behavior when `NUT_DEBUG_LEVEL` is set, see full explanation here: https://github.com/networkupstools/nut/issues/3116#issuecomment-3717646428

When running with these changes, I no longer see the disconnects after a couple minutes with my `CP1500PFCLCDa` 🎉 

## General points

  comments or text documents.

  case.

  or its equivalent.

  concept.

  the changed codebase.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -204,6 +204,8 @@ https://github.com/networkupstools/nut/milestone/12
      ignore any useful reports, and that we successfully use reasonably many
      of the existing mappings. Suggest how user can help improve the driver
      if too few data points were seen. [#3082, #3095]
+   * `openups-hid` had `nobattery` definition inverted, causing alarms -- now
+     fixed to use the same mapping helper as other subdrivers. [issue #3246]
 
  - `upsd` data server updates:
    * Sometimes "Data for UPS [X] is stale" and "UPS [X] data is no longer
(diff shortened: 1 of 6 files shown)


jimklimov:
Interesting catch, just shy of two decades old. I see this PR also adds `explorehidall` flag to `usbhid-ups` - can you please add documentation to the man page (and a mention in `NEWS.adoc`) about it please?

Also, I think its default value should be just zero, a `VAR_FLAG` mention sets it to `1` but otherwise there is no way to set it to `0`. So it is not un-settable on CPS devices where this could cause a problem, and it remains `-1` (so `true` in `if` clauses) by default on others. Maybe better redefine it as a `VAR_VALUE` and parse as zero/nonzero.

Also not sure if `fflush(stdout)` is needed there (at least nowadays, `upsdebugx` outputs are on `stderr`), and if either flush should be fenced by `if (should_output_debug) {...}` or remain unconditional.

jimklimov:
Regarding the disconnections you saw, are there any logs from around then - why did the driver decide to reconnect? Maybe the "correct" solution is to ensure the `cps-hid.c` subdriver polls certain data points that is currently lacks in the mapping table?

com6056:
Sorry for the delay, this actually didn't solve the issue. I've done some further digging and think I've found another potential solution though 🤞 Will push up the new changes when I can!
UPDATE: See #3259

jimklimov:
Aha... and here I was, bringing this PR into better shape :)

For the next one, please post from a dedicated feature branch (not an existing one like `master`) to simplify resync later on.

PR #3256, opened 2026-01-12
Python helper scripts: fix deprecation warning

systemcrash:
DeprecationWarning: codecs.open() is deprecated. Use open() instead.

### DIFF
--- a/scripts/augeas/gen-nutupsconf-aug.py.in
+++ b/scripts/augeas/gen-nutupsconf-aug.py.in
@@ -25,7 +25,6 @@ from __future__ import print_function
 import sys
 import re
 import glob
-import codecs
 
 # Return a sorted list of unique entries, based on the input 'list'
 def sortUnique(list):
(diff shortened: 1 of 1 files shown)


jimklimov:
Thanks, this seems like it should work with pythons we care about in legacy OS practice (2.6+ and 3.x), but I'll let the multiplatform CI say if this change is OK ir needs if/else try/except etc. sort of handling.

jimklimov:
So... more code is needed to make sure which python version we are running on (or just try/except for missing class properties or method signatures):
```
03:58:43  Regenerating Augeas ups.conf lens with 'python'...
03:58:43  Traceback (most recent call last):
03:58:43    File "./gen-nutupsconf-aug.py.in", line 82, in <module>
03:58:43      fd = open(filename, encoding='utf-8')
03:58:43  TypeError: 'encoding' is an invalid keyword argument for this function
03:58:43  + exit
script returned exit code 1
```

jimklimov:
Will be extended and addressed by #3300, thanks.

PR #3259, opened 2026-01-13
usbhid-ups: improve handling of transient LIBUSB_ERROR_IO failures

### LINKED ISSUE #3116: CyberPower CP1500AVRLCD3 doesn't reliably work unless NUT_DEBUG_LEVEL set

com6056:
Some devices (e.g., CyberPower CP1500PFCLCD) have firmware bugs that cause random I/O errors on specific HID reports during normal polling. Rather than triggering expensive reconnection attempts that can fail in daemon mode, skip the failing report and continue with remaining polls.

Add safety check to detect true device disconnection: if all polls fail during an update cycle (items_succeeded == 0), trigger reconnect as before.

This improves stability especially in daemon mode while still detecting real disconnections via other error codes or complete poll failure.

Fixes #3116

---

More in-depth details from my investigation in case it is helpful:

## Problem Description

CyberPower UPS devices (tested with CP1500PFCLCD, ProductID 0x0601) experience "Data stale" errors when running in daemon mode, but work perfectly when debug mode is enabled (`-D` flag or `NUT_DEBUG_LEVEL=1`).

## Root Cause

The device has firmware bugs that cause **transient** `LIBUSB_ERROR_IO` failures on certain HID reports during normal polling operations. Reports like:
- `0x1a` (input.sensitivity) 
- `0x19` (ups.realpower)
- Others during "Full update" cycles (every 12th poll with default `pollfreq`)

These reports **exist and work most of the time**, but randomly fail with I/O errors.

### Current Behavior (Broken)

When `LIBUSB_ERROR_IO` occurs during polling:
1. Driver immediately enters `reconnect.trying` state
2. Calls `reconnect_ups()` to close and reopen USB device
3. **In daemon/background mode**: USB reconnection fails (due to process being daemonized with `setsid()`, closed file descriptors, etc.)
4. Driver sets `dstate_datastale()` → "Data stale" error to upsd
5. Loop repeats on next poll → continuous reconnect failures

### Why Debug Mode "Worked"

The `-D` flag sets `foreground = 1`, preventing the driver from daemonizing:
- Process doesn't fork
- No `setsid()` call
- Original file descriptors remain open
- **USB reconnection succeeds in this environment**

So debug mode didn't fix the firmware bug—it just made the reconnection workaround actually work. But reconnecting on every transient error is expensive and unnecessary.

## Solution

**Skip transient errors instead of reconnecting:**

1. When `LIBUSB_ERROR_IO` occurs, log it and `continue` to next poll item instead of triggering reconnect
2. Track successful polls with `items_succeeded` counter
3. **Safety mechanism**: If zero items succeed during an UPDATE cycle, then trigger reconnect (true device failure)

### Why This Works

- **Transient errors on SOME reports** = firmware bug → skip and continue with remaining polls
- **Errors on ALL reports** = true disconnection → reconnect
- Most polls succeed despite occasional failures → driver stays stable
- True disconnections still detected via:
  - Other error codes (`LIBUSB_ERROR_NO_DEVICE`, `LIBUSB_ERROR_ACCESS`, etc.)
  - All polls failing (safety check)
  - upsd's `MAXAGE` timeout

## Testing

With this fix and `pollonly=true`:
- Driver runs indefinitely in daemon mode
- Occasional `LIBUSB_ERROR_IO` logged at debug level 3, but skipped
- Data collection continues normally
- No more reconnect loops or "Data stale" errors

## Related

This is likely the same underlying issue reported in other contexts (see instantlinux/docker-tools#67) where CyberPower devices experience intermittent failures. The difference is our specific model (0x0601) fails predictably during full update cycles, making the root cause easier to identify.

---

## General points

  comments or text documents.

  case.

  or its equivalent.

  concept.

  the changed codebase.

### DIFF
--- a/drivers/usbhid-ups.c
+++ b/drivers/usbhid-ups.c
@@ -2381,7 +2381,7 @@ static bool_t hid_ups_walk(walkmode_t mode)
 			 * all polls fail and safety check below triggers reconnect. */
 			upsdebugx(3, "Got LIBUSB_ERROR_IO on item '%s' ReportID=0x%02x - skipping",
 				item->info_type ? item->info_type : "(null)",
-				item->hiddata ? item->hiddata->ReportID : 0xFF);
+				item->hiddata ? item->hiddata->ReportID : 0xFFU);
 			interrupt_pipe_EIO_count++;
 			continue;
 


jimklimov:
Great analysis, thanks. That must have been a fun debugging session...

As for `setsid()` implications - I've re-checked, this method gets called from `common::background()` and indeed relatively late in general driver `main.c` startup (after device init and perhaps data dump, optionally before the update loop - unless deciding to remain foregrounded due to one or another reason).

This is probably a separate issue from this avoidance of reconnections in the first place, but did you have a chance to experiment whether reconnection in this driver actually works/fails? I wonder if the problem is coincidental, e.g. nothing wrong with the driver code, but it is the UPS firmware that gets stuck, reboots and is not responding just at the moment we try to connect back to it... I suppose pulling the USB cable and plugging it back while the driver is running can help in the investigation (unless this is one of those UPSes that power on the USB chip when it is connected, and power it off/cycle when not in use).

com6056:
https://github.com/networkupstools/nut/issues/3116#issuecomment-3750031798
> So far so good, been running _without_ `pollonly = true` for 35+ hours now with no disconnects 🎉

Regarding this:
> This is probably a separate issue from this avoidance of reconnections in the first place, but did you have a chance to experiment whether reconnection in this driver actually works/fails? I wonder if the problem is coincidental, e.g. nothing wrong with the driver code, but it is the UPS firmware that gets stuck, reboots and is not responding just at the moment we try to connect back to it... I suppose pulling the USB cable and plugging it back while the driver is running can help in the investigation (unless this is one of those UPSes that power on the USB chip when it is connected, and power it off/cycle when not in use).

Out of town right now but I can definitely dig in more when I'm back to see if I can somehow get the debug logging while in daemon mode to answer these questions!

jimklimov:
Disregard the netbsd runner fault, ran out of disk.

jimklimov:
Great thanks for your contribution!

Issue #3255, opened 2026-01-12
driver man link absent


systemcrash:
https://networkupstools.org/docs/man/blazer-common.html

404

jimklimov:
Odd, where was such a link published? This is a partial document that should be included into ser and usb dirver pages.

systemcrash:
https://networkupstools.org/docs/man/nutupsdrv.html (drivers)

jimklimov:
Also impacts `nut_usb_addvars`

jimklimov:
Oops, mis-linked from commit comments for #3253

PR #3260, opened 2026-01-13
Fix driver man page listing to not include snippets

jimklimov:
A few files are repetitively included into other man pages, but are not pages themselves. They ended up listed in the index, with links leading nowhere.

Closes: #3255


jimklimov:
Seems fixed according to https://github.com/networkupstools/nut/actions/runs/20950894268/artifacts/5109526426

Issue #2836, opened 2025-03-05
NUT late shutdown and Immutable Image OSes


jimklimov:
A chat with @poettering (feel free to correct me if I remembered something wrong here) at FOSDEM 2025 brought up an interesting consideration: as he argues in favor of immutable image based operating environments, whose life cycle is effectively managed by another immutable OS image living as initrd (and probably systemd at the heart of it all), the currently typical NUT shutdown integration would not be feasible/welcome there in the way it is done now...

Roughly speaking, what we do currently on numerous platforms is:
* A NUT server (with physical connection to the UPS) normally runs drivers and the data server, probably also an `upsmon` instance in `primary` role (whoever is actually fed by that UPS should have a copy; most of the time the system managing an UPS is also fed by it).
* If a power outage occurs, this server raises FSD (Forced Shut Down) state to tell everyone else to shut down, and after a while, a locally running `upsmon` shuts down its own operating system too.
* As part of such FSD handling, the `upsmon` creates a file in location specified by `POWERDOWNFLAG` configuration from its `upsmon.conf`.
* Running daemons, including NUT drivers for the UPS and the data server, are stopped as are any other services.
* (Some systems eventually kill all userland processes and remount read-only)
* In case of systemd-driven systems, the late shutdown hook script in `/usr/lib/systemd/system-shutdown/nutshutdown` kicks in, finds that `POWERDOWNFLAG` file, and runs the NUT driver program *again* to tell the UPS to power-off/power-cycle (so the UPS usually turns on automatically when the wall power returns - or if it already has, and all fed systems are guaranteed to fully restart) at/after the moment we know the power loss would not corrupt any data.

This last bit actually places a number of constraints on the environment:
* The `POWERDOWNFLAG` file location should still be mounted and at least readable;
* The NUT configuration files (e.g. `/etc/nut` or `/etc/ups`) should be on filesystems still mounted and at least readable, so that the correct driver is chosen and connects to the expected device(s);
* The NUT driver programs and any libraries they might dynamically link to (and possibly their resource files - maybe SNMP MIBs, etc.) should be on filesystems still mounted and at least readable (programs also executable).
[...]

PR #3262, opened 2026-01-14
Install sysusers.d config for user/group

bluca:
Ubuntu, debian and other distro create the nut user/group pair by default in their packages.

Add a sysusers.d config file and install it. This configuration file will create the configured user/group pair on first boot/factory reset/package installation, without requiring manual scripts or actions. This is very important for image-based OSes, but it is useful in other cases too, to allow using declarative configuration instead of scripts.

https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html

This is especially useful in combination with the existing tmpfiles.d config that is already installed, and makes use of the configured user/group. With the sysusers.d addition, it works out of the box without manual scripting.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -40,12 +40,6 @@ https://github.com/networkupstools/nut/milestone/12
    to patterns defined in link:docs/nut-names.txt[]
 
  - common code:
-   * Generally improved shell script portability (including `Makefile.am`
-     embedded recipes) when mixing back-quotes and double-quotes. Namely,
-     the classic Solaris `ksh` implementation (also used in current versions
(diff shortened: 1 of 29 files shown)


jimklimov:
Great thanks for your contribution!

jimklimov:
Oddly, OBS did not pick up from the PR some recipe discrepancies due to the new delivered file. Only complained once this got to master :\

jimklimov:
@bluca : can this be seen as a step towards #2836 that we happened to discuss last FOSDEM with you and Lennart?

bluca:
Yeah it's part of that whole workflow

PR #3273, opened 2026-01-20
drivers/tripplitesu.c: add 1s delay in command_send

TomZanna:
> [!IMPORTANT] 
> This PR is currently a draft. I am looking for feedback on the best implementation strategy (see below).

Intermittent communication timeouts were observed when multiple commands were sent in rapid succession. Specifically, when running the driver manually with maximum debug verbosity, a timeout occurs consistently at the second `AVL` message within the `init_comm` function.
Adding a 1-second delay at the beginning of `do_command()` ensures the hardware buffer is ready and prevents these "read timeout" errors.

Should we consider making this a configurable parameter in ups.conf (e.g., command_delay)? I'd appreciate the maintainers' insight on whether this hardware quirk is universal enough to justify a global delay for this driver.


jimklimov:
Thanks, the proposal looks reasonable. Hard to say if the problem is "universal", given that this is a driver aimed at certain models of UPSes by one vendor, so there are just so many devices impacted by any change here in any case.

I usually lean in favour of toggles, probably as a numeric value for that `usleep` argument, which would be e.g. `-1` by default to disable it, and maybe flip to `1E6` in case the device saw an I/O timeout. This would allow the problem to be probably fixed out of the box, with end users free to set a specific value or disable the feature by setting `0` depending on how the fix behaves in their particular case.

TomZanna:
First of all, I apologize for the radio silence! I’ve been a bit buried lately preparing for a university exam and managing a house move at the same time.

In the meantime, I’ve done some further testing and found a couple of interesting things:
- It turns out the `TSU_SHUTDOWN_RESTART` command can actually power on a UPS that is currently off. I’ll need to update the documentation to reflect this behavior.
- I’ve been sniffing the serial traffic from the official PowerAlert software and noticed it’s possible to specify custom delays for shutdown and reboot. I’m planning to integrate this into upscmd so we have better control over the timing.

PR #3288, opened 2026-01-28
drivers: powercom-hid: use offdelay/ondelay config options for shutdown commands

Vladdrako:
The powercom_shutdown_nuf(), powercom_stayoff_nuf(), and powercom_startup_nuf() functions were reading delay values from dstate_getinfo("ups.delay.shutdown") and dstate_getinfo("ups.delay.start"), which return the UPS's CURRENT delay values read via HID.

However, users configure their desired delays via the "offdelay" and "ondelay" driver options in ups.conf. These were being ignored.

This patch changes the functions to:
1. First try the value passed directly to the instant command
2. Then try getval("offdelay")/getval("ondelay") from driver config
3. Then try dstate_getinfo() for current UPS values
4. Finally fall back to sensible defaults (60s shutdown, 120s startup)

This ensures that user configuration is respected.

Tested on Powercom RPT-800AP (VendorID: 0x0d9f, ProductID: 0x0004).

Prevents daemon.crit usbhid-ups[10112]: instcmd: attempting a power state change operation [shutdown.return] [(null) (21504.000000)] on [powercom] with broken 21504 value

### DIFF
--- a/drivers/powercom-hid.c
+++ b/drivers/powercom-hid.c
@@ -28,7 +28,7 @@
 
 #include <ctype.h>	/* isdigit() */
 
-#define POWERCOM_HID_VERSION	"PowerCOM HID 0.73"
+#define POWERCOM_HID_VERSION	"PowerCOM HID 0.74"
 /* FIXME: experimental flag to be put in upsdrv_info */
 
 /* PowerCOM */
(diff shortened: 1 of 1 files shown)


jimklimov:
Thanks for the change and for testing it! I'll update the news entry and some comments.

Issue #3291, opened 2026-01-30
Fatal error Dynamic consumer of...


aarnold-sd68:
New version seem to be throwing an error that I've not seen on windows before.  When start up nut via nut.exe -N.  It seems to start upsdrvctl.exe with a couple of errors.  I was trying to fix the snmp-ups issue by upgrading to a newer version.   Version 2.8.4.3949 - is working and does not get this error.  

SvcMain: GetExitCodeProcess(upsdrvctl): daemon died, restarting: No error [The operation completed successfully. ]
Network UPS Tools upsdrvctl.exe - UPS driver controller 2.8.4.1284.9-1293+gb9711f8b7 (development iteration after 2.8.4) using a dynamic libnutprivate*
FATAL: Dynamic consumer of a NUT private library was not initialized correctly.

jimklimov:
Thanks, that must be a misfire of recent changes to reduce footprint of NUT by offloading repetitive code into shared libraries. Did you fully replace your installation with what was available in the 7z archive, or can there be some mix of older programs built in the semi-statically linked way and the new-styled ones?

The builds do pass CI testing which involves running a dummy driver, although I think `upsdrvctl` is not a part of it.

jimklimov:
Can you please run it with debug, e.g. `upsdrvctl -DDDDDD -d start` so it would show which data it filled for talking to the library  and what it might be missing?

jimklimov:
Ok, reproduced locally - upsdrvctl is not updated to that code structure, in that build at least; looking.

jimklimov:
Thanks, especially for finding this just as my fingers were tickling to cut a NUT release. Posting a new build to check soon :)

If all goes well, the fixed binaries should be under https://ci.appveyor.com/project/nut-travis/nut/builds/53463852

aarnold-sd68:
I tried it both ways.  I tend to just over write with the new copy, unless I have an issue, then I delete everything except etc and my script directory.

aarnold-sd68:
Seems to work now.  Thank you for fixing this so quickly.

jimklimov:
Fix merged

Issue #2800, opened 2025-02-09
NUT for Windows build is huge and sparse


jimklimov:
According to recent CI runs, the tarball for consumers is nearing 3GB in size (but compresses down to 100Mb over 10 minutes of 1-hour time slot on Appveyor):
````
Sun, Feb  9, 2025  1:26:51 PM
cd .inst
7z a ../NUT-for-Windows-x86_64-SNAPSHOT-%APPVEYOR_BUILD_VERSION%.7z NUT*
7-Zip 24.08 (x64) : Copyright (c) 1999-2024 Igor Pavlov : 2024-08-11
Scanning the drive:
34 folders, 482 files, 2840422626 bytes (2709 MiB)
Creating archive: ../NUT-for-Windows-x86_64-SNAPSHOT-2.8.2.2734-master.7z
Add new data to archive: 34 folders, 482 files, 2840422626 bytes (2709 MiB)
Files read from disk: 482
Archive size: 107839478 bytes (103 MiB)
Everything is Ok
Sun, Feb  9, 2025  1:35:42 PM
````

This is too much by whatever metric we can imagine (one sixth of the CI run time, huge storage footprint for whoever wants to try it out). Part of the problem may be duplication of dependency DLLs, however many NUT driver binaries are in 30-50Mb range each, as if they are all statically linked despite the presence of many shared objects in the loop?

* [ ] Check if builds ARE dynamic?
* [ ] Check if debug information is present (and can account for such size) in the binaries? Can it be stripped or externalized (PDB files etc.)?
  * Note it could well be there even by design of experimental CI builds, to help debug them as N4W is a recent newcomer to the ecosystem. But at what cost...

**Comment by jimklimov** [comment]
Actually, looking at it, the Linux builds for example are also not quite trim and slim, with drivers commonly weighing in at several megabytes.

As a wild guess, they all carry a copy of `libcommon*.a` and that could be off-loaded to a dynamically shared library?

CC @aquette @clepple : Any ideas on that - perhaps reasons why NUT common code was historically not arranged into delivered shared libraries?

**Comment by clepple** [comment]
I don't remember there being enough common code to really justify the space savings - but maybe that is due to the code not being shipped with symbols on Linux? For instance, NUT 2.7.4 on Ubuntu 22.04:
```
$ ls -lSrh /lib/nut/ | tail 
-rwxr-xr-x 1 root root  79K Mar 24  2022 solis
[...]

PR #3292, opened 2026-01-31
`upsdrvctl` did not play along with `ENABLE_SHARED_PRIVATE_LIBS` well

jimklimov:
Follow-up for #2800 and #3291.

Perhaps the NIT suite should use the tool instead of directly starting the driver programs; then this would have been caught. Maybe some other day though... => offloaded to #3294 (there are some hiccups that need a few more coding iterations to unravel)


jimklimov:
Adding upsdrvctl to NIT was not as trivial as hoped: something not found on Appveyor at least... Possibly DLLs not in the right path?..

PR #3304, opened 2026-02-06
Support providing CGI options for alternate template file names

jimklimov:
Closes: #2524


jimklimov:
To err on the safe side, want a config file like that for upsset, to have the admins confirm they want templates (maybe even list which names are allowed).

So also manpage for that, etc.

UPDATE: done, added keywords to existing `hosts.conf`

PR #3317, opened 2026-02-11
Remerge master as of 20260210 to FTY

jimklimov:
Another bump goes by... more recently fixed bugs to resync into the other code base, wannabe into the main trunk.


jimklimov:
GitHub auth engine crashed and denied checkouts :\

Issue #2525, opened 2024-07-09
Variable for STATUS includes spurious HTML


mikewhippsHarvest:
While editing the `upsstats-single.html` template to produce JSON output I discovered that the @STATUS@ template variable returns a spurious `<br>` element, as in `ONLINE<br>`

I haven't tested the output of other values for this variable.

This is running on a Raspberry Pi. Here's the driver info block output from `upsc`

driver.name: usbhid-ups
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.port: auto
driver.parameter.productid: ffff
driver.parameter.synchronous: auto
driver.parameter.vendorid: 0463
driver.version: 2.8.0
driver.version.data: MGE HID 1.46
driver.version.internal: 0.47
driver.version.usb: libusb-1.0.26 (API: 0x1000109)

jimklimov:
Looking at `git blame`, this `<br>` seems deliberate and is there for a long time, per https://github.com/networkupstools/nut/blame/0d61b63182aedc3be2388126bb05c9f804b352b0/clients/upsstats.c#L170-L173 => https://github.com/networkupstools/nut/commit/8ae7be7c0676c2c3ab3a3984f06572760bfaf05c

I'd say nowadays it should be more properly a `<br/>` tag (opened+closed), and for your question - only added before second+ status token. Maybe a space could be better, but likely less predictable for visual width/wrapping etc.
* Is your question about the `br` tag here due to #2524 - are you trying to parse info from HTML markup for now?

FWIW there is another break tag at https://github.com/networkupstools/nut/blame/0d61b63182aedc3be2388126bb05c9f804b352b0/clients/upsstats.c#L925 and I struggle to see why: it is a fixed end of a table cell... @aquette : any memories here? some crafty browsers liked it so? :)

mikewhippsHarvest:
I've created a template to output the data I want in JSON. Apart from limitations on the use of custom templates (for which I opened another issue) this is working well, but...

PRTG has limited capabilities for processing data it imports this way. I can import a single custom text message along with properly formatted numeric data. If I want to import a message like, for example, `UPS is ONLINE` the presence of the `<br>` tag messes up the formatting in the PRTG display, if it will import it at all. 

PRTG provides no facility to manipulate imported text, or substitute from a lookup table, so I have to use the value from the UPS (just `OL` or `OB`, etc.), or use `ONLINE<br>` as is.

[...]

Issue #3316, opened 2026-02-11
new potential upsstats visuals (another Gemini creation)


jawz101:
[upsstats.html](https://github.com/user-attachments/files/25225263/upsstats.html)
[upsstats-single.html](https://github.com/user-attachments/files/25225264/upsstats-single.html)
I just said to make something pretty and modern.  It says it did as much as it could without using javascript.  I am not a web coder but I figured I wanted to upload them if it is interesting.
<img width="938" height="356" alt="Image" src="https://github.com/user-attachments/assets/f1caba5e-4b15-435d-902a-a1289db789c1" />

<img width="923" height="395" alt="Image" src="https://github.com/user-attachments/assets/8f7f773c-f129-4e41-acae-5fb814e7dd52" />

jimklimov:
I can't deny, this does look pretty, thanks. Might go in hand with #3304 to do support several templated designs at all (so we can ship both "legacy" and "modern" templates).

The majority of "modern look" seems to be the rounded rectangles and maybe a softer selection of fonts?

jimklimov:
I think I can merge it with existing templates (there are also comments for whoever edits them manually) to ship the two design alternatives. At least, locally it seemed decent :)

jawz101:
Yeah, it was interesting to see how Gemini defined modern because I didn't have an idea going in.  I had been using Gemini to help me configure fedora Linux with cockpit and jellyfin installed so maybe it took some design cues from those programs from those other conversations.

I just had to coax it to keep some of the original design elements- such as vertical bar visuals with tick marks and labeled ranges. 
It did point out that the ranges were hardcoded in the original code which may not be as flexible for European something... I just said to do whatever makes sense for the widest variety of equipment nut supports.

I may just use stock, personally. I dunno.  I do Power BI dashboards at work and I think visuals that are too pretty distract from the message they convey.  In NUT's case, monitoring a UPS in a nuclear power plant's server room- unless critical looks jarring- I'd probably not register something bad has happened.

I may make another sometime and tell Gemini to combine the two so "modern" isn't as "soft" about the information.

jimklimov:
I am not sure how bar expressions like `(@VAR input.voltage@ - 60) / 100 * 100%` translate to a meaningful bar, European or not? Say with healthy American 120V inputs, you get `(120 - 60) / 100 * 100%` would be `60%` of the bar height, offset by the `tick`?

[...]

PR #3318, opened 2026-02-11
Introduce upsstats-modern-list.html.sample.in upsstats-modern-single.html.sample.in

jimklimov:
Thanks to #3316 and to Gemini offering the starting point for the updated template text.

I think there will be another iteration to reconcile design differences. The first commit replaces most of original HTML templates with the new proposal "as is" (just making sure bottom links to JSON and w3c badges remain in place).

Follows up for #3304 and #2524
Closes: #2525


jimklimov:
Oh nice, spellcheck the HTML. And life threw a GitHub outage into the mix.

jimklimov:
@jawz101 : I've finally got back from vacation and to this challenge, hopefully got the different voltage and temperature graphs to auto-adjust the scales and nominals. Can you please check if this branch works well with your setup (you'd also need to update the build of `upsstats.cgi` program)?

PR #3327, opened 2026-02-25
Add support for dual-byte command for APC SPM series (And potential SPRM SRVS series)

MysteryDove:
## Problem

Current apcsmart driver for uart apc serial devices only support single byte command, but newer model is having two bytes command.

Also for models with two bytes, the command F is for output frequency and a `0x9F 0xD3` is used for input frequency, so current logic is incorrect for these models.

These dual-byte commands are already reported by the UPS in the `a` command response (the command set / capability string), encoded in a previously ignored `.stuff` extension section using the format `prefix + 0x3A (colon) + sub-command bytes`. 

```
/*
	 * returned set is verified for validity above, so just extract
	 * what's interesting for us
	 *
	 * the known format is:
	 * ver.alerts.commands[.stuff]
*/
```

For now, the apcsmart driver ignore this section entirely in `getbaseinfo()`
```c
tail = strchr(++cmds, '.');
if (tail)
    *tail = 0;  /* everything after this is discarded */
```

## The Patch

1. Similar to `apc_vartab_t`, a new struct `apc_vartab_dual_t ` is added, the struct defines dual-byte command entries with prefix and sub fields, everything else should be the same as `apc_vartab_t`

2. Similar to existing `apc_vartab`, a `apc_vartab_dual` is added with necessary command

3. Additional logic to parse what previously called `stuff` in the code comment, swap input/output mapping only if `9F D3` command is detected

4. `protocol_verify_dual` `parse_cmdchars_dual` `poll_data_dual` function is added as tool function for processing 2 bytes command  and send uart commands using `ser_send_buf`

Just trying to keep the original logic and add new things after that, tested on my SPM2K and everything looks fine, newly added battery current and input frequency is working as well

baseinfo

test

### DIFF
--- a/drivers/apcsmart.c
+++ b/drivers/apcsmart.c
@@ -1168,11 +1168,12 @@ static void parse_cmdchars_dual(const unsigned char *data, size_t len)
 		/* detect prefix:subcmds pattern (byte followed by 0x3A colon) */
 		if (j + 1 < len && data[j + 1] == 0x3A) {
 			dual_prefix = data[j];
-			j++;	/* skip the colon */
 			upsdebugx(1, "%s: found dual-byte prefix 0x%02X",
 				__func__, dual_prefix);
 			continue;
 		}
+		if (data[j] == 0x3A)
+			continue;	/* skip literal colon separator */
 		if (dual_prefix)
 			protocol_verify_dual(dual_prefix, data[j]);
 	}


MysteryDove:
Hmm, interesting, seems like code scan does not like the way of mutating the counter in-loop. Will make a quick change.

MysteryDove:
@jimklimov Not sure if this patch is included in the 2.8.5 milestone?

jimklimov:
Should be, as it's merged to master and releases are snapshots of that eventually.

MysteryDove:
> Should be, as it's merged to master and releases are snapshots of that eventually.

The change is not shown in the 2.8.5 release note so I'm bit worried about that :D

jimklimov:
Aha, the PR did not include a change for NEWS.adoc - thanks for the reminder, I'll add.

jimklimov:
Aha, I've now got why you worry: due to fat-fingering or another malfunction, draft text attached to an RC tag got published at all, and as a "latest release" at that. I hope the CI will clear the current codebase to publish as 2.8.5 final by tomorrow though.

Issue #1358, opened 2022-04-08
Bug in snmp-ups.c causes bogus “unhandled ASN0x80” messages to be logged


smarsching:
There is a bug in `snmp-ups.c` that causes the following message to be written to syslog repeatedly (when using the `ietf` SNMP driver):

```
unhandled ASN 0x80 received from .1.3.6.1.2.1.33.1.7.7.1`
```

The problem is in `nut_snmp_get_int`: When this function processes a response of type `ASN_OBJECT_ID`, it tries to resolve this object ID. In case of the `upsTestId` (a.k.a. 1.3.6.1.2.1.33.1.7.1.0 as defined in `ietf-mib.c`), this does not work because the returned OID should be interpreted as a constant instead of a pointer to another value.

From the IETF UPS MIB:

```upsTestId OBJECT-TYPE
    SYNTAX     OBJECT IDENTIFIER
    MAX-ACCESS read-write
    STATUS     current
    DESCRIPTION
            "The test is named by an OBJECT IDENTIFIER which
            allows a standard mechanism for the initiation of
            tests, including the well known tests identified in
            this document as well as those introduced by a
            particular implementation, i.e., as documented in the
            private enterprise MIB definition for the device.

            Setting this variable initiates the named test. Sets
            to this variable require the presence of
            upsTestSpinLock in the same SNMP message.

            The set request will be rejected with an appropriate
            error message if the requested test cannot be
            performed, including attempts to start a test when
            another test is already in progress.  The status of
            the current or last test is maintained in
            upsTestResultsSummary. Tests in progress may be
            aborted by setting the upsTestId variable to

            upsTestAbortTestInProgress.

            Read operations return the value of the name of the
            test in progress if a test is in progress or the name
            of the last test performed if no test is in progress,
            unless no test has been run, in which case the well
            known value upsTestNoTestsInitiated is returned."
    ::= { upsTest 1 }
```

and 

```
upsWellKnownTests     OBJECT IDENTIFIER ::= { upsTest 7 }

upsTestNoTestsInitiated OBJECT-IDENTITY
    STATUS     current
    DESCRIPTION
            "No tests have been initiated and no test is in
            progress."
    ::= { upsWellKnownTests  1 }
```

So, the OID `1.3.6.1.2.1.33.1.7.7.1` represents a constant, that should not really be resolved.

[...]

PR #3337, opened 2026-03-03
snmp-ups: quiesce "unhandled ASN 0x80 received from ..." when we fall back to using the numeric value from last OID section

jimklimov:
Closes: #1358

Per analysis in the ticket (thanks @smarsching), we sometime get OIDs to something that represents a constant, that should not really be resolved. When we recurse from `nut_snmp_get_int` (for an `ASN_OBJECT_ID` PDU type) we first loudly `upslogx` an error in the `default` case (the `0x80` seems to be `ASN_CONTEXT`, with no more context to indicate that this specific query went to a constant... or should the lack of other bits mean exactly this?) and then fall back to returning the last section of the OID as the desired number.

With this PR, the original `nut_snmp_get_int()` method body became the implementation of `do_nut_snmp_get_int()` with an added option to log the `default` case loudly (`upslogx()`) or quietly (`upsdebugx()`), and the `nut_snmp_get_int()` signature called by others is a shortcut to call this implementation with loud debugging. Recursing into itself for `ASN_OBJECT_ID` is done in quiet mode.

This way, I hope, we still log OIDs that are out of place in e.g. mapping tables, but would not make infinite system log noise for returned references like these.

### DIFF
--- a/drivers/snmp-ups.c
+++ b/drivers/snmp-ups.c
@@ -1284,7 +1284,7 @@ static void nut_snmp_free(struct snmp_pdu ** array_to_free)
 }
 
 /* Return a NULL terminated array of snmp_pdu * */
-static struct snmp_pdu **nut_snmp_walk(const char *OID, int max_iteration)
+static struct snmp_pdu **nut_snmp_walk(const char *OID, int max_iteration, int log_unhandled_loudly)
 {
 	int status;
 	struct snmp_pdu *pdu, *response = NULL;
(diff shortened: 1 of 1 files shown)


smarsching:
@jimklimov Thank you very much for your efforts!

I didn’t find the time to test the changes this week, and next week I am out of office for a few days, so it most likely won’t happen before the week after.

jimklimov:
ACK, thanks for reaching out. This is merged by now, so when/if you do get to testing, you can at least follow the standard procedure as written, with `master` branch of the primary repository (e.g. not my fork).

PR #3330, opened 2026-02-26
Identify SSL support in upsd and libupsclient, add NIT tests for OpenSSL and Mozilla NSS

jimklimov:
Closes: #3328
Largely fulfills #1711 (for the C client tests, but not for other ecosystems like C++ and Python)

Issue #1711, opened 2022-11-21
CI: actively test SSL setups with NIT


jimklimov:
Generate keys and certs, configure NUT, try to connect.

Note there are different features and config keywords for OpenSSL and NSS, and possibly different tools available on the build box (`openssl` CLI, `keytool`, etc.) or absent altogether (skip not fail then).

Also note that over time we hopefully get SSL support in different clients/bindings (C, C++, Perl, Python...) and hope that the test suite will prod all of them.

jimklimov:
Initial implementation for C, C++, Python, Perl and Java addressed.
Further ideas posted into the tracker (see references above), so this issue can be closed as achieved.

PR #3344, opened 2026-03-08
Make NUT OpenSSL handshake more reliable

jimklimov:
Following up from findings in PR #3330 to address issue #1711, this PR would hopefully make the handshake more reliable on all systems supported by NUT (CI farm at least). Apparently Linux mostly handles the SSL handshake quickly enough to succeed by the first time we ask for results, and other platforms may need a few more cycles. Tests with OpenSSL in that PR failed so frequently, they had to be neutered by default.

Prepared with AI (Claude Sonnet 4.6) as a test to see how it fares... quite well, despite not seeing failure logs from CI server and mostly relying on test case descriptors which convey compilers, OSes, etc. Had troubles preparing a diff-patch file though.

Findings seem to match what @clepple proposed but could not commit to coding the fix quickly - review welcome. I think this addresses a similar problem to that explored in https://github.com/networkupstools/nut/compare/ssl_accept_nbio_v285?expand=1 (variant of an older branch, rebased over recent NUT master code to pre-resolve merge conflicts), but at least that branch did not fully solve the issues the NIT test reproduced all too well across the board. I'd love a comment from @clepple on whether that fix may bring some other benefits (e.g. by setting socket options), or may be abandoned if this PR works?

### DIFF
--- a/tests/NIT/Makefile.am
+++ b/tests/NIT/Makefile.am
@@ -27,11 +27,11 @@ ABS_TOP_BUILDDIR = @ABS_TOP_BUILDDIR@
 # Early in issue #1711 OpenSSL tests failed more often than we liked, so
 # this toggle was added.  It may still be useful for CI on some platforms.
 # Developers can run the script directly or override the make variable.
-if WITH_OPENSSL
-WITHOUT_SSL_TESTS = true
-else !WITH_OPENSSL
+### if WITH_OPENSSL
+### WITHOUT_SSL_TESTS = true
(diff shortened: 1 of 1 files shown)


vibingarthur:
CI was failing (shellcheck). Attempt #1 to fix.

Skipping: External CI (ci.networkupstools.org) requires external access to diagnose. Please check the shellcheck errors directly in the CI logs.

jimklimov:
Please no vibing flood.

Issue #1252, opened 2022-01-06
Incorrect Status from Liebert PSI5-1500RM120


airforceixi:
Running NUT 2.7.4 on Fedora 35 connecting to a Liebert PSI5-1500RM120. Does not provide any valid status information. Here are some outputs:

# upsc ups@localhost
```
device.mfr: Vertiv Co.
device.model: Liebert PSI5
device.serial: ###
device.type: ups
driver.name: usbhid-ups
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.port: auto
driver.parameter.productid: 0002
driver.parameter.synchronous: no
driver.parameter.vendorid: 10AF
driver.version: 2.7.4
driver.version.data: Belkin/Liebert HID 0.17
driver.version.internal: 0.41
ups.mfr: Vertiv Co.
ups.model: Liebert PSI5
ups.productid: 0002
ups.serial: ###
ups.status: OB
ups.vendorid: 10af
```

/etc/ups/ups.conf
```
[ups]
driver = usbhid-ups
port = auto
productid = 0002
vendorid = 10AF
```

usbhid-ups -DDDDD -a ups
```
Network UPS Tools - Generic HID driver 0.41 (2.7.4)
USB communication driver 0.33
   0.000000     send_to_all: SETINFO driver.parameter.port "auto"
   0.000031     send_to_all: SETINFO driver.parameter.productid "0002"
   0.000040     send_to_all: SETINFO driver.parameter.vendorid "10AF"
   0.000052     debug level is '5'
   0.001548     upsdrv_initups...
   0.006497     Checking device (8087/0024) (002/002)
   0.006534     Failed to open device, skipping. (Permission denied)
   0.006546     Checking device (1D6B/0002) (002/001)
   0.006560     Failed to open device, skipping. (Permission denied)
   0.006566     Checking device (5986/02D2) (001/006)
   0.006575     Failed to open device, skipping. (Permission denied)
   0.006583     Checking device (0A5C/21E6) (001/005)
   0.006599     Failed to open device, skipping. (Permission denied)
   0.006605     Checking device (147E/2020) (001/004)
   0.006617     Failed to open device, skipping. (Permission denied)
   0.006623     Checking device (10AF/0002) (001/014)
   0.009188     - VendorID: 10af
   0.009203     - ProductID: 0002
   0.009207     - Manufacturer: Vertiv Co.
   0.009212     - Product: Liebert PSI5
[...]

Issue #1028, opened 2021-05-03
Trouble with Liebert PSI5-1500RT120LI


darconeous:
I recently purchased a [Vertiv Liebert PSI5-1500RT120LI](https://www.vertiv.com/en-us/products-catalog/critical-power/uninterruptible-power-supplies-ups/psi5-1500rt120li-vertiv-liebert-psi5-lithium-ion-ups-1500va1350w-120v-line-interactive-avr-/) (Admittedly not listed as an officially supported UPS), and after reading [this thread](https://www.mail-archive.com/nut-upsuser@alioth-lists.debian.net/msg01090.html) I assumed that simply adding something like this to my `/etc/nut/ups.conf` would do the trick:

```
[ups]
driver = usbhid-ups
port = auto
desc = "Libert PSI5 (Rack)"
productid = 0002
```

And this did kinda work, except that it only reports that it is always "on battery". I haven't tried running it down to see if it ever gets to the "LB" condition.

I'm a software developer and I'm not afraid to try hacking around to get this to work, but I'm not sure where to start. Any advice or pointers would be welcome.

Here is the output from running `usbhid-ups -DDDDD -a ups` manually:

```
$ sudo -u nut /lib/nut/usbhid-ups -DDDDD -a ups
Network UPS Tools - Generic HID driver 0.41 (2.7.4)
USB communication driver 0.33
   0.000000	send_to_all: SETINFO driver.parameter.port "auto"
   0.000150	send_to_all: SETINFO driver.parameter.productid "0002"
   0.000256	debug level is '5'
   0.001567	upsdrv_initups...
   0.026776	Checking device (1D6B/0003) (002/001)
   0.055284	- VendorID: 1d6b
   0.055336	- ProductID: 0003
   0.055377	- Manufacturer: unknown
   0.055416	- Product: unknown
   0.055452	- Serial Number: unknown
   0.055487	- Bus: 002
   0.055522	- Device release number: 0510
   0.055556	Trying to match device
   0.055601	Device does not match - skipping
   0.055997	Checking device (10AF/0002) (001/003)
   0.062778	- VendorID: 10af
   0.062824	- ProductID: 0002
   0.062862	- Manufacturer: Vertiv Co.
   0.062901	- Product: Liebert PSI5
   0.062940	- Serial Number: 0000000000000000
   0.062986	- Bus: 001
   0.063023	- Device release number: 0002
   0.063062	Trying to match device
   0.063154	Device matches
[...]

PR #3345, opened 2026-03-09
libusb1: auto-select interface 1 for Liebert PSI5 / PowerSure PST (10af:0002)

dcgrove:
## Problem

The Liebert PSI5 / PowerSure PST (USB VID:PID `10af:0002`) is a composite HID device with two interfaces:

| Interface | Class | Report descriptor | Contents |
|-----------|-------|-------------------|----------|
| 0 | Vendor-specific | 27 bytes | No usable data |
| 1 | HID Power Device Class | 662 bytes | Full UPS telemetry |

`usbhid-ups` claims interface 0 by default. Because that interface returns no AC-present bit, the driver permanently reports `ups.status: OB` (on battery) even when the UPS is running on mains power.

This is a long-standing issue that has been independently reported multiple times:
- #1252 (2021)
- #1028
- #3340 (opened 2026-03-05, same root cause)

## Root cause

`libusb1.c` iterates USB interfaces sequentially and claims the first one it encounters (`hid_rep_index = 0`). For this device, that is the vendor-specific interface with no usable Power Device Class data.

## Fix

When the device is identified as `10af:0002` and the user has not already overridden `usb_hid_rep_index` in `ups.conf`, automatically set `hid_rep_index = 1` so that the HID Power Device Class interface is claimed.

The fix follows the same pattern as the existing Eaton OEM special-case block and fully respects the `usb_hid_rep_index` manual override.

```c
/* Liebert PSI5 / PowerSure PST (10af:0002): the HID Power Device
 * Class descriptor is on interface 1.  Interface 0 exposes only a
 * 27-byte vendor-specific report that contains no usable UPS data,
 * which caused NUT to incorrectly report the UPS as on battery
 * (OB) even when AC power was present.
 * See GitHub issues #1252 and #3340. */
if ((curDevice->VendorID == 0x10af) && (curDevice->ProductID == 0x0002)) {
    if (!getval("usb_hid_rep_index"))
        usb_subdriver.hid_rep_index = 1;
}
```

## Testing

Tested on a Liebert PSI5 1100VA connected via USB to a Raspberry Pi 4 running Debian bookworm (libusb-1.0, NUT built from current master).

| | `ups.status` |
|---|---|
| Before patch | `OB` (false positive — UPS was on mains) |
| After patch | `OL CHRG` (correct) |

Note: `belkin-hid.c` already includes the `10af:0002` device table entry (added in #2369), so no change is needed there.

## Workaround for existing installations

[...]

Issue #3340, opened 2026-03-05
The parameters "usb_hid_rep_index = 1" and "usb_hid_desc_index = 1" are incorrectly handled if the driver 'usbhid-ups' does not start in debug mode.


philipperollin:
Hello Gentlemen
I'm not a programmer, but I know how to compile Linux software. I own a Vertiv Edge 3kVA UPS that I'm trying to use with NUT UPS 2.8.4. After numerous attempts and searches, I've come to the following conclusion: if the parameter "debug_min = 1" is not set, the parameters "usb_hid_rep_index = 1" and "usb_hid_desc_index = 1" are ignored, and the reported data is incorrect.

Below is a file with the essential configuration and results.

```
#ups.conf
[ups]
driver = "usbhid-ups"
port = "auto"
vendorid = "10AF"
productid = "0002"
product = "Vertiv Edge"
serial = "2014010030ED18C"
vendor = "Vertiv Co."
usb_hid_rep_index = 1
usb_hid_desc_index = 1
pollonly = 1
debug_min = 1
```

```
#Bad data without debug_min
battery.charge: 100
battery.charge.warning: 20
battery.runtime: 6480
battery.type: PbAc
device.mfr: Vertiv Co.
device.model: Vertiv Edge
device.serial: 2014010030ED18C
device.type: ups
driver.debug: 0
driver.flag.allow_killpower: 0
driver.name: usbhid-ups
driver.parameter.interrupt_pipe_no_events_tolerance: -1
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.pollonly: 1
driver.parameter.port: auto
driver.parameter.product: Vertiv Edge
driver.parameter.productid: 0002
driver.parameter.serial: 2014010030ED18C
driver.parameter.synchronous: auto
driver.parameter.usb_hid_desc_index: 1
driver.parameter.usb_hid_rep_index: 1
driver.parameter.vendor: Vertiv Co.
driver.parameter.vendorid: 10AF
driver.state: quiet
driver.version: 2.8.4
driver.version.data: Belkin/Liebert HID 0.22
driver.version.internal: 0.67
driver.version.usb: libusb-1.0.26 (API: 0x01000109)
input.frequency: 50.0
input.voltage: 231.5
input.voltage.nominal: 230
output.frequency: 50.0
output.voltage: 231.6
output.voltage.nominal: 230
ups.mfr: Vertiv Co.
ups.model: Vertiv Edge
[...]

PR #3349, opened 2026-03-10
`belkin-hid.c`: auto-select interface 1 for Liebert PSI5 / PowerSure PST (10af:0002)

jimklimov:
Aims to supplant PR #3345 by addressing the issue detailed there at the more proper location.

Fixes: #1252
Fixes: #1028

Not sure if really related to #3340 or a different issue (could simplify the config there though).

### DIFF
--- a/drivers/arduino-hid.c
+++ b/drivers/arduino-hid.c
@@ -155,24 +155,36 @@ static int arduino_claim(HIDDevice_t *hd)
 		case POSSIBLY_SUPPORTED:
 			/* by default, reject, unless the productid option is given */
 			if (getval("productid")) {
-				if (!getval("usb_hid_ep_in"))
+				if (!getval("usb_hid_ep_in")) {
+					upsdebugx(1, "%s: defaulting usb_hid_ep_in=4 for this model", __func__);
 					usb->hid_ep_in=4;
-				if (!getval("usb_hid_ep_out"))
+				}
+				if (!getval("usb_hid_ep_out")) {
+					upsdebugx(1, "%s: defaulting usb_hid_ep_out=5 for this model", __func__);
 					usb->hid_ep_out=5;
-				if (!getval("usb_hid_rep_index"))
+				}
+				if (!getval("usb_hid_rep_index")) {
+					upsdebugx(1, "%s: defaulting usb_hid_rep_index=2 for this model", __func__);
 					usb->hid_rep_index = 2;
+				}
 				return 1;
 			}
 			possibly_supported("Arduino", hd);
 			return 0;
 
 		case SUPPORTED:
-			if (!getval("usb_hid_ep_in"))
+			if (!getval("usb_hid_ep_in")) {
+				upsdebugx(1, "%s: defaulting usb_hid_ep_in=4 for this model", __func__);
 				usb->hid_ep_in=4;
-			if (!getval("usb_hid_ep_out"))
+			}
+			if (!getval("usb_hid_ep_out")) {
+				upsdebugx(1, "%s: defaulting usb_hid_ep_out=5 for this model", __func__);
 				usb->hid_ep_out=5;
-			if (!getval("usb_hid_rep_index"))
+			}
+			if (!getval("usb_hid_rep_index")) {
+				upsdebugx(1, "%s: defaulting usb_hid_rep_index=2 for this model", __func__);
 				usb->hid_rep_index = 2;
+			}
 			return 1;
 
 		case NOT_SUPPORTED:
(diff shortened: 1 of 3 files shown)


dcgrove:
## Compatibility note: this approach breaks NUT 2.8.0 (Debian bookworm stable)

I tested the `belkin_claim()` approach from this PR against NUT 2.8.0 (the version in Debian bookworm, which is also what Raspberry Pi OS ships) and found it fails silently — the driver starts but reports no device found. Here's why.

### Root cause: dual use of `hid_rep_index` in 2.8.0's `libusb1.c`

In NUT 2.8.0, `libusb1.c` uses `hid_rep_index` for **two distinct purposes**:

1. As the **USB configuration descriptor index** passed to `libusb_get_config_descriptor()`:
   ```c
   // drivers/libusb1.c ~line 331 (NUT 2.8.0)
   upsdebugx(2, "Reading first configuration descriptor");
   ret = libusb_get_config_descriptor(device,
       (uint8_t)usb_subdriver.hid_rep_index,   // ← config index, not interface!
       &conf_desc);
   ```

2. As the **USB interface number** to claim (later in the same function):
   ```c
   libusb_claim_interface(udev, usb_subdriver.hid_rep_index);
   ```

In NUT **master**, these have been correctly separated: `libusb_get_config_descriptor()` now uses a dedicated `usb_config_index` field (defaulting to 0), so `hid_rep_index` is free to hold only the interface number. The `belkin_claim()` approach in this PR works correctly on master for exactly that reason.

### What goes wrong on 2.8.0

The call order is:

```
libusb_open_device()
  → match_function_subdriver()      ← calls belkin_claim(), which sets hid_rep_index = 1
  → libusb_get_config_descriptor(device, hid_rep_index=1, ...)   ← tries config #1, not #0!
      → returns LIBUSB_ERROR_NOT_FOUND (-5)  (device only has one config, index 0)
  → libusb_claim_interface(udev, hid_rep_index=1)  ← succeeds (interface 1 exists)
  → retrieve HID descriptor → fails (conf_desc is invalid/null from the -5 error)
  → device rejected, driver exits with "No matching HID UPS found"
```

Debug output from a 2.8.0 build with this PR's patch applied:

```
[D1] belkin_claim: defaulting usb_hid_rep_index=1 for Liebert PSI5/PST 10af:0002
[D2] Device matches
[D2] Reading first configuration descriptor
[D2] result: -5 (Entity not found)        ← config descriptor #1 doesn't exist
[D2] Claimed interface 1 successfully     ← interface 1 is fine, but conf_desc is bad
[D2]   Couldn't retrieve descriptors      ← HID descriptor fetch fails
[D2] libusb1: No appropriate HID device found
libusb1: Could not open any HID devices: insufficient permissions on everything
```

### What works on 2.8.0

Setting `hid_rep_index = 1` **after** `libusb_get_config_descriptor()` returns (i.e., in `libusb1.c` itself, right after the config read) avoids the conflict:

```c
// drivers/libusb1.c — after libusb_get_config_descriptor() call
/* Now we have matched the device we wanted. Claim it. */

/* Liebert PSI5/PowerSure PST (10af:0002): the HID Power Device Class
 * descriptor is on interface 1; interface 0 exposes only a 27-byte
 * vendor-specific report with no usable data. */
if ((curDevice->VendorID == 0x10af) && (curDevice->ProductID == 0x0002)) {
    usb_subdriver.hid_rep_index = 1;
}
```

This is the approach in PR #3345 (targeting master's `libusb1.c`, which at that point has already used `usb_config_index` for the descriptor read).

### Suggestion

If a 2.8.x backport or stable-branch fix is ever needed, the `libusb1.c` placement is the safe one for that era of code. The `belkin_claim()` placement in this PR is correct and clean for master.

For what it's worth, I've been running the `libusb1.c`-based fix on a Raspberry Pi 4 (Debian bookworm, NUT 2.8.0) with a Liebert PSI5 1100VA since this thread started and it's rock-solid — `ups.status: OL CHRG` throughout.

jimklimov:
Thanks, good points. But why really care about 2.8.0 if it's old and set in stone 4 years ago or so?

I think the assumption here is that NUT versioning follows proper SEMVER (MAJOR-MINOR-PATCH releases), then 2.8 feature set is a reasonable baseline to code a patch against.

Alas, at the moment we're diverged from that, and whole X.Y.Z numbers are distinct releases. Hope we'll get back to truer semvers after some planned refactoring/hardening works are complete (a decade in the making)...

jimklimov:
So if this fix is confirmed working on master, it is good to be merged to master (eventually to be part of v2.8.5 release).

@dcgrove : was this change only analyzed by Claude, or you actually built and tested the PR source branch with your UPS?

jimklimov:
Per https://github.com/networkupstools/nut/pull/3345#issuecomment-4036224259 :
> Just had Claude reinstall with PR #3349 and post a comment on that PR with the results. Feel free to close/reject/accept whatever you want with this one if 3349 is better.

Aha, ok :)

Issue #3331, opened 2026-02-28
Improvement request: capability disparity of OpenSSL and Mozilla NSS builds: provide same functionality


jimklimov:
Currently NUT OpenSSL builds allow the NUT `upsd` data server to certify itself with a PEM certificate+chain+key file, and NUT clients using the `libupsclient` to initiate `STARTTLS` and essentially trust anyone. With additional setup in client code (as currently only done in `upsmon`, see more at issue #3329), the trusted server range can be constrained to a directory with a list of PEM files.

With Mozilla NSS, the server and clients (currently only `upsmon` again) can use the specified trust database locations, with certificates uploaded there. Clients can present themselves with a certificate, and server can be configured to only talk to clients that present a certificate (any, or those validated by the server database). Clients can also be configured to require that a particular server should respond with a particular certificate subject name.

While the original difference to support the two libraries was largely due to licensing, they now also incur different capabilities and setup for the server and for the (remote) client depending on the build.

This issue is about trying to level the field, and having OpenSSL builds able to do all the same tricks as NSS builds (so the only practical difference is about cert/key setup).

jimklimov:
After experiments with PR #3330 it generally seems that OpenSSL session init is much more likely to fail, with various problems reported on different platforms. NIT code added by that PR for now will not test the OpenSSL variants in NUT CI runs, but that is of course not tenable long-term (focus of this issue).

A bit more context from @clepple in https://github.com/networkupstools/nut/issues/246#issuecomment-3970310763

PR #3351, opened 2026-03-16
Make NUT OpenSSL dialogs more reliable

jimklimov:
Follow-up to #3344, this PR applies similar logic of retrying (in case of `SSL_ERROR_WANT_WRITE` or `SSL_ERROR_WANT_READ` return codes) to read and write commands during the normal protocol dialog, not just during handshake.

Without this, `make check-NIT` failed on Windows at least with OpenSSL builds - `upsc` established the connection but failed actual queries (possibly it broke at the moment of sending/reading the `"OK STARTTLS"` confirmation). Apparently it fared better on all other NUT CI farm systems, regardless of their performance (some builders are co-located on quite congested hosts).

CC @clepple : this picks up the trick from your `ssl_accept_nbio` branch, of `return 0` from `ssl_error()` methods in case of these retryable return codes. The rest of non-blockable support (is it needed or not, toxic or not, after progress made in recent PRs) remains questionable to me. For some clarity, I'll bump the `ssl_accept_nbio_v285` branch after merging this PR (if it passes) to see how it fares in CI.

Also FYI CC @aquette - with this, SSL support should be at least functional on various platforms. Previously much of it happened to work if conditions were just right, and crashed elsewhere. Converging OpenSSL and NSS code to behave similarly, so it is a preference and not a choice driven by technical constraints, is a separate matter (#3331).

UPDATE: Huh, after a series of refactors into different smaller PRs and rebases, the big description here remained attached to a small change for tests in that area.

### DIFF
--- a/tests/NIT/nit.sh
+++ b/tests/NIT/nit.sh
@@ -920,8 +920,9 @@ case "${WITH_SSL_CLIENT}${WITH_SSL_SERVER}" in
                         # OpenSSL CA trust "database" should include hashes
                         # of CA PEM certificates as symlinks to actual files:
                         CERTHASH="`openssl x509 -subject_hash -in rootca.pem | head -1`"
-                        ln -s rootca.pem "${CERTHASH}".0
-                        ln -s rootca.pem "${CERTHASH}"
+                        # NOTE: Symlinking may be prohibited or not implemented on some platforms (e.g. Windows) or file systems
+                        ln -fs rootca.pem "${CERTHASH}".0 || ln -f rootca.pem "${CERTHASH}".0 || cp -f rootca.pem "${CERTHASH}".0
+                        ln -fs rootca.pem "${CERTHASH}" || ln -f rootca.pem "${CERTHASH}" || cp -f rootca.pem "${CERTHASH}"
                         ls -l "${TESTCERT_PATH_ROOTCA}"/rootca.pem "${TESTCERT_PATH_ROOTCA}/${CERTHASH}"*
                         ;;
                     *)


user01010111:
Following up on the nonblocking support left open here and the logic deferred in #3388: on master fe4e6fec3807017418c6fbc19b007a660d8e409c, accepted Linux sockets are still blocking. In a bounded two-client test, a paused TLS handshake or incomplete TLS record held up another client's ordinary query for over six seconds; closing the paused peer restored service. Normal TLS and fragmented plaintext controls passed. The five-second retry count cannot bound a blocking TLS call. The `ssl_accept_nbio_v285` branch also retains synchronous retries.

Before drafting a transport change, which direction would you prefer: keep synchronous operations with an actual elapsed deadline, or resume each connection's TLS and output progress from `mainloop()`? What handshake/write timeout and, if output is queued, pending-output limit should apply? I want to preserve the intended behaviour for slow clients and avoid duplicating a planned continuation.

Investigation assisted by OpenAI Codex (`gpt-6-astra`).

jimklimov:
Good question. I think, as long as we do define timeouts, they should be honoured. Whether a server is busy processing some other dialog and can not answer a new call, or some other reason, if the caller says they want *some* answer within 10 seconds, they should get it then.

At the moment, most NUT clients have a configurable "network timeout for initial connections" (as opposed to timeouts for R/W afterwards). Unless we want to add more variables and command-line options (or somehow extend common `-W` beyond a single number), it can be argued that establishing an SSL session is also an initial connection - for that transport and security level.

user01010111:
Thanks, that clarifies the client side: establishing TLS can use the existing initial-connection timeout, and a busy server should not extend that budget. The C++ timeout follow-up is already scoped to that path.

The server-side stalls are a separate question. A client timing out does not prevent a paused TLS exchange or blocked response write from holding up `upsd`'s other clients. Your reply does not appear to select a server-side operation deadline or output-buffer limit, so I have kept those decisions open.

There is a smaller independent correction: when a plaintext POSIX read returns `EINTR`, preserve the connection and partially parsed command, then return to the event loop so pending signals can be handled. This follows the existing driver-read pattern. Retrying a blocked write in place would need more care because it can delay shutdown or reload handling.

In `server/upsd.c`, immediately after the plaintext `read()` in `client_readline()`:

```c
#ifndef WIN32
		if (ret < 0 && errno == EINTR) {
			/* Let the main loop handle pending signals before reading again. */
			return;
		}
#endif	/* !WIN32 */
```

This handles an interrupted plaintext read; the remaining question is how `upsd` should keep serving other clients when a TLS exchange or response write stalls.

The focused real-server regression fails before this correction and passes after it on Linux and native macOS. It injects read results and checks fragmented commands, repeated interruptions, reload, shutdown, permanent read errors and EOF.

For the remaining server work, would you prefer bounded synchronous operations or per-connection TLS/output progress in the event loop, and what server-side duration and pending-output limit should apply?

Investigation and local changes assisted by OpenAI Codex (`gpt-6-astra`).

user01010111:
@jimklimov - The round-robin note restored in #3682 ties back to the server-side stalls reported above. In the earlier two-client Linux test, a paused TLS exchange prevented another client receiving a response until the paused peer closed.

The client timeout behaviour is clarified, but the server-side approach remains open: bounded synchronous operations or per-connection progress, with appropriate duration and pending-output limits. Could you confirm the preferred direction, or whether a dedicated issue would be a better place to track this?

jimklimov:
Collected my thoughts on this into a separate ticket indeed, let's follow up in #3700

Issue #3329, opened 2026-02-26
Feature request: more NUT clients should be SSL-capable with proper certificate trust and all


jimklimov:
During work on #3228/#1711 I found that only `upsmon` currently has code and configuration options to set up SSL connections. Other clients just "try ssl" in their `upscli_init()` or similar calls, and happily go on in plaintext (to check: maybe they do STARTTLS trusting any server with NUT protocol support, which is marginally better than plaintext?). This may be behind the infamous "Init SSL without certificate database" messages from `upsc` - the value is never known by client nor passed to the library, and could be especially unfortunate IF it never does SSL in the first place (to check, as noted).

This relaxedly unsafe may be okay-ish for read-only anonymous streams (`upsc`, `upslog`, `upsstats.cgi`, `upsimage.cgi`, `nut-scanner`, possibly `dummy-ups` as a read-only relay -- though even there MITM attacks may be a concern for monitoring systems that base decisions on what they get from a NUT data server), but is nowadays frowned upon for authenticated sessions and so logins/passwords, variable setting and instant command sending (`upsrw`, `upscmd`, `upsset.cgi` at least; perhaps NUT-Monitor UI also, although that is rather a matter to solve with issue #1600; `dummy-ups` if it ever becomes a writing relay for INSTCMD/SETVAR) even in a local network.

This issue tracks that we have this problem, and should devise some common set of CLI arguments (especially for clients currently without config files).

Probably a more flexible and future-proof idea (especially given the difference of OpenSSL vs NSS functionality at the moment) is to define a config file for such auth details (excerpt from `upsmon.conf`?) that could be used if present, and add one option to the clients to let them select an alternative file (e.g. to use a different client cert per client program, for cases such as one NUT installation monitoring multiple "unrelated" servers).

See also:
* #1600
* #1599
* #1711
* #3328

Might be related:
* https://github.com/networkupstools/nut/issues/246#issuecomment-1310035946

CC @clepple @aquette 

UPDATE: In the test suite now updated to try SSL, at least `upsc` does report e.g.
````
Connected to NUT server localhost in SSL
Certificate verification is disabled
````

...so at least the basic crypto setup (when the server is willing) does work. Still, the client currently can't/doesn't care about server identity. Exporting `SSL_CERT_DIR` (for OpenSSL-capable builds) does not change this.

[...]

Issue #1600, opened 2022-08-19
PyNUT does not talk SSL


jimklimov:

jimklimov:
Addressed by #3352, #3372 (both released as part of NUT v2.8.5) and later work in v2.8.6 time frame.

PR #3352, opened 2026-03-16
Enhance PyNUTClient with SSL (NUT STARTTLS) support and test it

jimklimov:
Partially addresses issues:
* #1600 (no NSS key/cert db support here yet)
* #1711
* #3329 (NUT Monitor is now potentially SSL capable, to be propagated into its code/config)

AI DISCLAIMER: Prepared with contributions from IntelliJ Junie


jimklimov:
Seems python-3.13+ includes much stricter rules about certificate validation, and we generate something not up to its spec (not all extensions it wants) at least for the CA certificate.

PR #3419, opened 2026-04-14
`nutdrv_qx`: only claim USB device was supported out of  the box if `subdriver_command` was assigned 

### LINKED ISSUE #3410: nutdrv_qx: No subdriver selected for Tecnoware Era Plus 2000 (VID 0665, PID 5161)
## Description

The nutdrv_qx driver fails to autodetect the correct subdriver for 
Tecnoware Era Plus 2000 UPS without explicitly specifying vendorid 
and productid in ups.conf.

The device is correctly recognized by the OS via lsusb, but the driver 
exits with "No subdriver selected" without the explicit parameters.

## Device

- **Model:** Tecnoware Era Plus 2000
- **VendorID:** 0665
- **ProductID:** 5161
- **lsusb:** `Bus 001 Device 003: ID 0665:5161 INNO TECH USB to Serial`

## NUT Version

2.8.5

## Configuration that fails

[tecnoware]
    driver = nutdrv_qx
    port = auto
    desc = "Technoware Era Plus 2000"

## Error

Network UPS Tools 2.8.5 release - Generic Q* USB/Serial driver 0.49
USB communication driver (libusb 1.0) 0.53
No subdriver selected

## Configuration that works

[tecnoware]
    driver = nutdrv_qx
    port = auto
    desc = "Technoware Era Plus 2000"
    vendorid = 0665
    productid = 5161

## Result after fix

Using protocol: Voltronic-QS 0.10
Autodetected 1 as number of battery packs [24/27.20]
Connected to UPS [tecnoware]: nutdrv_qx-tecnoware

## upsc output (after fix)

battery.charge: 100
battery.voltage: 27.2
battery.voltage.high: 26.00
battery.voltage.low: 20.80
battery.voltage.nominal: 24.0
device.type: ups
driver.debug: 0
driver.name: nutdrv_qx
driver.parameter.pollfreq: 30
driver.parameter.port: auto
driver.parameter.productid: 5161
driver.parameter.vendorid: 0665
driver.state: quiet
driver.version: 2.8.5
driver.version.data: Voltronic-QS 0.10
driver.version.internal: 0.49
driver.version.usb: libusb-1.0.28 (API: 0x0100010A)
input.voltage: 237.1
input.voltage.fault: 237.1
output.current.nominal: 5.0
output.frequency: 50.1
output.frequency.nominal: 50
output.voltage: 237.1
output.voltage.nominal: 230
ups.beeper.status: enabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.firmware.aux: PM-V
ups.load: 7
ups.productid: 5161
ups.status: OL
ups.type: offline / line interactive
ups.vendorid: 0665

## Suggestion

Consider adding Tecnoware Era Plus 2000 (VID 0665 / PID 5161) to the 
nutdrv_qx subdriver autodetection list so that explicit vendorid/productid 
are not required.

## Reference

Related issue: #2253

jimklimov:
Closes: #3410

@DjMayone: hopefully this small change should do it. See https://github.com/networkupstools/nut/wiki/Building-NUT-for-in%E2%80%90place-upgrades-or-non%E2%80%90disruptive-tests about testing this :)

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -48,6 +48,11 @@ https://github.com/networkupstools/nut/milestone/13
  - `apc_modbus` driver updates:
     * Fixed string join not doing zero termination. [PR #3413]
 
+ - `nutdrv_qx` driver updates:
+    * Only claim a USB device as "supported" during discovery out of the box
+      if `subdriver_command` was assigned (`1A86:7523` is used by CH340/341
+      USB chips not only in UPSes). [issue #3410]
+
  - NUT client libraries:
     * Complete support for actions documented in `docs/net-protocol.txt`
       was implemented in C++, Python and PERL bindings in-tree, and for Java
--- a/drivers/nutdrv_qx.c
+++ b/drivers/nutdrv_qx.c
@@ -58,7 +58,7 @@
 #	define DRIVER_NAME	"Generic Q* Serial driver"
 #endif	/* QX_USB */
 
-#define DRIVER_VERSION	"0.49"
+#define DRIVER_VERSION	"0.50"
 
 #ifdef QX_SERIAL
 #	include "serial.h"
@@ -2567,7 +2567,7 @@ static qx_usb_device_id_t	qx_usb_id[] = {
 	{ USB_DEVICE(PHOENIXTEC_VENDORID,	0x0601),	NULL,		NULL,			&phoenix_subdriver },	/* Online Zinto A */
 	{ USB_DEVICE(UNITEK_VENDORID,	0x0001),	NULL,		NULL,			&cypress_subdriver },	/* Unitek Alpha 1200Sx */
 	{ USB_DEVICE(GE_VENDORID,	0x00c9),	NULL,		NULL,			&phoenix_subdriver },	/* GE EP series */
-	{ USB_DEVICE(QINHENG_VENDORID,	0x7523),	NULL,		NULL,			NULL },	/* Ippon Innova TAE series, using QinHeng Electronics CH340 serial converter; no specific "USB subdriver" handler defined at the moment */
+	{ USB_DEVICE(QINHENG_VENDORID,	0x7523),	NULL,		NULL,			NULL },	/* NOTE: VID:PID may be used by non-UPS devices with CH340/341 chips! But also Ippon Innova TAE series, using QinHeng Electronics CH340 serial converter; no specific "USB subdriver" handler defined at the moment */
 	{ USB_DEVICE(STMICRO_VENDORID,	0x0035),	NULL,		NULL,			&sgs_subdriver },	/* TS Shara UPSes; vendor ID 0x0483 is from ST Microelectronics - with product IDs delegated to different OEMs */
 	{ USB_DEVICE(NONAME0001_VENDORID,	0x0000),	"MEC",		"MEC0003",		&fabula_subdriver },	/* Fideltronik/MEC LUPUS 500 USB */
 	{ USB_DEVICE(NONAME0001_VENDORID,	0x0000),	NULL,		"MEC0003",		&fabula_hunnox_subdriver },	/* Hunnox HNX 850, reported to also help support Powercool and some other devices; closely related to fabula with tweaks */
@@ -2611,7 +2611,8 @@ static int qx_is_usb_device_supported(qx_usb_device_id_t *usb_device_id_list, US
 		if (usbdev->fun != NULL)
 			(*usbdev->fun)(device);
 
-		return SUPPORTED;
+		if (subdriver_command)
+			return SUPPORTED;
 
 	}
 


DjMayone:
Thank you for the quick fix! 

Based on the PR description, this looks like it should solve my issue — 
the driver was claiming 1A86/7523 (Acemagic T9 Plus internal LED 
controller) even though no subdriver was assigned, and stopping there 
instead of continuing to scan for the actual UPS (0665/5161).

I would be happy to test the fix if you can guide me on how to apply 
it safely on Home Assistant OS, otherwise I will wait for the next 
official release and confirm then.

jimklimov:
Haven't really used HA yet, so can't say for sure. There are contributed Wiki articles, e.g. https://github.com/networkupstools/nut/wiki/Building-NUT-integration-for-Home-Assistant - but I think they rely on a packaged NUT and you'd have to change the instruction a bit to supply the custom build into the container instead.

DjMayone:
Thank you for the pointer! I had a look at the wiki article, 
but the process looks quite complex for my level of experience 
with Docker and HA internals.

Since my current workaround (specifying vendorid and productid 
explicitly) works fine, I'll wait for the fix to land in an 
official NUT release and then confirm it works without 
the workaround.

Thanks again for looking into this so quickly!

jimklimov:
Fair enough. But just so you know - keep in mind that many months may pass till the next release, 2.8.5 was last week...

DjMayone:
Further investigation reveals what I believe is the root cause of the issue.

When comparing the kernel drivers bound to each USB device:

CH341 (1A86/7523 - Acemagic T9 Plus LED controllers):
- Both devices are claimed by the kernel driver 'ch341'
- They appear as /dev/ttyUSB0 and /dev/ttyUSB1
- The kernel has already bound these devices before NUT starts scanning

Tecnoware Era Plus 2000 (0665/5161):
- Managed by 'usbfs' (direct userspace access)
- No kernel driver has claimed it
- Freely accessible by NUT via libusb

This means that when nutdrv_qx scans the USB bus and encounters 
1A86/7523, it tries to claim a device that is already bound to the 
kernel ch341 driver. Instead of detecting this condition and moving 
on to the next device, the driver appears to get stuck and exits 
with "No subdriver selected".

The Tecnoware UPS (0665/5161) on the other hand sits on usbfs and 
is perfectly accessible — but NUT never reaches it.

This was confirmed by reading the kernel driver bindings directly:

/sys/bus/usb/devices/1-4/1-4:1.0/driver -> ch341      ← LED controller #1
/sys/bus/usb/devices/1-7/1-7:1.0/driver -> ch341      ← LED controller #2
/sys/bus/usb/devices/1-2/1-2:1.0/driver -> usbfs      ← Tecnoware UPS

Additional context on the CH341 chips:
- They are used by the Acemagic T9 Plus motherboard for LED ring control
- They communicate via serial at 10000 baud (non-standard)
- bcdDevice differs between the two chips: 8132 and 8134
- References:
  https://www.reddit.com/r/MiniPCs/comments/18icusg/t9_plus_n100_how_to_control_led/
  https://github.com/thekief/s1t-ledcontrol
  https://minipcunion.com/viewtopic.php?t=4424

Hope this additional technical detail helps identify the exact 
point in the code where the fix should be applied!

jimklimov:
Was the recent post re-tested with a build of the current code base (e.g. custom build of the `master` branch) with the fix from this PR applied? Or is it (an AI?) reiteration of findings with the older code?

DjMayone:
Ciao Jim,
Good question! To clarify:

The recent findings (kernel driver bindings, USB descriptor analysis) 
are new investigations done on the live system, not a reiteration of 
previous findings. I am using AI assistance to guide the debugging 
process, but all commands are run directly on the actual hardware and 
results verified by me.

However, these tests were done with the original NUT 2.8.5 build 
(without the fix from this PR applied). We attempted to test a custom 
build of the PR code, but ran into compatibility issues when trying to 
deploy it in the Home Assistant OS container environment.

The key new finding is that the CH341 devices (1A86/7523) are already 
claimed by the kernel ch341 driver, and when NUT scans without 
vendorid/productid, it causes ttyUSB1 to permanently disconnect 
(the kernel releases the ch341 driver but never reattaches it).

So to directly answer your question: no, the fix from this PR has not 
yet been tested on this system with the current code base. We are 
waiting for it to land in an official release.
Thank you!

jimklimov:
Aha, thanks for the clarifications. 

From this angle, one thing that comes to mind is to revise the `udev.rules` or what you have as equivalent on that system, with partial configuration files coming from a NUT package, so that the devices which on your system would be real serial ports and not UPS USB-to-serial converters would not be owned by `nut` user and so it would not be able to attach to them (and detach someone else). Possibly it is these permissions that somehow block their re-attachment as well.

Another approach might be to look at it through the container lens, and ensure only needed `/dev` or `/sys` filesystem nodes are visible (passed-through) in the container operating environment for the NUT HA plugin, and others just do not exist as far as it is concerned. I can't really say *how* to achieve that on HA, other than this sounds reasonable and should be possible :D

DjMayone:
Further update: we tested the container isolation approach you suggested.

By running nutdrv_qx in a temporary container with only the Tecnoware 
device passed through (--device=/dev/bus/usb/001/003), the driver 
worked perfectly without specifying vendorid/productid:

- All CH341 devices (1A86/7523) were skipped with 
  "Failed to open device, skipping: No such device"
- The Tecnoware UPS (0665/5161) was found and claimed successfully
- Protocol matched: Voltronic-QS 0.10
- Driver entered normal update loop

This confirms your hypothesis: the issue is purely about device 
visibility in the container environment. When the CH341 devices 
are not accessible to the container, nutdrv_qx has no problem 
finding and using the actual UPS.

The challenge now is how to achieve this in the Home Assistant 
add-on environment, since the NUT add-on container has the entire 
/dev filesystem mounted (read-only) from the host, and this is 
managed by HA Supervisor with no easy way to restrict it.

As a reference, from the container inspect:
  Mounts: Source="/dev" Target="/dev" ReadOnly=true
  DeviceCgroupRules: "c 189:* rwm" (all USB devices)

So the fix in the driver code (PR #3419) remains the most practical 
solution for HA users, as the container isolation approach would 
require changes to how HA Supervisor manages device access for add-ons.

We are using AI assistance to guide the investigation, but all 
tests are run directly on the actual hardware and results verified.

PR #3425, opened 2026-04-29
drivers: fix CPS HID input/output voltage decoding for 0764:0601

jeje232:
## Summary

This patch fixes incorrect CPS HID voltage decoding for devices with:

- vendor ID: `0764`
- product ID: `0601`

In my case, the UPS LCD showed about 229V input/output, while NUT reported values around 22-26V.

## Problem

The raw HID reports already contain the correct voltage information in the payload bytes.

Examples observed during debugging:

- `0f e5 08` -> little-endian payload `0x08e5` -> `2277` -> `227.7V`
- `0f f6 08` -> little-endian payload `0x08f6` -> `2294` -> `229.4V`
- `12 0c 09` -> little-endian payload `0x090c` -> `2316` -> `231.6V`

At the same time, the UPS front panel displayed about 229V.

So the device appears to send the correct raw voltage, but NUT decodes it incorrectly for CPS voltage report IDs.

## Fix

The patch corrects the raw value reconstruction for CPS HID voltage report IDs:

- `0x0F` (`UPS.Input.Voltage`)
- `0x12` (`UPS.Output.Voltage`)

This allows the final decoded values to match the actual mains voltage shown by the UPS.

## Validation

Tested on a real CyberPower/PowerWalker-family UPS with:

- `vendorid = 0764`
- `productid = 0601`

After the fix:

- `input.voltage` is reported around 230V
- `output.voltage` is reported around 230V
- values match the UPS LCD

## Notes

While debugging this, I also found that the current `cps_fix_report_desc()` logic assumes the presence of `HVT(0x10)`, but this object was not present on my device. The decisive issue for this case was the voltage raw-value decoding path.


jimklimov:
Why were existing comments, debug printouts and blank lines removed, and other blanks added? Was this coded by yourself, or by an AI assistant (the removals look like a misconfiguration of the latter)?

Also, `libhid.c` is (or aims to be) generic, with minimal awareness about specific devices. If it is possible to wedge the `debug_maybe_fix_cps_voltage_raw_value()` fix into `cps_fix_report_desc()` so it only ever impacts these devices, please do so.

PR #3430, opened 2026-05-08
Fix NUT-Monitor tray tooltip text

ostrich:
## Summary

NUT-Monitor reused the rich-text status label string as the Qt tray tooltip. On KDE Plasma 6, that tooltip text is exposed as literal HTML, so the tray tooltip showed tags such as `<font>`, `<b>`,
and `<br>`.

This change builds a separate plain-text tooltip for the Qt5 and Qt6 monitor variants while preserving the existing rich-text formatting in the main window status labels.

## Testing

- Built and installed locally
- Verified KDE Plasma 6 tray tooltip shows plain status text without HTML tags
- Verified no regressions in the main window status display


jimklimov:
Thanks for your contribution!

PR #3429, opened 2026-05-07
Fix NSS bundling for Windows clients, do not abort `upsmon` if SSL was not required (and we failed to set it up)

jimklimov:
Closes: #3420

* Windows: The 7-zip archive from AppVeyor (and proto area made by our Makefile) did not deliver some libraries apparently needed by `nss3.dll` but not referenced from it in a way visible to `objdump`, `ldd` and similar tools. Hopefully `strings | grep` would fare better to catch such cases.
* All platforms: we no check more diligently for (in)ability to set up SSL, so even the `NoDB` mode in NSS builds failed if the library could not load (due to trouble above). Currently `upsmon` aborted the start-up if it could not initialize SSL (also in older releases), but now this happened with runs that did not require the SSL at all.
  * The method did `return -1` in some failure cases earlier too, so this is not a newly introduced change of behavior. Just there is now more abilities and more error-checking inside.

Issue #3420, opened 2026-04-15
Because SSL is not set up upsmon will not connect on Windows


andrew-nprime:
I have been using CI snapshot builds on Windows with an APC UPS and for version 2.8.4.x it worked fine and I never thought about setting up SSL.  However, 2.8.4.3467 build is the most recent one I have tried that allows upsmon to connect to upsd without SSL, despite SSL not being configured.  I did try to set up SSL for it, but so far have failed.  I did get the SSL setup to work with NSS on my Ubuntu server, however, so my set up shouldn't be too far off.  So any help getting the SSL feature to work on Windows would be great, or allowing the upsmon to connect to upsd without using SSL would also be very helpful.

Newer upsmon that fails to connect:

```
C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.5.4570-master\NUT-for-Windows-x86_64-SNAPSHOT-2.8.5.4570-master\mingw64>sbin\upsmon.exe -DDDD
Network UPS Tools upsmon 2.8.5.106-106+g68fb8bdb7 (development iteration after 2.8.5) using a dynamic libnutprivate*
   0.001874     [D3] setproctag: starting first tagging as 'init'...
   0.004593     [D3:libupsclient] setproctag: starting first tagging as 'init'...
   0.005429     [D1:libupsclient:init] upscli_init_default_connect_timeout: upscli_default_connect_timeout=10.000000 sec assigned from: default_secs
   0.006298     [D1:init] Just failed to send signal, no daemon was running
   0.007341     [D1:init] loadconfig: Loading C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.5.4570-master\NUT-for-Windows-x86_64-SNAPSHOT-2.8.5.4570-master\mingw64\sbin\..\etc/upsmon.conf
   0.008859     UPS: UPS1@nprimeserver01 (primary) (power value 1)
   0.009460     Using power down flag file C:\Temp\killpower
   0.010411     [D1:init] debug level is '4'
   0.010551     [D1:init] will use a shutdown command (SHUTDOWNCMD): 'C:\WINDOWS\system32\shutdown.exe -s -t 0'
   0.011343     [D1:init] will use custom notification command (NOTIFYCMD): 'C:\NUT-for-Windows-x86_64-SNAPSHOT-2.8.5.4570-master\NUT-for-Windows-x86_64-SNAPSHOT-2.8.5.4570-master\mingw64\sbin\message.exe'
[...]

PR #3432, opened 2026-05-11
Refine `scripts/Windows/dllldd.sh` helper script

jimklimov:
Follow-up to PR #3429 for issue #3420

Try to optimize the helper script here and there so it does less work and completes faster. Spawning processes in MSYS2/MinGW emulation is catastrophically slow :\


jimklimov:
Results analysis (comparison to earlier merged iterations) at https://github.com/networkupstools/nut/pull/3429#issuecomment-4420414188

PR #3437, opened 2026-05-12
Fix libusb threads when backgrounded (and split `background()` method into parent/child phases)

dougnazar:
This is needed in particular with `libusb` on linux as it will create
a thread to handle either `netlink` or `udev` messages. If this is created
before we background, the driver will hang during exit trying to join
the non-existant thread in the new process. Shows this warning while
stopping the driver and is delayed until it switches to `SIGKILL`.

`Stopping /run/nut/usbhid-ups-xxxx.pid failed, retrying harder: Success`

I picked closing and then re-opening as the simpler to implement & test,
but has the biggest impact on each driver. Quite possibly some/all might
need to be modified to allow re-opening like the `usb-hid` driver.

The other option I considered was breaking up the backgrounding into stages,
creating the background process early, but leaving the forground process
running to report startup messages until initialization is finished.

Or perhaps there is another option?


jimklimov:
I haven't seen the described behavior, I think. Is this something that changed in recent libusb releases and/or Linux distros/kernel?

At what point does libusb start that thread - some method we call (so might defer that until after backgrounding), or somehow automatically during library load?

dougnazar:
Looks like the thread creation was added in `v1.0.16` (*6853291ed Add hotplug support to the Linux backend.*) and changed to `pthread_join` for `v1.0.17` (*3107f30ba linux_netlink: Remove use of pthread_cancel*). I'm currently running `v1.0.21`.

It happens as soon as libusb is initialized, and cancelled when closed.  The call trace is:
```
upsdrv_callbacks.upsdrv_initups();
	upsdrv_initups()  [usbhid-ups.c]
		ret = comm_driver->open_dev(&udev, &curDevice, subdriver_matcher, &callback);
			nut_libusb_open()
				libusb_init()
					libusb_init_context()
						usbi_backend.init();
							op_init()
								linux_start_event_monitor()
									linux_netlink_start_event_monitor()
									or
									linux_udev_start_event_monitor()
```


Edit: It just occurred to me that this might only happen on sysvinit style systems. I'm guessing that systemd handles backgrounding itself and wont see this issue.

jimklimov:
Thanks for the details, never knew that nuance.

As for systemd vs. init systems, I suppose this would have popped up in BSDs etc. as well? Otherwise, the difference is not so much in systemd itself, as in the `*.service` unit definitions which let the NUT daemons stay "foregrounded" since systemd does the forking. In services for NUT v2.7.x IIRC this was not the case, so there was double (or more) forking with systemd, backgrounding/detachment from terminal, possibly the root/nonroot split in upsmon, which blurred the lines about which PID(s) to monitor as part of the service unit.

Previously the only option to stay foregrounded was to hack the init script or similar, to add `-D` for debugging (which coincidentally also left the process foregrounded). With NUT v2.8.0 this was resolved to keep debugging and fore-/back-grounding separable (`-F`, `-B`) with `-D` just retaining legacy behavior about this by default; later also `-FF` to stay foregrounded and still save the PID file (so signals can be sent easily).

FWIW, the `retrying harder` wording comes from `upsdrvctl`.

jimklimov:
Looking around for context, the idea about `upsdrv_initups` happening long before backgrounding is:
* sometimes we do not background at all, e.g. to run jut to handle `killpower` request quickly or to `dump_data` or to debug with default other settings, or if the platform won't let us (e.g. Windows)
* if the device was not found etc. we do not want to start the driver but rather fail early - hard(er) to do so visibly when detached - but this might be solvable with an `upsmon`-style pipe I suppose (keep a parent process around for a while, initups in the child, signal back to parent whether it should exit with success or error; maybe even pass the text message to the parent to print in case of error), all this complexity subject to whether we want to background at all. I think this PR should morph towards something like this approach?.. (UPDATE: I re-read the original post and this is essentially your "other option", so that path looks good to me)

As for the libusb threads, I wonder if they should not be inherited by the child process? I guess when the parent exits, the library closes the context and reaps the threads - so there's none to `join` for the child (and also maybe this is what breaks driver reconnections for some people)? Maybe there's a way to just NOT cleanly close libusb when the parent exits?

To this end, currently `drivers/main.c` goes like this (in broader strokes, since at least 2008; `exit_upsdrv_cleanup()` was separated to wrap `upsdrv_cleanup()` in 2023):
```
        dstate_setinfo("driver.state", "init.starting");

        atexit(exit_cleanup);
...
        dstate_setinfo("driver.state", "init.device");
        upsdrv_callbacks.upsdrv_initups();
        dstate_setinfo("driver.state", "init.quiet");

        /* UPS is detected now, cleanup upon exit */
        atexit(exit_upsdrv_cleanup);

        /* now see if things are very wrong out there */
        if (upsdrv_callbacks.upsdrv_info->status == DRV_BROKEN) {
                fatalx(EXIT_FAILURE, "Fatal error: broken driver. It probably needs to be converted.\n");
        }
...
        if (do_forceshutdown) {
                dstate_setinfo("driver.state", "fsd.killpower");
                forceshutdown();
        }
...
        switch (foreground) {
                case 0:
                        background();
...
```

There are several potential `exit()` points between that `atexit(exit_upsdrv_cleanup);` and the possible `background();` call, so this hook would be useful there to un-initialize the device when the mono-process exits before it gets to backgrounding. I think `exit_upsdrv_cleanup()` and some global variables to hint about initial/current PID of the process can be used to just not-call `upsdrv_cleanup()` if this is a parent process and we were just about to process a `background()` call (so some global variable about that, not the `foreground` intent variable).

I don't know if that would actually help (e.g. `libusb` may well have its own `atexit` registrations to reap those threads and close `netlink` etc. listeners), but this looks like something that can be tested quickly - and if it does not help, then go to re-architecting how the backgrounding works.

dougnazar:
> message to the parent to print in case of error), all this complexity subject to whether we want to background at all. I think this PR should morph towards something like this approach?.. (UPDATE: I re-read the original post and this is essentially your "other option", so that path looks good to me)
> 

Okay, this didn't turn out as complicated as I thought it might be. I deconstructed `background()` into a couple helper functions and then re-used those. There is a slight change in that the logging changes used to happen for both parent and child but I needed to move that to happen only for the child. We exit the parent immediately so it should not matter.

 
> As for the libusb threads, I wonder if they should not be inherited by the child process? I guess when the parent exits, the library closes the context and reaps the threads - so there's none to `join` for the child (and also maybe this is what breaks driver reconnections for some people)? Maybe there's a way to just NOT cleanly close libusb when the parent exits?
> 

It is my understanding that `fork()` doesn't copy threads (at least on Linux). Also, the parent uses `_exit()` which bypasses all `atexit()` functions.

jimklimov:
At least for WIN32 builds, this PR is not right in some nuance. Probably a missed or misplaced `ifdef`, so POSIX code remains visible there - see Appveyor CI complaints:

```
common.c: In function 'background':
common.c:815:17: error: variable 'pid' set but not used [-Werror=unused-but-set-variable]
  815 |         int     pid;
      |                 ^~~
common.c: In function 'background_child':
common.c:852:35: error: initialization of 'HANDLE' {aka 'void *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
  852 |                 TYPE_FD devnull = open("/dev/null", O_RDWR);
      |                                   ^~~~
common.c:853:29: error: ordered comparison of pointer with integer zero [-Werror=pedantic]
  853 |                 if (devnull < 0)
      |                             ^
common.c:856:26: error: passing argument 1 of 'dup2' makes integer from pointer without a cast [-Wint-conversion]
  856 |                 if (dup2(devnull, STDIN_FILENO) != STDIN_FILENO)
      |                          ^~~~~~~
      |                          |
      |                          HANDLE {aka void *}
In file included from C:/msys64/mingw64/include/fcntl.h:8,
                 from ../include/common.h:46,
                 from common.c:23:
C:/msys64/mingw64/include/io.h:349:24: note: expected 'int' but argument is of type 'HANDLE' {aka 'void *'}
  349 |   int __cdecl dup2(int _FileHandleSrc,int _FileHandleDst) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |                    ~~~~^~~~~~~~~~~~~~
common.c:858:26: error: passing argument 1 of 'dup2' makes integer from pointer without a cast [-Wint-conversion]
  858 |                 if (dup2(devnull, STDOUT_FILENO) != STDOUT_FILENO)
      |                          ^~~~~~~
      |                          |
      |                          HANDLE {aka void *}
C:/msys64/mingw64/include/io.h:349:24: note: expected 'int' but argument is of type 'HANDLE' {aka 'void *'}
  349 |   int __cdecl dup2(int _FileHandleSrc,int _FileHandleDst) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |                    ~~~~^~~~~~~~~~~~~~
common.c:861:34: error: passing argument 1 of 'dup2' makes integer from pointer without a cast [-Wint-conversion]
  861 |                         if (dup2(devnull, STDERR_FILENO) != STDERR_FILENO)
      |                                  ^~~~~~~
      |                                  |
      |                                  HANDLE {aka void *}
C:/msys64/mingw64/include/io.h:349:24: note: expected 'int' but argument is of type 'HANDLE' {aka 'void *'}
  349 |   int __cdecl dup2(int _FileHandleSrc,int _FileHandleDst) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |                    ~~~~^~~~~~~~~~~~~~
common.c:865:23: error: passing argument 1 of 'close' makes integer from pointer without a cast [-Wint-conversion]
  865 |                 close(devnull);
      |                       ^~~~~~~
      |                       |
      |                       HANDLE {aka void *}
C:/msys64/mingw64/include/io.h:345:25: note: expected 'int' but argument is of type 'HANDLE' {aka 'void *'}
  345 |   int __cdecl close(int _FileHandle) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |                     ~~~~^~~~~~~~~~~
At top level:
make[3]: *** [Makefile:1189: libcommon_la-common.lo] Error 1
```

dougnazar:
> At least for WIN32 builds, this PR is not right in some nuance. Probably a missed or misplaced `ifdef`, so POSIX code remains visible there - see Appveyor CI complaints:

I was more waiting to see if there were any issues with the rest of the patch. Didn't want to push another version just for the WIN32 fixes. Anyways, that should all be cleaned up now.

jimklimov:
NOTE: CI faults on freebsd15 may be due to baby troubles of the updated CI build agent, rather than a problem of tested code per se.

jimklimov:
Thanks! Let's hope nothing explodes :D

PR #3447, opened 2026-05-17
ragtech: new driver for Ragtech UPS (USB CDC-ACM, family 10)

juslex:
## Summary

Adds a new `ragtech` driver supporting the Ragtech "Easy Pro" family of
line-interactive UPS units (also sold under the **NEP**, **TORO**,
**INNERGIE** and **OneUP** brands in Brazil). Devices expose a USB CDC-ACM
serial interface (`VID 0x04D8` / `PID 0x000A`, Microchip PIC firmware) and
speak a proprietary 6-byte register-access protocol.

- Single standalone serial driver in `drivers/ragtech.c` (no shared
  sub-driver layer — Ragtech is not a Megatec/Qx variant).
- 20 models from the OEM "family 10" device table populate `ups.model`,
  `ups.realpower.nominal`, output voltage/current scaling and battery
  voltage scaling.
- `input.voltage.nominal` switches between 115V and 220V dynamically based
  on the measured input (TI vs M2 variants).
- Reverse-engineered from `strace` of the OEM `supsvc` binary and the OEM
  `devices.xml`, cross-checked against
  [antunesls/UPS_ESP32_tinySrv](https://github.com/antunesls/UPS_ESP32_tinySrv).

### Protocol (decoded from OEM traffic)

Every command on the wire is 6 bytes: `0xAA OPCODE ADDR_HI ADDR_LO VALUE CHECKSUM`
where `CHECKSUM = (ADDR_HI + ADDR_LO + VALUE) & 0xFF`.

| OPCODE | Meaning |
|---|---|
| `0x01` | Write byte |
| `0x02` | AND mask (atomic bit-clear) |
| `0x03` | OR mask (atomic bit-set) |
| `0x04` | Read range |

### Instant commands

- `shutdown.stayoff` — byte-for-byte identical to the OEM `supsvc` sequence.
  **Verified on hardware**: cuts output and stays off until manual power-on
  in both AC and battery modes.
- `shutdown.stop` — aborts an armed countdown. **Verified on hardware.**
- `shutdown.return` — implemented as a `shutdown.stayoff` with a
  `LOG_WARNING`. The Easy 2000 TI firmware does **not** auto-restart on
  mains return; the operator must press the power button. The driver
  surfaces this consistently to `upsmon` callers rather than silently
  failing.
- `test.battery.start.deep` — bit writes matching the OEM `fullDischarge`
  action. Implemented but **not yet verified end-to-end** (would empty the
  battery during testing).

### Safety: shutdowns are opt-in by default

Because the firmware does not auto-restart, an unintended `shutdown.return`
(e.g. from `upsmon` on a low-battery event) would leave the UPS off until
somebody physically presses the power button. To avoid silent-fail
integrations, every `shutdown.*` and `test.battery.start.deep` instcmd is
gated behind `allow_shutdown` in `ups.conf`:

- Without the flag: instcmds are not registered, direct socket requests are
  rejected with `STAT_INSTCMD_INVALID`, and `upsdrv_shutdown` exits with
  failure rather than cutting output.
- With the flag: behaviour as designed.

The default-deny posture is conservative on purpose; a future Ragtech model
with working auto-restart could carry a `firmware_returns` capability in
the model table to flip the default.

### Wire-level notes

- `ser_get_buf_len()` is **all-or-nothing**: it discards already-collected
  bytes on timeout. CDC-ACM fragments replies into 1–8 byte chunks, so the
  driver uses `ser_get_buf()` in a manual loop.
- `ser_flush_in()` does not call `tcflush()`; under `O_NONBLOCK` USB CDC
  bytes inside the tty layer survive its select+read loop. The driver uses
  `ser_flush_io()` (which calls `tcflush(TCIOFLUSH)`) before each TX.
- The driver does **not** call `ser_set_speed()` on the CDC-ACM endpoint —
  `tcsetattr(TCSANOW)` pulses DTR on some Linux tty drivers, which the
  Ragtech firmware can interpret as a shutdown signal. CDC-ACM ignores
  baud at the wire anyway.
- DTR/RTS are forced to 0 immediately after `ser_open` (the UPS reads
  non-zero levels as a remote-shutdown signal).
- `tcdrain()` + ~50 ms of slack after every `ser_send_buf()` to let the
  firmware compose the reply.

## Files changed

- `drivers/ragtech.c` — new driver (single C file).
- `drivers/Makefile.am` — registered in `SERIAL_DRIVERLIST`, `ragtech_SOURCES`, `ragtech_LDADD`.
- `docs/man/ragtech.txt` — new AsciiDoc man page.
- `docs/man/Makefile.am` — three registrations (`.txt`, `MAN_SECTION_CMD_SYS`, `.html`).
- `data/driver.list.in` — 20 entries for the family-10 models (level "1",
  reverse engineering).
- `docs/nut.dict` — added `INNERGIE`, `NEP`, `OneUP`, `Ragtech`, `TORO` and
  bumped the word count.
- `NEWS.adoc` — new-driver entry under 2.8.6.

## Driver status

`DRV_EXPERIMENTAL` at version `0.04`. Only the Easy 2000 TI has been
validated end-to-end with hardware; the other 19 models share the same
register layout and scaling table but have not been hardware-tested yet.

## Test plan

- [x] Polling status and `ups.*` variables on Easy 2000 TI — verified.
- [x] `shutdown.stayoff` — verified on hardware (AC and battery modes).
- [x] `shutdown.stop` — verified on hardware (aborts armed countdown).
- [x] `iout_calib` override via `ups.conf` — verified.
- [x] `allow_shutdown` default-deny posture — verified (instcmds not
      registered, `upsdrv_shutdown` exits with failure).
- [ ] `shutdown.return` end-to-end — not verified (firmware does not
      auto-restart; surfaces as `stayoff` + warning).
- [ ] `test.battery.start.deep` — implemented but not verified
      (deliberately not exercised during RE — would discharge battery).
- [ ] Other 19 family-10 models — community testing requested.

Welcoming feedback on the `shutdown.return` design (currently warns and
stays off), and on whether to fold the model discrimination into a
per-model scaling table once a second model is verified.


juslex:
@jimklimov I used the claude code to initially get the communication information made by other projects, especially https://github.com/antunesls/UPS_ESP32_tinySrv. After that, I started new tests and ended up discovering more information. I'm currently running a version of nut with this proposed PR and it's working perfectly. I'm not a developer, I used the claude code for that. But I'm available to try to help. Thank you!

Rafael-Fioramonte:
Test report — QDW 1200 CBU-TI (data point, possibly unrepresentative hardware)
Tried the ragtech driver (v0.06, commit 92cd420a6257a6ba1aad2898b56fa925b1a62a02) against a unit badged Ragtech QDW 1200 CBU-TI — Quadri line, not in the family-10 table. USB ID matches: 04d8:000a, iManufacturer "Ragtech Sistemas de Energia".
Symptom: ragtech -a … -DDD ends with ragtech_read(0x0080,30): short RX (0/30) — no reply to the initial poll. A usbmon capture of the run (attached) shows two 6-byte Bulk-OUT packets, both delivered successfully at the USB layer:

ff fe 00 8e 01 8f (sent first — does not begin with 0xAA, unclear to me if intentional)
aa 04 00 80 1e 9e (the valid read-range request for reg 0x80)

Every subsequent Bulk-IN completes -2 / 0 bytes — the device returns nothing to either packet. USB transport itself works; the UPS just doesn't answer a correctly-formed frame.
Caveats — please weigh accordingly: the UPS's comms board was replaced by a repair shop ~1 year ago and may not be original; the installed board is custom — riveted in, no legible markings. On an earlier run I heard a relay click in the UPS on port open and power to it's outlets was cut (brief 10 seconds then self-recovered); I couldn't re-confirm it this time when I used usbmon as the battery was depleted, but the usb port was still powered. So this hardware may not represent a genuine Quadri — reporting as a data point, not asserting a driver bug.
Attached: usbmon trace, full -DDD log, lsusb -v, build/version info.
[ragtech-report.tar.gz](https://github.com/user-attachments/files/28182172/ragtech-report.tar.gz)

juslex:
> Test report — QDW 1200 CBU-TI (data point, possibly unrepresentative hardware) Tried the ragtech driver (v0.06, commit [92cd420](https://github.com/networkupstools/nut/commit/92cd420a6257a6ba1aad2898b56fa925b1a62a02)) against a unit badged Ragtech QDW 1200 CBU-TI — Quadri line, not in the family-10 table. USB ID matches: 04d8:000a, iManufacturer "Ragtech Sistemas de Energia". Symptom: ragtech -a … -DDD ends with ragtech_read(0x0080,30): short RX (0/30) — no reply to the initial poll. A usbmon capture of the run (attached) shows two 6-byte Bulk-OUT packets, both delivered successfully at the USB layer:
> 
> ff fe 00 8e 01 8f (sent first — does not begin with 0xAA, unclear to me if intentional) aa 04 00 80 1e 9e (the valid read-range request for reg 0x80)
> 
> Every subsequent Bulk-IN completes -2 / 0 bytes — the device returns nothing to either packet. USB transport itself works; the UPS just doesn't answer a correctly-formed frame. Caveats — please weigh accordingly: the UPS's comms board was replaced by a repair shop ~1 year ago and may not be original; the installed board is custom — riveted in, no legible markings. On an earlier run I heard a relay click in the UPS on port open and power to it's outlets was cut (brief 10 seconds then self-recovered); I couldn't re-confirm it this time when I used usbmon as the battery was depleted, but the usb port was still powered. So this hardware may not represent a genuine Quadri — reporting as a data point, not asserting a driver bug. Attached: usbmon trace, full -DDD log, lsusb -v, build/version info. [ragtech-report.tar.gz](https://github.com/user-attachments/files/28182172/ragtech-report.tar.gz)

Hey, thanks for the detailed report. Looking at your usbmon: both OUT packets are delivered fine (status 0, 6/6 on the handshake and on the read range), and every IN completes only after the port closes with -2 / 0 bytes. So the USB layer is happy and the firmware on your unit just never replies. Not really a driver bug, the device is staying silent.

This PR only covers Easy Pro / family 10 (the model table we built from the OEM devices.xml). The QDW is from the Quadri line, which uses a different family byte at register 0x9A and presumably a different register map too. Bytes on the wire might look similar but we don't have a decode for Quadri yet, so the model table doesn't claim to support it.

A couple of things in your report line up with it being different firmware. The family-10 units don't auto-recover after a coordinated shutdown (once they cut output they stay off until the physical button is pressed), but yours came back on its own in 10 seconds. That's closer to what aftermarket / non-OEM boards tend to do, which matches your caveat about the comms board having been replaced.

About the handshake bytes, the 0xAA SOF only applies to the read/write/AND/OR opcodes. The wake-up probe sent before that is a different thing, that's why it doesn't start with 0xAA. The driver source could be clearer about that, I'll add a note.

If you want to push the QDW case further, the most useful thing would be a capture of the OEM Ragtech Windows tool (Supervise) talking to your unit, USBPcap or similar. That's basically how the family-10 map got reverse engineered. With one good capture we'd have a baseline to start a separate Quadri PR from later on.

Really appreciate the effort here, keeping the tarball around for when more Quadri data shows up.

jimklimov:
NOTE: CI faults on freebsd15 may be due to baby troubles of the updated CI build agent, rather than a problem of tested code per se. There are however several other platforms whose complaints are more on-target.

jimklimov:
Some of the troubles that had upset CI about the earlier iteration are tacked in new PR #3453 more generally, others posted here (similar issues introduced by this PR and so exposing some more general problems).

jimklimov:
Thanks!

Issue #2453, opened 2024-05-22
Weird connectivity issues with a generic ViewPower UPS


hmv47:
I run Debian 11 with NUT 2.8.0 and connect a [generic ViewPower UPS](https://www.eltrox.pl/p/35383-ups-zasilacz-awaryjny-volt-polska-micro-1000va-600w-1x9ah). It gets identified:
```
Bus 001 Device 002: ID 0665:5161 Cypress Semiconductor USB to Serial
```
I tried running `nut-scanner`:
```
Scanning USB bus.
No start IP, skipping NUT bus (old connect method)
Scanning NUT bus (avahi method).
Failed to create client: Daemon not running
[nutdev1]
        driver = "nutdrv_qx"
        port = "auto"
        vendorid = "0665"
        productid = "5161"
        product = ""
        bus = "001"
```
But it doesn't get detected at all with `nutdrv_qx`. So I tried `blazer_usb` and it _sort of_ worked:
```
[nutdev1]
    driver = blazer_usb
    subdriver = cypress
    protocol = mustek
    port = auto
    desc = "UPS"
    vendorid = 0665
    productid = 5161
```
`upsc` sometimes return status from the UPS, but very often I get the stale data error.

Additionally, after reboots, randomly, USB device is not recognized properly e.g. it's shown as this:
```
Bus 001 Device 002: ID ffff:0000
```
And doesn't work at all with either driver.
I'm pretty positive this is a hardware problem and probably a flaky USB2Serial converter in the UPS, but still I'm opening this issue; I'd like to see if anyone else has had similar problems, or maybe knows a solution.

hmv47:
I experimented a bit more, with the latest 2.8.2 version
**/usr/local/ups/bin# nut-scanner**
```
Scanning USB bus.
No start IP, skipping NUT bus (old connect method)
Scanning NUT simulation devices.
[nutdev-usb1]
	driver = "nutdrv_qx"	# alternately: blazer_usb
	port = "auto"
	vendorid = "0665"
	productid = "5161"
	product = ""
	# bus = "001"
	# device = "009"
	# busport = "001"
```
**/usr/local/ups/bin/nutdrv_qx -a nutdev-usb1 -u root -DD**
```
Network UPS Tools - Generic Q* USB/Serial driver 0.36 (2.8.2)
USB communication driver (libusb 1.0) 0.47
   0.000000	[D1] upsdrv_makevartable...
[...]

Issue #993, opened 2021-03-17
Not work with IPPON Back Comfo Pro II


bdn76:
`
# nut-scanner 
[nutdev1]
	driver = "blazer_usb"
	port = "auto"
	vendorid = "0665"
	productid = "5161"
	product = "USB to Serial"
	serial = "20100826"
	vendor = "INNO TECH"
	bus = "001"
`

`
# /usr/lib/nut/blazer_usb -a nutdev1
Network UPS Tools - Megatec/Q1 protocol USB driver 0.12 (2.7.4)
No supported devices found. Please check your device availability with 'lsusb'
and make sure you have an up-to-date version of NUT. If this does not help,
try running the driver with at least 'subdriver', 'vendorid' and 'productid'
options specified. Please refer to the man page for details about these options
(man 8 blazer_usb).
`

Try to add subdriver = "ippon" without success.

Issue #598, opened 2018-08-22
Communications with UPS lost


kosfango:
Hi All,

I have FreeBSD installed:
FreeBSD server 11.1-BETA2 FreeBSD 11.1-BETA2 #0 r320065
and nut-2.7.4_8 prot port collections

I see in logs a lot of lines:
[skip]
Aug 22 18:47:27 server blazer_usb[77549]: Communications with UPS re-established
Aug 22 18:47:27 server upsd[77551]: UPS [Ippon] data is no longer stale
Aug 22 18:47:28 server upsmon[1374]: Communications with UPS ippon@localhost established
Aug 22 18:47:42 server blazer_usb[77549]: Communications with UPS lost: status read failed!
Aug 22 18:47:44 server upsd[77551]: Data for UPS [Ippon] is stale - check driver
Aug 22 18:47:48 server upsmon[1374]: Poll UPS [ippon@localhost] failed - Data stale
Aug 22 18:47:48 server upsmon[1374]: Communications with UPS ippon@localhost lost
Aug 22 18:47:53 server blazer_usb[77549]: Communications with UPS re-established
Aug 22 18:47:53 server upsd[77551]: UPS [Ippon] data is no longer stale
Aug 22 18:47:54 server upsmon[1374]: Communications with UPS ippon@localhost established

When communication is lost:
# upsc ippon@localhost
Error: Data stale

When all is fine I see next:
# upsc ippon@localhost
battery.charge: 100
battery.voltage: 13.10
battery.voltage.high: 13.00
battery.voltage.low: 10.40
battery.voltage.nominal: 12.0
device.type: ups
driver.name: blazer_usb
driver.parameter.pollinterval: 2
driver.parameter.port: /dev/cuau1
driver.parameter.synchronous: no
driver.version: 2.7.4
driver.version.internal: 0.12
input.current.nominal: 2.0
input.frequency: 49.9
input.frequency.nominal: 50
input.voltage: 225.4
input.voltage.fault: 225.8
input.voltage.nominal: 220
output.voltage: 225.8
ups.beeper.status: enabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.load: 13
ups.productid: 5161
ups.status: OL
ups.temperature: 25.0
ups.type: offline / line interactive
ups.vendorid: 0665

but sometimes I see more shortly output:
# upsc ippon@localhost
battery.voltage: 13.10
battery.voltage.high: -1.08
battery.voltage.low: -0.87
device.type: ups
[...]

PR #3448, opened 2026-05-18
drivers/libusb1.c+nutdrv_qx.c: fix per-close libusb_exit() deadlock; escalate persistent LIBUSB_ERROR_OVERFLOW to USB reset

Pyker:
## Summary

This PR has been revised since the initial submission. The original
single-commit clear-halt-and-retry approach turned out to be insufficient
in soak: some firmware-stuck states recover only after a full USB-level
device reset, and triggering that reset exposed a longstanding deadlock
in `drivers/libusb1.c`'s teardown path. The PR now ships two commits
that together address both halves:

1. **`8574a2ee9`** (`drivers/libusb1.c`): use an explicit libusb
   context, drop the `libusb_exit()` call from `nut_libusb_close()`.
   This was a misuse (the documented contract is "call only after
   closing all open devices and before your application terminates");
   on certain firmware it wedged indefinitely when the 1->0 refcount
   transition tore the context down with orphaned URBs still pending.
   The bug class is reachable from any reconnect path; the OVERFLOW
   escalation in commit 2 just makes it deterministic on the 0665:5161
   family.

2. **`6ea5fcea5`** (`drivers/nutdrv_qx.c`): counter-gated
   escalation. On `qx_command()` receiving `LIBUSB_ERROR_OVERFLOW`,
   increment a per-driver counter; after 3 consecutive overflows, call
   `libusb_reset_device()` to recover the firmware out of its stuck
   state, then reuse the existing fallthrough reconnect path. Any clean
   read resets the counter, so isolated transient overflows don't
   trigger the reset.

Commit order is bisect-friendly: every intermediate state on the branch
is independently safe. Commit 1 fixes the latent deadlock without
changing OVERFLOW behavior; commit 2 introduces the escalation only
after the deadlock cure is in place.

## Background

The byte-level root cause analysis for the OVERFLOW symptom is in
[this comment on #598][analysis]: firmware in the 0665:5161 Cypress
USB-Serial bridge family (Salicru SPS, Ippon, Belkin F6C1200-UNV,
ViewPower, various Voltronic Power UPSes) occasionally emits an
interrupt-IN frame larger than the declared `wMaxPacketSize=8`. xHCI
flags this as `LIBUSB_ERROR_OVERFLOW` and discards the data; every
subsequent read on the endpoint also returns OVERFLOW until a USB-level
reset is performed.

The original clear-halt approach unsticks the kernel-side endpoint state
but does not always recover the firmware's TX FIFO. After ~2.5 days of
soak (see prior status update in this thread), I observed cases where
the device returned OVERFLOW for dozens of consecutive reads and only a
`libusb_reset_device()` cleared it. This is exactly the same recovery
as running `usb_resetter --reset-device` externally.

Calling `libusb_reset_device()` from inside the driver then exposed a
second bug: the existing reconnect path's `nut_libusb_close()` called
`libusb_exit(NULL)`, which on the 1->0 refcount transition wedged on
orphaned URBs that the reset had left undrained. This was traced via
gdb (`futex_do_wait` inside libusb's sync primitives, called from
`libusb_exit`, called from `nut_libusb_close`, called from the
fallthrough reconnect path in `qx_command`).

The libusb fix benefits more than just #598: any reconnect-heavy path
through the libusb1 backend now releases cleanly. Likely relevant to
**#993** ("Not work with IPPON Back Comfo Pro II", same 0665:5161) and
**#2453** ("Weird connectivity issues with a generic ViewPower UPS",
same 0665:5161); plausibly relevant to other long-standing USB
reconnect reports in the same backend.

## Validation

- **Authorize-toggle reproduction** (forces the exact close/reopen path
  that escalation would): 8 cycles of `echo 0 | tee
  /sys/bus/usb/devices/.../authorized; echo 1 | tee ...` produced clean
  reconnects each time, driver PID stable throughout. Pre-patch, the
  same toggle wedged the driver on the second cycle (gdb backtrace
  showed `libusb_exit` futex wait).
- **Natural OVERFLOW in soak**: with `pollinterval=10` on cypher's
  Salicru SPS 1500 ONE BL, a real OVERFLOW landed mid-poll. Logs show
  `Got OVERFLOW on EP 0x81 (1/3), retrying on next poll`, the counter
  incremented but did not reach the escalation threshold, the next poll
  succeeded, and steady-state polling resumed (counter-gated logic
  working as designed).
- **22h+ of post-fix uptime** with the patched binary, same driver PID
  throughout, no watchdog-triggered restarts.

## Test plan

- [x] Build `nutdrv_qx` with the patch (`make -C drivers nutdrv_qx`
      succeeds on Linux x86_64, libusb-1.0)
- [x] Authorize-toggle reproduction (8 clean reconnects, same PID)
- [x] At least one natural `Got OVERFLOW on EP 0x81 (N/3)` log line
      captured in soak, with subsequent poll success and counter reset
- [x] Driver PID stable for >20h post-install
- [ ] Maintainer review

## Related issues

- Fixes (the open root-cause symptom for): **#598** ("Communications
  with Ippon UPS lost (blazer_usb)")
- Likely fixes: **#993** (IPPON Back Comfo Pro II, same 0665:5161),
  **#2453** (ViewPower flaky USB2Serial, same 0665:5161)

[analysis]: https://github.com/networkupstools/nut/issues/598#issuecomment-4472810671


Pyker:
Status update after ~2.5 days soaking this on the originally-affected hardware (Salicru SPS 1500 ONE BL, `0665:5161`, libusb-1.0, Linux 7.0.x).

**The endpoint-halt-clear in this PR turns out to be necessary but not sufficient.** It does cleanly recover *isolated* overflows: I captured one where `usb_clear_halt(0x81)` plus a single re-read returned valid data on the same poll. But it does nothing for the *sustained* lockup that is the actual chronic hang. In one captured event the firmware returned `LIBUSB_ERROR_OVERFLOW` on every read for ~25 seconds (76 consecutive overflows); the clear-halt-and-retry fired on all of them and not one recovered. Only a USB-level device reset (an external `usb_resetter --reset-device`, i.e. `libusb_reset_device`) cleared it.

**Why the driver never self-recovers:** in `qx_command()`, `LIBUSB_ERROR_OVERFLOW` is grouped with `LIBUSB_ERROR_TIMEOUT` and `default` (all `break`, i.e. "transient, retry next poll"), so a wedged endpoint never reaches the `usb_reset()` + reconnect path that `PIPE`/`ETIME`/`IO`/`NO_DEVICE` already use. From git history this grouping appears to date to the original blazer import and has only been mechanically renamed since (`-EOVERFLOW` -> `ERROR_OVERFLOW` -> `LIBUSB_ERROR_OVERFLOW`); it does not look like a deliberate "overflow is benign" choice.

**Revised approach** (under soak now; I'll replace the commit here once it's validated): route `LIBUSB_ERROR_OVERFLOW` to `usb_reset()` + reconnect, gated behind a small consecutive-overflow counter that any clean read resets. The first couple of overflows are still retried on the next poll, so genuine one-off transients never trigger a reset; only a sustained run escalates. This reuses the existing recovery machinery rather than the in-loop retry currently in this PR.

A couple of things I'd value maintainer input on:
1. Is changing the shared `LIBUSB_ERROR_OVERFLOW` handling acceptable for all `nutdrv_qx` devices, or would you prefer it scoped behind a subdriver flag? Argument for leaving it global: a compliant device never overruns `wMaxPacketSize`, so the path is inert for healthy hardware, while for this firmware it converts an indefinite hang into a bounded auto-recovery.
2. Any preference on the escalation threshold, and on applying the same change to `blazer_usb.c` (structurally identical switch, and the driver #598 was originally filed against)?

Leaving this as draft until the soak confirms an in-driver reset clears the wedge as reliably as the external one does.

jimklimov:
Thanks for the details and revision. As for the questions - I think yes to both, better nail this exceptional situation widely.

Pyker:
Force-pushed with a revised approach. The PR now ships two commits instead of one:

1. `8574a2ee9` (`drivers/libusb1.c`): fix the per-close `libusb_exit()` misuse that wedges teardown on orphaned URBs after a device reset or disconnect. Generalizes the 2018 partial fix in 4f84b7f92.
2. `6ea5fcea5` (`drivers/nutdrv_qx.c`): counter-gated escalation for persistent `LIBUSB_ERROR_OVERFLOW`, calling `libusb_reset_device()` only after 3 consecutive overflows.

The PR body has been rewritten with the full background, validation evidence (authorize-toggle reproduction + natural OVERFLOW in soak with counter-gating working as designed), and the bisect-friendly commit ordering rationale.

Marked ready-for-review.

jimklimov:
Thanks, I hope to take a look in the coming week. A quick question so far: any idea if similar problem and fix applies to libusb-0.1 as well, or is their data/logic model significantly different?

Pyker:
Thanks for taking a look, no rush at all on the schedule.

Short answer: no, the equivalent fix doesn't apply to `drivers/libusb0.c`. The misuse pattern can't exist there for structural reasons.

**The API model difference:**

libusb-0.1 has no context object. `usb_init()` (declared in `<usb.h>` as `void usb_init(void)`) sets up internal global state once; there's no symmetric `usb_exit()` function in the libusb-0.1 API at all (grep'ed the `libusb-compat 0.1.8` header on my box: zero matches). Cleanup happens implicitly at process exit.

`drivers/libusb0.c`'s `nut_libusb_close()` (around line 1005) reflects this: it only calls the native `usb_close(udev)` (per-handle fd close) and returns. There's no per-close global-state teardown call to misuse, and no compat macro that would route it through `libusb_exit` (the macros in `drivers/usb-common.h:380-385` cover `usb_claim_interface`, `usb_clear_halt`, `usb_close`, `usb_set_configuration`, `usb_release_interface`, and `usb_reset`, but not `usb_init`: the signatures don't line up trivially, which is why the two backends are sibling translation units rather than one being a macro-thin wrapper).

For installations that link against `libusb-compat-0.1` (the shim providing the libusb-0.1 API on top of libusb-1.0), the compat layer's own `usb_close()` doesn't call `libusb_exit()` either, so the deadlock isn't reachable through that route.

**The failure mode is structurally impossible in libusb-0.1:**

The libusb-1.0 deadlock is "context teardown waits on internal sync primitives for outstanding URBs to drain after `libusb_reset_device()` orphaned them." libusb-0.1 has no async URB API; all transfers are blocking syscalls into the kernel. There are no URBs to orphan and no internal sync primitives that would wait on them.

**The qx_command counter-gated escalation already works for both backends:**

The escalation calls `usb_reset(udev)` using the libusb-0.1 native name. `drivers/usb-common.h:385` has the forward-mapping `#define usb_reset libusb_reset_device` for libusb-1.0 builds, so the same source compiles correctly against either backend. The escalation's subsequent fallthrough to the reconnect path calls `usb->close_dev() -> nut_libusb_close() -> usb_close()` on the libusb-0.1 side, which is safe.

So the only file change needed for the deadlock fix is the one already in this PR (`drivers/libusb1.c`); the `nutdrv_qx.c` half delivers the OVERFLOW recovery to both backends transparently.

(For completeness on error-code mapping: `drivers/usb-common.h:461-473` maps `LIBUSB_ERROR_OVERFLOW` to `-EOVERFLOW` for libusb-0.1 builds, so `qx_command()`'s switch fires identically when running against the libusb-0.1 backend.)

Pyker:
Quick soak update: the counter-gated escalation has now fired naturally on the originally-affected hardware (Salicru SPS 1500 ONE BL, `0665:5161`), and the W-fixed close/reopen path recovered cleanly without any process restart. Timeline:

```
11:04:58  Full update...
11:04:59  Got OVERFLOW on EP 0x81 (1/3), retrying on next poll
11:04:59  Got OVERFLOW on EP 0x81 (2/3), retrying on next poll
11:04:59  OVERFLOW on EP 0x81 persisted for 3 polls; resetting device
11:05:05  Device reset handled
11:05:05  Claimed interface 0 successfully
11:05:05  upsdrv_updateinfo finished
11:05:07+ ...steady-state 2s polling resumed...
```

Same driver PID throughout (no process exit), recovery completed in 6 seconds (inside NUT's 15s `MAXAGE`, so `upsd` never marked data stale and no downstream consumer saw an interruption). The local soak watchdog never had anything to detect.

One small wording nuance worth surfacing in case it matters for review: the commit message describes the counter as "retried on the next poll", but in practice consecutive reads within a single `Full update` cycle (status + ratings + alarms + battery in quick succession) can all hit OVERFLOW and trip escalation in well under a second, as happened here. Faster detection than the wording implies; behavior is desirable, but I can tweak the commit message if you'd prefer it to match the actual semantics more precisely.

Yesterday's authorize-toggle reproduction proved the disconnect path; this natural event proves the escalation path. Together they cover every codepath the fix touches.

jimklimov:
Thanks!

jimklimov:
NOTE: CI faults on freebsd15 may be due to baby troubles of the updated CI build agent, rather than a problem of tested code per se.

Pyker:
I noticed the failures but can't really check them because it asks me to login (and I imagine it would reject my GitHub account).

jimklimov:
I'll re-check after PR #3453 gets merged (for baseline to be green again with the updated CI workers), to rule out any code problems due to this PR.

Otherwise, I think the CI server should let in any authenticated GH user with read access (to see build logs, etc).

Pyker:
Ah, you're right, it does allow me to sign-in. It does seem related to the workers themselves rather than the code so I'll just wait.

jimklimov:
Those faults with mingw builds seem to have been due to OOM on server, not code as such.

jimklimov:
Thanks!

PR #3456, opened 2026-05-31
Use non-blocking connect() for AF_UNIX sockets to ensure we don't stall

dougnazar:
If a driver has become unresponsive, connect() will block after the listen() backlog has been exhausted, causing the main upsd to also become unresponsive.

### DIFF
--- a/common/common.c
+++ b/common/common.c
@@ -5174,6 +5174,53 @@ char *xstrdup(const char *string)
 	return p;
 }
 
+/* Try to connect to addr, using select() for timeout since AF_UNIX won't timeout normally */
+#ifndef WIN32
+int select_connect(int fd, const struct sockaddr_un *addr, size_t addrlen, const time_t d_sec, const suseconds_t d_usec)
+{
+	int rc;
+	int err = 0;
+	int flags = fcntl(fd, F_GETFL, 0);
+	fcntl(fd, F_SETFL, flags | O_NONBLOCK);
+
+	rc = connect(fd, (const struct sockaddr *)addr, addrlen);
+	if (rc == -1 && errno != EINPROGRESS) {
+		err = errno;
+	} else if (rc == -1) {
+		fd_set w_fds;
+		fd_set e_fds;
+		struct timeval tv;
+
+		FD_ZERO(&w_fds);
+		FD_SET(fd, &w_fds);
+		FD_ZERO(&e_fds);
+		FD_SET(fd, &e_fds);
+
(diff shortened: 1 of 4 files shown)


jimklimov:
This does look reasonable, maybe worth a mention in `NEWS.adoc`, maybe `UPGRADING.adoc` to (to highlight behavioral differences that might help or bite, comparing to older releases).

Just in case, CCing other maintainers for a second opinion ;)

jimklimov:
I wonder if the 1 second timeout here is right (hard-coded if I read that code correctly)? Maybe it should be a configurable value.

Specifically, what if the counterpart is busy or waiting between loop cycles, so will get to picking up the phone only in a few seconds? Did you test this, with various timing settings?

dougnazar:
I believe this will only be an issue once the listen() queue is filled, in which case I was thinking better to fail fast as this is in the main loop.

I've only extensively tested the failure mode on Linux (as that's where I have the failing driver), and in that case it never actually makes it to the select() call. Once in non-blocking mode the connect() call will change to returning EAGAIN if the listen() queue is full (and success if the queue is not full).

The rest was written with the help of various man pages (freebsd, aix, etc) and stack overflow answers, but I don't know that connect() will ever return an EINPROGRESS for an AF_UNIX socket.

dougnazar:
Just for fun, ran a little test on a FreeBSD vm I had here. With a listen() backlog of 5, and then just sleeping, the client returned:
```
connect: 0 rc=0, errno=0
connect: 1 rc=0, errno=0
connect: 2 rc=0, errno=0
connect: 3 rc=0, errno=0
connect: 4 rc=0, errno=0
connect: 5 rc=0, errno=0
connect: 6 rc=0, errno=0
connect: 7 rc=0, errno=0
connect: 8 rc=-1, errno=61
connect: 9 rc=-1, errno=61
```
So on FreeBSD at least, connect() just errors with ECONNREFUSED.

Edit: AIX does the same, errno 79 ECONNREFUSED
Edit 2: Only linux requires O_NONBLOCK to have the connect() fail, but has no impact on FreeBSD or AIX.

jimklimov:
Many systems complain about `unknown type name 'socklen_t'` - something amiss regarding includes/defines?

dougnazar:
I changed it to use `size_t` for the api as we just need the value and you're making an effort to exclude processing `sys/socket.h` if not needed.

jimklimov:
Thanks!

PR #3479, opened 2026-06-11
drivers: fix Windows state pipe permissions

mbv06:
### Summary
Set explicit Windows security attributes for driver state pipes by using a NULL DACL, allowing upsd to connect regardless of the default DACL inherited from the driver's launch context.

### Problem
I don't face this issue with a normal interactive launch, but did reproduce it in an automated/restricted launch context.

The driver started successfully and reported:

```text
Listening on named pipe \\.\pipe\richcomm_usb-test
```

while upsd failed to attach to the same state pipe and reported the driver as disconnected:
```
Can't connect to UPS [test] (richcomm_usb-test) named pipe \\.\pipe\richcomm_usb-test
ERR DRIVER-NOT-CONNECTED
```
// A direct PowerShell/.NET connection attempt to the driver pipe from the same restricted context also failed with `Access to the path is denied.`

Since the pipe was created with default Windows security attributes, access could depend on the process security context. 

### Testing
After the change, the same workflow succeeds on Windows 11: `upsd` connects to the driver pipe and `upsc` can read UPS data.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -77,6 +77,8 @@ https://github.com/networkupstools/nut/milestone/13
  - common code:
    * Revised 'dstate' machinery to track socket connections closed mid-way
      through a call, to avoid access after `free()`. [#3302]
+   * Windows driver state named pipes now use explicit security attributes
+     instead of relying on default settings from the launch context. [PR #3479]
 
  - NUT client libraries:
     * Complete support for actions documented in `docs/net-protocol.txt`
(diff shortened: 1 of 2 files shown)


jimklimov:
Thanks, looks good!

jimklimov:
@mbv06 : Hello, I am looking to generalize this solution, to see if that would help with issue #3525. Do you know if the security attribute should only be considered with listeners (`CreateNamedPipe`) or also with senders (`CreateFile` on the other side to submit messages)?

Issue #1352, opened 2022-04-01
[CI] Find a way to avoid certain CLI flags not supported by compilers


jimklimov:
A few options probed in `m4/nut_compiler_family.m4` are not really supported by some versions of compilers that the NUT CI build farm uses (whether by front-ends like `gcc` or back-ends like `cc1`) and are reported as problematic during "real builds". During the test operations done in the m4 script however, they seem fine and are accepted - only to begin complaining when e.g. C++11 support is probed just after.

They seem to also be reported as errors when `-Werror` mode is used (so warnings are fatal) but were not see to cause actual build failures, just confusion about why they failed (and noise in the log or terminal).

Seen with `-Wno-reserved-identifier` and `-Wno-unknown-warning-option` on gcc-4.x.

jimklimov:
It seems the problem was between `AX_CHECK_COMPILE_FLAG()` and `AC_LANG_PUSH()/AC_LANG_POP()` and/or `AC_MSG_CHECKING()`, in that they saved `-Werror` from fourth argument of the former macro into the resulting `CFLAGS` or `CXXFLAGS` and seem to have ignored the "if failed" third argument of the latter.

Maybe not all of this analysis is spot-on, but anyhow the proposed change to raise a flag and apply the update or revert of the standard variable after the lang pop did help ignore the unrecognized compiler parameters.

Another part of the issue is that the initial check seems to ignore `-Werror` and succeeds compilation with whatever `-W*` arguments, and so the inner check (if first step was successful, try to link and/or run) is still needed. The initial check does fail for wholly unknown arguments like `-Q*` not supported by some compiler versions.

jimklimov:
It still misfires for some cases, e.g.:

````
configure:11094: Starting check compile flag for '-Wno-reserved-identifier'; now CFLAGS='-g -O2' and CXXFLAGS='-g -O2'
configure:11107: checking whether C compiler accepts -Wno-reserved-identifier
configure:11127: /usr/lib/ccache/gcc-4.4.4 -std=gnu99 -c -g -O2  -Wno-reserved-identifier  conftest.c >&5
configure:11127: $? = 0
configure:11145: result: yes
configure:11150: checking whether the flag '-Wno-reserved-identifier' is still supported in CC linker mode
[...]

PR #3489, opened 2026-06-20
Fix minor issues to compile on AIX

dougnazar:
Just for fun I've been playing with getting it to compile on my 7.2 install under QEMU. Mainly minor changes, compiles clients & drivers before failing trying to write to the source directory (mounted read only).

* Add check for older IBM compilers.
* Check enable version of warnings to see if valid
* Use lsgroup on AIX to check for valid groups
* Fix %zu format check
* Switch to using -t to specify date as -d not always availble
* Ignore unused parameters
* AIX doesn't provide defines for those speeds.
* Provide function declaration when needed

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -137,13 +137,13 @@ install-fanout-cleanup: install-recursive @dotMAKE@
 touch-include-all-nut_version-generated.timestamp:
 	@[ -s include/nut_version.h ]
 	@touch -r include/nut_version.h -d '-10 seconds' include/.all.nut_version-generated.timestamp 2>/dev/null && exit ; \
-	 touch -d '1970-01-01T00:00:00' include/.all.nut_version-generated.timestamp && exit ; \
+	 touch -t '197001010000.00' include/.all.nut_version-generated.timestamp && exit ; \
 	 touch include/.all.nut_version-generated.timestamp
 
 touch-clients-all-libupsclient_version-generated.timestamp:
 	@[ -s clients/libupsclient-version.h ]
 	@touch -r clients/libupsclient-version.h -d '-10 seconds' clients/.all.libupsclient_version-generated.timestamp 2>/dev/null && exit ; \
-	 touch -d '1970-01-01T00:00:00' clients/.all.libupsclient_version-generated.timestamp && exit ; \
+	 touch -t '197001010000.00' clients/.all.libupsclient_version-generated.timestamp && exit ; \
 	 touch clients/.all.libupsclient_version-generated.timestamp
 
 touch-docs-man-all-linkman-generated.timestamp:
@@ -153,7 +153,7 @@ touch-docs-man-all-linkman-generated.timestamp:
 	 else \
 	    touch -r docs/man/linkman-driver-names.txt -d '-10 seconds' docs/man/.all.nut_linkman-generated.timestamp 2>/dev/null && exit ; \
 	 fi ; \
-	 touch -d '1970-01-01T00:00:00' docs/man/.all.nut_linkman-generated.timestamp && exit ; \
+	 touch -t '197001010000.00' docs/man/.all.nut_linkman-generated.timestamp && exit ; \
 	 touch docs/man/.all.nut_linkman-generated.timestamp
 
 # Verbosity for fanout rule tracing; 0/1 (or "default" that may auto-set
--- a/configure.ac
+++ b/configure.ac
@@ -1703,12 +1703,13 @@ AC_CACHE_CHECK([for %zu as PRIuSIZE],
     [ac_cv_printfmt_zu],
     [AX_RUN_OR_LINK_IFELSE(
         [AC_LANG_PROGRAM([$CODE_STRINGINCL
+#include <stdio.h>
 #include <stdlib.h>
 ],
             [[size_t n = 123;
 char buf[64];
 snprintf(buf, sizeof(buf), "%zu", n);
-return !strcmp(buf, "123");
+return strcmp(buf, "123");
 ]])],
         [ac_cv_printfmt_zu=yes], [ac_cv_printfmt_zu=no]
     )])
--- a/configure.ac
+++ b/configure.ac
@@ -525,6 +525,7 @@ AS_IF([test ! -d "${PIDPATH}" -o -h "${PIDPATH}" && test -d "/run" -a ! -h "/run
 
 AC_CHECK_PROGS([GETENT], [getent], [])
 AC_CHECK_PROGS([ID], [id], [])
+AC_CHECK_PROGS([LSGROUP], [lsgroup], [])
 
 PROBE_OS_USER="false"
 PROBE_OS_GROUP="false"
@@ -534,20 +535,24 @@ AS_IF([test x"${GETENT}" != x], [
 	],[
 	AS_IF([test x"${ID}" != x], [
 		PROBE_OS_USER="${ID} -u "
-		AS_IF([test -r "/etc/groups"], [
-			dnl FIXME: For command prefix usage like used in this
-			dnl  script, we might be better off defining a shell
-			dnl  function and referring to it. Grep would catch
-			dnl  also user names that have a group as secondary!
-			PROBE_OS_GROUP="cat /etc/groups | ${GREP} -w "
+		AS_IF([test x"${LSGROUP}" != x], [
+			PROBE_OS_GROUP="${LSGROUP} "
 			], [
-			dnl This shows groups of a USER with specified name!
-			dnl But to probe for cases where string names are
-			dnl same (e.g. "nut" or "ups") this might be good
-			dnl enough.
-			PROBE_OS_GROUP="${ID} -g "
+			AS_IF([test -r "/etc/groups"], [
+				dnl FIXME: For command prefix usage like used in this
+				dnl  script, we might be better off defining a shell
+				dnl  function and referring to it. Grep would catch
+				dnl  also user names that have a group as secondary!
+				PROBE_OS_GROUP="cat /etc/groups | ${GREP} -w "
+				], [
+				dnl This shows groups of a USER with specified name!
+				dnl But to probe for cases where string names are
+				dnl same (e.g. "nut" or "ups") this might be good
+				dnl enough.
+				PROBE_OS_GROUP="${ID} -g "
+				])
+			AC_MSG_WARN([Can not PROPERLY check existence of group accounts on this system, but can try best-effort])
 			])
-		AC_MSG_WARN([Can not PROPERLY check existence of group accounts on this system, but can try best-effort])
 		],[
 		AC_MSG_WARN([Can not check existence of user and group accounts on this system])
 		])
--- a/m4/nut_compiler_family.m4
+++ b/m4/nut_compiler_family.m4
@@ -111,6 +111,10 @@ dnl NOTE: this option should not be passed via the fourth argument of the macro,
 dnl or it ends up in the flags too, possibly during the "pop"; have to use the
 dnl GOOD_FLAG instead :\
     COMPILERFLAG="$1"
+dnl GCC will silently accept warning disable options, but then if there is an issue, warn about the option, i.e.
+dnl "cc1: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics"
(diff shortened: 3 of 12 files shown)


jimklimov:
Sounds cool, thanks! I'll take a closer look later.

One point about writing to source dirs - this is possible in a few developer-oriented recipes (like maintenance of spell-checking dictionary), but normally should not happen and `make distcheck` verifies that.

m4/nut_compiler_family.m4
```diff
@@ -111,6 +111,10 @@ dnl NOTE: this option should not be passed via the fourth argument of the macro,
 dnl or it ends up in the flags too, possibly during the "pop"; have to use the
 dnl GOOD_FLAG instead :\
     COMPILERFLAG="$1"
+dnl GCC will silently accept warning disable options, but then if there is an issue, warn about the option, i.e.
+dnl "cc1: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics"
+dnl Instead try the enable option to check if it is valid
```
jimklimov, line 116: This and the changes below look like they may close a long-standing and annoying issue #1352 (if they work), thanks!

configure.ac
```diff
@@ -1698,12 +1703,13 @@ AC_CACHE_CHECK([for %zu as PRIuSIZE],
...
     [AX_RUN_OR_LINK_IFELSE(
         [AC_LANG_PROGRAM([$CODE_STRINGINCL
+#include <stdio.h>
 #include <stdlib.h>
 ],
             [[size_t n = 123;
 char buf[64];
 snprintf(buf, sizeof(buf), "%zu", n);
-return !strcmp(buf, "123");
+return strcmp(buf, "123");
```
jimklimov, line 1712: Facepalm :) Thanks!

jimklimov:
Thanks, so the changes did look promising, but at least the ones about possibly-unsupported compiler warnings need some more work, as the builds no longer ignore some misplaced flags that they did before (something we might impact with more careful coding of the numerous `*FLAGS` env-/make-vars) and no longer ignore transgressions by third-party code (not much we can do about), e.g.:

* third-party header mess:
```
In file included from phoenixcontact_modbus.c:26:
/usr/include/modbus/modbus.h:157:16: error: identifier '_modbus' is reserved because it starts with '_' at global scope [-Werror,-Wreserved-identifier]
  157 | typedef struct _modbus modbus_t;
      |                ^
```

* system-provided header mess:
```
upsclient.c:1496:5: error: identifier '__i' is reserved because it starts with '__' [-Werror,-Wreserved-identifier]
                                FD_ZERO(&fds);
                                ^
/usr/include/sys/select.h:88:26: note: expanded from macro 'FD_ZERO'
#define FD_ZERO(fdsetp)         __FD_ZERO (fdsetp)
                                ^
/usr/include/bits/select.h:27:18: note: expanded from macro '__FD_ZERO'
    unsigned int __i;                                                         \
                 ^
```

* linker options in compiler commands (might be faults of either our hacks in m4 or `c_build.sh` to get non-default library locations usable, or of system's pkg-config data):
  * OmniOS:
```
configure:21152: WARNING: Optional C library routine strlwr not found
configure:21157: checking for strsep(s1,s2)
configure:21217: /usr/lib/ccache/clang-17 -o conftest  -I/tmp/jenkins-nut/workspace/nut_nut_PR-3489/tmp/include -std=gnu11 -m32 -Werror -Werror=implicit-function-declaration   -I/tmp/jenkins-nut/workspace/nut_nut_PR-3489/tmp/include -L/tmp/jenkins-nut/workspace/nut_nut_PR-3489/tmp/lib -m32 -R/opt/ooce/lib conftest.c  >&5
error: unknown remark option '-R/opt/ooce/lib' [-Werror,-Wunknown-warning-option]
configure:21217: $? = 1
configure: program exited with status 1
configure: failed program was:
...
```
  * NetBSD:
```
configure:21855: result: no
configure:21863: WARNING: Optional C library routine strsep not found
configure:21877: checking for strstr(s1,s2)
configure:21939: /usr/lib/ccache/clang-18 -o conftest -isystem /usr/pkg/include  -I/var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3489@3/tmp/include -std=gnu11 -m64 -I/usr/pkg/include -Werror -Werror=implicit-function-declaration   -I/var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3489@3/tmp/include -L/var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3489@3/tmp/lib -m64 -R/usr/pkg/lib conftest.c  >&5
error: unknown remark option '-R/usr/pkg/lib' [-Werror,-Wunknown-warning-option]
configure:21939: $? = 1
configure: program exited with status 1
configure: failed program was:
...
```

Makefile.am
```diff
@@ -137,13 +137,13 @@ install-fanout-cleanup: install-recursive @dotMAKE@
 touch-include-all-nut_version-generated.timestamp:
 	@[ -s include/nut_version.h ]
 	@touch -r include/nut_version.h -d '-10 seconds' include/.all.nut_version-generated.timestamp 2>/dev/null && exit ; \
-	 touch -d '1970-01-01T00:00:00' include/.all.nut_version-generated.timestamp && exit ; \
+	 touch -t '197001010000.00' include/.all.nut_version-generated.timestamp && exit ; \
```
jimklimov, line 140: Can not yet say if all systems would be okay with this change of spelling, or we would need to try it both ways. Deferring the decision to CI build passing everywhere :)

m4/nut_compiler_family.m4
dougnazar: This has an issue on Clang, in that the defines from `confdefs.h` that are added to the test file conflict with `-Wreserved-identifier`. Trying to see a way around that besides moving those tests.

jimklimov:
@dougnazar : If the puzzle about compiler warnings does prove to be complicated, feel free to e.g. rebase this PR without that commit (most of the other changes seem portable at first glance), and address the warnings as a separate PR (billed as a solution to #1352), so we can merge the working improvements faster :)

m4/nut_compiler_family.m4
dougnazar: Found the magical incantations that make it work, with both current and older versions of `autoconf`. We'll see once the CI finishes, but it's succeeded on the previous failing builds.

jimklimov:
NetBSD with clang disliked it already, many scenarios still queued:
```
configure:21841: result: no
configure:21849: WARNING: Optional C library routine strsep not found
configure:21863: checking for strstr(s1,s2)
configure:21925: /usr/lib/ccache/clang-19 -o conftest -isystem /usr/pkg/include  -I/var/shm/jenkins-nutci-netbsd-11-amd64/nut_nut_PR-3489/tmp/include -std=gnu99 -m64 -I/usr/pkg/include -Werror -Werror=implicit-function-declaration   -I/var/shm/jenkins-nutci-netbsd-11-amd64/nut_nut_PR-3489/tmp/include -L/var/shm/jenkins-nutci-netbsd-11-amd64/nut_nut_PR-3489/tmp/lib -m64 -R/usr/pkg/lib conftest.c  >&5
error: unknown remark option '-R/usr/pkg/lib' [-Werror,-Wunknown-warning-option]
configure:21925: $? = 1
configure: program exited with status 1
configure: failed program was:
```

dougnazar:
I'm actually a little confused on this one. It seems to be complaining about `-R/usr/pkg/lib` but that's being passed in via `LDFLAGS` on the command line. It does cause the checks for the `-Wno-unknown-warning-option` to fail since we're adding `-Werror` so that option doesn't get added. Removing `-Werror` would fix that but I think fixing the `LDFLAGS` is the right thing to do. I think we should have actually failed earlier.

jimklimov:
I think the deeper problem may be about pkg-config files passing C/C++ "flags" and "libs". I haven't seen a separate spot for `LDFLAGS` there, nor significant regard for compiler/linker/toolkit quirks.

Not sure OTOH if/how autoconf tests differentiate or fence that or not, but at least here with `-o conftest` this was a "build aand link" operation.

dougnazar:
This is from before `pkg-config`. It's part of the `configure` command-line.

`/var/shm/jenkins-nutci-netbsd-11-amd64/nut_nut_PR-3489/configure ... 'LDFLAGS=-L/var/shm/jenkins-nutci-netbsd-11-amd64/nut_nut_PR-3489/tmp/lib -m64 -R/usr/pkg/lib' ...`

I think the intention was to set the rpath, however it's passed to `clang` not the linker and `clang` thinks it's a remark option. It should be `-Wl,-R,/usr/pkg/lib` or `-Wl,-rpath,/usr/pkg/lib` to pass it through to the linker.

Also, I believe that before my patch, it would fail the tests for several flags (`-Qunused-arguments`, `-Wno-reserved-identifier`, `-fdiagnostics-color=always`) before it added `-Wno-unknown-warning-option` since the `-Werror` already existed when testing the options when linking.

I'll push a version that will revert the linker portion of the check, to make it same as before, but I think the CI command line should be fixed too. (I'll just wait a bit to see if anything else crops up).

jimklimov:
Sounds right, my bad with those `-R` values. Probably some more may be lurking in fallback code then. Thanks!

UPDATE: Seems that in #3495 was the only explicit one in code (a few more in docs/comments though). There are some lurking in loops, like additions for "/opt/ooce" (OmniOS).

FWIW, having `-R/...` in `*.pc` files is a recurrent theme in a few ecosystems (maybe `nspr` is my local hotfix, but others should not be I think) :\
```
jim@nutci-oi:~$ grep -rE ' -R */' {/usr,{/local,}}/lib*/pkgconfig
/usr/lib/pkgconfig/menu.pc:Libs:  -L/usr/gnu/lib -m32 -D_XPG6 -R/usr/gnu/lib -lmenu
/usr/lib/pkgconfig/nspr.pc:Libs: -L/usr/lib/mps -R/usr/lib/mps -lplds4 -lplc4 -lnspr4 -lpthread -ldl -lposix4
/usr/lib/pkgconfig/panel.pc:Libs:  -L/usr/gnu/lib -m32 -D_XPG6 -R/usr/gnu/lib -lpanel
/usr/lib/pkgconfig/ncurses++.pc:Libs:  -L/usr/gnu/lib -m32 -D_XPG6 -R/usr/gnu/lib -lncurses++
/usr/lib/pkgconfig/ncurses.pc:Libs:  -L/usr/gnu/lib -m32 -D_XPG6 -R/usr/gnu/lib -lncurses  -lpcre2-posix
/usr/lib/pkgconfig/form.pc:Libs:  -L/usr/gnu/lib -m32 -D_XPG6 -R/usr/gnu/lib -lform

root@nutci-slackware-15-amd64:~# grep -rE ' -R */' {/usr,{/local,}}/lib*/pkgconfig
/usr/lib64/pkgconfig/sdl.pc:Libs.private: -lSDL  -lpthread  -lm -ldl  -L/usr/lib64 -R/usr/lib64 -lX11 -lXext -lXrandr -lXrender -lvga -lpthread
/usr/lib64/pkgconfig/serf-1.pc:Libs.private: -L/usr/lib64 -R/usr/lib64 -laprutil-1 -lldap -llber -ldb-4   -lexpat       -lsqlite3  -L/usr/lib64/  -lmysqlclient -L/usr/lib64/ -lmariadb  -ldb-4   -lldap -llber -L/usr/lib64 -R/usr/lib64 -lapr-1 -luuid -lrt -lcrypt  -lpthread -ldl  -lz
```

jimklimov:
It seems this PR only failed on platforms impacted by fixes posted later in #3495 and #3498.

dougnazar:
Yeah, I have hazy memories of having to fix the `pkg-config` files, but that was quite some time ago. I don't see any `-R` (but lots of `-Wl,--rpath,` in any of my systems. And all those are pointing to system lib directories.

I've pushed a new version of the warning check, where the linking portion of the test is the same as before the patch, so it'll work the same as before. I've checked the following scenarios:

* **gcc**: now properly ignores `-Wno-reserved-identifier -Wno-unknown-warning-option`
* **clang**: correctly adds `-Wno-reserved-identifier -Wno-unknown-warning-option`
* **clang with LDFLAGS='-R/blah'**: fails `-Qunused-arguments -Wno-reserved-identifier` but adds `-Wno-unknown-warning-option` (as before)

PR #3491, opened 2026-06-21
Fix building with a read-only nfs mounted source

dougnazar:
First one had me very confused for a while. Couldn't figure out how it switched from the build directory to the source directory.

* Use full path for generated files
* Don't try to copy permissions


jimklimov:
Looks reasonable, thanks! Waiting for CI agents to concur across the board :)

PR #3496, opened 2026-06-22
`upsimage.cgi`: adjust scaling for small ranges (single-digit)

### LINKED ISSUE #3469: # upsimage.cgi crashes with SIGFPE (division by zero) on small per-cell `battery.voltage` values
## Summary
_Disclaimer: I had this report by Claude.ai while trying to debug some issues with my new UPS monitoring.
I'm not a developer so I could not full verify this report before posting it nor full validate its suggestions. I tested overrides on my own computers before post it, is the only warranty I can provide about it.
 Please, be careful with suggested fix._

`upsimage.cgi` terminates with a **floating point exception (SIGFPE, integer division by zero)** instead of rendering a PNG when it is asked to draw `battery.voltage` for a UPS whose driver reports the voltage of a **single 2 V cell** (typical of many Megatec / Q* devices) rather than the whole battery assembly.

Because the crash happens *before* any HTTP header is emitted, the web server reports it as `End of script output before headers`, and the embedding `upsstats.cgi` page shows a broken image. A CGI program should never abort on input data: `upsimage.cgi` already has a `noimage()` path for unusable input, but this particular case bypasses it and divides by zero.

## Environment

- NUT version: # Network UPS Tools upsc 2.8.5 release
- OS: Arch Linux
- Web server: Apache/2.4.67 (`mod_cgid`)
- Driver: `nutdrv_qx`, `protocol = q1`
- Device: VulTech GS-3KVAS-RK (3 kVA online, lead-acid bank reported per 2 V cell)

## Apache error log

```
[cgid:error] [pid …] [remote …] End of script output before headers: upsimage.cgi,
referer: https://…/ups/upsstats.cgi?host=vultech-3kva@<host>
```

## Steps to reproduce

A device/driver that publishes per-cell battery voltage, e.g. `upsc` reports:

```
battery.voltage: 2.25
battery.voltage.nominal: 2.0
battery.voltage.high: 2.27
battery.voltage.low: 1.75
battery.charge: 96
```

Then invoke the CGI directly:

```sh
QUERY_STRING='host=vultech-3kva@<host>&display=battery.voltage&width=90' \
  /usr/share/nut/cgi/upsimage.cgi > /tmp/out.png ; echo "exit=$?"
```

Result:

```
Floating point exception (core dumped)
exit=136          # 128 + 8 = SIGFPE
/tmp/out.png: empty
```

`display=battery.charge` (and other large-scale variables) render fine; only the small-magnitude `battery.voltage` triggers the crash.

## Expected behavior

The CGI should render a PNG in all cases — either a sensible bar, or the `noimage()` placeholder for unusable input — and never crash.

## Root cause

Path: `main()` → `draw_battvolt()` → `drawgeneralbar()` → `drawbar()` → `drawscale()` (all in `clients/upsimage.c`).

1. In `main()`, the min/nominal/max values are converted with **`atoi()`**, which truncates the decimals of a sub-10 V, per-cell reading:
   - `min = atoi("1.75") = 1`
   - `nom = atoi("2.0")  = 2`
   - `max = atoi("2.27") = 2`

2. `draw_battvolt(var=2.25, min=1, nom=2, max=2, deviation=0)`:
   - `nom`, `min`, `max` are all `!= -1`, so none of the fallback/auto-nominal logic runs;
   - `deviation = (int)(-nom * 0.05) = (int)(-0.1) = 0` → reset to `-1`;
   - calls `drawgeneralbar(2.25, 1, 2, 2, -1, …)`.

3. In `drawgeneralbar()`, with `min=1, max=2`, `(max - min) <= 50` so `step10 = 10`, and the scale bounds collapse:
   ```c
   lo = ((min - deviation) / step10) * step10 = ((1 - 1) / 10) * 10 = 0;
   hi = ((max + deviation + step10/2) / step10) * step10 = ((2 + 1 + 5) / 10) * 10 = 0;
   ```
   so `drawbar()`/`drawscale()` are called with `lvllo == lvlhi == 0`.

4. In `drawscale()`:
   ```c
   range = lvlhi - lvllo;                       /* = 0 */
   ...
   y = scale_height * (lvlhi - level) / range;  /* integer division by zero -> SIGFPE */
   ```

So any input where the rounded `lo` and `hi` coincide (which is guaranteed for the ~2 V per-cell scale, since the axis is quantized to multiples of `step10`) divides by zero. The header-printing function `drawimage()` is only reached at the very end, so the process dies before emitting `Content-type`, hence Apache’s `End of script output before headers`.

## Suggested fixes

- **Minimal / robust:** guard against `range == 0` in `drawscale()` (and/or in `drawgeneralbar()` after computing `lo`/`hi`). If the rounded range collapses, either widen it (e.g. `hi = lo + step10`) or fall back to `noimage("battery.voltage out of drawable range")` so the CGI degrades gracefully instead of aborting.
- **Contributing factor:** `main()` uses `atoi()` for the min/nominal/max bounds, discarding the fractional part. For small absolute values this both loses precision and helps drive the range to zero; `strtod()`/`atof()` would be more faithful, though the integer-quantized scale (multiples of `step10 = 10`) is inherently unsuitable for sub-10 V magnitudes.

## Notes / workaround

This is independent of the driver-side question of *whether* the device should report per-cell voltage. On the operator side it can be worked around with `nutdrv_qx` by scaling the reading to the whole bank:

```ini
battery_voltage_reports_one_pack
override.battery.packs = <nominal_bank_voltage / 2>
override.battery.voltage.nominal = <bank nominal>
override.battery.voltage.high    = <bank high>
override.battery.voltage.low     = <bank low>
```

…but `upsimage.cgi` should not SIGFPE on the unscaled data regardless, which is what this report is about.

## Related
- [#3467](https://github.com/networkupstools/nut/issues/3467) nutdrv_qx: battery_voltage_reports_one_pack silently zeroes battery.charge/runtime guesstimation when voltage thresholds are set to whole-bank values 
- [#3468](https://github.com/networkupstools/nut/issues/3468) [DDL] Device support report: VulTech GS-3KVAS-RK (nutdrv_qx, Q1) — works, with notes

jimklimov:
Closes: #3469

Avoid lack of output when the range of a variable is small (e.g. 0-10) and division by zero occurred instead.

Practical testing needed (not done so far).

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -127,6 +127,10 @@ https://github.com/networkupstools/nut/milestone/13
     * The `libupsclient` (C) and `libnutclient` (C++) API were updated to
       report the ability to check `CERTIDENT` information. [#3331]
 
+ - `upsimage.cgi` client updates:
+    * Adjust scaling for small (single-digit) value ranges to avoid division
+      by zero and not rendering anything. [issue #3469]
+
  - `upsmon` client updates:
     * Introduced support for `CERTFILE` option, so the client can identify
       itself to the data server also in OpenSSL builds. [issue #3331]
--- a/clients/Makefile.am
+++ b/clients/Makefile.am
@@ -156,7 +156,7 @@ upssched_LDADD += \
 	$(NETLIBS)
 
 upsimage_cgi_SOURCES = upsimage.c upsclient.h upsimagearg.h cgilib.c cgilib.h
-upsimage_cgi_LDADD = $(LDADD) $(LIBGD_LDFLAGS)
+upsimage_cgi_LDADD = $(LDADD) $(LIBGD_LDFLAGS) -lm
 
 upsset_cgi_SOURCES = upsset.c upsclient.h cgilib.c cgilib.h
 
--- a/clients/upsimage.c
+++ b/clients/upsimage.c
@@ -42,6 +42,7 @@
 #include "upsclient.h"
 #include "cgilib.h"
 #include <stdlib.h>
+#include <math.h>
 #include <gd.h>
 #include <gdfontmb.h>
 
@@ -152,7 +153,8 @@ static void drawscale(
 	int step, int step5, int step10,	/* steps for minor, submajor and major dashes */
 	int redlo1, int redhi1,			/* first red zone start and end */
 	int redlo2, int redhi2,			/* second red zone start and end */
-	int grnlo, int grnhi)			/* green zone start and end */
+	int grnlo, int grnhi,			/* green zone start and end */
+	int scale)				/* scaling factor for decimal precision */
 {
 	int	col1, col2, back_color, scale_num_color, ok_zone_maj_color,
 		ok_zone_min_color, neutral_zone_maj_color,
@@ -180,6 +182,11 @@ static void drawscale(
 	gdImageColorTransparent(im, back_color);
 
 	range = lvlhi - lvllo;
+	if (range <= 0) {
+		/* Prevent division by zero on collapsed ranges */
+		lvlhi = lvllo + step10;
+		range = step10;
+	}
 
 	/* draw scale to correspond with the values */
 	for (level = lvlhi; level >= lvllo; level -= step) {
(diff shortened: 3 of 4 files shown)


mfrakka:
Tested on: Arch Linux, NUT 2.8.5, nutdrv_qx Q1, VulTech GS-3KVAS-RK (device reports battery.voltage as single 2 V cell)
Test command:
QUERY_STRING='host=vultech-3kva@192.168.150.1&display=battery.voltage&width=90' ./upsimage.cgi > out.png
echo "exit=$?" && xxd out.png | head -4

Result with stock 2.8.5: exit=136 (SIGFPE), empty output, Apache logs End of script output before headers.
Result with PR https://github.com/networkupstools/nut/pull/3496 patch: exit=0, 798-byte output starting with Pragma: no-cache / Content-type: image/png followed by valid PNG signature (89 50 4e 47). Image renders correctly in browser.

Patch confirmed working. ✅

Issue #3501, opened 2026-06-23
New device: GoldenMate 1000VA/800W LiFePO4 UPS (06DA:FFFF, Phoenixtec/BMS)


bshor:
## Device

**Brand/Model:** GoldenMate 1000VA/800W Lithium UPS Battery Backup (GOLDENMATE 1000VA/800W Sinewave UPS System with LiFePO4 Battery, 230Wh, Upgraded Pro Ver. with Communication Port)
**USB ID:** `06DA:FFFF` (Phoenixtec Power Co., Ltd)
**bcdDevice:** `0001`
**Manufacturer string:** `-BMS-`
**Product string:** `Smart-Battery`
**Serial:** `J7152BCA698BE5C6805M251`
**NUT version:** `2.8.1` (Debian trixie, aarch64 / Raspberry Pi 4)
**Driver matched:** `Phoenixtec/Liebert HID 0.41` (generic fallback)

## Current behavior

The device is detected and partially functional via `usbhid-ups` using the Phoenixtec/Liebert HID subdriver. Basic status (`OL`, `CHRG`), battery charge percentage, and voltage are reported correctly.

**Known issues:**
- `battery.runtime` (`RunTimeToEmpty`) reports `599940` seconds (~7 days) — clearly bogus, likely a unit/scaling issue with the LiFePO4 chemistry not being handled correctly
- `ups.load` and `input.voltage` are not exposed — the device's HID descriptor doesn't include output/input power paths
- HID descriptor has `LogMax < LogMin` errors in ReportID `0x01` and `0x02` (firmware issue, worked around by NUT)
- `file_report_buffer: expected 9 bytes, but got 512` on every poll — report size mismatch
- `nut-scanner` marks device as `###NOTMATCHED-YET###` (bcdDevice `0001` not in USB ID table)
- Device reports chemistry as `Lion` (HID index 4) but is actually **LiFePO4**

## upsc output

```
battery.charge: 76
battery.runtime: 599940
battery.type: Lion
battery.voltage: 13.20
device.mfr: -BMS-
device.model: Smart-Battery
device.serial: J7152BCA698BE5C6805M251
device.type: ups
driver.name: usbhid-ups
driver.version: 2.8.1
driver.version.data: Phoenixtec/Liebert HID 0.41
experimental.battery.capacity: 100
experimental.battery.capacity.nominal: 100
ups.delay.shutdown: 20
ups.delay.start: 30
ups.mfr: -BMS-
[...]

PR #3502, opened 2026-06-23
idowell-hid: add GoldenMate 1000VA/800W LiFePO4 (USB ID 06da:ffff)

bshor:
## Summary

GoldenMate ships the same `-BMS-` firmware with an identical 291-byte HID descriptor across multiple USB vendor/product IDs. The `075d:0300` variant was already added to this subdriver in #3015. This PR adds `06da:ffff` (Phoenixtec VID), which was previously falling through to the generic Phoenixtec/Liebert HID subdriver.

**Device:** GoldenMate 1000VA/800W Sinewave UPS with LiFePO4 Battery (230Wh), USB ID `06da:ffff`, manufacturer string `-BMS-`, product string `Smart-Battery`.

**Problem:** Without this change the device matches `Phoenixtec/Liebert HID 0.41` and reports `battery.runtime = 599940` seconds — a firmware placeholder value. The iDowell subdriver mappings (which were already extended for the identical `075d:0300` firmware) handle this device correctly.

**Evidence that HID descriptors are identical:** Both devices report the same 291-byte report descriptor (`05 84 09 04 a1 01 ...`), the same 28 HID objects, and the same `LogMax < LogMin` quirks in ReportID 0x01 and 0x02.

Tracked in #3501.

## Changes

- `drivers/idowell-hid.c`: add `06da:ffff` to the USB device table
- `data/driver.list.in`: split GoldenMate entry into two lines with correct model names, USB IDs, support level 2, and issue links

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -82,6 +82,12 @@ https://github.com/networkupstools/nut/milestone/13
  - `usbhid-ups` driver updates:
     * When reconnecting, report success more visibly (not only in debug), and
       do not reset driver state to "quiet" when it will loop trying. [PR #3423]
+    * `idowell-hid` subdriver now also supports GoldenMate 1000VA/800W LiFePO4
+      battery packs (`0x06da:0xffff`), which carry the same `-BMS-` firmware and
+      HID descriptor as the existing `0x075d:0x0300` device. The shared
+      Phoenixtec vendor ID (`0x06da`) is gated by device strings so the other
+      devices on that ID still fall through to `liebert-hid` / `mge-hid`.
+      [issue #3501, PR #3502]
 
  - common code:
     * Revised 'dstate' machinery to track socket connections closed mid-way
--- a/data/driver.list.in
+++ b/data/driver.list.in
@@ -545,7 +545,8 @@
 
 "Gemini"	"ups"	"1"	"UPS625/UPS1000"	""	"safenet"
 
-"GoldenMate"	"ups"	"1"	"UPS 1000VA Pro"	"USB"	"usbhid-ups"
+"GoldenMate"	"ups"	"2"	"UPS 1000VA Pro (USB ID 075d:0300)"	"USB"	"usbhid-ups"	# https://github.com/networkupstools/nut/issues/3015
+"GoldenMate"	"ups"	"2"	"1000VA/800W LiFePO4 (USB ID 06da:ffff)"	"USB"	"usbhid-ups"	# https://github.com/networkupstools/nut/issues/3501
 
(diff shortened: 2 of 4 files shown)


jimklimov:
Hello, thanks for the contribution, but it would not work the way it is originally posted, I think (might hijack devices not intended for this subdriver) - see also `mge-hid.c` which also deals with `PHOENIXTEC_VENDORID`.

To make the PR acceptable, you would need to:
* actually define the vendor ID in the subdriver source (this is used by parser into udev rules, nut-scanner helper data, etc.), and refer to it from the `...device_table[]`:
```
/* Phoenixtec Power Co., Ltd */
#define PHOENIXTEC_VENDORID     0x06da

...
    { USB_DEVICE(PHOENIXTEC_VENDORID, 0xffff), NULL },
```
* Revise the `idowell_claim()` method to do something like what `mge_claim()` does to pick assumed-handled devices by vendor/product strings, and defer others to `liebert-hid` as the default sink.
* Bump the `IDOWELL_HID_VERSION`
* Adding a bullet point in `NEWS.adoc` would be a nice touch :)
* While at it, adding a Signed-off-by line into the commit comment would be another nice touch, see https://github.com/networkupstools/nut/wiki/Code-contributions,-PRs,-PGP-and-DCO

Looking forward to getting this merged (mergeable) :)

bshor:
Thanks for the detailed review! I've force-pushed a revision that addresses all the points:

- **Named vendor ID:** added `#define PHOENIXTEC_VENDORID 0x06da` and use it for the `0x06da:0xffff` device-table entry instead of the bare literal.
- **Guarded claim:** reworked `idowell_claim()` to mirror `mge_claim()`'s approach — on the shared `PHOENIXTEC_VENDORID` it now only claims the GoldenMate by its firmware strings (manufacturer `-BMS-` / product `Smart-Battery`) and returns 0 otherwise, so AEG PROTECT NAS still goes to `mge-hid` and everything else on `0x06da` falls through to `liebert-hid` as the default sink. The `0x075d` iDowell VID is unique, so it stays claimed unconditionally. (Probe order is `mge -> idowell -> liebert`, so this intercepts only GoldenMate before liebert.)
- **Version:** bumped `IDOWELL_HID_VERSION` to `0.21`.
- **NEWS.adoc:** added a bullet under the `usbhid-ups` driver updates.
- **DCO:** the commit is now `Signed-off-by`.

I don't have a second `0x06da` (AEG/Liebert) device on hand to confirm the deferral empirically, but the string gate is conservative — please let me know if you'd prefer a tighter match (e.g. requiring both strings).

jimklimov:
Looks great, thanks!

Out of curiosity, was AI involved in this PR or is it all a manually investigated and made change?

bshor:
Definitely AI assisted -- Claude Code Opus. I'm overseeing it, but it's handling most of the work.

Issue #3510, opened 2026-06-28
nhs_ser: hardcoded CRTSCTS prevents communication on USB/CDC-ACM (and 3-wire serial) NHS UPSes


jpastoreli:
Hi NUT team, and thank you for `nhs_ser` — it is the only driver that speaks the NHS
binary protocol, and it turned out to be the correct driver for my unit. I hit a
communication blocker that I was able to diagnose down to a single line, and I have a
fix to propose. I also have a couple of secondary observations about status/info
decoding on a newer model, with raw data available if useful.

## Environment

- **NUT version:** built from the `v2.8.5` release tarball. The driver reports
  `DRIVER_VERSION "0.04"`, which is identical in current `master`.
- **OS:** Debian 13 (Proxmox VE 9.2.3), kernel `7.0.6-2-pve`, x86-64.
- **Build:** `./configure --with-drivers=nhs_ser --with-serial ...`
  (the Debian/Proxmox `nut-server` package does not ship `nhs_ser`, so a source build
  is currently required to use it at all).

## Device

- **NHS Mini Gamer** — line-interactive (standby) topology, 600 VA, 1 × 7 Ah / 12 V battery.
- USB communication interface enumerating as **CDC-ACM**:
  `idVendor=0x0925 idProduct=0x1241` ("MCU VIRTUAL COM DEMO", NXP Semiconductors),
  bound by `cdc_acm` → `/dev/ttyACM0`.

---

## Problem 1 (blocker): hardcoded `CRTSCTS` deadlocks the port

With the stock driver the port opens, but no packet is ever assembled:

```
upsdrv_initups: Communication started on port /dev/ttyACM0, baud rate 2400
upsdrv_updateinfo: packet reading did not finish, not interpreting yet      (repeats forever)
```

With `-x debug_pkt_raw` enabled, **no bytes are printed at all** — the driver is
reading nothing.

### Diagnosis

In `openfd()` the driver forces hardware flow control:

```c
tty.c_cflag |= CRTSCTS;   /* Enable hardware flow control */
```

The CDC-ACM virtual port exposes no RTS/CTS lines, so the kernel blocks on flow
control and the driver never receives data.

### Proof

A raw read of the *same* port at the *same* speed, with flow control **disabled**,
receives the NHS stream cleanly:

```sh
stty -F /dev/ttyACM0 2400 cs8 -cstopb -parenb raw -echo -ixon -crtscts
od -An -tx1z /dev/ttyACM0
```

[...]

PR #3511, opened 2026-06-28
drivers: nhs_ser: do not force CRTSCTS (breaks CDC-ACM and 3-wire ser…

jpastoreli:
Disables the hardcoded CRTSCTS in nhs_ser openfd(), which blocks
   communication on NHS UPSes with no RTS/CTS lines (USB/CDC-ACM and the
   official 3-wire serial cable).

   Full diagnosis, raw packet captures and device details: Refs #3510

### DIFF
--- a/drivers/nhs_ser.c
+++ b/drivers/nhs_ser.c
@@ -595,7 +595,7 @@ static int openfd(const char * portarg, int BAUDRATE) {
 	tty.c_cflag |= CS8;		/* 8 bits per byte */
 
 	/* CTS / RTS */
-	tty.c_cflag |= CRTSCTS;	/* Enable hardware flow control */
+    tty.c_cflag &= ~CRTSCTS;	/* NHS serial / CDC-ACM has no RTS/CTS flow control */
 
 	/* Enable Read and disable modem control */
 	/* // tty.c_cflag |= (CLOCAL | CREAD); */


jimklimov:
Generally looks reasonable; deferring to @Challado as original author per #2692 to make sure it works with the model it was originally posted for too.

The PR would benefit from also a note in `NEWS.adoc` about the driver update, and a `DRIVER_VERSION` number bump.

Challado:
> Generally looks reasonable; deferring to @Challado as original author per #2692 to make sure it works with the model it was originally posted for too.
> 
> The PR would benefit from also a note in `NEWS.adoc` about the driver update, and a `DRIVER_VERSION` number bump.

As expected, testing on 3 units that I know the problem appears fail on communication.

Challado:
Re-tested with other 2 units, same problem @jimklimov. I'll see that I can make something about these code and commit a more elegant solution.

Challado:
@jimklimov 
Quick update: I reworked the proposed fix into a more general `nhs_ser` change instead of just removing `CRTSCTS` unconditionally.

The current implementation keeps the existing driver behavior as the default, but makes the relevant serial communication settings configurable from `ups.conf`.

Default behavior is intentionally preserved:

* `baud = 2400`
* `serial_data_bits = 8`
* `serial_parity = none`
* `serial_stop_bits = 1`
* `serial_flow_control = hardware`
* `serial_read_timeout_ms = 100`
* `serial_send_pace_us = 0`

So existing installations should continue to behave as before, including the historical RTS/CTS default.

For the NHS Mini Gamer / CDC-ACM case described in this issue, the fix is now simply:

```ini
serial_flow_control = none
```

I also changed the implementation to use the standard NUT serial helpers instead of keeping direct serial I/O paths in the driver:

* `TYPE_FD_SER`
* `ser_open_nf()`
* `ser_set_speed_nf()`
* `ser_close()`
* `ser_get_char()`
* `ser_send_buf()`
* `ser_send_buf_pace()`
* `ser_flush_io()`

The driver now only keeps local `termios` handling for the settings not covered directly by those helpers: data bits, parity, stop bits and flow control.

I deliberately reduced the scope after testing the first approach. The earlier version exposed too many low-level `termios` and Linux-specific details, so I removed those and kept only conventional serial-port options which are useful for this driver:

* baud rate
* data bits
* parity
* stop bits
* flow control
* per-byte read timeout
* optional transmit pacing

I also removed the Linux-specific `TIOCGSERIAL` / `TIOCSSERIAL` handling from the proposed change. That does not help CDC-ACM devices and seemed unnecessary for solving this issue.

All new options are registered with `addvar(VAR_VALUE, ...)` and validated through `getval()` during driver initialization. Invalid values fail early with a clear configuration error instead of being silently accepted.

The NHS packet protocol handling itself is unchanged: DATA/HWINFO framing, checksums, packet decoding, status handling and reconnect logic remain the same.

I have built the updated driver successfully. My next step is to test it on the affected NHS Mini Gamer device with:

```ini
serial_flow_control = none
```

and also verify that the default configuration still behaves the same on devices which worked with the previous hardcoded RTS/CTS behavior.

jimklimov:
@Challado @jpastoreli  : was the updated fix pushed anywhere and posted as a PR?

Challado:
Hi Jim,

Yes, I pushed the updated implementation to my fork:

https://github.com/Challado/nut

The latest commit contains the corrected and simplified implementation:

https://github.com/Challado/nut/commit/eb0465d4f191068c921b6f80eee656d08193bc32

The preceding commit (`ae8fc51`) was the first implementation and
contained unnecessary complexity; it is corrected by the latest
commit. The final branch state is the intended version.

The change preserves the historical `nhs_ser` serial defaults, but
allows devices which do not use RTS/CTS to configure:

    serial_flow_control = none

It also migrates the serial-port lifecycle and data transfer to the
standard NUT serial helper functions.

I have not opened a separate Pull Request for this implementation yet.
The current comparison against the upstream master branch is available
here:

https://github.com/networkupstools/nut/compare/master...Challado:master

I will be away on vacation for a few days, but I will review any
additional feedback or necessary adjustments when I return.

Best regards,
Lucas

jimklimov:
Thanks, i'll take a look - and gave a good vacation!

Challado:
Jim,
I have created a new pull at code 3543
https://github.com/networkupstools/nut/pull/3543 with significant
improvements on implementation. Please see and fell free to comment
anything.

Em sáb., 18 de jul. de 2026 às 15:11, Jim Klimov ***@***.***>
escreveu:

PR #3521, opened 2026-07-03
nutdrv_qx: Voltronic-QS: Add QI command

Steve-Tech:
Hi, this PR adds the `QI` command for battery charge, runtime, input frequency, output current, and input transfer low/high values to the Voltronic-QS driver.

I am unsure of the consequences of using this with an unsupported UPS, hence marking this PR as a draft. A decompilation of Voltronic's ViewPower software shows this is only used for `H`/`I`/`J` protocols only. I am hoping someone can test this on a `V` protocol UPS, but otherwise let me know if there's a way to disable this for the `V` protocol, or if it's fine anyway.

I found this command while collecting a USB packet capture of a Powershield PSDR800 while using the manufacturer's NetGuard software, which seems to be based off of Voltronic's ViewPower.

Thanks,
Steve

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -81,6 +81,9 @@ https://github.com/networkupstools/nut/milestone/13
       allowing the Megatec payload to be successfully reconstructed. [PR #3485]
     * In `Voltronic-QS-Hex` subdriver, the format string sanity check added in
       recent NUT releases caught a mismatch in `ups.load` reading. [issue #3532]
+    * Modified `voltronic-qs` subdriver to request the QI command, supplying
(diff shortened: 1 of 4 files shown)


jimklimov:
Thanks! I suppose this should be safe for those devices which do not support the command, they should just ignore it somehow. At least, while probing for protocol support, the subdrivers spew all kinds of strings into the UPSes and I don't think anyone got bricked by that yet :D

Steve-Tech:
Oh cool! Thank you for reviewing & merging this, and thanks heaps for maintaining NUT!

PR #3530, opened 2026-07-15
NUT-Monitor: Add automatic reconnection mechanism for Py2Gtk2 as well as Py3Qt{5,6} variants

jimklimov:
Replicate the solution from #3523; testing needed

Should help UI on antique platforms that can not for some reason be updated; in fact I currently struggle to find testbeds with GUI that would be so far behind ;)


jimklimov:
Seems to work. I found a host where Py2Gtk could still run (NetBSD 9.2), connected to a NUT server from NUT-Monitor, and stopped the server.

Within a few seconds, the client status bar was cycling the reconnection messages every couple of seconds:

* image

* image

and it was satisfied soon after I started the data server back:

* image

NOTE: The status bar message remains "Reconnected" even after a couple of hours, as opposed to "Connected..." seen initially.

Double-checked that at least Py3Qt5 variant worked similarly at least once, more details in #3523.

Kudos to @developersorli for the original work on Python3 versions, and IntelliJ Junie for porting it to Python2 :)

jimklimov:
Rebased and updated to address reconnection issues seen with Py3 variants, logged in #3523.

Per testing, reconnection now works for all variants, with Py3Qt* tested over both X11+SSH (with no system notification handler to talk to) and in a natively windowed environment (with two relevant notifications about loss and appearance of connection, no flood).

PR #3547, opened 2026-07-25
riello_usb, riello_ser: report NAK on RE as failure

### LINKED ISSUE #3546: riello_usb / riello_ser: `get_ups_extended()` treats a NAK as success, publishing bogus zeroed `output.L1.power`/`.realpower`/`.current`
When the UPS answers the extended-data command (`RE`) with a NAK, `get_ups_extended()` returns `0` — the same value it returns on success. The caller in `upsdrv_updateinfo()` therefore sets `getextendedOK = 1` and publishes `output.L1.power`, `output.L1.realpower` and `output.L1.current` (plus L2/L3 on three-phase) from `DevData` fields that were **never written**. On a device that does not implement `RE`, these variables are reported as a constant `0` for the lifetime of the driver.

The result is not a missing variable — it is a variable that looks like a valid measurement and is silently wrong. Anything consuming NUT data (dashboards, alerting, `upslog`) has no way to tell it apart from a genuine zero-load reading.

Present in 2.8.1 and still present in `master` at the time of writing, in **both** `drivers/riello_usb.c` and `drivers/riello_ser.c`.

## Environment
- NUT 2.8.1-5 (Debian 13, kernel 6.12.95)
- Driver: `riello_usb` 0.11
- Device: RPS S.p.a. USV5, 800 VA / 480 W
- Firmware: `SWM036-01-02`

## Evidence

Driver run with `-DDDDD` for ~20 s, 10 full update cycles. Every cycle logs:

```
[D3] Get extended Ko: command not supported
```

The raw response to `RE` is a 12-byte frame whose command byte is `0x15` (NAK):

```
02 22 20 15 32 30 30 30 30 3E 39 03
         ^^ NAK
```

Yet `upsc` reports the extended variables as valid data:

```
output.L1.current: 0
output.L1.power: 0
output.L1.realpower: 0
```

These three values are never sourced from the device. They are the initial contents of `DevData.Iout1`, `DevData.Pout1VA` and `DevData.Pout1W`.

For contrast, the `RS` (status) frame from the same device parses perfectly, which shows communication is otherwise healthy. Reconstructed from the HID reports, with the 4-byte trailing CRC verified independently (computed `2373` == observed `2373`, sum mode), the framing is 7 header + 36 payload + 4 CRC + ETX = 48 bytes:

```
byte 12-14  31 3F 35     -> Finp   501   -> input.frequency  50.10
byte 15-17  30 3E 3D     -> Uinp1  237   -> input.voltage    237
byte 18-20  31 3F 34     -> Fout   500   -> output.frequency 50.00
byte 21-23  30 3E 37     -> Uout1  231   -> output.voltage   231
byte 24-25  30 30        -> Pout1    0   -> ups.load         0
byte 26-31  3F 3F 3F ... -> bypass 4095  -> unsupported sentinel
byte 32-35  30 30 38 39  -> Ubat   137   -> battery.voltage  13.7
byte 36-42  3F ...       -> BatCap / BatTime / Tsystem: unsupported sentinels
```

## Root cause

`drivers/riello_usb.c` (master, line ~581):

```c
static int get_ups_extended(void)
{
	...
	/* optional */
	if (!wait_packet && foundnak) {
		upsdebugx (3, "Get extended Ko: command not supported");
		return 0;          /* <-- indistinguishable from success */
	}

	upsdebugx (3, "Get extended Ok: read byte: %d", recv);

	riello_parse_re(&bufIn[0], &DevData);

	return 0;
}
```

The sole caller, `upsdrv_updateinfo()` (line ~1318):

```c
	if (get_ups_extended() == 0)
		getextendedOK = 1;
	else
		getextendedOK = 0;
	...
	if (getextendedOK) {
		dstate_setinfo("output.L1.power", "%u", DevData.Pout1VA);
		dstate_setinfo("output.L1.realpower", "%u", DevData.Pout1W);
		dstate_setinfo("output.L1.current", "%u", DevData.Iout1);
		...
	}
```

The `/* optional */` comment shows the intent: a NAK on `RE` must not be treated as a communication failure. That intent is correct — but the return value also has to tell the caller that no data arrived, and currently it does not. Since `getextendedOK` is the only consumer of the return value, the two concerns can be separated without side effects.

`drivers/riello_ser.c` (line ~352) has the identical pattern (with `/* optonal */` [sic]) and the same single-caller structure at line ~1187.

## Proposed fix

Return a non-zero value on NAK in both drivers. `getextendedOK` then correctly becomes `0` and the uninitialised variables are simply not published. `get_ups_extended()` has exactly one caller in each driver and that caller only compares against `0`, so returning `-1` cannot be mistaken for a comms failure anywhere else.

```diff
--- a/drivers/riello_usb.c
+++ b/drivers/riello_usb.c
@@ static int get_ups_extended(void)
 	/* optional */
 	if (!wait_packet && foundnak) {
 		upsdebugx (3, "Get extended Ko: command not supported");
-		return 0;
+		return -1;
 	}
```

```diff
--- a/drivers/riello_ser.c
+++ b/drivers/riello_ser.c
@@ static int get_ups_extended(void)
 	/* optonal */
 	if (!wait_packet && foundnak) {
 		upsdebugx (3, "Get extended Ko: command not supported");
-		return 0;
+		return -1;
 	}
```

I am happy to open a PR with this if the approach looks right.

## Side note (not part of this report)

On this particular unit `ups.load` is also stuck at `0` while there is real load on the outputs — the UPS genuinely answers `"00"` in the `RS` load field (bytes 24-25 above), rather than the `0x3F` "unsupported" sentinel it uses for bypass, battery capacity, runtime and temperature. That looks like a firmware limitation of this model rather than a NUT bug, and NUT has no way to distinguish it. Mentioned only in case it is useful to whoever maintains the Riello device compatibility notes.

josesj:
Fixes #3546

## What

`get_ups_extended()` returns `0` both on success and when the UPS answers the
extended-data command (`RE`) with a NAK. `upsdrv_updateinfo()` therefore sets
`getextendedOK = 1` and publishes `output.L1.power`, `output.L1.realpower` and
`output.L1.current` (plus L2/L3 on three-phase) from `DevData` fields that were
never written. On a device that does not implement `RE`, these are reported as a
constant `0` for the lifetime of the driver -- indistinguishable from a genuine
zero-load reading.

The `/* optional */` comment marks the intent correctly: a NAK on `RE` must not
be treated as a comms failure. This PR keeps that, but makes the return value
also tell the caller that no data arrived, so the variables are simply not
published.

## How

Return `-1` instead of `0` on the NAK path in both drivers. `get_ups_extended()`
has exactly one caller in each driver and that caller only compares against `0`. 
Driver versions bumped.

## Hardware

Developed and tested against:

- RPS S.p.a. USV5, 800 VA / 480 W, firmware `SWM036-01-02`
- NUT 2.8.1 on Debian 13, driver `riello_usb`

Before the fix, every cycle logs `Get extended Ko: command not supported` (raw
`RE` reply is a 12-byte frame with command byte `0x15`, NAK) while `upsc` still
reports:

```
output.L1.current: 0
output.L1.power: 0
output.L1.realpower: 0
```

After the fix those three variables are absent, and every other variable is
unchanged (`input.voltage`, `input.frequency`, `output.voltage`,
`output.frequency`, `battery.voltage`, `ups.load`). 

`riello_ser.c` carries the identical pattern and the same single-caller
structure, so it is changed for symmetry. I have no serial Riello unit, so that
half is reviewed but **not** tested on hardware.

## AI disclosure

Claude Code was used to help analyse the driver flow and draft the
issue report. The frame decoding and the CRC check were verified by hand 
against the raw USB captures, the fix was reviewed and tested
on real hardware by me, and I take responsibility for the content of this
change.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -125,6 +125,9 @@ https://github.com/networkupstools/nut/milestone/13
     * Completed support for NUT standard USB configuration options, and fixed
       some problems possible when iterating devices. [issue #1768, PR #3477]
 
+ - `riello_usb`, `riello_ser` driver updates:
+	* riello_usb, riello_ser: report NAK on RE as failure [issue #3546]
+
  - `usbhid-ups` driver updates:
     * When reconnecting, report success more visibly (not only in debug), and
       do not reset driver state to "quiet" when it will loop trying. [PR #3423]
--- a/drivers/riello_ser.c
+++ b/drivers/riello_ser.c
@@ -49,7 +49,7 @@
 #include "riello.h"
 
 #define DRIVER_NAME	"Riello serial driver"
-#define DRIVER_VERSION	"0.17"
+#define DRIVER_VERSION	"0.18"
 
 #define DEFAULT_OFFDELAY   5  /*!< seconds (max 0xFF) */
 #define DEFAULT_BOOTDELAY  5  /*!< seconds (max 0xFF) */
@@ -372,7 +372,7 @@ static int get_ups_extended(void)
 	/* optonal */
 	if (!wait_packet && foundnak) {
 		upsdebugx (3, "Get extended Ko: command not supported");
-		return 0;
+		return -1;
 	}
 
 	upsdebugx (3, "Get extended Ok: received byte %u", buf_ptr_length);
--- a/drivers/riello_usb.c
+++ b/drivers/riello_usb.c
@@ -36,7 +36,7 @@
 #include "riello.h"
 
 #define DRIVER_NAME	"Riello USB driver"
-#define DRIVER_VERSION	"0.18"
+#define DRIVER_VERSION	"0.19"
 
 #define DEFAULT_OFFDELAY   5  /*!< seconds (max 0xFF) */
 #define DEFAULT_BOOTDELAY  5  /*!< seconds (max 0xFF) */
@@ -600,7 +600,7 @@ static int get_ups_extended(void)
 	/* optional */
 	if (!wait_packet && foundnak) {
 		upsdebugx (3, "Get extended Ko: command not supported");
-		return 0;
+		return -1;
 	}
 
 	upsdebugx (3, "Get extended Ok: read byte: %d", recv);


jimklimov:
Thanks!

Issue #3136, opened 2025-10-19
Eaton 9PX Resource unavailable, HID length mismatch


desertwitch:
I have an Unraid user reporting their newly purchased Eaton 9PX fails to be detected.
He's tried NUT versions 2.7.4 up until `Network UPS Tools 2.8.4.39-39+g009fbbd1b`.
He's tried various USB ports, cables, hubs - all without success.

```
[ups]
driver = usbhid-ups
port = auto
debug_min = 6
```

```
[nutdev-usb1]
	driver = "usbhid-ups"
	port = "auto"
	vendorid = "0463"
	productid = "FFFF"
	product = "Eaton 9PX"
	serial = "REMOVED"
	vendor = "EATON"
	# bus = "001"
	# device = "012"
	# busport = "003"
```

We can see the following on every driver startup in their logs:
```
Oct 19 08:56:04 Serveur rc.nut:    0.354822#011[D2:132754] Checking device 7 of 8 (0463/FFFF)
Oct 19 08:56:04 Serveur rc.nut:    0.355709#011[D2:132754] - VendorID: 0463
Oct 19 08:56:04 Serveur rc.nut:    0.355710#011[D2:132754] - ProductID: ffff
Oct 19 08:56:04 Serveur rc.nut:    0.355710#011[D2:132754] - Manufacturer: EATON
Oct 19 08:56:04 Serveur rc.nut:    0.355711#011[D2:132754] - Product: Eaton 9PX
Oct 19 08:56:04 Serveur rc.nut:    0.355711#011[D2:132754] - Serial Number: REMOVED
Oct 19 08:56:04 Serveur rc.nut:    0.355712#011[D2:132754] - Bus: 001
Oct 19 08:56:04 Serveur rc.nut:    0.355712#011[D2:132754] - Bus Port: 011
Oct 19 08:56:04 Serveur rc.nut:    0.355713#011[D2:132754] - Device: 004
Oct 19 08:56:04 Serveur rc.nut:    0.355713#011[D2:132754] - Device release number: 0202
Oct 19 08:56:04 Serveur rc.nut:    0.355714#011[D2:132754] Trying to match device
Oct 19 08:56:04 Serveur rc.nut:    0.355714#011[D2:132754] match_function_subdriver (non-SHUT mode): matching a device...
Oct 19 08:56:04 Serveur rc.nut:    0.355715#011[D3:132754] match_function_regex: matching a device...
[...]

PR #3550, opened 2026-07-28
drivers/libusb{0,1}.c: bound rdlens loop by element count, not sizeof

0x-a6:
`nut_libusb_open()` iterates the two candidate HID report descriptor lengths:

```c
	int32_t rdlen1, rdlen2, rdlens[2];
	size_t j;
	...
	for (j = 0; j < sizeof(rdlens); j++) {
		rdlen = rdlens[j];
```

`sizeof(rdlens)` is the size in **bytes** (8), not the element count (2), so the loop reads `rdlens[2..7]` — 24 bytes past the end of the array — and uses that stack garbage as `rdlen`. The "ran out of candidates" test after the loop (`if (j >= sizeof(rdlens))`) has the same bug, so it never fires as intended either.

The out-of-range indices are only reached when neither real candidate satisfies the caller — i.e. exactly when a device is in a bad state and both descriptor reads fail, which is when you least want undefined behaviour. Garbage that happens to survive the `rdlen` sanity checks is then passed to `libusb_control_transfer()` and on to the HID parser callback, which is where it segfaults.

### How it showed up

`usbhid-ups` segfaulting against an EcoFlow DELTA 3 Plus (`3746:ffff`) on an NVIDIA Jetson Orin Nano (aarch64, NUT built from `v2.8.5`) after the device's USB connection had wedged. `usbreset` on the device cleared the wedge, and the crash went away with it — the descriptor reads then succeeded on the first candidate and the loop never ran past index 0.

### Fix

Use `SIZEOF_ARRAY()` from `include/common.h`, which both files already include. `drivers/libusb0.c` carries the identical defect in the same two places and is fixed the same way.

Regression from aba6f43544b17d2f16e69ca19f162c6c19748fc7 ("if the tried `rdlen` did not succeed, fall back to the other value we had in mind" [#3136]), first released in v2.8.5.

Happy to add a `NEWS.adoc` entry under the 2.8.6 section if you'd like one — I left it out to keep the diff to the fix, since the entry would want this PR's number.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -171,6 +171,8 @@ https://github.com/networkupstools/nut/milestone/13
       chunks populated only partially with bytes we intended. These buffers are
       now more diligently pre-zeroed as their siblings were in other code paths,
       to avoid sending host stack garbage to a device. [#3529]
+    * USB drivers: fixed an out-of-bounds read of the candidate HID report
+      descriptor lengths, which could cause a segmentation fault. [PR #3550]
 
  - NUT client libraries:
     * Complete support for actions documented in `docs/net-protocol.txt`
--- a/drivers/libusb0.c
+++ b/drivers/libusb0.c
@@ -651,7 +651,7 @@ static int nut_libusb_open(usb_dev_handle **udevp,
 				}
 			}
 
-			for (j = 0; j < sizeof(rdlens); j++) {
+			for (j = 0; j < SIZEOF_ARRAY(rdlens); j++) {
 				rdlen = rdlens[j];
 				if (rdlen < 0)
 					continue;
@@ -716,7 +716,7 @@ static int nut_libusb_open(usb_dev_handle **udevp,
 				break;
 			}
 
-			if (j >= sizeof(rdlens)) {
+			if (j >= SIZEOF_ARRAY(rdlens)) {
 				/* Ended the loop without success */
 				goto next_device;
 			}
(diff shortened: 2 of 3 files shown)


jimklimov:
My bad, thanks for the catch and fix! A news entry would be appropriate, perhaps with a new block like `- Fix fallout of development in ...` seen for some earlier releases. Maybe without reference to specific commit (rather to PR number) to avoid the hassle with `nut.dict` :)

0x-a6:
Added to common code, but happy to put it into it's own block.

jimklimov:
Thanks, ok for now - I've actually added such a block in #3551 so will (hopefully remember to) relocate your entry there after both PRs get merged.

strasharo:
Hit this independently on a NUT 2.8.5.1 (post-2.8.5, "development iteration") build, Alpine-based image, against an Eaton 5E 650i (0463:ffff) on Arch Linux/x86_64.

`usbhid-ups` segfaulted right as the UPS re-enumerated on the USB bus after a driver restart (bus/device number changed under it while probing). Two crashes back-to-back, then a third `upsdrvctl` retry succeeded once the topology settled - matches the "only reached when both descriptor reads fail" trigger described above exactly.

Core dump backtrace confirms the same site:
```
#0 nut_libusb_open (...) at libusb1.c:811
811		rdlen = rdlens[j];
```
with `rdlens = {549, -1}` and the crashing read walking off the end of the 2-element array, same as described. Confirms this isn't hardware-specific - same failure mode on a different UPS model and a different trigger path (post-restart USB re-enumeration race rather than a wedged/wall-power device).

jimklimov:
Thank you for the cross-check and sorry about the confusion I introduced. Should be fixed on master branch now.

PR #3559, opened 2026-08-04
win32: return partial serial data on timeout

### LINKED ISSUE #3452: nutdrv_qx (Megatec) errors out after reading UPS data on Windows over USB CDC/COM
When using the `nutdrv_qx` driver with the `megatec` protocol on Windows (via a USB CDC/COM serial connection), the driver successfully reads exactly 47 bytes of data sent by the UPS in response to `Q1`. However, immediately after reading the 47th byte, it hangs for exactly 1 second, resulting in an I/O error (`err 0`) and a failure to preprocess the answer.

My hypothesis is that the Win32 serial wrapper (`w32_serial_read`) might be failing to correctly catch the `\r` terminator to close the read operation early, or there is an issue with the buffer/timeout configurations on Windows compared to Linux.

### To Reproduce
1. Run NUT 2.8.5 on Windows.
2. Connect a UPS using the Megatec protocol over a virtual serial port (USB CDC/COM).
3. Execute the driver directly with maximum debug level: 
   `.\nutdrv_qx -a tsshara -DDDD`
4. Observe the log where it reads exactly 47 bytes, waits exactly 1 second, and throws a timeout/I/O error.

### Expected behavior
The read operation should complete successfully upon receiving the `\r` terminator (the 47th byte) and pass the buffer to `qx_process` without hitting the 1-second timeout.

### Logs
```text
   1.435648     [D3:tsshara] send: 'Q1'
   1.435839     [D4:tsshara] w32_serial_read : ulen 511, vmin_ -1, vtime_ 0, hEvent 0000000000000118
[...]
   1.639599     [D4:tsshara] w32_serial_read : total characters read = 46
   1.643466     [D4:tsshara] w32_serial_read : characters are available on input buffer
   1.643553     [D4:tsshara] w32_serial_read : Reading 1 characters
   1.643579     [D4:tsshara] w32_serial_read : total characters read = 47
   2.643838     [D4:tsshara] w32_serial_read : err 0
   2.644013     [D3:tsshara] read: Input/output error (-1)
   2.644143     [D4:tsshara] qx_process: failed to preprocess answer [input.voltage]
```

### Environment / Desktop
- OS: Windows (x86_64)
- NUT Version: 2.8.5 (64-bit build for x86_64, built with gcc from MSYS2)
- Installation method: Windows binaries
- Device: TS Shara UPS via Serial (USB CDC/COM mapped to `\\.\COM3`)
- Driver / Protocol: `nutdrv_qx` / `megatec`

### Additional context
- The same problem does not happen on Ubuntu, where I tested with the same configs.
- By using PuTTY to connect directly to `COM3` and sending `Q1\r`, the UPS responds immediately with:
`(231.0 231.0 116.0 011 60.0 27.2 41.0 00001000`. This string length is exactly 46 characters + `\r` (47 bytes). The driver log clearly shows it successfully reading byte by byte up to that exact 47th byte (at `1.643s`), at which point it halts and times out exactly 1.000 second later (at `2.643s`).

*Disclaimer:* I have used AI to help me write this issue, since is my ever first issue and english is not my main language.

Marcelo-PX:
## Summary

Fix the Win32 serial read wrapper so that data already accumulated before a
timeout is returned to the caller instead of being discarded as an I/O error.

When `nutdrv_qx` receives a complete Q1 response over a Windows COM port, the
subsequent timeout currently falls through to the generic error path even
though valid bytes were already received.

This change cancels the pending overlapped operation, resets its state, clears
the Win32 error status and returns the accumulated byte count when the timeout
occurs after receiving data.

Fixes #3452.

## Testing

Tested on Windows x86_64 with a UPSBrasil 3 kVA UPS using:

- `nutdrv_qx`
- Q1 protocol
- USB-to-serial adapter on COM4
- NUT 2.8.5 plus the current development branch

Verified:

- the complete Q1 response is accepted without the previous EIO failure;
- continuous polling remains stable;
- `upsd`, `upsc` and `upsmon` receive the UPS data;
- the Windows NUT service starts automatically after reboot;
- communication recovers after disconnecting and reconnecting the USB serial
  adapter without restarting the driver process;
- UPS status changes between `OL` and `OL BYPASS` are reported correctly.

The broader `ci_build.sh` run stopped on an unrelated GCC 16 `-Werror` warning
in `nut-scanner`, after the affected driver and libraries had compiled.

## AI disclosure

ChatGPT was used to assist with diagnosing the Win32 timeout behavior,
reviewing the proposed change, and drafting the issue and pull request text.

The submitted change was manually reviewed, compiled and tested against real
hardware by the contributor.

## General points

  comments or text documents.

- [x] Use of coding helper tools and AI should be disclosed in the commit
  or PR comments (it is interesting to know which ones do a decent job).
  As with other contributions, a human is responsible and thanked for the
  quality and content of the change, and is presumed to have the right to
  post that code to be published further under the project's license terms.

- [x] Especially with involvement of AI, including modern IDE coding aid,
  please be sure to revise that proposed code and documentation changes
  follow NUT code style guide -- this helps portability across the decades
  worth of supported systems. Notably, avoid Unicode characters where ASCII
  text is expected.

  case.

  or its equivalent.

  VID/PID combinations, or raised discussions when this is not the case.

  `scripts/upower/95-upower-hid.hwdb` file.

- [ ] Proposed NUT data mapping is aligned with existing
  `docs/nut-names.txt` file.

- [ ] Updated `data/driver.list.in` if applicable.

  structure layout and alignment in memory, endianness, or use of generic
  `int` where language or libraries dictate the use of `size_t` or
  `ssize_t`.

- [x] Coding style, including whitespace for indentations, follows precedent
  in the code of the file and the guidance in `docs/developers.txt`.

- [ ] Added a bullet point into `NEWS.adoc`, possibly also `UPGRADING.adoc`
  if there is something packagers or custom-build users should take into
  account.

- [ ] Updated `docs/acknowledgements.txt`.

- [ ] Added or updated manual page information in `docs/man/*.txt` files and
  corresponding recipe lists in `docs/man/Makefile.am`.

- [ ] Passed `make spellcheck` and updated `docs/nut.dict` if needed.

  against real hardware.

- [ ] Address NUT CI farm build failures for the PR.

- [ ] Revise suggestions from LGTM.COM analysis about new issues.

### DIFF
--- a/common/wincompat.c
+++ b/common/wincompat.c
@@ -898,6 +898,15 @@ int w32_serial_read(serial_handler_t *sh, void *ptr, size_t ulen, DWORD timeout)
 							errno = 0;
 							return 0;
 						}
+						CancelIo(sh->handle);
+						sh->overlapped_armed = 0;
+						ResetEvent(sh->io_status.hEvent);
+						upsdebugx(4,
+							"w32_serial_read : timeout after receiving %d characters, returning partial data",
+							tot);
+						SetLastError(ERROR_SUCCESS);
+						errno = 0;
+						return tot;
 					default:
 						goto err;
 				}


jimklimov:
I wonder if it is correct to return partial data, without even a hint like setting `errno`? In case of `nutdrv_qx`, ASCII `\r` ending the line (or not) can be a good clue; also expected-length checks, to parse or discard a really short read. I hope other drivers are similarly protected (or should become if not).

marcelodotnet:
> I wonder if it is correct to return partial data, without even a hint like setting `errno`? In case of `nutdrv_qx`, ASCII `\r` ending the line (or not) can be a good clue; also expected-length checks, to parse or discard a really short read. I hope other drivers are similarly protected (or should become if not).

That makes sense.

In my test with `nutdrv_qx`, the UPS returned the complete Q1 response including
the final `\r`, and the driver parsed it correctly. My goal was only to avoid
discarding valid bytes that had already been received on Windows.

But I agree that `w32_serial_read()` is shared by other serial drivers, so
returning partial data with `errno` cleared could hide a truncated response.

Would it be better to return the byte count but keep `errno = ETIMEDOUT` and
`SetLastError(WAIT_TIMEOUT)`, so the caller can detect that the read ended by
timeout? Or do you think this validation should be handled specifically by
`nutdrv_qx`, checking the final `\r` or the expected response length?

jimklimov:
TL;DR: Ultimately, probably what you did is in fact the correct way to do it for the case where `tot > 0` (and code already handled `if (!tot)` just above your addition, although maybe setting non-zero `errno` *there* would be right).

I thought returning the byte count *and* setting the `errno` (and/or equivalent) would be the most honest way forward. Having said that, I am not sure OTOH how the POSIX systems go about this (if there is some sort of standard on that, possibly with `select` or similar methods), so I asked AI for a second opinion.

A quick "Chat with GPT" confirmed it also does not know about any precedents directly like what I first suggested either. Portable readers just loop `while read() > 0` if they expect more data than a single buffer's worth (or expect to collect a specific packet size), so stopping at EOF or error cases that either way filled zero bytes in the buffer. Quoting from it directly: 

POSIX does not define a mechanism whereby `read()` both returns a positive byte count *and* reports an error via `errno`. In fact, the POSIX API is intentionally designed so that a successful return (`>= 0`) means `errno` is unspecified and should be ignored.

Specifically, `read()` returns:
* `> 0`: that many bytes were read successfully.
* `0`: EOF.
* `-1`: failure, with `errno` indicating the reason.

POSIX explicitly treats `errno` as meaningful *only if the function indicates failure*.

*Partial reads are not errors*

A partial read is already a normal, expected outcome for many kinds of file descriptors:

* pipes
* terminals
* sockets
* some devices
* interrupted reads
* nonblocking descriptors

The standard already allows
```
ssize_t n = read(fd, buf, 4096);
```
to return
```
127
```
without implying anything is wrong.

The application is expected to call `read()` again if it wants more data.

*Timeouts*

POSIX generally treats timeouts as one of two things:

1. No data transferred

If no bytes have been transferred before the timeout expires, the operation fails:
```
read() -> -1
errno = EAGAIN
```
or
```
ETIMEDOUT
```
depending on the API and object.

For example, sockets with `SO_RCVTIMEO` commonly behave this way.

2. Some data transferred

If some bytes have already been copied into the user's buffer, virtually every implementation considers the operation *successful* and simply returns the byte count.

The timeout is effectively forgotten.

This mirrors how `EINTR` works.

For example, POSIX says (paraphrasing):

* interrupted before reading *anything* → return `-1`, `errno=EINTR`
* interrupted after reading some bytes → return the bytes read

Timeouts are commonly implemented analogously.

*Why not both?*

Suppose
```
read(fd, buf, 4096)
```
copies
```
1500 bytes
```
and then times out.

If it returned
```
1500
errno = ETIMEDOUT
```
what should portable code do?

Typical code is
```
n = read(fd, buf, size);
if (n < 0) {
    perror("read");
}
```
It never even looks at `errno` after success.

Changing that convention would break decades of software.

*Existing precedents*

I can't think of a POSIX interface that returns both

* successful primary result, and
* mandatory secondary error via `errno`.

There are *other* APIs that return additional status separately, e.g. `recvmsg()` returns

* byte count
* flags

where flags may contain
```
MSG_TRUNC
MSG_CTRUNC
MSG_EOR
```
These are not reported through `errno` but extra OUT-args.

If a datagram is larger than the supplied buffer, it returns `buffer_size` and sets `MSG_TRUNC` instead of failing with `errno=EMSGSIZE`.

Similarly,
```
poll()
select()
epoll_wait()
kevent()
```
return readiness information separately from errors.

Some APIs have out-parameters for auxiliary status. But POSIX generally avoids "success plus errno".

jimklimov:
Huh, why closed?

UPDATE: refactored via #3560

PR #3565, opened 2026-08-12
nutdrv_qx: add richcomm-svc USB transport for SVC V-1000F

### LINKED ISSUE #3562: nutdrv_qx: session initialization required for SVC V-1000F Richcomm variant (0925:1234)
## Summary

The SVC V-1000F UPS uses the Megatec protocol over a Richcomm USB-to-UART bridge with USB ID `0925:1234`. The current `armac` USB communication subdriver in NUT 2.8.5 claims the device, but receives no usable reply to `Q1` and eventually reports `Device not supported!`.

This is related to #1978, #2219, #3132 and PR #2154, but the existing `armac` implementation does not support this session-initialized Richcomm variant.

Analysis of the USB communication used by the vendor's Windows companion software showed that this variant needs two initialization steps before it accepts Megatec commands:

1. Configure the USB-UART bridge for 2400 baud.
2. Send a 64-byte session-opening sequence.

After adding these steps locally to `armac_command()`, the UPS responds to `Q1`, reports valid data through `upsc`, and accepts the standard NUT quick battery-test instant command.

## Hardware

- Vendor/brand: SVC
- Model: V-1000F
- USB Vendor ID: `0925`
- USB Product ID: `1234`
- USB manufacturer string: `RICHCOMM`
- USB product string seen on this hardware: `UPS USB MON V1.4`
- The same VID:PID is also widely reported as `UPS USB Mon V2.0`

The device presents a vendor-specific HID interface used as a USB-to-UART bridge. As already discussed in PR #2154, this VID:PID is shared by Richcomm devices with different communication details, so VID:PID alone is not enough to select the exact behavior safely.

## Environment

- Linux Mint 21.3
- Kernel: `6.8.0-136-generic`
- Architecture: `x86_64`
- NUT source: tag `v2.8.5`, commit `0e051f9f6853e9ae0087b388b70840c9651a5eed`
- `nutdrv_qx` internal version: `0.49`
- NUT USB communication driver: `0.53`, libusb 1.0

## Configuration before the local patch

```ini
[svc1000]
    driver = nutdrv_qx
    port = auto
    subdriver = armac
    protocol = megatec
    vendorid = 0925
    productid = 1234
    novendor
    norating
    desc = "SVC V-1000F"
```

## Failure with unmodified NUT 2.8.5

The USB device matches and interface 0 is claimed successfully, but `Q1` gets no UART payload:

```text
[D2] Device matches
[D2] Claimed interface 0 successfully
[D4] armac command Q1
[D1] interrupt read error: Input/Output Error (-1)
[D4] qx_process: failed to preprocess answer [input.voltage]
Device not supported!
```

The same UPS was also rejected by the older NUT 2.8.0 `armac` implementation.

### 1. UART configuration

Before the first Qx command, the companion software sends a class-interface `SET_REPORT` control transfer:

```text
bmRequestType = 0x21
bRequest      = 0x09
wValue        = 0x0200
wIndex        = 0
data          = 80 d0 00 00
```

This configures the bridge for the 2400-baud serial link used by the UPS.

### 2. Session-opening sequence

It then sends this 64-byte sequence as sixteen 4-byte `SET_REPORT` transfers with the same request fields:

```text
40 66 af f8 41 8a d3 1c 65 ae f7 40 89 d2 1b 64
ad f6 3f 88 d1 1a 63 ac f5 3e 87 d0 19 62 ab f4
3d 86 cf 18 61 aa f3 3c 85 ce 17 60 a9 f2 3b 84
cd 16 5f a8 f1 3a 83 cc 15 5e a7 f0 39 82 cb 14
```

After these transfers, regular Megatec commands such as `Q1\r` work through the existing ARM/MAC HID framing.

## Empty HID reports

This device can return idle HID reports before a UART response. The useful-byte count is in the low six bits of byte 0. A report with a zero count before any UART data should not immediately terminate the read.

The local test patch ignores up to five leading idle reports. This is also needed for instant commands which take effect without returning a textual Megatec response.

## Result with the local patch

`upsc svc1000@localhost` reports stable values such as:

```text
ups.status: OL
input.voltage: 218.0
output.voltage: 218.0
battery.voltage: 27.0
ups.load: 15
```

The existing Megatec mapping and parser are sufficient after the USB session is initialized; no new Qx protocol dialect was needed.

## Instant-command test

The standard command was sent through `upsd`/`upscmd`:

```text
test.battery.start.quick
SUCCESS
```

Observed status sequence:

```text
OL -> OB (approximately 6 seconds) -> OL
```

The UPS transferred to battery and returned to line power normally. No deep battery test or load/shutdown command was attempted. `test.battery.stop` was available but was not needed because the quick test ended normally.

## Proposed upstream direction

The Qx dialect is still Megatec, while the special behavior belongs to the Serial-over-USB transport layer. A dedicated explicit USB communication subdriver, tentatively named `richcomm-svc`, may therefore be safer than enabling this sequence for every `0925:1234` device:

```ini
subdriver = richcomm-svc
protocol = megatec
```

An alternative would be an explicit opt-in quirk/configuration flag in `armac`. Automatic activation based only on `0925:1234` may affect other Richcomm variants which already work with the current control-transfer or interrupt-endpoint ARM/MAC implementations.

The final implementation should also reset its session state after USB reconnect and retry initialization when communication becomes stale.

Would the maintainers prefer:

1. a dedicated USB communication subdriver sharing common ARM/MAC framing code; or
2. an explicit session-initialization quirk inside the existing `armac` subdriver?

I have a small proof-of-concept patch against NUT 2.8.5 and physical access to the SVC V-1000F for further testing. I can rework it against current `master` after guidance on the preferred structure.

The implementation is independently written from observed device communication. No proprietary binaries or decompiled vendor source code will be submitted.

alliancesko:
`nutdrv_qx: add richcomm-svc USB transport for SVC V-1000F`

## Summary

This adds an explicit `richcomm-svc` USB communication subdriver to
`nutdrv_qx` for older session-initialized Richcomm USB-to-UART bridges.

The implementation was developed and tested with an SVC V-1000F reporting
USB ID `0925:1234` and product string `UPS USB MON V1.4`. This device uses the
existing Megatec protocol and Armac command framing, but does not answer Qx
commands until its bridge has been configured for 2400 baud and sent a
64-byte session-opening sequence.

Closes #3562.

Commit tested: `4e1aa23b32376658f455110cf9d1c880c0176ff2`.

## Implementation

- Add `richcomm-svc` as an explicit `nutdrv_qx` USB communication subdriver.
- Share the existing Armac command framing instead of adding another Qx
  protocol dialect.
- Configure the bridge for 2400 baud and send its session-opening reports
  before the first command.
- Ignore up to five leading empty HID reports before a UART response.
- Use the 6-byte interrupt reads required by the V-1000F; the 8-byte reads
  used by the current normal `armac` transport cause `LIBUSB_ERROR_OVERFLOW`
  on this hardware.
- Reset the transport initialization state after USB reconnection.
- Reject failed and short initialization transfers.
- Bump the internal `nutdrv_qx` version from `0.54` to `0.55`.
- Document the subdriver, tested hardware, mandatory configuration and known
  test limits in the man page, HCL and NEWS.

The mode is deliberately not selected automatically. USB ID `0925:1234` is
shared by devices with different communication details, including hardware
which already works with the normal `armac` transport.

## Configuration

```ini
[svc-v1000f]
    driver = nutdrv_qx
    port = auto
    subdriver = richcomm-svc
    protocol = megatec
    vendorid = 0925
    productid = 1234
```

## Hardware verification

Tested on a physical SVC V-1000F under the submitter's control:

- polling and Megatec protocol detection;
- stable `Q1`, `I` and `F` responses;
- live NUT values including `OL`/`OB`, input and output voltage, battery
  voltage, load, frequency and temperature;
- `test.battery.start.quick`, with the observed sequence `OL -> OB -> OL`;
- fresh initialization after physically disconnecting and reconnecting USB.

Deep battery tests and load or shutdown commands were not attempted.

The implementation is based on independently observed USB communication. It
does not include copied or decompiled proprietary source code.

## Checks

- `make -j4 -C drivers nutdrv_qx`
- `make -j4 check` — 6 tests passed, 0 failed
- documentation spellcheck and man-source sanity check
- `git diff --check`
- `nutdrv_qx --help` lists `richcomm-svc`
- physical read-only test passed with the v2.8.5 proof of concept
- physical read-only test of the final `master`-based binary passed on the
  SVC V-1000F: Megatec 0.10 was detected and live status data was read

## AI assistance disclosure

The protocol investigation, patch development, documentation and PR drafting
were assisted by OpenAI Codex. The submitter directed the work, performed the
tests on physical hardware, reviewed the resulting behavior and remains
responsible for the contribution.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -83,6 +83,11 @@ https://github.com/networkupstools/nut/milestone/13
       reconnection behavior. [issue #3510, PR #3543]
 
  - `nutdrv_qx` driver updates:
+    * Added the `richcomm-svc` USB communication subdriver for older
+      session-initialized Richcomm bridges. It configures the bridge before
+      reusing the existing Armac command framing and Megatec protocol. Polling
+      and a quick battery test were verified with an SVC V-1000F
+      (`0925:1234`, `UPS USB MON V1.4`). [issue #3562]
     * Only claim a USB device as "supported" during discovery out of the box
       if `subdriver_command` was assigned (`1A86:7523` is used by CH340/341
       USB chips not only in UPSes). [issue #3410]
--- a/data/driver.list.in
+++ b/data/driver.list.in
@@ -1453,6 +1453,7 @@
 "SuperPower"	"ups"	"2"	"HP360, Hope-550"	""	"blazer_ser"
 
 "SVC"	"ups"	"2"	"VP-1250-LCD 1250VA"	"(USB ID 0925:1234, armac subdriver, megatec protocol)"	"nutdrv_qx"	# https://github.com/networkupstools/nut/pull/3485
+"SVC"	"ups"	"1"	"V-1000F"	"USB (USB ID 0925:1234, product UPS USB MON V1.4)"	"nutdrv_qx port=auto vendorid=0925 productid=1234 protocol=megatec subdriver=richcomm-svc"	# https://github.com/networkupstools/nut/issues/3562 ; polling and quick battery test verified on physical hardware; deep battery tests and load or shutdown commands untested
 
 "SVEN"	"ups"	"2"	"Power Pro+ series"	"USB"	"blazer_usb (USB ID ffff:0000)"
 "SVEN"	"ups"	"2"	"Power Pro+ series"	"USB"	"blazer_usb (USB ID 05b8:0000)"
--- a/docs/man/nutdrv_qx.txt
+++ b/docs/man/nutdrv_qx.txt
@@ -487,7 +487,7 @@ include::nut_usb_addvars.txt[]
 
 *subdriver =* 'string'::
 Select a USB communication subdriver to use.
-You have a choice between *ablerex*, *armac*, *cypress*, *fabula*, *fuji*, *gtec*, *hunnox*, *ippon*, *krauler*, *omron*, *phoenix*, *phoenixtec*, *sgs* and *snr*.
+You have a choice between *ablerex*, *armac*, *cypress*, *fabula*, *fuji*, *gtec*, *hunnox*, *ippon*, *krauler*, *omron*, *phoenix*, *phoenixtec*, *richcomm-svc*, *sgs* and *snr*.
 +
 Run the driver program with the `--help` option to see the exact list of
 `subdriver` values it would currently recognize.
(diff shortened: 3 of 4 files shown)


jimklimov:
Thanks, looks great!

alliancesko:
> Thanks, looks great!

It works for me. I need to check with other people now. Glad to help. I had problems with similar UPS for a long time. With the release of version 2.8.5, things started moving forward.

PR #3567, opened 2026-08-13
Some follow-up for NUT Auth and revise Jenkinsfile-dynamatrix test-case counts

jimklimov:
Mostly small follow-up from #3331, and a testbed for dynamatirx refactoring so we can achieve similar coverage faster.


jimklimov:
The mix of `Jenkinsfile` changes in this PR and jenkins-dynamatrix in its repo, iterated together, led to non-instantiation of several slow build scenario groups that were expected to be seen (and correctly removed scenarios which this PR made redundant). Those groups are further recovered in work on PR #3577, again with bits of solution in both Jenkins library and consumer repos.

PR #3414, opened 2026-04-14
apc_modbus: simplify error handling with read retries

EchterAgo:
Replace the complex `_apc_modbus_handle_error` function with a simpler retry mechanism built into `_apc_modbus_read_registers`.

The new approach:
- Retries register reads on `ETIMEDOUT` errors up to `modbus_retries` times (configurable, default 3)
- On non-timeout errors or after retry exhaustion, closes the connection for reconnection on the next update cycle
- Removes the platform-specific (WIN32/POSIX) timeout detection and the flush-based recovery that didn't work anyway (flush is already done in `_apc_modbus_reopen` upon reconnection)

Also adds a `modbus_retries` driver option to configure the number of read retry attempts, and improves logging for connection open/close events.

This change was inspired by a patch by @marcan to do the same and from testing the behaviour of `apcupsd`, noticing that on my USB unit it times out on read and only succeeds on the first retry.

## General points

- [x] Described the changes in the PR submission or a separate issue, e.g.
  known published or discovered protocols, applicable hardware (expected
  compatible and actually tested/developed against), limitations, etc.

- [x] There may be multiple commits in the PR, aligned and commented with
  a functional change. Notably, coding style changes better belong in a
  separate PR, but certainly in a dedicated commit to simplify reviews
  of "real" changes in the other commits. Similarly for typo fixes in
  comments or text documents.

- [x] Use of coding helper tools and AI should be disclosed in the commit
  or PR comments (it is interesting to know which ones do a decent job).
  As with other contributions, a human is responsible and thanked for the
  quality and content of the change, and is presumed to have the right to
  post that code to be published further under the project's license terms.

- [x] Please star NUT on GitHub, this helps with sponsorships! ;)

## Frequent "underwater rocks" for driver addition/update PRs

- [x] Revised existing driver families and added a sub-driver if applicable
  (`nutdrv_qx`, `usbhid-ups`...) or added a brand new driver in the other
  case.

- [x] Did not extend obsoleted drivers with new hardware support features
  (notably `blazer` and other single-device family drivers for Qx protocols,
  except the new `nutdrv_qx` which should cover them all).

- [x] For updated existing device drivers, bumped the `DRIVER_VERSION` macro
  or its equivalent.

- [x] For USB devices (HID or not), revised that the driver uses unique
  VID/PID combinations, or raised discussions when this is not the case
  (several vendors do use same interface chips for unrelated protocols).

- [x] For new USB devices, built and committed the changes for the
[...]

PR #3571, opened 2026-08-14
apc_modbus: three fixes for the #3414 read-retry loop

d01:
Three defects in the register-read retry loop added by #3414, found while
getting `apc_modbus` running reliably on an APC Smart-UPS X1500 (051d:0003,
FW "UPS 16.0") over USB. Each is device-independent; the X1500 is just the
hardware that makes all three visible at once. Compile-tested on master and
running in production (backported) on 2.8.4.

### 1. The loop ignores `exit_flag`

A driver asked to stop keeps working through its remaining retries — up to
`retries × response_timeout`, which with non-default settings comfortably
exceeds the 5 s upsdrvctl allows for SIGTERM before escalating to SIGKILL.
On the X1500 a SIGKILL mid-exchange leaves the device unable to serve Modbus
until its USB cable is physically reseated, so every clean shutdown wedged
the UPS. Journal signature before the fix:

```
Stopping /run/nut/apc_modbus-smx.pid failed, retrying harder: Success
Main process exited, code=killed, status=9/KILL
```

### 2. The loop retries only ETIMEDOUT

It gives up on `EMBBADSLAVE`/`EMBBADCRC`/`EMBBADDATA` — the one case where a
retry is most obviously right, because a reply arrived that belongs to an
*earlier* request, and reading it has already consumed it. On packetised
transports (rtu_usb) a deferred reply stays queued on the endpoint rather
than ageing off the wire, so this is routine: the first driver start after a
host reboot collects a reply queued before the reboot (the self-powered UPS
keeps its state), fails, and systemd restart-loops the driver, degrading the
device further with every failed start.

### 3. Every attempt is logged at LOG_ERR

The pre-#3414 error log was left inside the new loop, so attempts the next
retry recovers are still reported as errors — the exact case the loop was
added for. Measured here in a degraded state: ~310 LOG_ERR/min while the
connection was never closed once and every read succeeded; ~890k lines/day,
42% of the journal (doubled by upsdrvctl forwarding driver stderr alongside
the syslog write). Attempts now log at `upsdebugx(1)`; the error logs once,
after the loop, with the attempt count.

### Context

The wider investigation (the X1500's post-reset behaviour, its reply
queueing, and the transport-level fixes in networkupstools/libmodbus#12) is
written up separately; these three fixes stand on their own and are the
device-independent part. Related: #2609, #3414.

### Note on authorship

Developed with assistance from Claude (Anthropic); commits carry a
Co-Authored-By trailer to that effect. All findings were verified on real
hardware.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -70,6 +70,11 @@ https://github.com/networkupstools/nut/milestone/13
     * Added `modbus_retries` option to configure the number of read retry
       attempts on timeout errors, simplifying error recovery. [PR #3414]
     * Increased default TCP response timeout to 2000 ms. [PR #3418]
+    * Three fixes for the read-retry loop added in PR #3414:
+      the loop ignored `exit_flag`, causing `upsdrvctl` to kill the driver
+      and wedge the device if it was killed mid-exchange; now we retry on more
+      failure causes than just a timeout; quieter retry logging to avoid
+      severe syslog/journal thrashing. [PR #3571, #3414]
 
  - `dummy-ups` driver updates:
     * Added `authconf` driver parameter for repeater mode to control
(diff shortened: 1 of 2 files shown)


jimklimov:
Great catches, thanks!

jimklimov:
Some (apparently older) systems complain:
```
apc_modbus.c:1060:22: error: use of undeclared identifier 'EMBBADSLAVE'
                &&  saved_errno != EMBBADSLAVE
```

Probably coincides with them using an older `mobus.h`

PR #3577, opened 2026-08-18
`configure.ac`: revise `NUT_ARG*` methods to remember if the option was "given" by caller or defaulted, address valgrind findings

### LINKED ISSUE #3513: configure error with nut-2.8.5
With nut-2.8.5 when running the following configure build command in the IPFire build

    configure \
				--prefix=/usr \
				--sysconfdir=/etc/nut \
				--with-usb \
				--with-user=root \
				--with-group=nut \
				--with-wrap=no \
				--with-udev-dir=/etc/udev \
				--disable-static \
				--without-docs \
				--with-dev

The following error message is shown

    checking whether compiler suggests a MULTIARCH value... checking if requested user to run as... specified
    checking setting for User to run as... "root"
    checking if requested group membership of user to own state files... default
    checking setting for Group of user to own state files... "nut"
    configure: error: If you specify --with-user, you also must specify --with-group
    make: *** [nut:87: /usr/src/log/nut-2.8.5] Error 1

With nut-2.8.4 there is no error message with exactly the same configure command

    checking whether compiler suggests a MULTIARCH value... checking if requested user to run as... specified
    checking setting for User to run as... "root"
    checking if requested group membership of user to run as... specified
    checking setting for Group of user to run as... "nut"
    checking whether to pre-set warnings (from 'auto')... gcc-medium
    checking whether to make warnings fatal... no

Something appears to have changed in the configure script but I haven't been able to figure what is causing the problem.

jimklimov:
…and fix passing one default and one custom user/name value

Closes: #3513

Follow-up to #3567 in refining the CI recipes.

Fallout of #3140 (NUT v2.8.5): In that PR, the logic was changed to track if the caller gave us the user and group values - and then should have specified both. That check was flawed in case that (exactly) one of those specified values was same as the default that the `configure` script would apply, then it got treated as not-set and rejected the build configuration.

### DIFF
--- a/Jenkinsfile-dynamatrix
+++ b/Jenkinsfile-dynamatrix
@@ -1004,7 +1004,7 @@ set | sort -n """
                 runAllowedFailure: true,
                 mergeMode: [ 'excludeCombos': 'merge', 'dynamatrixAxesCommonEnv': 'replace' ], // NOTE: We might want to replace other fields, but excludeCombos must be merged to filter compiler versions vs. language standards as centrally defined!
                 excludeCombos: dynacfgPipeline.excludeCombos_DEFAULT_CPPUNIT
-                    + [dynacfgPipeline.axisCombos_STRICT_C] // TODO: revise to match the purpose of this filter
+                    //+ [dynacfgPipeline.axisCombos_STRICT_C] // TODO: revise to match the purpose of this filter
                     + [dynacfgPipeline.axisCombos_COMPILER_NOT_GCC]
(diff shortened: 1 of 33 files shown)


jimklimov:
Faults with GCC Strict-C89 builds (recently re-enabled) on newer platforms (OmniOS latest, Fedora 43) are due to bad NSS headers, fix proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=2064644 and mozilla/nss#34 (auto-closed by bot) and will be applied locally on those agents.

PR #3586, opened 2026-08-25
drivers/hidparser.c: bound the HID usage and path stacks [GHSA-f444-v6f2-2vcj]

CaffinatedSquirrel-OG:
Fixes the heap out-of-bounds write reported privately as GHSA-f444-v6f2-2vcj, plus the same-class problem in the path stack that I flagged in the same report. Opening this in the open at @jimklimov's direction on the advisory thread.

Both commits touch only `drivers/hidparser.c`. Nothing else in the tree changes.

WHAT THE FIRST COMMIT FIXES

`ITEM_USAGE` stored into `pParser->UsageTab[pParser->UsageSize]` with no check against `USAGE_TAB_SIZE`. The one `USAGE_TAB_SIZE` test in the file sits after the parsing loop and only calls `upslogx()`, so it reports the problem after the writes have already landed. `UsageTab` is `HIDNode_t[50]` and it's the last member of the heap-allocated `HIDParser_t`, so a report descriptor with more than 50 usage items writes 32-bit values from the descriptor past the end of the allocation. About 820 bytes of it, linearly, with the values under the descriptor's control.

The fix checks before the store and returns -1, which is the parser's existing "stop here" return. `Parse_ReportDesc()` already breaks out of its loop on `ret < 0`, so whatever parsed before the bad item is kept and the rest of the descriptor is dropped. That seemed better than clamping, which would silently mis-associate usages with the following main item.

Allowing `UsageSize` to reach `USAGE_TAB_SIZE` exposes a one-past-the-end read in the two unstack loops, which shifted `UsageTab[j+1]` down for every `j < UsageSize`. The last slot has no successor, so it now shifts in 0. `ResetLocalState()` zeroes unused slots anyway, so this matches what was already there.

WHAT THE SECOND COMMIT FIXES

The `Path` handling has the same shape - `upslogx()` after the loop instead of a test at the store - in the three places that append to `Data.Path.Node[]`. And `ITEM_END_COLLECTION` decremented `Path.Size` unconditionally, so an `End Collection` with no matching `Collection` wrapped the unsigned counter to 255 and the "remove index if any" test on the very next line read far past `Path.Node[PATH_SIZE]`.

This one stays inside the `HIDParser_t` allocation, so a sanitizer won't flag it. That's why it's folded in here rather than reported as its own vulnerability. It isn't only an overread though: on an unbalanced `End Collection` the current parser returns success with wrong paths - first node of every path missing, `pathsize` 2 where it should be 3 - after logging "HID path too long". A device quietly reporting wrong values is harder to catch than a crash, so it seemed worth fixing at the same time.

HOW IT WAS TESTED

There's a harness in my tree that builds the real `drivers/hidparser.c` twice under ASAN, pristine and patched, and runs five phases against both. I ran it against `0d949e373` this morning, all five pass:

1. Negative control. Unpatched build on the malicious descriptor: `AddressSanitizer: heap-buffer-overflow hidparser.c:181 in HIDParse`, `WRITE of size 4`, reached through the real `Parse_ReportDesc()` at `hidparser.c:734`. Exit 134. This phase runs first on purpose - a harness that has never been shown to fire proves nothing when it stays quiet.
2. Patched build, same descriptor: refused cleanly, no ASAN report.
3. Patched build, unbalanced `End Collection`: refused cleanly.
4. Positive control. A well-formed descriptor must parse to byte-identical output before and after. It does: same 3 items, same offsets, same `pathsize` 3 paths.
5. Positive control at the boundary. Exactly `USAGE_TAB_SIZE` usages must still be accepted, not rejected off by one. It is.

Phases 4 and 5 are the ones that matter for regression risk. A patch that "fixed" this by refusing every descriptor would pass 1 through 3 and fail these two.

I've left the harness out of this PR because wiring it into `tests/` properly is a bigger change than the fix and I didn't want to put that in front of a security fix, especially right after you spent weeks getting the build farm green. Happy to send it as a follow-up in whatever form suits `tests/Makefile.am` if you want it.

REACHABILITY

The descriptor comes off the wire in `nut_libusb_open()` in `drivers/libusb0.c` / `libusb1.c` and the SHUT equivalent in `libshut.c`, and goes straight into `Parse_ReportDesc()`. So the trigger is a malicious USB device, or a UPS running firmware someone else controls. Physical vector, not remote. I've scored and described it that way throughout and I'm not claiming code execution.

NOTES

- Both commits are signed off per `LICENSE-DCO`.
- No `NEWS.adoc` entry yet, since the entries in that file carry PR numbers. Say the word and I'll push one referencing this PR, or leave it to you if you'd rather write it in your own words.
- `drivers/hidparser.c` and `drivers/hidparser.h` are byte-identical between `26177060b`, where I originally found this on 4 August, and today's master, so nothing about the analysis has drifted.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -321,6 +321,10 @@ https://github.com/networkupstools/nut/milestone/13
       exiting right away or remaining in data stale mode indefinitely, also
       using `reconnect_trying()` for consistent reporting. They now track
       serial port file descriptor validity a bit more diligently. [PR #3541]
+    * The `hidparser.c` code (used primarily in USB drivers) risked a heap
+      out-of-bounds write if the device would report too many descriptors.
+      A similar problem could lead to "HID path too long" messages and loss
+      of readings from the device. [PR #3586]
 
  - NUT client libraries:
     * Complete support for actions documented in `docs/net-protocol.txt`
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -323,7 +323,7 @@ https://github.com/networkupstools/nut/milestone/13
       serial port file descriptor validity a bit more diligently. [PR #3541]
     * The `hidparser.c` code (used primarily in USB drivers) risked a heap
       out-of-bounds write if a device reported a HID report descriptor with
-      more usage items than the parser's fixed stack could hold. A similar
+      more usage items than the fixed stack in the parser could hold. A similar
       problem could lead to "HID path too long" messages and loss of
       readings from the device. [PR #3586]
 
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -322,9 +322,10 @@ https://github.com/networkupstools/nut/milestone/13
       using `reconnect_trying()` for consistent reporting. They now track
       serial port file descriptor validity a bit more diligently. [PR #3541]
     * The `hidparser.c` code (used primarily in USB drivers) risked a heap
-      out-of-bounds write if the device would report too many descriptors.
-      A similar problem could lead to "HID path too long" messages and loss
-      of readings from the device. [PR #3586]
+      out-of-bounds write if a device reported a HID report descriptor with
+      more usage items than the parser's fixed stack could hold. A similar
+      problem could lead to "HID path too long" messages and loss of
+      readings from the device. [PR #3586]
 
  - NUT client libraries:
     * Complete support for actions documented in `docs/net-protocol.txt`
--- a/drivers/hidparser.c
+++ b/drivers/hidparser.c
@@ -176,6 +176,19 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 			break;
 
 		case ITEM_USAGE:
+			/* The usage stack is a fixed-size array. Refuse to walk past
+			 * it instead of merely complaining about it after the fact
+			 * (see the USAGE_TAB_SIZE report at the end of this loop).
+			 * -1 is the established "no more items" return, and
+			 * Parse_ReportDesc() breaks out of its loop on ret < 0, so
+			 * everything parsed up to here is kept and the remainder of
+			 * the descriptor is dropped. */
+			if (pParser->UsageSize >= USAGE_TAB_SIZE) {
+				upslogx(LOG_ERR, "%s: HID Usage stack overflow, "
+					"aborting report descriptor parsing", __func__);
+				return -1;
+			}
+
 			/* Copy global or local UPage if any, in Usage stack */
 			if ((pParser->Item & SIZE_MASK) > 2) {
 				pParser->UsageTab[pParser->UsageSize] = pParser->Value;
@@ -197,7 +210,12 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 				int	j;
 
 				for (j = 0; j < pParser->UsageSize; j++) {
-					pParser->UsageTab[j] = pParser->UsageTab[j+1];
+					/* With UsageSize now allowed to reach
+					 * USAGE_TAB_SIZE, the last slot has no
+					 * successor to shift down. Unused slots
+					 * are zero (see ResetLocalState). */
+					pParser->UsageTab[j] = (j + 1 < USAGE_TAB_SIZE)
+						? pParser->UsageTab[j+1] : 0;
 				}
 
 				/* Remove Usage */
@@ -249,7 +267,12 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 				int j;
 
 				for (j = 0; j < pParser->UsageSize; j++) {
-					pParser->UsageTab[j] = pParser->UsageTab[j+1];
+					/* With UsageSize now allowed to reach
+					 * USAGE_TAB_SIZE, the last slot has no
+					 * successor to shift down. Unused slots
+					 * are zero (see ResetLocalState). */
+					pParser->UsageTab[j] = (j + 1 < USAGE_TAB_SIZE)
+						? pParser->UsageTab[j+1] : 0;
 				}
 
 				/* Remove Usage */
--- a/drivers/hidparser.c
+++ b/drivers/hidparser.c
@@ -202,6 +202,11 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 
 		case ITEM_COLLECTION:
 			/* Get UPage/Usage from UsageTab and store them in pParser->Data.Path */
+			if (pParser->Data.Path.Size >= PATH_SIZE) {
+				upslogx(LOG_ERR, "%s: HID path too long, "
+					"aborting report descriptor parsing", __func__);
+				return -1;
+			}
 			pParser->Data.Path.Node[pParser->Data.Path.Size] = pParser->UsageTab[0];
 			pParser->Data.Path.Size++;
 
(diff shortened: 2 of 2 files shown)


CaffinatedSquirrel-OG:
Thanks for writing the NEWS entry, and for pushing it straight to the branch - that saves a round trip.

One wording thing worth catching before it ships. The entry says the write happens "if the device would report too many descriptors", which reads as a count of descriptors. It's really a single report descriptor carrying too many usage items - more than `USAGE_TAB_SIZE`, so more than 50 - inside the one descriptor the device hands over. A device with several descriptors is fine as long as no individual one is over-long in that sense.

Suggested swap, if you want it:

```asciidoc
    * The `hidparser.c` code (used primarily in USB drivers) risked a heap
      out-of-bounds write if a device reported a HID report descriptor with
      more usage items than the parser's fixed stack could hold. A similar
      problem could lead to "HID path too long" messages and loss of
      readings from the device. [PR #3586]
```

Happy to push that myself if you'd rather not bother with it, or to leave it alone if you think the shorter phrasing reads better for who actually reads NEWS. Your file, your call.

jimklimov:
Your fix to NEWS makes sense, I guess I was distracted when I wrote my variant :\

CaffinatedSquirrel-OG:
Pushed as 3fc83f0b. It's the exact text from the suggestion above and nothing else, so the diff is just those four lines.

That restarted CI, so the Jenkins check needs another pass before this goes green again.

CaffinatedSquirrel-OG:
Sorry, my NEWS change broke the Tarballs job. Fixed in 759a3e7e.

The dist build runs aspell over `NEWS.adoc` against `docs/nut.dict`, and the dictionary has `parsers` but not the possessive `parser's`, which my wording introduced:

```
5851:& parser's 25 33: parsers, parer's, passer's, purser's, ...
make[5]: *** [docs/Makefile:1904: ../NEWS.adoc-spellchecked] Error 1
```

Your original text never used the word, so this only showed up once mine landed. It's the only word flagged - I reran aspell locally with the same arguments the Makefile uses, and the file is clean now.

So the line reads slightly differently from what you approved:

```asciidoc
      more usage items than the fixed stack in the parser could hold. A similar
```

I reworded rather than adding `parser's` to `docs/nut.dict`. The dictionary already carries a fair number of possessives so an entry would fit fine, and I'd have no objection if you prefer that - it just felt wrong to edit a shared file as a side effect of a security fix. Happy to swap it if you'd rather keep the possessive.

jimklimov:
That seems good, thanks.

jimklimov:
Took another pass at reviewing the code, and I wonder if any normal USB HID devices would overflow, or only specially crafted attackers? Namely, here you report the problem with `upslogx()` with would be pretty loud (syslog at least) every 5 to 30 secs or so (data update loop). If this is a situation we expect somewhat frequently, `upsdebugx(1,...)` may be preferable.

And also I wonder if it is possible to use some sliding window or `realloc()` to read the remaining data on a real device with more than 50 items reported -- if such exist. We can probably use the RD byte size from device discovery as an estimate of how much (tops) to expect. But this can be addressed later, if anyone in the wild actually has devices whose reports get truncated this way - now verbosely.

CaffinatedSquirrel-OG:
Short answers first, then the detail.

The logging can't repeat, and it doesn't run in the update loop. Each of the five `upslogx()` calls is followed immediately by `return -1`, and `Parse_ReportDesc()` breaks out of its item loop on `ret < 0` (`hidparser.c:782`). So a bad descriptor gets one line, then parsing stops.

The parse is device-open work rather than poll work. `upsdrv_initups()` passes `&callback` at `usbhid-ups.c:1843`, but `reconnect_ups()` passes NULL at `:2613`, and all three backends return early on a NULL callback before they even fetch the descriptor (`libusb1.c:707`, `libusb0.c:554`, `libshut.c:463`). A device flapping every five seconds won't re-parse.

One exception turned up while I was checking. `apcmicrolink` does re-pass its callback in `microlink_usb_reset_and_reopen()` (`apcmicrolink-usb.c:377`), so that driver can re-parse. It's gated on `MLINK_USB_RESET_AFTER_SEC` (60) at `apcmicrolink.c:3181` and only fires while the tunnel is already unresponsive and already logging `LOG_NOTICE` on each attempt, so the worst case is a line a minute in a state that's noisy anyway. `apc_modbus.c:1997` is open-time only.

It's your log, though. If you'd rather split them, I'd keep `LOG_ERR` on the two that mean malformed or hostile, usage stack overflow (`:187`) and unbalanced End Collection (`:249`), and drop the three "HID path too long" ones (`:206`, `:233`, `:282`) to `upsdebugx(1, ...)`. A deep path is the more plausible "real device we're just not deep enough for" case. Say which you want and I'll push it.

On real devices, this is mostly what I said on the advisory back in August: `UsageSize` is local state, cleared at every collection, end collection and exhausted main item (`:241`, `:261`, `:330`), so the 50 is what's pending for one main item - not a descriptor-wide budget. One thing I didn't mention then is that the parser has no case for Usage Minimum (0x18) or Usage Maximum (0x28), so both land in `default: break`. That means the descriptors which legitimately carry hundreds of usages never touch `UsageTab` at all. I still have no telemetry from real hardware and can't rule it out, but the shape of what it would take points at crafted rather than a UPS in the wild.

Your idea about using the RD byte size is better than an estimate. It's an exact bound, and I think it makes both the sliding window and the realloc loop unnecessary.

A `Usage` item costs at least one byte on the wire. `ItemSize[]` is `{0, 1, 2, 4}` (`hidparser.c:34`), so prefix `0x08` with size bits `00` is a one-byte item that still pushes onto the stack. That caps pending usages at `n`. `Parse_ReportDesc()` already has `n` at `:721`, right where it `calloc()`s the parser at `:768`, so sizing `UsageTab` in that same allocation is exact by construction and can't truncate. The cost stays small: the backends cap descriptors at `MAX_REPORT_SIZE` 0x1800 (`libusb1.c:61`), which puts the ceiling at 24 KB once per device open, and a realistic descriptor runs a few hundred bytes.

One thing to confirm before doing it that way. `apc_modbus.c:1997` passes `device->hid_report_descriptor_len` from libmodbus's USB layer rather than NUT's own `rdbuf`, and I haven't traced whether that one is capped. Clamping the sizing to `MAX_REPORT_SIZE` regardless would sidestep the question.

`PATH_SIZE` I'd leave as it is. Collection nesting depth is a structural limit, not a capacity guess.

Happy to do the dynamic `UsageTab` as a follow-up once this lands. Feels cleaner to keep them apart, given this one is the security fix and that one changes behaviour.

jimklimov:
Sounds good, sorry if my challenges and double-checks feel annoying - but second thoughts do often help get better results (or a better deliberated foundation for good choices already made) :)

> If you'd rather split them, I'd keep LOG_ERR on the two that mean malformed or hostile, usage stack overflow (:187) and unbalanced End Collection (:249), and drop the three "HID path too long" ones (:206, :233, :282) to upsdebugx(1, ...). A deep path is the more plausible "real device we're just not deep enough for" case. Say which you want and I'll push it.

Makes sense, keep two and "quiesce a bit" those three.

Dynamic change for behavior can indeed wait for another PR. Thanks!

CaffinatedSquirrel-OG:
Pushed as `39dcd1b9a`. `:187` (usage stack overflow) and `:249` (unbalanced End
Collection) keep `LOG_ERR`; the three `PATH_SIZE` ones at `:206`, `:233` and
`:282` are now `upsdebugx(1, ...)`. The refusals themselves are unchanged, only
where they get reported.

No challenge taken at the double-checks, by the way. The "does a real device do
this" question is the one I couldn't answer from the code alone, and it's the
right one to ask before we start rejecting descriptors.

I'll open the dynamic `UsageTab` sizing as a separate PR once this one lands.

jimklimov:
> I'll open the dynamic UsageTab sizing as a separate PR once this one lands.

Thanks! Good to go, on that one :)

PR #3588, opened 2026-08-27
drivers/hidparser.c: size the HID usage stack from the descriptor length

CaffinatedSquirrel-OG:
Follow-up to #3586, as offered there. This is the dynamic `UsageTab` sizing you agreed could wait for its own PR, plus one bound I tripped over while doing it.

Both commits touch `drivers/hidparser.c`; the second also changes a comment in `drivers/hidtypes.h`. Nothing else in the tree changes.

WHAT THE SECOND COMMIT DOES

This is the one you asked for, so I'll take it first.

The usage stack was `HIDNode_t UsageTab[50]` inside `HIDParser_t`. Your question on #3586 was whether a real device could legitimately report more than that and get truncated. It could. The 50-usage ceiling wasn't new - before #3586 the same descriptor walked off the end of the allocation instead of stopping - but #3586 turned it into a clean truncation, and a truncation is still a device that doesn't work.

`Parse_ReportDesc()` already knows the descriptor length, which is the estimate you suggested using. Every Usage item costs at least the one prefix byte that introduces it, so an n-byte descriptor can push at most n usages, and an n-slot stack always holds all of them. I checked the one path that looked like it might break that: when `ReportCount` is greater than 1 the parser re-enters `HIDParse()` without advancing `Pos`, but `Count` is only ever set inside the `ITEM_FEATURE`/`ITEM_INPUT`/`ITEM_OUTPUT` case, so the replay never reaches `ITEM_USAGE`. Every usage push follows a `Pos++`.

So the ceiling is gone for any descriptor that fits in the length we were handed. `USAGE_TAB_SIZE` stays as the floor, so short descriptors keep the capacity they've always had, and I kept the overflow check as a backstop - it's unreachable now for a descriptor that fits, but it costs one comparison and the invariant seemed worth stating in code rather than only in a commit message.

`UsageSize` goes from `uint8_t` to `size_t` in the same commit. At its old width it wrapped at 256 no matter how big the table was, so growing the table alone wouldn't have raised the real limit.

Cost is one `calloc()` of 4 bytes per descriptor byte, freed before `Parse_ReportDesc()` returns. That's at most 24kB against the 6144-byte `MAX_REPORT_SIZE` the libusb and libshut backends cap at, and at most 8kB against the 2048-byte buffer `apc_modbus` gets from libmodbus. It happens once per device open, not in the data update loop - same reasoning as the logging discussion on #3586.

One second-order cost you'll spot and I'd rather raise myself: `ResetLocalState()` zeroes the whole table, which used to be 200 bytes and can now be 24kB. It fires on every Collection, End Collection and Main item, so a maximum-size descriptor could run it a few thousand times - order of milliseconds once per device open, on my reading, but it is a real change.

I left it zeroing the whole table on purpose, because that's exactly the old semantics. The tighter version is to zero only the first `UsageSize` slots: the two unstack loops already shift a 0 into the last slot, so slots at or above `UsageSize` are always zero and re-zeroing them is redundant. I didn't want to make the reset depend on that invariant without you weighing in, since if it ever broke, the failure would be a stale usage silently reappearing in a later path rather than anything loud. Say which you prefer and I'll push it.

WHAT THE FIRST COMMIT DOES

While sizing the stack from the length I checked what else trusts it, and the item loop doesn't. It tests `Pos < ReportDescSize` before reading an item prefix, but nothing tests that the 1, 2 or 4 payload bytes the prefix promises are actually there. A descriptor whose last byte is a size-bearing prefix reads up to 4 bytes past the buffer the caller gave us.

I traced where that lands, because I didn't want to describe it as worse than it is. In every path today it's a short read just off the end of a fixed buffer:

- `libusb1.c` / `libusb0.c` / `libshut.c` - `rdbuf` is a 6144-byte stack array and `libusb1.c:823` rejects `rdlen > sizeof(rdbuf)`, so the read leaves the array only for `rdlen` in [6141, 6144].
- `apc_modbus.c:1997` - the buffer is libmodbus's 2048-byte stack array in `_modbus_rtu_usb_connect()`, and the length is the actual transfer count from `libusb_control_transfer()`, so the equivalent window is [2045, 2048].

Both windows are device-supplied, so both are reachable, but neither is dramatic and I'm not calling this a vulnerability. It's the parser reading past the length it was given, which it shouldn't do regardless of who owns the buffer.

Reported at debug level rather than `LOG_ERR`. A descriptor ending mid-item is malformed, but a truncated USB transfer is a plausible way for a real device to produce one, so it belongs with the `PATH_SIZE` refusals rather than with the two you kept loud.

HOW IT WAS TESTED

Same harness as #3586, re-baselined against master after ab82520df, extended to 11 phases. All 11 pass. Full output is in the first comment below.

The re-baselining matters and phase 1 exists to enforce it: "unpatched" now already contains your merged fix, so phase 1 asserts the baseline refuses the 200-usage descriptor. If someone runs this against a pre-#3586 clone it fails there instead of quietly comparing against the wrong tree.

Every descriptor is now copied into an exact-sized heap buffer before it reaches the parser. Passing a static array would have hidden the over-read inside the array's own slack, which is exactly the kind of silence that makes a harness worthless.

- 2, negative control. Unpatched, descriptor ending on a `Usage` prefix: `AddressSanitizer: heap-buffer-overflow hidparser.c:165 in HIDParse`, `READ of size 1`, `0 bytes after 7-byte region`, through the real `Parse_ReportDesc()` at `hidparser.c:781`.
- 3. Patched, same input: refused at debug level, no ASAN report.
- 4, negative control. Unpatched, 200 usages plus a Main item: hits the fixed ceiling, `parse=NULL`. The truncation you asked about, reproducing.
- 5. Patched, same input: `parse=OK items=1`. This is the phase the PR is for.
- 6, regression. Patched, the original #3586 overflow input: no ASAN report. Sizing the stack dynamically doesn't reintroduce what you merged the fix for.
- 7 and 8, positive controls. Well-formed descriptor and a full-but-legal 50-usage stack must produce byte-identical output before and after. Both do.
- 9 and 10, regressions. Unbalanced `End Collection` still refused; over-deep path still refused and still at debug level.
- 11. The new allocation must actually be freed. `leaks(1)` on a non-ASAN build: `0 leaks for 0 total leaked bytes`. LeakSanitizer isn't available on macOS, so this is the local substitute rather than the same check your Linux CI would run.

Phases 7 and 8 are the ones that matter for regression risk, same as last time. A patch that "fixed" this by refusing everything would sail through the rest and fail those two.

NOTES

- Both commits are signed off per `LICENSE-DCO`.
- No `NEWS.adoc` entry yet, since those carry PR numbers. Say the word once this has one and I'll push it, or write it yourself if you'd rather - either way I'll run `aspell` against `docs/nut.dict` locally first this time.
- The test harness still isn't in this PR. Same reason as before: wiring it into `tests/Makefile.am` is a bigger change than the fix. The offer stands if you want it as its own PR.
- One thing I noticed and did not touch: `HIDParser_t.Pos` is `uint16_t` while `ReportDescSize` is `size_t`. Nothing today can hand the parser more than 6144 bytes so it can't bite, and if it ever did the outer `MAX_REPORT` bound stops it rather than looping - it would just re-parse from the start and produce nonsense. Left alone to keep this PR to what you asked for. Happy to widen it in a third commit if you want it closed off.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -326,6 +326,11 @@ https://github.com/networkupstools/nut/milestone/13
       more usage items than the fixed stack in the parser could hold. A similar
       problem could lead to "HID path too long" messages and loss of
       readings from the device. [PR #3586]
+    * The `hidparser.c` code was further updated to use not a fixed 50-entry
+      "usage table" buffer for resource descriptor parsing, but one aligned
+      with the actual size detected during device discovery. This should
+      avoid potential truncation introduced by the fix above (and avoid
+      the overflows it addressed) in normal device interactions. [#3588]
 
  - NUT client libraries:
     * Complete support for actions documented in `docs/net-protocol.txt`
--- a/drivers/hidparser.c
+++ b/drivers/hidparser.c
@@ -159,13 +159,30 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 	while ((Found < 0) && (pParser->Pos < pParser->ReportDescSize)) {
 		/* Get new pParser->Item if current pParser->Count is empty */
 		if (pParser->Count == 0) {
+			uint8_t	ItemDataSize;
+
 			pParser->Item = pParser->ReportDesc[pParser->Pos++];
 			pParser->Value = 0;
-			for (i = 0; i < ItemSize[pParser->Item & SIZE_MASK]; i++) {
+			ItemDataSize = ItemSize[pParser->Item & SIZE_MASK];
+
+			/* The item prefix promises ItemDataSize more bytes, and
+			 * nothing so far has checked that the descriptor still
+			 * holds them. A descriptor ending part-way through an
+			 * item would read up to 4 bytes past the end of the
+			 * caller's buffer. Pos is <= ReportDescSize here - the
+			 * loop condition tested it before the increment above -
+			 * so the subtraction cannot wrap. */
+			if ((size_t)ItemDataSize > pParser->ReportDescSize - pParser->Pos) {
+				upsdebugx(1, "%s: truncated HID item, "
+					"aborting report descriptor parsing", __func__);
+				return -1;
+			}
+
+			for (i = 0; i < ItemDataSize; i++) {
 				pParser->Value += (uint32_t)(pParser->ReportDesc[(pParser->Pos)+i]) << (8*i);
 			}
 			/* Pos on next item */
-			pParser->Pos += ItemSize[pParser->Item & SIZE_MASK];
+			pParser->Pos += ItemDataSize;
 		}
 
 		switch (pParser->Item & ITEM_MASK)
--- a/drivers/hidparser.c
+++ b/drivers/hidparser.c
@@ -54,8 +54,9 @@ typedef struct {
 	uint8_t		Count;				/* Store local report count	*/
 
 	uint16_t	UPage;				/* Global UPage			*/
-	HIDNode_t	UsageTab[USAGE_TAB_SIZE];	/* Usage stack			*/
-	uint8_t		UsageSize;			/* Design number of usage used	*/
+	HIDNode_t	*UsageTab;			/* Usage stack			*/
+	size_t		UsageTabSize;			/* Slots allocated in UsageTab	*/
+	size_t		UsageSize;			/* Design number of usage used	*/
 } HIDParser_t;
 
 /* return 1 + the position of the leftmost "1" bit of an int, or 0 if
@@ -89,7 +90,8 @@ static inline unsigned int hibit(unsigned long x)
 static void ResetLocalState(HIDParser_t* pParser)
 {
 	pParser->UsageSize = 0;
-	memset(pParser->UsageTab, 0, sizeof(pParser->UsageTab));
+	memset(pParser->UsageTab, 0,
+		pParser->UsageTabSize * sizeof(*pParser->UsageTab));
 }
 
 /*
@@ -193,14 +195,15 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 			break;
 
 		case ITEM_USAGE:
-			/* The usage stack is a fixed-size array. Refuse to walk past
-			 * it instead of merely complaining about it after the fact
-			 * (see the USAGE_TAB_SIZE report at the end of this loop).
-			 * -1 is the established "no more items" return, and
-			 * Parse_ReportDesc() breaks out of its loop on ret < 0, so
-			 * everything parsed up to here is kept and the remainder of
-			 * the descriptor is dropped. */
-			if (pParser->UsageSize >= USAGE_TAB_SIZE) {
+			/* The usage stack is sized from the descriptor length and
+			 * every Usage item costs at least its one prefix byte, so
+			 * this cannot fire for a descriptor that fits in the length
+			 * we were handed. It stays as a backstop. -1 is the
+			 * established "no more items" return, and Parse_ReportDesc()
+			 * breaks out of its loop on ret < 0, so everything parsed up
+			 * to here is kept and the remainder of the descriptor is
+			 * dropped. */
+			if (pParser->UsageSize >= pParser->UsageTabSize) {
 				upslogx(LOG_ERR, "%s: HID Usage stack overflow, "
 					"aborting report descriptor parsing", __func__);
 				return -1;
@@ -229,14 +232,14 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 
 			/* Unstack UPage/Usage from UsageTab (never remove the last) */
 			if (pParser->UsageSize > 0) {
-				int	j;
+				size_t	j;
 
 				for (j = 0; j < pParser->UsageSize; j++) {
-					/* With UsageSize now allowed to reach
-					 * USAGE_TAB_SIZE, the last slot has no
+					/* With UsageSize allowed to reach
+					 * UsageTabSize, the last slot has no
 					 * successor to shift down. Unused slots
 					 * are zero (see ResetLocalState). */
-					pParser->UsageTab[j] = (j + 1 < USAGE_TAB_SIZE)
+					pParser->UsageTab[j] = (j + 1 < pParser->UsageTabSize)
 						? pParser->UsageTab[j+1] : 0;
 				}
 
@@ -305,14 +308,14 @@ static int HIDParse(HIDParser_t *pParser, HIDData_t *pData)
 
 			/* Unstack UPage/Usage from UsageTab (never remove the last) */
 			if(pParser->UsageSize > 0) {
-				int j;
+				size_t j;
 
 				for (j = 0; j < pParser->UsageSize; j++) {
-					/* With UsageSize now allowed to reach
-					 * USAGE_TAB_SIZE, the last slot has no
+					/* With UsageSize allowed to reach
+					 * UsageTabSize, the last slot has no
 					 * successor to shift down. Unused slots
 					 * are zero (see ResetLocalState). */
-					pParser->UsageTab[j] = (j + 1 < USAGE_TAB_SIZE)
+					pParser->UsageTab[j] = (j + 1 < pParser->UsageTabSize)
 						? pParser->UsageTab[j+1] : 0;
 				}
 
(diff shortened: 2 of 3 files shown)


CaffinatedSquirrel-OG:
Full harness output, as promised in the description. Run today against `a32520584` with the two commits on this PR applied to a copy, so the clone itself is never modified.

Local paths are trimmed to `.` and `<nut-clone>`; nothing else is edited.

<details>
<summary>verify_fix.sh - 11 phases, 11 pass</summary>

```

=== build ===
applied 0001-drivers-hidparser.c-refuse-truncated-HID-items-inste.patch
applied 0002-drivers-hidparser.c-size-the-HID-usage-stack-from-th.patch
built ./.verify/test_pristine
built ./.verify/test_patched

=== phase 1: BASELINE CHECK -- unpatched tree must already carry the #3586 fix ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=malicious desc_bytes=1000
[upslogx] HIDParse: HID Usage stack overflow, aborting report descriptor parsing
parse=NULL
PASS  baseline is post-#3586: refuses the 200-usage descriptor, no overflow

=== phase 2: NEGATIVE CONTROL -- unpatched, truncated item (expect ASAN over-READ) ===
./verify_fix.sh: line 108: 73082 Abort trap: 6           "$BUILD/test_$1" "$2" > "$BUILD/$1_$2.txt" 2>&1
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==73082==ABORTING
PASS  unpatched build reports an out-of-bounds READ past the descriptor (exit 134)

=== phase 3: patched, truncated item (expect clean refusal at debug level) ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=truncated desc_bytes=7
[upsdebugx:1] HIDParse: truncated HID item, aborting report descriptor parsing
parse=NULL
PASS  truncated item is refused at debug level and nothing is read past the end

=== phase 4: NEGATIVE CONTROL -- unpatched, 200 usages + Main item (expect truncation) ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=usages200 desc_bytes=1008
[upslogx] HIDParse: HID Usage stack overflow, aborting report descriptor parsing
parse=NULL
PASS  unpatched build truncates a 200-usage descriptor to nothing

=== phase 5: patched, 200 usages + Main item (expect a real parse) -- the point of the PR ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=usages200 desc_bytes=1008
parse=OK items=1
item[0] rid=1 type=0x80 attr=0x82 off=0 sizebits=8 logmin=0 logmax=0 pathsize=1 path=41414100
replen[1]=1
PASS  a 200-usage descriptor now parses instead of being truncated

=== phase 6: REGRESSION -- patched, the original #3586 overflow input ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=malicious desc_bytes=1000
parse=NULL
PASS  the #3586 input no longer overflows and still does not with a dynamic stack

=== phase 7: POSITIVE CONTROL -- well-formed descriptor must still parse ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=wellformed desc_bytes=52
parse=OK items=3
item[0] rid=1 type=0x80 attr=0x82 off=0 sizebits=8 logmin=0 logmax=100 pathsize=3 path=00840004,00840024,00850066
item[1] rid=1 type=0x80 attr=0x82 off=8 sizebits=16 logmin=0 logmax=65535 pathsize=3 path=00840004,00840024,00850068
item[2] rid=1 type=0x80 attr=0x82 off=24 sizebits=1 logmin=0 logmax=1 pathsize=3 path=00840004,00840024,00850044
replen[1]=4
PASS  well-formed descriptor parses identically before and after the patch

=== phase 8: POSITIVE CONTROL -- exactly USAGE_TAB_SIZE usages must be accepted ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=boundary50 desc_bytes=258
parse=OK items=1
item[0] rid=1 type=0x80 attr=0x82 off=0 sizebits=8 logmin=0 logmax=0 pathsize=1 path=41414100
replen[1]=1
PASS  a full usage stack (USAGE_TAB_SIZE entries) parses unchanged

=== phase 9: REGRESSION -- unbalanced End Collection still refused ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=unbalanced desc_bytes=53
[upslogx] HIDParse: unbalanced HID End Collection, aborting report descriptor parsing
parse=NULL
PASS  unbalanced End Collection is still refused

=== phase 10: REGRESSION -- over-deep path still refused, still at debug level ===
USAGE_TAB_SIZE=50 PATH_SIZE=10 HIDNode_t=4
case=deeppath desc_bytes=56
[upsdebugx:1] HIDParse: HID path too long, aborting report descriptor parsing
parse=NULL
PASS  over-deep path is still refused and still reported at debug level

=== phase 11: the new usage-stack allocation must be freed ===
./verify_fix.sh: line 285: 73127 Abort trap: 6           ASAN_OPTIONS=detect_leaks=1 "$BUILD/test_patched" usages200 > "$BUILD/patched_leakcheck.txt" 2>&1
Process 73136: 0 leaks for 0 total leaked bytes.

PASS  no leak reported by leaks(1)

=== result ===
ALL CHECKS PASSED
artifacts in ./.verify
```

</details>

CaffinatedSquirrel-OG:
Heads up that the `GHA-01: Tarballs` red on this PR isn't from this PR. `distcheck` fails on `ecoflow_cdc_protocol_utest`, which shares no code with `hidparser.c` - its only sources are `ecoflow_cdc_protocol_utest.c`, `ecoflow-cdc-protocol.c` and `-lm`.

It's already on master. Building that test at four points with nothing else changed:

```
ab82520df  PASS            Merge pull request #3586
09e41aa90  FAIL  5 checks  tests/ecoflow_cdc_protocol_utest.c: revise use of fabs() => f_equal() [#3569]
ba9bccfa4  FAIL  5 checks
a32520584  FAIL  5 checks  current master, and this PR's base
```

The five that fail are exactly the five `09e41aa90` converted from a `0.01` tolerance to `f_equal`. The three that already used `f_equal` - temperatures 26 and 28, frequency 60 - still pass, because those land on exactly representable values.

`f_equal` is an absolute comparison against `FLT_EPSILON`:

```c
#define f_equal(x, y) ( fabsf((float)(x) - (float)(y)) <= FLT_EPSILON )
```

`FLT_EPSILON` is 1.19e-07, the gap between 1.0 and the next float. At the magnitudes these particular checks use, one ULP of `float` is already far bigger than that - 3.05e-05 at 258 - so `f_equal` there is effectively "bit-identical or fail".

The decoded values aren't bit-identical, and the old tolerance was absorbing that:

```
output_power  decodes to 258.7681884766  vs literal 258.768  delta 1.83e-04
input_power   decodes to  29.2039623260  vs literal  29.204  delta 3.81e-05
```

So those deltas are roughly 1500x and 320x what `f_equal` allows, against the `0.01` that used to be there.

The `NUT CI farm - Jenkins checks` check is red here too. I can't read `ci.networkupstools.org` without an account, so I don't know whether that's the same test or something else, and I'd rather say so than guess. Jenkins hasn't reported on master's `a32520584` yet, so there's nothing to compare against there either.

Happy to send a separate PR putting a tolerance back on those five if you want it from me. Whether that's a plain tolerance again or a relative-epsilon helper in `nut_float.h` is your call, since the second one reaches further than this test. Leaving it alone from here either way - it has nothing to do with these commits.

jimklimov:
@CaffinatedSquirrel-OG : I have a feeling I am looking at AI talking to itself here ;) Is yours watching the PR and analyzing how it goes?

The master branch was fixed after you initiated the PR, I'll update yours once its quick checks pass. Sorry about that mishap.

As for a question in the top post - erring on the safe side and zeroing out the whole table is better.

jimklimov:
I'll investigate tomorrow why "Stash source" stage fails, in different PR builds. For some reason it looks for CI library commit hash in NUT sources and fails :\

jimklimov:
Jenkins build relaunched.

jimklimov:
Ubuntu-trusty also fixed on master (later). Distcheck on FreeBSD 15 is occasionally annoying, part of fallout of the recent build agents' upgrade (maybe its `bmake` is flawed, not our recipes -- at least there were issues with its variant in NetBSD 11 pre-release) :\

Issue #2502, opened 2024-07-01
CircleCI deprecated the XCode version NUT tests used, newer is needed


jimklimov:
````
Job was rejected because resource class macos.m1.medium.gen1, image xcode:12.5.1 is not a valid resource class
````

It seems our builds were landing by default on x86 MacOS builders, which are obsoleted since last weekend. And the Apple M1 builders do not have an XCode 12.5.1 image.

* https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718/4
* https://circleci.com/docs/using-macos/#supported-xcode-versions-silicon
* https://github.com/EmbeddedRPC/erpc/pull/376/files

In a way, this matter follows up from #1419 and #823 efforts.

jimklimov:
Kudos, really, to CircleCI for a "Re-run job with SSH" feature that I've missed originally. It was very helpful for fixing and improving the build scenarios.

PR #3590, opened 2026-08-29
Consolidate repeated CppUnit warning suppression blocks

user01010111:
## Summary

- move the repeated CppUnit compiler-diagnostic setup and teardown into paired
  private test headers;
- retain the `-Wzero-as-null-pointer-constant` suppression only for the common
  test runner which needs it;
- correct the Clang diagnostic setup to use separate `push` and `ignored`
  pragmas and require matching push/pop capability support;
- distribute both new headers through `tests/Makefile.am`.

The eight CppUnit translation units previously carried near-identical warning
blocks. Keeping the policy in one test-local location removes roughly 340
duplicated lines while preserving the existing translation-unit-wide scope and
GCC capability guards. The Clang guard is normalised so paired diagnostic
push/pop directives are only used when that capability is available. No public
header, dependency, configure probe or runtime behaviour is changed.

This follows the warning-suppression work associated with #2502/#2506 and the
later Clang updates in #2756.

## Validation

- `git diff --check`: passed.
- `check-source-nonascii`, residual-block and balanced include-pair checks:
  passed.
- Independent review against the relevant `docs/developers.txt` source-style
  and portability requirements found no remaining issues.
- Debian 12/glibc with CppUnit 1.15.1:
  - `autogen.sh` and all four GCC/Clang, nutconf enabled/disabled configure
    combinations: passed;
  - GCC 12.2 hard warnings with `-Werror`: both CppUnit binaries built and
    `cppunittest` passed 34 tests with nutconf and 11 without;
  - Clang 14 hard warnings compiled all nine affected object targets with
    nutconf enabled and all five with nutconf disabled, then both full builds
    of the patched tree stopped on conversion diagnostics in unchanged
    `common/common.c` under `-Weverything -Werror`; runtime was consequently
    not run in these two lanes;
  - Clang 14 medium warnings with `-Werror`: both binaries built and
    `cppunittest` passed 34 tests with nutconf and 11 without;
  - `stylecheck`: passed;
  - `distcheck-light`: passed its nested 6/6 suite, including `cppunittest`;
    direct compiler lanes used `--with-doc=skip`, while `distcheck-light`
    reconfigured with `--with-doc=auto` after preparing 270 faked man-page
    files;
  - the generated distribution archive contains both new headers.
- macOS 27.0 on arm64 with Apple Clang 21.0.0:
  - a pristine checkout at the target revision also fails a full Clang-hard
    `-Werror` build in unchanged common sources, establishing a baseline
    limitation rather than a regression from this patch;
  - every changed CppUnit target compiled separately with Clang-hard warnings
    and `-Werror`;
  - with Clang-hard warnings and `-Werror` disabled, the full build,
    `stylecheck`, explicit NIT and `make check` passed; `cppunittest` passed all
    34 tests, NIT passed 26/26, and each of its four `cppnit` scenarios passed
    5 tests;
  - `distcheck-light` was not run because the strict full build is blocked by
    the unchanged baseline and its docs-auto configuration differs from the
    direct build.
- Ubuntu 24.04.4 on a Raspberry Pi 4 Model B Rev 1.4, aarch64, with GCC/G++
  13.3.0:
  - Autotools regeneration and configuration with GCC-hard warnings,
    `-Werror`, GNU C++11, CppUnit, nutconf and NIT: passed;
  - the full build, both CppUnit targets, `cppunittest` (34 tests),
    `make check` with NIT 26/26, `stylecheck` and `distcheck-light`: passed;
  - the generated distribution archive contains both new headers.
- A native Windows 11/MSYS2 MinGW64 run with GCC 16.2 reported both CppUnit
  targets built, `cppunittest` `OK (34)`, NIT 25 passed and 1 skipped because
  `nut-scanner` was unavailable, four `cppnit` scenarios at `OK (5)` each,
  and `stylecheck` passing. The full build/check and strict-warning matrix were
  incomplete due diagnostics outside this patch; no MSVC, installer or Windows
  service-runtime result is claimed.

Native Intel and older macOS, BSD, OpenIndiana, other ARM systems and
upstream CI remain untested. No install or packaging result is claimed.

## Scope and compatibility

- The change is confined to eight CppUnit sources, two private test headers and
  `tests/Makefile.am`.
- The code remains C++11-compatible and follows the compiler-pragma capability
  probes already provided by NUT.
- No generated files or unrelated formatting changes are included.

## AI assistance disclosure

I used OpenAI Codex with `gpt-5.6-sol` for planning, repository analysis,
implementation, review, drafting and macOS validation, and `gpt-5.6-terra` for
Raspberry Pi validation. Both models used high reasoning where applicable. I
used an additional Windows validation session with OpenAI Codex `gpt-5.6-sol`,
high reasoning. Daybreak and Codex Security were not used for that validation. I
reviewed the resulting changes and validation evidence and remain responsible
for the contribution.


jimklimov:
Thanks, would be interesting to see how it goes. My earlier experiments and docs-reading implied that pragmas can not be carried via macros, or not in all compiler implementations (e.g. only starting from some year). If this would get proven wrong - great, a lot of other codebase can get simplified then.

user01010111:
Thanks. This change does not expand pragmas from macros. The new headers contain ordinary `#pragma GCC diagnostic` and `#pragma clang diagnostic` directives; the existing `HAVE_PRAGMA_*` macros only guard which direct pragmas are emitted. The includes place those directives around the same CppUnit-dependent portions of each translation unit as the repeated blocks did previously.

The current GCC and Clang matrices are green, with Jenkins still running. If a supported compiler treats the header include boundary differently, I’ll adjust the approach.

jimklimov:
Right, I misread the change on the first pass - thought it was similar to my experiments with macros that would resolve to setting those pragmas. Using header files is an interesting trick in this regard, which can still be productized for other cases, and now proven to work for various platforms that we can check in NUT CI farm (with a caveat that some builders or setting combos are incapable of C++11 and/or lack libcppunit) :)

Thanks!

PR #3579, opened 2026-08-20
feat(apcmicrolink): add USB HID transport with HID-PDC fallback

nmbro:
## General points

- [x] Described the changes in the PR submission or a separate issue
- [x] There may be multiple commits in the PR (3 commits: initial USB transport, a libusb-0.1 build fix, and a `port=auto` convention rework in response to review)
- [x] Use of coding helper tools and AI disclosed: **Claude Code was used** for development assistance
- [x] Please star NUT on GitHub, this helps with sponsorships! ;)

## Frequent "underwater rocks" for driver addition/update PRs

- [x] Revised existing driver families and added a brand new driver
- [x] Did not extend obsoleted drivers with new hardware support features
- [x] For updated existing device drivers, bumped the `DRIVER_VERSION` macro (0.02)
- [x] For new USB devices, built and committed changes for `scripts/upower/95-upower-hid.hwdb` (N/A - custom HID protocol, not standard PDC)
- [x] Proposed NUT data mapping aligned with `docs/nut-names.txt` (uses experimental.hid_fallback.active)
- [x] Updated `data/driver.list.in`

## General documentation updates

- [x] Added bullet point into `NEWS.adoc`
- [x] Updated `docs/nut.dict` for spellcheck (added: SCL, descriptor's, replug, unswitched)
- [x] Added/updated manual page in `docs/man/apcmicrolink.txt`
- [x] Passed `make distcheck` locally

## Additional work

- [ ] Propose PR for NUT DDL with device data dumps (will follow after merge)
- [ ] Address NUT CI farm build failures (awaiting CI run)

## Summary

Adds USB HID transport for APC Smart-UPS models with Microlink protocol but no serial port (e.g., SCL500RMI1UC). Key features:

- **HID-PDC fallback**: If Microlink handshake fails (common on startup), driver falls back to standard HID Power Device data and automatically upgrades to full Microlink when available
- **Per-outlet-group commands**: load.cycle, shutdown.* commands for each outlet group
- **USB transport tuning**: 1s timeout vs 100ms serial, 10 retries vs 3 (device interleaves HID reports)
- **Udev requirement documented**: Kernel usbhid must be unbound for this device

Tested on APC Smart-UPS 500 (SCL500RMI1UC).

### Update following @jimklimov's review

[...]

PR #3591, opened 2026-08-29
apcmicrolink: async USB reads, session stability, outlet-group and beeper fixes

nmbro:
Three commits against the existing experimental `apcmicrolink` driver: a rework
of how USB reads are done, a set of long-run session-stability fixes that came
out of it, and one framing bug found via issue #3587.

## Why

The USB HID tunnel transport added in #3579 worked, but a session would drop
after an inconsistent interval and then fail to recover. Running it against an
APC Smart-UPS 500 (SCL500RMI1UC) over multi-day sessions turned up several
distinct causes rather than one.

## What changed

**Reads are no longer issued per call.** The driver used to start a fresh
interrupt-IN transfer each time it wanted a byte, so it was only listening while
it happened to be blocked waiting for a reply. This device can go several
seconds between replies and pushes unrelated HID reports on the same pipe in the
meantime, so replies arriving outside that window were simply lost. A single
transfer is now kept permanently outstanding, serviced by a dedicated pump
thread; only that thread touches libusb's event loop, and everything else takes
a lock to drain the queue it fills. Builds without libusb-1.0 or pthreads, and
any failure to start the listener, fall back to the previous synchronous read.

This needs the libusb-1.0 context, which is exposed as
`nut_libusb_get_context()` in `drivers/libusb1.c` (guarded by `WITH_LIBUSB_1_0`;
libusb-0.1 has no equivalent object to hand out).

**A USB bus reset is now triggered only on a genuine disconnect.** It used to
fire off a retry count, which hit just as often for a live-but-stalled device
that a reset never helped. It now fires only when a read, write or transfer
completion has actually reported `NO_DEVICE`.

**Three smaller session fixes:** the cached page 0 contents are no longer wiped
on every session (re)start; the read queue is no longer flushed on every session
retry (that was discarding replies that had already arrived -- an isolated test
put it at roughly half of otherwise-valid replies); and the authentication
challenge is randomized instead of being a fixed `0x00 0x00`.

**A page 0 that contradicts itself is now rejected.** Page 0 announces the frame
width every later frame is parsed with, and any checksum-valid copy of it was
accepted. A corrupt copy is therefore unrecoverable: the bogus width becomes the
length the frame parser demands, nothing checksum-validates again, and the
session dies with the device having done nothing wrong. A genuine page 0 always
arrives in a frame of exactly the width it announces, so the two are now
required to agree. This is in the shared framing layer, so it applies to the
serial transport too.

**Diagnostics:** warn once when built against libusb-0.1 (untested for this
driver, and on one system its interrupt-IN read did not honor its timeout and
hung the driver); back off and log, rate-limited, when the kernel's `usbhid`
driver has reclaimed the interface and every submission fails instantly with
`EBUSY` (one observed incident spun at roughly 14,000 failed submissions/sec for
over half an hour); and warn once if a device ever sets the page 0 implicit
byte-stuffing bit, which this parser does not implement and previously ignored
silently.

**Two further descriptor usages mapped:** `experimental.battery.serial` (the
battery pack's own serial, distinct from `ups.serial`; confirmed against the
compartment label on a real unit) and `microlink.diag.slave_password_echo` (the
register the auth challenge is written to and echoed back -- troubleshooting
visibility only, and explicitly not proof that authentication was accepted,
since a device that blindly echoes any register write would look identical).

## Hardware

Developed and tested against an APC Smart-UPS 500 (SCL500RMI1UC) over USB, by a
single tester, including deliberate fault injection: USB replug, power cycles,
driver restarts mid-handshake, and forced fallback. No other USB Microlink model
has been tested, and USB mode remains experimental relative to the rest of this
already-experimental driver.

## Limitations, and what this does not fix

The page 0 fix came out of issue #3587 (`apcmicrolink` failing to start on an
SMX1500RM2U). **It does not make that device work.** It removes a
self-inflicted unrecoverable state -- the driver no longer wedges its own parser
-- but that device stops serving pages at 0x3C while its descriptor does not
begin until page 0x48, so startup still will not complete there. I do not have
that hardware; the fix was derived from the debug log in that issue and verified
by checking the guard against the two captured page 0 frames (the good one is
accepted, the corrupt one rejected). Discussion continues in #3587.

`experimental.battery.serial` is in the `experimental.*` namespace as
`docs/nut-names.txt` prescribes. Happy to raise it on nut-upsdev to standardize
if that seems worthwhile.

Use of coding helper tools and AI disclosed: **Claude Code was used** for
development assistance. The code and documentation were reviewed against the
NUT style guide, specifically for non-ASCII punctuation and for compiler pragmas
(none are introduced here).

## Checklist

General points:

- [x] Changes described above; hardware, limitations and untested paths called out
- [x] Each commit a separate functional change; no style-only changes mixed in
- [x] AI use disclosed, here and in each commit message
- [x] Revised for NUT code style; ASCII only (`tools/check-source-nonascii.pl` clean)

Driver PRs:

- [x] Updated the existing driver rather than adding a new one or a sub-driver
- [x] Bumped `DRIVER_VERSION` (0.02 -> 0.03)
- [x] No new VID/PID: 051d:0003 was already added by #3579, including
      `scripts/upower/95-upower-hid.hwdb` and `data/driver.list.in`
- [x] Data mapping aligned with `docs/nut-names.txt`; non-standard points use
      `experimental.*` or the driver's existing `microlink.*` namespace

General C code:

- [x] No assumptions about integer sizes, alignment or endianness; `size_t`
      used where the language or libraries expect it
- [x] `upsdebugx()`/`upslogx()`/`fatalx()` and NUT allocation helpers throughout
- [x] Coding style follows precedent in these files
- [x] No new files, so no `Makefile.am` recipe changes were needed

Documentation:

- [x] `NEWS.adoc` bullets added
- [x] `docs/man/apcmicrolink.txt` updated: the USB MODE section described a
      periodic reset on an unresponsive tunnel, which is no longer what happens,
      and it gains a "libusb backend" subsection
- [x] `make spellcheck` passes, no `docs/nut.dict` changes needed
- [ ] `UPGRADING.adoc` not touched -- nothing here is breaking
- [ ] `docs/acknowledgements.txt` not touched -- not vendor-backed work

Verification on this branch: `make distcheck` passes (with real asciidoc man
page generation, test suite 10/10), `make spellcheck` passes, and the driver
builds without warnings.

## Update: stability and mapping fixes from review

Several more commits landed during review, all under this same PR:

* USB mode: the standard-HID-PDC fallback now engages whenever the Microlink
  tunnel stops producing usable data, not only when it fails to answer during
  startup. A device that answers every poll while reporting an all-zero state
  previously left `ups.status` empty indefinitely. The driver hands back to
  Microlink data once that source looks plausible again, with hysteresis so
  the two do not swap on alternate polls.
* The Microlink tunnel is now polled for a full page pass per update cycle,
  instead of one record per `pollinterval`. The device does not acknowledge
  its authentication handshake until roughly 20 exchanges after the response
  is sent, and serves no measurement page until it does, so the slower
  cadence could leave every measured value reading 0 on a device that was
  answering normally. Publishes `experimental.microlink.diag.auth_status`
  and `experimental.microlink.diag.auth_refused`, and warns when the
  handshake goes unacknowledged.
* The driver now closes the Microlink session with a STOP when it exits.
  Without it the device could answer a later client's session request
  mid-walk rather than at page 0, and would not re-synchronize until
  re-enumerated.
* `2:4.7.28` and `2:4.7.49` are percentages of the nominal ratings rather
  than absolute readings. `ups.load` is now published from the former, with
  `ups.realpower` and `ups.power` derived against the nominal ratings,
  instead of both reading two orders of magnitude low. `ups.test.result`
  now comes from the battery-scope usage that a self test actually updates,
  and `ups.test.interval` is derived from the self-test schedule enumeration.
* Newly mapped: `input.sensitivity`, `ups.beeper.status`,
  `experimental.battery.firmware` and
  `experimental.statistics.battery.transfers`.
* Driver-specific diagnostic variables moved from `microlink.*` to
  `experimental.microlink.*`.
* Two `-Werror` warnings that only surface on specific compilers
  (`-Wcovered-switch-default` on Clang/CodeQL, `-Wswitch-default` under
  `fightwarn`) were resolved on the async transfer callback's `switch` by
  guarding a pragma per-warning instead of picking one side; see the commit
  message and `m4/ax_c_pragmas.m4` for the detection this relies on.
* Documented the Microlink descriptor format the driver parses, for anyone
  extending its variable maps (NUT Developer Guide).

## Update: outlet-group-count and beeper fixes

Three more commits, cherry-picked in from separate branches after further
testing against an APC Smart-UPS X1500 (SMX1500RM2U, issue #3587):

* `outlet.group.count` (and everything gated on group presence: command
  registration, `outlet.group.N.*` variables, and the "all groups" target
  mask for `load.*`/`shutdown.*`) previously returned/behaved as if no
  groups existed whenever the unswitched "main" group was absent, even with
  switched groups present. The SMX1500RM2U has three switched groups and no
  main group, so this silently dropped all outlet-group functionality on
  it. Fixed by checking group presence per index instead of gating
  everything on group 0; verified unchanged against a real SCL500RM1UC
  capture (which does have a group 0).
* `ups.beeper.status` matched two exact enum values (0xC1/0xC2), derived
  from one device that happened to always set the same extra bits alongside
  the enabled/disabled bit. The SMX1500RM2U reports 0x09 (enabled plus an
  unrelated PowerChute UI flag bit), which matched neither value and so
  published a raw number instead of a label. Decompiling PowerChute's own
  alarm-status class confirmed this register is a bitmask, not an enum;
  switched the map to `MLINK_DESC_BITFIELD_MAP` with the real bit values.
  Verified against captures from both devices.
* `beeper.mute` wrote a "user interface command" register with no confirmed
  effect on real hardware (no observable change reported on the
  SMX1500RM2U). Decompiling PowerChute shows its actual beeper on/off
  control writes a persistent value to the same usage `ups.beeper.status`
  reads -- not the transient mute/re-arm action NUT documents `beeper.mute`
  as. Replaced with `beeper.enable`/`beeper.disable`, which write that
  usage directly. `test.panel.start` is untouched; it already works via the
  original register.

None of this makes the SMX1500RM2U's page-0 gap (still stuck at page 0x3C,
descriptor starting at 0x48 -- see the original PR description above)
usable; these are independent fixes uncovered while working through that
device's debug logs. Discussion continues in #3587.

`NEWS.adoc` updated to cover all three. `DRIVER_VERSION` left at `0.03` for
now since the PR is still open (no released version to distinguish it
from); happy to bump before merge if maintainers prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LB4rN1EMA4s2TumED2ZWB2


nmbro:
got a fix for the "missing enum cases" ready, I'm just waiting for the CI pipeline to run through before I push it - in case I need to touch up some more things

jimklimov:
A CI run across so many systems can take about half a day (and now competes with some other PRs), so perhaps better you push the fix and I cancel the earlier build as it had started "just recently", relatively speaking.

nmbro:
I'm investigating my unit going into a locked out state again.

it's a special kind of lock out this time - I'm getting all of the static data from the microlink tunnel, but the dynamic data: ie. battery status, ups status on the microlink are just all 0

and it looks like my AUTH is not getting accepted by the UPS for some reason. microlink sure is a fun and fantastic proprietary protocol

nmbro:
Pushed a follow-up fixing the CI failures and the lockout I mentioned above.

**CI**: two `-Werror` diagnostics account for all 66 failing checks (the
`slow build ... abortion verdict` entries are the same two, once per toolchain):
`-Wswitch-enum` on the `transfer->status` switch, and an unused `now` parameter
on `!WITH_USB` builds.

**The lockout was our polling cadence, not the UPS.** The driver fetched one
tunnel record per `pollinterval`. This device does not acknowledge the
slave-password handshake until roughly 20 further exchanges, and serves no
measurement page until it does, so the acknowledgement never arrived and every
measured value read 0 while static pages came through fine. It now polls a full
page pass per update cycle, and publishes and warns on `AUTH_STATUS`.

Also: the HID-PDC fallback now engages when the tunnel answers but reports
nothing usable, not only at startup; `upsdrv_cleanup()` sends `STOP`;
`2:4.7.28`/`2:4.7.49` are percentages, so `ups.load` is direct and
`ups.realpower`/`ups.power` derived; `microlink.*` diagnostics moved to
`experimental.microlink.*`; and `docs/apcmicrolink-descriptors.txt` documents
the descriptor format, which obsoletes the "no new files" checklist item above.

One question: `apc-hid` uses `low`/`medium`/`high` for `input.sensitivity`. I
publish `normal`/`reduced`/`low`, matching the device and PowerChute. Happy to
remap for consistency with the sibling driver.

Use of coding helper tools and AI disclosed: **Claude Code was used** for
development assistance.

jimklimov:
I'll have to check precedents in other drivers about `input.sensitivity`, whether they are diverse or not, as `docs/nut-names.txt` does not currently state if that item is "opaque" or has specific expected values.

nmbro:
@jimklimov anything I need to fix? except for potentially the alignment of the input.sensitivity wording?

jimklimov:
Regarding sensitivity, I'm now going through drivers for precedents. TL;DR: seems we've got no standard to adhere to here, so matching the device docs is as good a shot in the dark as any. I'll update the docs.

* drivers/apc-ats-mib.c:static info_lkp_t apc_ats_sensitivity_info[] : `high`, `low`
* drivers/apc-hid.c:static info_lkp_t apc_sensitivity_info[]: `low`, `medium`, `high`
* drivers/apc-mib.c:static info_lkp_t apcc_sensitivity_modes[]: `auto`, `low`, `medium`, `high`
* drivers/apcsmart-old.h, drivers/apcsmart_tabs.c [?] (needs more digging)
* drivers/apcupsd-ups.h [?] (needs more digging, maybe passed as is from apcupsd?)
* drivers/belkin-hid.c:static const char *belkin_sensitivity_conversion_fun(): `reduced`, `low`, `normal` 
* drivers/belkinunv.c: `normal/medium/low` (per comments, may need to dig into `voltsens[]` for more)
* drivers/cps-hid.c:static info_lkp_t cps_sensitivity_info[]: `low`, `normal`, `high`
* drivers/delta_ups-hid.c:static info_lkp_t delta_ups_sensitivity_info[]:  `normal`, `reduced`, `low`
* drivers/dummy-ups.h: passed as is from capture file or relayed device
* drivers/eaton-ats16-nm2-mib.c:static info_lkp_t eaton_ats16_nm2_sensitivity_info[]:  `normal`, `high`, `low`
* drivers/eaton-ats16-nmc-mib.c:static info_lkp_t eaton_ats16_nmc_sensitivity_info[]:  `normal`, `high`, `low`
* drivers/eaton-ats30-mib.c:static info_lkp_t eaton_ats30_input_sensitivity[]: `high`, `low`
* drivers/mge-hid.c:static info_lkp_t mge_sensitivity_info[]: `normal`, `high`, `low`
* drivers/mge-xml.c:static const char *mge_sensitivity_info(): `normal`, `high`, `low`
* drivers/microdowell.c: `H`, `M`, `L`
* drivers/tripplitesu.c: sensitivity[]: `Normal`, `Reduced`, `Low` (cased)

nmbro:
Think I found a fix for the beeper settings in microlink too, going to push as well

PR #3597, opened 2026-08-30
Fix NHS serial framing for embedded packet markers

### LINKED ISSUE #3592: nhs_ser: unconditional 0xFF resync discards HWINFO packets that contain 0xFF in the payload (Premium PDV Senoidal 1500VA)
Hi @Challado, @jimklimov

TL;DR: my unit transmits HWINFO packets with `0xFF` in byte 11 of the payload, which trips the driver's packet assembler and makes it discard them. The driver never obtains hwinfo, keeps re-requesting it, and sleeps 6 seconds inside every `upsdrv_updateinfo()` call.

## Environment

- **NUT:** current `master` (`1da36c4794f42dcaeb93a32b6e8f0b2f95233b5c`, tarball fetched 2026-08-29), NHS Nobreak Drivers 0.05`.
- **OS:** Debian 13 (trixie), kernel `6.18.34+rpt-rpi-v8`, aarch64, Raspberry Pi 4 Model B Rev 1.5, gcc 14.2.0.
- **Build:** `./configure --with-drivers=nhs_ser,dummy-ups --with-user=nut --with-group=nut --sysconfdir=/etc/nut --with-statepath=/run/nut --with-altpidpath=/run/nut` then `make -C drivers nhs_ser`

## Device

- **NHS Premium PDV Senoidal Rack 3U 1500 VA**, part `91.B1.015202`, 4 × 9 Ah / 12 V.
- CDC-ACM: `idVendor=0x0925 idProduct=0x1241`, `NXP SEMICONDUCTORS` / `MCU VIRTUAL COM DEMO`, `bcdDevice 1.01` → `/dev/ttyACM0`.

```ini
[nhs]
    driver = nhs_ser
    port = /dev/serial/by-id/usb-NXP_SEMICONDUCTORS_MCU_VIRTUAL_COM_DEMO-if00
    serial_flow_control = none
    va = 1500
    ah = 9
    vbat = 12.0
```

`serial_flow_control = none` is applied correctly (`stty` confirms `-crtscts`), and DATA packets decode fine. #3543 works on this model.

## Symptom

`Incoming packet size not recognized, discarding!` continuously.
`upsc` is missing `ups.model`, `ups.mfr`, `ups.firmware` and `device.serial` entirely, with `battery.packs: 0`.

I first noticed the Home Assistant integration constantly flapping all entities to `unavailable` and back. The driver was discarding the `hwinfo` packets and sleeping 6 seconds between retries. This pushed past the 15-second `MAXAGE` and made NUT consider the UPS stale. As a workaround I set `MAXAGE 45` in `/etc/nut/upsd.conf` and the Home Assistant entities stopped flapping. `upsc` was still missing the fields but that was a minor issue.

## Reproduction on master

I did a 3-minute run with `-DDDDD -x debug_pkt_raw -x debug_pkt_hwinfo -x debug_pkt_data`:

| Observation | Count |
|---|---|
| `Incoming packet size not recognized, discarding!` | 17 |
| DATA packets parsed (length 21) | 26 |
| unrecognised fragments of length 39 | 10 |
| unrecognised fragments of length 7 | 7 |
| init packets sent | 43 |
| `Increase checktime` | 42 |
| hwinfo ever parsed | **never** |

`Incoming packet does not seem to end` never fired, so `DATAPACKETSIZE` is not the limit being hit.

## Root cause

I captured the raw serial stream directly (driver stopped, `termios` set to 2400 8N1, no flow control), sending the driver's own init strings. 698 bytes: 30 clean 21-byte DATA frames, plus one HWINFO reply to each init string.

**The two HWINFO packets, verbatim:**

```
long init  (50 bytes, [1]=0x32=50):
ff 32 53 55 00 34 00 04 54 9c b0 ff 78 dc 3b 00 39 31 42 31 30 31 35 32
30 30 30 30 32 33 30 30 00 00 00 00 00 00 00 00 00 00 00 00 00 00 15 33
b2 fe

short init (18 bytes, [1]=0x12=18):
ff 12 53 55 00 34 00 04 54 9c b0 ff 78 dc 3f 00 24 fe
```

Both are valid:

| Packet | Length | `[1]` declares | Checksum stored | Recomputed | |
|---|---|---|---|---|---|
| HWINFO long | 50 | 50 | `0xb2` @ `[48]` | `0xb2` over `[1..47]` | MATCH |
| HWINFO short | 18 | 18 | `0x24` @ `[16]` | `0x24` over `[1..15]` | MATCH |
| DATA | 21 | 21 | `0xe1` @ `[19]` | `0xe1` over `[1..18]` | MATCH |

Both HWINFO packets carry `0xFF` at payload offset 11, which trips the assembler that treats any `0xFF` as start-of-packet.

```c
if (chr == 0xFF) {          /* DataPacket start */
    datapacketstart = true;
    memset(datapacket, 0, sizeof(datapacket));
    datapacket_index = 0;
}
```

So the remainder of the packet is re-read as a new packet whose "length" byte is `0x78`.
- 50-byte HWINFO `becomes` one **39-byte** fragment
- 18-byte HWINFO `becomes` one **7-byte** fragment

Neither match the expected length so both are discarded.
> This could technically happen to DATA packets too, in case a measurement or checksum byte lands on `0xFE`/`0xFF`, but I have not observed it on this unit.

### The cascade

Because hwinfo never parses, `lastpkthwinfo.checksum_ok` stays false and the driver re-requests on every poll, then `usleep(checktime)` sleeps for 6 seconds*. The effective poll cadence becomes ~6 seconds.
>`checktime` saturated at `max_checktime` (6 seconds).

`upsd` intermittently marked the UPS `stale` at the default `MAXAGE` of 15, which explained the flapping entities in the HA integration.

### HWINFO Data

The driver already knows how to read it, the 50-byte packet decodes cleanly under `mount_hwinfo()` mappings:

| Field | Offset | Value |
|---|---|---|
| model | `[3]` | 85 |
| hardware version | `[4]` | 0 |
| software version | `[5]` | 52 |
| **numbatteries** | `[7]` | **4** (correct — 4 × 9 Ah pack) |
| serial | `[16..31]` | `91B1015200002300` (matches part `91.B1.0152`) |

> #### Secondary: `debug_pkt_raw` can't see this
> `pdatapacket()` is called from inside `mount_datapacket()` and `mount_hwinfo()`. Since packets that fail the size dispatch never reach either, `-x debug_pkt_raw` prints nothing for them. I bypassed the driver to obtain the bytes above. Dumping the buffer in the `else` branch would make this class of bug self-diagnosing.

## Proposed fix

Make the framing length-aware rather than marker-driven: once `datapacketstart` is true, stop re-triggering on `0xFF` and use the declared length at `[1]` to bound the read instead of scanning for `0xFE`. This is model-agnostic and doesn't change any field or status-bit mappings.

Happy to test a patch or a branch on this hardware and report back, and to run any further captures that would help.

Thanks for the driver, and for the work on #3543.

Challado:
## Summary

- make NHS serial packet framing honor the total length declared in byte 1, allowing `0xFF` and `0xFE` inside packet payloads
- preserve partial packet state across normal serial-read timeouts and validate packet length and end marker before dispatch
- add compatibility initialization fallback and optional model description override
- retain per-model protocol metadata and bypass alarm behavior, and correct input voltage minimum/maximum publication
- document the driver changes in `NEWS.adoc`

## Testing

Hardware validation is requested from the reporter of #3592, who supplied captures demonstrating valid 18-byte and 50-byte HWINFO frames containing `0xFF` in the payload.

Fixes #3592


jimklimov (approved):
Thanks, looks reasonable upon at least cursory review.

Reading the PR was complicated by unrelated style changes however (removing braces, gluing together multi-line debugs or large `if` clauses into 200+ char long single-line monsters) - do you insist on those?

Challado:
Hi Jim,

Thank you for the review and approval.

The refactoring was intentional. I am also taking this opportunity to
teach some people on my development team about device-driver
development at this level.

To be honest with you, the code became easier for me to maintain this
way than it was before. I had written much of the original code in a
hurry, and I was already refactoring it in a separate internal Git
repository. I decided to bring the work together in this PR.

With the way I program, this structure is more intelligible to me.
There were many unnecessary loop exits and many early returns that
made me lose track of the execution flow and cost me time while
debugging. At least with the current structure, I do not get as lost
when following the code.

To be completely honest, C stopped being my main area more than 20
years ago. I sincerely apologize that these changes made the PR more
difficult to review, but I intend to continue using this programming
style so that I can keep helping with this driver without so much
difficulty reading and rereading the code several times just to find
my place again.

I hope you understand, and thank you again for taking the time to review it.

Best regards,
Lucas Bocchi

jimklimov:
Ok, makes sense - especially if, for the time being, it is you that would be occasionally nagged and summoned to maintain these files :D until others learn the ropes...

PR #3598, opened 2026-08-31
common/snprintf.c: preserve wide integer values

### LINKED ISSUE #1602: Fallback snprintf.c emits a warning if built
From `common/snprintf.c`:
````
      case 'p':
        strvalue = va_arg (args, void *);
        /* FIXME: in 64-bit (and Windows-targeted) builds this code yields:
         * warning: cast from pointer to integer of different size
         * so probably prints a truncated pointer value. Should check
         * if sizeof(void*) == sizeof(long) and output e.g. two pieces
         * (lower and upper long for bytes of the pointer).
         */
        fmtint (buffer, &currlen, maxlen, (long) strvalue, 16, min, max, flags);
        break;
````

So far only noticed on Windows MSYS2 builds (and even that - when an antivirus precluded writing `conftest.exe` so the system `snprintf` was deemed missing). Most platforms do not even enable and build this source file.

### PR DESCRIPTION
**Pull request by user01010111**
## Summary

- Carry integer magnitudes as `unsigned LLONG` through the fallback formatter, with signedness tracked separately.
- Retrieve `%o`, `%u`, `%x` and `%X` arguments directly into unsigned storage instead of passing them through a possibly narrower signed `long`.
- Convert `%p` through `size_t` before widening, avoiding pointer sign extension on conventional 32-bit and LLP64 targets.
- Retain unsigned negation for negative signed values so `LLONG_MIN` is handled without signed overflow.
- Extend `TEST_SNPRINTF` with focused `UINT_MAX`, `ULONG_MAX`, `ULLONG_MAX`, top-bit hexadecimal and full-width octal comparisons.

Fixes #1602

## Review follow-up

Wider ABI testing prompted by review found that the previous PR head regressed the narrow unsigned conversions on systems where `long` is narrower than `long long`. On native Windows LLP64, `%u` with `UINT_MAX` was widened through negative `long` and printed as `18446744073709551615`; the `%o`, `%x`, `%X` and corresponding `unsigned long` cases failed in the same way.

Casting later inside `fmtint()` cannot repair a value already changed by the earlier signed conversion. This revision instead keeps unsigned values unsigned from `va_arg()` through digit conversion. For signed decimal formats, the caller supplies the converted unsigned value and its sign separately; `fmtint()` negates only the unsigned value when producing the magnitude. Plain `-value` is not used because it overflows for `LLONG_MIN`.

## Validation

- Native Windows 11 x64, MSYS2 MinGW64 GCC 16.2.0 Rev3, LLP64:
  - exact PR head failed 8 of 13 focused integer/pointer comparisons;
  - the revised source passed all 13;
  - the expanded built-in comparison passed 196/196;
  - a native forced-fallback configure confirmed `snprintf` and `vsnprintf` unavailable and `HAVE_LONG_LONG_INT=1`;
  - `make -C common V=1 snprintf.lo` passed with GNU99, `-pedantic` and `-Werror`, producing a PE x86-64 object.
- Native Linux ILP32 GCC:
  - exact PR head failed nine focused cases and emitted `-Wpointer-to-int-cast`;
  - the revised source passed 13/13 and the built-in comparison passed 196/196.
- Native Linux LP64:
  - focused and built-in comparisons passed with GCC 16.2.1 and Clang 22.1.8;
  - GNU89 and strict C99/pedantic builds passed with warnings as errors;
  - Clang UBSan passed, including `LLONG_MIN`.
- Forced-fallback project object builds passed with GCC and Clang.
- `make stylecheck`, the non-ASCII source check and `git diff --check` passed.
- `-Wconversion -Wsign-conversion` still reports 11 pre-existing diagnostics in this legacy file; this revision removes the additional signed-to-unsigned diagnostic introduced by the previous PR head and adds none.

No upstream CI result is claimed for this revision until it is published.

## General C checklist

- [x] Integer widths and signedness are carried explicitly without assuming `long` matches pointers or `long long`.
- [x] The change follows the existing fallback implementation and in-file test placement; no new helper, dependency or file is introduced.
- [x] Coding style and whitespace follow `common/snprintf.c` and `docs/developers.txt` precedent.
- [x] No build, distribution or documentation list needs updating.

## AI assistance

OpenAI Codex `gpt-5.6-sol` with xhigh reasoning was used for repository analysis, implementation, review, drafting and local/native validation. Anthropic Claude Fable 5 (`claude-fable-5`) with xhigh reasoning performed an independent adversarial review and produced a prototype used as design evidence. The human contributor reviewed the complete diff and validation evidence and remains responsible for the contribution.

### DIFF
--- a/common/snprintf.c
+++ b/common/snprintf.c
@@ -118,7 +118,8 @@ static void dopr (char *buffer, size_t maxlen, const char *format,
 static void fmtstr (char *buffer, size_t *currlen, size_t maxlen,
                     char *value, int flags, int min, int max);
 static void fmtint (char *buffer, size_t *currlen, size_t maxlen,
-                    LLONG value, int base, int min, int max, int flags);
+                    unsigned LLONG uvalue, int negative,
+                    int base, int min, int max, int flags);
 static void fmtfp (char *buffer, size_t *currlen, size_t maxlen,
                    LDOUBLE fvalue, int min, int max, int flags);
 static void dopr_outch (char *buffer, size_t *currlen, size_t maxlen, char c );
@@ -175,6 +176,7 @@ static void dopr (char *buffer, size_t maxlen, const char *format, va_list args)
 {
   char ch;
   LLONG value;
+  unsigned LLONG uvalue;
   LDOUBLE fvalue;
   char *strvalue;
   int min;
@@ -313,39 +315,40 @@ static void dopr (char *buffer, size_t maxlen, const char *format, va_list args)
           value = va_arg (args, LLONG);
         else
           value = va_arg (args, int);
-        fmtint (buffer, &currlen, maxlen, value, 10, min, max, flags);
+        fmtint (buffer, &currlen, maxlen, (unsigned LLONG)value,
+                (value < 0), 10, min, max, flags);
         break;
       case 'o':
         flags |= DP_F_UNSIGNED;
         if (cflags == DP_C_SHORT)
 #ifdef C89PLUS
-          value = (unsigned short int)va_arg (args, unsigned int);
+          uvalue = (unsigned short int)va_arg (args, unsigned int);
 #else
-          value = va_arg (args, unsigned short int);
+          uvalue = va_arg (args, unsigned short int);
 #endif
         else if (cflags == DP_C_LONG)
-          value = (long)va_arg (args, unsigned long int);
+          uvalue = va_arg (args, unsigned long int);
         else if (cflags == DP_C_LLONG)
-          value = (LLONG)va_arg (args, unsigned LLONG);
+          uvalue = va_arg (args, unsigned LLONG);
         else
-          value = (long)va_arg (args, unsigned int);
-        fmtint (buffer, &currlen, maxlen, value, 8, min, max, flags);
+          uvalue = va_arg (args, unsigned int);
+        fmtint (buffer, &currlen, maxlen, uvalue, 0, 8, min, max, flags);
         break;
       case 'u':
         flags |= DP_F_UNSIGNED;
         if (cflags == DP_C_SHORT)
 #ifdef C89PLUS
-          value = (unsigned short int)va_arg (args, unsigned int);
+          uvalue = (unsigned short int)va_arg (args, unsigned int);
 #else
-          value = va_arg (args, unsigned short int);
+          uvalue = va_arg (args, unsigned short int);
 #endif
         else if (cflags == DP_C_LONG)
-          value = (long)va_arg (args, unsigned long int);
+          uvalue = va_arg (args, unsigned long int);
         else if (cflags == DP_C_LLONG)
-          value = (LLONG)va_arg (args, unsigned LLONG);
+          uvalue = va_arg (args, unsigned LLONG);
         else
-          value = (long)va_arg (args, unsigned int);
-        fmtint (buffer, &currlen, maxlen, value, 10, min, max, flags);
+          uvalue = va_arg (args, unsigned int);
+        fmtint (buffer, &currlen, maxlen, uvalue, 0, 10, min, max, flags);
         break;
       case 'X':
         flags |= DP_F_UP;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by user01010111** [comment]
The CentOS 8/ppc64le OBS failure occurs during RPM database initialisation, before `%prep` or any source compilation:

`BDB0091 DB_VERSION_MISMATCH: Database environment version mismatch`

CentOS 9/ppc64le builds this commit successfully. Could the CentOS 8/ppc64le preinstall image (`a0769ebbc600e42fa9742d61cacaedd7`) be refreshed before rerunning the target?

**Comment by jimklimov** [comment]
OBS scenarios are on Open Build System, following their dependency tree etc. - as often as they regenerate them. I gather there is a regular inability to start the (emulated?) builders for this platform, which is mostly worked around by persistent restarting of the build in their Web-UI...

File: common/snprintf.c [hunk 1]
```diff
@@ -522,7 +523,7 @@ static void fmtint (char *buffer, size_t *currlen, size_t maxlen,
   {
     if( value < 0 ) {
       signvalue = '-';
-      uvalue = -value;
+      uvalue = -uvalue;
```
jimklimov, line 531: This one change does not look right.

jimklimov (changes requested):
I think the signed number support got broken here, the rest seems OK, thanks.

common/snprintf.c
user01010111: Thanks, Jim. This is intentional unsigned arithmetic. After the negative `LLONG` value is converted to the corresponding unsigned type, C defines unary minus modulo `2^N`, so this produces its magnitude without overflowing for `LLONG_MIN`. Using `-value` here would be undefined for `LLONG_MIN`. The added `%lld` comparison for `LLONG_MIN` exercises this path and passed against the system formatter.

Happy to add a short explanatory comment if that would make the intent clearer?

jimklimov: I think yes, to avoid same questions later on. "Bits magic, not maths magic" :)

user01010111: Thanks, Jim. Fair point: unsigned negation in shared formatting code is unusual enough to deserve scrutiny. I've added a short comment so future readers do not have to reconstruct the modulo arithmetic to see why it is intentional.

common/snprintf.c
jimklimov: Still, I wonder how many platforms actually build this fallback implementations.

From experience with annoying compiler warnings, I'd actually expect this one with neither range-checks nor explicit casting to raise questions from static analysis.

Wouldn't it be clearer somewhat like this?
```
if (flags & DP_F_UNSIGNED) {
    uvalue = (unsigned LLONG)value;
} else {
    if (value < 0) {
        signvalue = '-';
        uvalue = -value;
    } else {
        uvalue = value;
        if (flags & DP_F_PLUS)  /* Do a sign (+/i) */
            signvalue = '+';
        else
            if (flags & DP_F_SPACE)
                signvalue = ' ';
    }
}
```

user01010111: Thanks, Jim. I took your concern back for an independent adversarial review and broader ABI testing. It confirmed the reasoning for unsigned negation, but it also found a separate regression in my previous diff that vindicates your underlying concern about the signed/unsigned structure.

The problem occurred earlier in `dopr()`: `%o`, `%u`, `%x` and `%X` values, including the `long` variants, were still retrieved through signed `long`. On Windows LLP64, `long` is 32-bit while `LLONG` is 64-bit, so values above `LONG_MAX` became negative and were then sign-extended before reaching `fmtint()`. Native Windows testing confirmed eight failures; for example, `%u` with `UINT_MAX` printed `18446744073709551615` instead of `4294967295`. The same issue appeared on ILP32, where a high pointer could also be sign-extended.

Your proposed separation pointed in the right direction. The one expression I could not safely use was `uvalue = -value`: when `value == LLONG_MIN`, that is signed overflow and therefore undefined behaviour. Converting to unsigned and negating there remains deliberate, defined modulo arithmetic.

I have pushed a rework that moves the separation up to the argument handling: unsigned formats are retrieved directly into `unsigned LLONG`, signed decimal values pass their sign separately, `%p` is converted through `size_t`, and `fmtint()` operates only on the unsigned magnitude. This also removes the additional `-Wsign-conversion` diagnostic you anticipated.

Against the exact previous PR head, native Windows LLP64 failed 8/13 focused cases and the rework passed 13/13. The expanded built-in comparison passed 196/196 on Windows, Linux LP64 and native ILP32, with GCC, Clang and UBSan coverage; the forced-fallback project object build also passed with warnings as errors on Windows, GCC and Clang.

Thank you for pressing on this. My explanation of the disputed line was sound, but my surrounding test coverage was not, and without your scrutiny this PR could have introduced a Windows regression while trying to fix one.

For transparency, I asked Anthropic Claude Fable 5 (`claude-fable-5`, xhigh reasoning) to adversarially review both positions, then used OpenAI Codex `gpt-5.6-sol` (xhigh reasoning) to implement and independently revalidate the result. I reviewed the complete diff and evidence and remain responsible for the contribution.

jimklimov: Sounds great, thanks! I guess we have all missed bits here and there in this recent flood of reviews, luckily not all blind spots had overlapped :)

Issue #3595, opened 2026-08-30
Refactor pragma usage in C/C++ code by using dedicated include files


jimklimov:
PR #3590 has demonstrated the possibility of using headers to offload repetitive code like pragma start/end incantations we use to quiesce certain warnings (like when compilers both require all pre-known `switch/case` possibilities to be covered including a `default`, and that it would never be visited for an enum with all values listed explicitly).

This issue is a relatively low-hanging fruit about identifying pragma combos that are frequently used, and replacing such instances with header files to begin and end a block by a couple of dedicated include files per combo (not sure "header files" would be the right term if we `#include` them many times in the middle of C source code). This would significantly shorten actual codebase in some files and make it more readable and maintainable.

Kudos to @user01010111 for being the first to try this idea in practice!

biergaizi:
Strongly approve this PR.

These pragmas were added during a project refactor campaign called "fightwarn" with the goal of suppressing all compiler warnings. Unfortunately, because it was rushed, the fixes themselves are often "brute-force" in nature. If the compiler says there's an integer overflow, the fix was to cast all possible integer values instead of fix the location of the overflow. If the compiler says there's a buffer overflow, the fix (even to a compiler false-positive) was to add range-checked in all locations where the value is used, rather than fixing the logic.

 For example, one commit was introduced to the driver I maintain to suppress an integer overflow warning:

```c
crc16_recv = (uint16_t)((uint16_t)(ident_response_end[0]) << 8) | (uint16_t)(ident_response_end[1]);
```

I later reverted this change because it was clearly an overfix.

These GCC pragmas are my most disliked commits during the fix. If these changes were subject to review of the maintainer of the respective drivers, I would have rejected them in my driver immediately. Sharing them into a standard header is a choice that I would make from the beginning, if the  I don't understand why the copy-and-paste approach was used in the first place.

PR #3599, opened 2026-08-31
include: share covered-switch pragmas [#3595]

user01010111:
## Summary

This moves the repeated GCC and Clang warning suppression around defensive
`default:` clauses into `nut-pragmas-covered-switch-default.h` and its paired
`-end.h` header. The headers are intentionally repeat-includable and are
distributed as private build inputs.

The change replaces 41 identical regions across 21 C and C++ translation
units. It does not alter any `switch`, `case`, or `default` body.

Related to #3595. This is one exact warning-family slice and does not close
the broader issue.

## Scope

This deliberately leaves the following distinct cases local:

- `include/nutipc.hpp`
- the mixed `-Wswitch-enum` block in `drivers/apcmicrolink.c`
- eight GCC-only regions in `drivers/tripplite_usb.c`,
  `drivers/powerp-bin.c`, `drivers/adelsystem_cbi.c`, and
  `drivers/generic_modbus.c`

No other warning-suppression family is generalised in this PR.

## Validation

- GCC 16.2.1, hard warning profile with `-Werror`: full clean build passed.
- Clang 22.1.8, medium warning profile with `-Werror`: full clean build
  passed.
- All 21 changed consumer files compiled in both builds, including the
  available Modbus, Avahi, neon, libusb, nutconf, C, and C++ paths.
- CppUnit passed with GCC and Clang: 34 tests each.
- Aggregate test suites passed with GCC and Clang: 9 of 9 each.
- NIT passed: 26 passed, 0 failed, 0 skipped.
- GCC and Clang C and C++ probes confirmed repeated inclusion and restoration
  of the warning state after each closing include.
- `make stylecheck`, `make check-source-nonascii`, and whitespace checks
  passed.
- A bounded `distcheck-light` run with `CHECK_FILES_QUICK_TARGETS=` and
  `SPELLCHECK_ERROR_FATAL=no` passed, including packaging, install,
  uninstall, and distclean checks. The candidate-specific quick checks were
  run separately as listed above.

The Clang hard warning profile stops on 14 warnings treated as errors: ten
existing conversion warnings in `common/common.c` and four existing global
constructor warnings in `common/nutwriter.cpp`. Compiling pristine source at
the same revision reproduced the same diagnostics.

An ordinary `distcheck-light` run is also stopped by existing Aspell findings
in unchanged `NEWS.adoc` and `docs/man/apcmicrolink.txt`. Those files match
the base revision byte-for-byte.

Local validation did not exercise Powerman, IPMI, GPIO, Modbus RTU USB,
macOS, or Windows paths. Upstream CI remains responsible for broader platform
coverage.

## AI assistance

OpenAI Codex gpt-5.6-sol was used for repository analysis, implementation,
review, drafting, and validation with high reasoning. The human contributor
remains responsible for the contribution.

### DIFF
--- a/clients/nutclient.cpp
+++ b/clients/nutclient.cpp
@@ -4792,30 +4792,11 @@ std::vector<std::string> TcpClient::explode(const std::string& str, size_t begin
 			state = QUOTED_STRING;
 			break;
 
-#if (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_PUSH_POP) && ( (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT) || (defined HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE) )
-# pragma GCC diagnostic push
-#endif
-#ifdef HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_COVERED_SWITCH_DEFAULT
-# pragma GCC diagnostic ignored "-Wcovered-switch-default"
-#endif
-#ifdef HAVE_PRAGMA_GCC_DIAGNOSTIC_IGNORED_UNREACHABLE_CODE
-# pragma GCC diagnostic ignored "-Wunreachable-code"
-#endif
-/* Older CLANG (e.g. clang-3.4) seems to not support the GCC pragmas above */
-#ifdef __clang__
-# pragma clang diagnostic push
-# pragma clang diagnostic ignored "-Wunreachable-code"
-# pragma clang diagnostic ignored "-Wcovered-switch-default"
-#endif
+#include "nut-pragmas-covered-switch-default.h"
 		default:
 			/* Must not occur. */
 			break;
-#ifdef __clang__
-# pragma clang diagnostic pop
-#endif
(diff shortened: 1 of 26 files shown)


biergaizi:
Strongly approve this PR.

These pragmas were added during a project refactor campaign called "fightwarn" with the goal of suppressing all compiler warnings. Unfortunately, because it was rushed, the fixes themselves are often "brute-force" in nature. If the compiler says there's an integer overflow, the fix was to cast all possible integer values instead of fix the location of the overflow. If the compiler says there's a buffer overflow  (even to a compiler false-positive), the fix was to add range-checked in all locations where the value is used, rather than fixing the logic. 

 For example, one commit was introduced to the driver I maintain to suppress an integer overflow warning:

```c
crc16_recv = (uint16_t)((uint16_t)(ident_response_end[0]) << 8) | (uint16_t)(ident_response_end[1]);
```

I later reverted this change because it was clearly an overfix.

These GCC pragmas are my most disliked commits during the fix. If these changes were subject to review of the maintainer of the respective drivers, I would have rejected them in my driver immediately. Sharing them into a standard header is a choice that I would make from the beginning, if the  I don't understand why the copy-and-paste approach was used in the first place.

jimklimov:
I dislike the screenfulls of these pragmas sprinkled over the code too, and kick myself for not thinking about the include-file approach earlier :)

As for keeping the warnings quiet, they are mostly (hopefully) due to sort-of-overkill in range checks that may be pointless in specific-bitness builds, but may vary between platforms (OTOH, like checking `int i <= INT16_MAX` which depends on bit-width of a native `int`, maybe always true, maybe not).

I hope there was not too much brute-forcing per se, although it is possible that attention meandered and some corners got cut as hundreds of very similar cases were addressed by hand or mass replacement patterns...

PR #3604, opened 2026-08-31
huawei-ups2000: fix broken instant command invocations and undefined Modbus behaviors.

### LINKED ISSUE #3593: huawei-ups2000: shutdown commands never power off the UPS on firmware V2R1C1SPC50 — the unit ignores Modbus function 0x10 writes; only 0x06 works
### Environment

- **UPS**: Huawei UPS2000-A (1kVA-3kVA line-interactive), firmware **V2R1C1SPC50** (device identification: `HUAWEI/UPS2000A/V2R1C1SPC50`)
- **NUT**: 2.8.5 (Windows x64 snapshot 4499), driver `huawei-ups2000` **v0.13**
- **Link**: USB (Exar XR21V1410 bridge, internal RS485) → COM3, Modbus RTU 9600 8N1, slave 1
- Note: issue is firmware-side, not OS-side — applies to Linux too.

### Summary

`upscmd shutdown.return` (and, per code inspection, `shutdown.stayoff` / `shutdown.reboot` / `shutdown.reboot.graceful`) **never powers the UPS off** on this unit: the command reports success (rc=0) but the UPS keeps running on battery until mains returns. Direct Modbus probing shows the root cause: **this firmware does not honor value-changing Modbus function 0x10 writes** (no response, register unchanged), while **function 0x06 single-register writes work reliably**. The driver sends **all** its writes through `modbus_write_registers()` (function 0x10) — including "single-register" writes (`ups2000_write_register()` is `ups2000_write_registers(addr, 1, …)`) — so every driver-initiated change is silently dropped.

### Evidence

**A. Direct register tests (COM3, raw Modbus, same unit, 2026-08-29):**

| Write method | Register(s) | Value change | Response | Effect |
|---|---|---|---|---|
| 0x06 | 11044 (start.auto) | 1 → 0 | ACK | persisted; UPS stays OFF when mains returns |
| 0x06 | 11044 (start.auto) | 0 → 1 | ACK | UPS auto-restarts on mains return |
| 0x06 | 11049 (shutdown delay, 0.1 min) | 0 → 5 | ACK | UPS fully powered off after ~30 s (11024: 3→0) |
| 0x10 (1 reg) | 11046 (beeper) | 0 → 1 | **no response** | no effect (readback stays 0) |
| 0x10 (1 reg) | 11044 (start.auto) | 0 → 1 | **no response** | no effect (readback stays 0) |
| 0x10 (2 regs) | 11047 + 11048 (the pair `shutdown.return` writes) | 0,0 | **no response** | no effect |
| 0x10 (1 reg) | 11046 | 0 → 0 (no-op) | ACK echo | n/a |

**B. End-to-end:** 7 test rounds of `upscmd -w -t 10 -u admin shutdown.return` (instcmds=ALL, admin user): rc=0 every time, UPS never powered off. After the command, register 11044 still read 0 — i.e. the driver's `setvar("ups.start.auto","yes")` also failed to write, consistent with the 0x10 path being dead. Only the direct 0x06 writes (11044=0 + 11049=delay) powered the UPS off, in a controlled sequence: battery mode → write → UPS off after the delay → mains return → stays off.

**C. Driver source (master, `drivers/huawei-ups2000.c`):**
- `ups2000_write_register()` → `ups2000_write_registers(ctx, addr, 1, &val)` → libmodbus `modbus_write_registers()` = **function 0x10** (never 0x06).
- `shutdown.return` / `shutdown.reboot` / `shutdown.reboot.graceful` → `ups2000_shutdown_guaranteed_return()`: `setvar("ups.start.auto","yes")` then `ups2000_write_registers(11047, 2, {offdelay*10/60, ondelay/60})` (2-register 0x10).
- `shutdown.stayoff` → `setvar("ups.start.auto","no")` + `ups2000_write_register(11049, offdelay*10/60)` (single-register 0x10).
- Beeper/instcmd direct writes (lines ~1524/1537) also go through `ups2000_write_register` → 0x10.

### Root cause

Firmware V2R1C1SPC50 does not act on Modbus function 0x10 writes that change a register value (no response; register unchanged). Since the driver uses 0x10 for **every** write, no driver-initiated register change (shutdown delays, start.auto, beeper, load.on) takes effect on this firmware, even though upscmd may report success.

### Workaround (confirmed working)

Power the UPS off directly over the wire with **function 0x06** writes:
1. write 11044 = 0 (start.auto = no → stays off when mains returns)
2. write 11049 = seconds × 10 / 60 (0.1-minute units; e.g. 60 s → 10) → UPS powers off after the delay

Verified repeatedly; also verified 11044 persists across a full power-off + mains cycle.

### Suggested fix

In the driver, for single-register writes use libmodbus `modbus_write_register()` (function 0x06) instead of `modbus_write_registers()` — that alone fixes beeper, start.auto, and `shutdown.stayoff`. For the 2-register guaranteed-return path (11047/11048), write them as two separate 0x06 calls, or fall back to the stayoff-style 11044 + 11049 0x06 writes, with readback verification. A runtime readback check would also make the driver robust to firmware variance (units where 0x10 works vs. units like this one).

### Offer

Happy to test a patched driver on this exact unit (including a readback-verified power-off round-trip). I can also provide the full register map I validated (11024 status, 11044 start.auto, 11046 beeper, 11049 shutdown delay, 11000-11028 / 12000-12033 telemetry).

---

*AI-assisted notice: this report was researched and written with the assistance of an AI coding agent; every claim was validated on real hardware (power-cut drills, direct Modbus probing) and reviewed by a human before posting.*

### LINKED ISSUE #3603: huawei-ups2000: bypass.start, shutdown.return, shutdown.reboot, and shutdown.reboot.graceful are broken in all NUT versions
After reading the bug report #3593, it prompted me to perform a review of `huawei-ups2000` driver again as its original author. Unfortunately, I found a serious regression introduced during the "fightwarn" campaign in e9f02e242acf09e95a32413a26dceb44b112547d that completely broke `bypass.start`, `shutdown.return`, `shutdown.reboot`, and `shutdown.reboot.graceful`.

Unlike the cited bug, this bug affects all NUT versions, all UPS hardware, and all firmware revisions.

Any attempt to use them will fail, with an error logged to `syslog`, for example:

    huawei-ups2000: instcmd: command [bypass.start] reg1 is negative
    huawei-ups2000: instcmd: command [shutdown.return] reg1 is negative
    huawei-ups2000: instcmd: command [shutdown.reboot] reg1 is negative
    huawei-ups2000: instcmd: command [shutdown.reboot.graceful] reg1 is negative

Because the regression was introduced shortly *after* the initial driver was merged into the upstream, as a result, instant commands *do no work in any version of NUT*, with the exception of the earliest development build. This problem was not discovered during my test for the same reason: it's a regression that occurred after the Pull Request was merged. 

## Contributing Factors and Context

This regression was introduced due to a combination of two factors:

1. During the development, I mistakenly believed that GCC `-Wextra` warnings were turned on, but in fact they were turned off! I didn't notice it. As a result, many unsafe operations such as signness confusions existed in my original Pull Request.

2. The merge of the Pull Request coincided with NUT's 2.8.0 refactoring, during a project refactor campaign called "fightwarn" with the goal of suppressing all compiler warnings. Unfortunately, because it was rushed, the fixes themselves are often "brute-force" in nature. If the compiler says there's an integer overflow, the fix was to cast all possible integer values instead of fix the location of the overflow. If the compiler says there's a buffer overflow, the fix (even to a compiler false-positive) was to add range-checked in all locations where the value is used, rather than fixing the logic.

For example, one commit was introduced in `huawei-ups2000` to suppress an integer overflow warning with three typecasts, when one casting was necessary.

```c
crc16_recv = (uint16_t)((uint16_t)(ident_response_end[0]) << 8) | (uint16_t)(ident_response_end[1]);
```

For another example, one commit introduced a brute-force range check to a `crc16()` function, this was suboptimal since changing `crc16()` to accept `size_t` instead of `uint16_t` would be safer. 

```c
if (ident_response_len < IDENT_RESPONSE_CRC_LEN
|| (((uintmax_t)(ident_response_len) - IDENT_RESPONSE_CRC_LEN) > UINT16_MAX)
) {
        fatalx(EXIT_FAILURE, "response header shorter than CRC "
                             "or longer than UINT16_MAX!");
}

crc16_calc = crc16(ident_response, ident_response_len - IDENT_RESPONSE_CRC_LEN);
```

I later submitted a Pull Request, which partially improved these brute-force changes.

However, I did not notice all changes, and missed the critical instant command regression.

## Root Cause Analysis

In `huawei-ups2000`, all instant commands are driven by a lookup table to consolidate the logic to a central dispatcher.

```c
/*
 * A lookup table of all instant commands "cmd" and their
 * corresponding registers "reg". For each instant command,
 * it's handled by...
 *
 * 1. One register write, by writing "val1" to "reg1", the
 * simplest case.
 *
 * 2. Two register writes, by writing "val1" to "reg1", and
 * writing "val2" to "reg2". One after another.
 *
 * 3. Calling "*handler_func" and passing "reg1". This is
 * used to handle commands that needs additional processing.
 * If "reg1" is not necessary or unsuitable, "-1" is used.
 */
#define REG_NULL  -1, -1
#define FUNC_NULL NULL

static struct ups2000_cmd_t {
        const char *cmd;
        const int16_t reg1, val1, reg2, val2;
        int (*const handler_func)(const uint16_t);
} ups2000_cmd[] =
{
        { "test.battery.start.quick", 2028,  1, REG_NULL, FUNC_NULL },
        { "test.battery.start.deep",  2021,  1, REG_NULL, FUNC_NULL },
        { "test.battery.stop",        2023,  1, REG_NULL, FUNC_NULL },
        { "beeper.enable",            1046,  0, REG_NULL, FUNC_NULL },
        { "beeper.disable",           1046,  1, REG_NULL, FUNC_NULL },
        { "load.off",                 1045,  0, 1030, 1,  FUNC_NULL },
        { "bypass.stop",              1029,  1, 1045, 0,  FUNC_NULL },
        { "load.on",                  1029, -1, REG_NULL, ups2000_instcmd_load_on                  },
        { "bypass.start",             REG_NULL, REG_NULL, ups2000_instcmd_bypass_start             },
        { "beeper.toggle",            1046, -1, REG_NULL, ups2000_instcmd_beeper_toggle            },
        { "shutdown.stayoff",         1049, -1, REG_NULL, ups2000_instcmd_shutdown_stayoff         },
        { "shutdown.return",          REG_NULL, REG_NULL, ups2000_instcmd_shutdown_return          },
        { "shutdown.reboot",          REG_NULL, REG_NULL, ups2000_instcmd_shutdown_reboot          },
        { "shutdown.reboot.graceful", REG_NULL, REG_NULL, ups2000_instcmd_shutdown_reboot_graceful },
        { NULL, -1, -1, -1, -1, NULL },
};
```

The comment explains three cases, and it's important to note Case 3 here:

>  3. Calling `*handler_func` and passing `reg1`. This is used to handle commands that needs additional processing. If `reg1` is not necessary or unsuitable, `-1` is used. 

In other words, not all `*handler_func` actually reads `reg1`, it's an optional argument despite that the function signature always includes it. Because declaring the function as a variadic function or a `(void *)` function would be over-engineering, a simple unsigned 16-bit value is used.

When the `*handler_func` doesn't use `reg1`, `reg1` is silently ignored. But, during the fightwarn refactor campaign, to make the code "safe". The compiler generates a warning because it saw that the signed `-1` is sometimes passed to `*handler_func` as an unsigned value, becoming 65535. As a result, an incorrect range check was added to the function, which refuses to invoke `*handler_func` if `reg1` is negative.

```patch
--- a/drivers/huawei-ups2000.c
+++ b/drivers/huawei-ups2000.c
@@ -1495,13 +1495,18 @@ static int instcmd(const char *cmd, const char *extra)
 
        if (cmd_action->handler_func) {
                /* handled by a function */
-               status = cmd_action->handler_func(cmd_action->reg1);
+               if (cmd_action->reg1 < 0) {
+                       upslogx(LOG_WARNING, "instcmd: command [%s] reg1 is negative", cmd);
+                       return STAT_INSTCMD_UNKNOWN;
+               } else {
+                       status = cmd_action->handler_func((uint16_t)cmd_action->reg1);
+               }
        }
```

Because we have four instant commands that don't take any registers:

```c
        { "bypass.start",             REG_NULL, REG_NULL, ups2000_instcmd_bypass_start             },
        { "shutdown.return",          REG_NULL, REG_NULL, ups2000_instcmd_shutdown_return          },
        { "shutdown.reboot",          REG_NULL, REG_NULL, ups2000_instcmd_shutdown_reboot          },
        { "shutdown.reboot.graceful", REG_NULL, REG_NULL, ups2000_instcmd_shutdown_reboot_graceful },
```

These calls are now always rejected.

The correct fix was to fix the logic, not to suppress the symptom: make `reg1` an unsigned value in the LUT.

I'll submit a patch soon to fix it.

## Comments

This regressions shows the undesirable effects of:

1. Suppressing compiler warnings for the sake of the suppressing warnings.
2. Not notifying the respective maintainers to review changes in drivers in which they're responsible.

    In some projects, a "code ownership" model is used: no change is made to any subsystems until the subsystem maintainer approve it (I believe the early days of NetBSD did that). In some projects, everyone is free to change whatever subsystems they see fit (which is the Linux kernel's model). I believe NUT would be better if a middle ground position is taken: a mechanism to notify driver maintainers to review changes would reduce these incidents. To prevent unresponsive maintainers from slowing down the project, the review can be made optional, a single GitHub CC may be sufficient.

biergaizi:
## huawei-ups2000: fix broken instant command invocations, close #3603

A serious regression was introduced during the "fightwarn" campaign that completely broke `bypass.start`, `shutdown.return`, `shutdown.reboot`, and `shutdown.reboot.graceful`. Any attempt to use them will fail, with an error logged to `syslog`:

    huawei-ups2000: instcmd: command [bypass.start] reg1 is negative
    huawei-ups2000: instcmd: command [shutdown.return] reg1 is negative
    huawei-ups2000: instcmd: command [shutdown.reboot] reg1 is negative
    huawei-ups2000: instcmd: command [shutdown.reboot.graceful] reg1 is negative

Because the regression was introduced shortly after the initial driver was merged into the upstream, these instant commands do no work in any NUT version (with the exception of the earliest development build). This problem was not discovered during pre-merge test for the same reason: it's a regression that occurred after the Pull Request was merged.

In huawei-ups2000, all instant commands are driven by a lookup table to consolidate the logic to a central dispatcher. Some commands are dispatched to a `(*handler_func)(uint16_t reg1)` function, which is used to handle commands that needs additional processing.

Not all `*handler_func` actually reads `reg1`, it's an optional argument. Because declaring the function to accept a variadic or a `(void *)` would be over-engineering, a simple unsigned 16-bit value is used, `*handler_func` may or may not use it.

During the project-wide "fightwarn" refactor campaign, it was noticed that `reg1` is stored in the lookup table as `int16_t` (`-1` indicates an unneeded register), but `*handler_func` accepts a `uint16_t`. In case that `*handler_func` doesn't need the register ID from the lookup table, `reg1` is implicitly converted from `-1` to `65535`. This is harmless, since `reg1` is not actually used in this case.

Unfortunately, in an attempt to make the code "safe" by suppressing compiler warning, an incorrect range check was added to the function, which refuses to invoke `*handler_func` if `reg1` is negative.

As a result, all instant commands that don't use `reg1` are broken. These commands include `bypass.start`, `shutdown.return`, `shutdown.reboot`, and `shutdown.reboot.graceful`.

To fix this problem, fix the implicit type conversion correctly by change the `reg1` data type in the lookup table from `int16_t` to `uint16_t`, and remove the "negative reg1" check. As defensive programming. all `*handler_func` that actually need `reg1` would check and abort if `reg1` is 0.

Fixes: e9f02e242acf ("drivers/huawei-ups2000.c: instcmd(): range-check and cast for ups2000_write_register() and handler_func()")

## huawei-ups2000: use modbus_write_register, not modbus_write_registers, fix https://github.com/networkupstools/nut/issues/3593.

On some Huawei UPS2000 models, no instant commands or variable writes are possible, because `modbus_write_registers()` doesn't work correctly.
    
Upon investigation, Huawei UPS2000 doesn't support multiple-register writes at all (Modbus command `0x10`), the official datasheet only supports single-register writes (Modbus command `0x06`). Multiple-register writes worked on some models, but this turned out to be undefined behavior, it doesn't work with all models.
    
This commit switches `modbus_write_registers()` to `modbus_write_register()` to fix the problem.

## huawei-ups2000: bump driver version to v0.14.

Dump driver versions to reflect changes.

## docs/man/huawei-ups2000.txt: document bug #3593 and bug #3603.

Two bugs #3593, #3603 can prevent users from sending any instant commands or changing any variables at all. Document these known bugs in the man page.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -117,6 +117,21 @@ https://github.com/networkupstools/nut/milestone/13
       authentication configuration discovery. It accepts `default`, `none`,
       or a specific authconf file path. [issue #3329]
 
+ - `huawei-ups2000` driver updates:
+    * Fixed a Modbus compatibility problem that made changing any variables
+      or sending any instant commands impossible with some UPS models. This
+      was caused by the use of `modbus_write_registers()` (Modbus command
+      `0x10`), which is not officially supported, making any register write
+      impossible on some models. All uses of `modbus_write_registers()` have
+      been changed to `modbus_write_register()` (Modbus command `0x06`) to
+      avoid this problem. [issue #3593, PR #3604].
+
+    * Fixed a serious problem previously existed in all NUT versions that
+      made many instant commands unusable, including `bypass.start`,
+      `shutdown.return`, `shutdown.reboot`, and `shutdown.reboot.graceful`.
+      Any attempt to use them would fail, with an error logged to `syslog`:
+      `reg1 is negative`. [issue #3603, PR #3604].
+
  - `nhs_ser` driver updates:
     * Modernized serial communication and added validated settings for baud
       rate, data bits, parity, stop bits, flow control, per-byte read timeout,
--- a/UPGRADING.adoc
+++ b/UPGRADING.adoc
@@ -87,6 +87,16 @@ The new `-A filename` option defaults to trying to use a `nutauth.conf` file
   device, please add `cypress_drain_quirk = 0` in your `ups.conf` and let us
   know. [issue #2534, PR #3563]
 
+- The `huawei-ups2000` driver fixed two serious problems that affected all
+  previous NUT versions. All `huawei-ups2000` users are strongly recommended
+  to upgrade to NUT 2.8.6, all previous versions are considered too unreliable
+  to use. Due to a Modbus compatibility problem, it was not possible to *write
+  to any registers at all* on some UPS models, meaning that changing any
+  *variable* or sending any *instant command* was impossible. Furthermore, due
+  to a broken instant command handler, some instant commands would always fail
+  with a `reg1 is negative` error (even with a compatible UPS), including
+  `bypass.start`, `shutdown.return`, `shutdown.reboot`, and
+  `shutdown.reboot.graceful`. [issue #3593, #3603, PR #3604].
 
 Changes from 2.8.4 to 2.8.5
 ---------------------------
--- a/docs/man/huawei-ups2000.txt
+++ b/docs/man/huawei-ups2000.txt
@@ -375,6 +375,16 @@ to endless read failures and permanent data stalls.
 Driver v0.12 (NUT v2.8.5) fixed support for Microsoft Windows,
 Windows was previously unsupported due to a file locking bug.
 
+Driver v0.14 (NUT v2.8.6) fixed two serious problems that affected all
+previous NUT versions, upgrading is strongly recommended. Due to a broken
+instant command handler, it was impossible to send some instant commands
+to the UPS, including *bypass.start*, *shutdown.return*, *shutdown.reboot*,
+and *shutdown.reboot.graceful*, any attempt to use them will fail with an
+"reg1 is negative" error in the syslog. Due to a separate Modbus compatibility
+problem, the negative impact is worse on some UPS models: it was not possible
+to *write to any registers at all*, meaning that changing any *variable* or
+sending any *instant command* is impossible.
+
 See *Historical bugs* at the bottom of this manual for links.
 
 Battery status has a non-fatal read failure
@@ -491,6 +501,12 @@ https://github.com/networkupstools/nut/issues/3244
 
 * Commit: call ser_close() before modbus_connect(), fix #3244:
 https://github.com/networkupstools/nut/commit/db2e148503d654726d5520d68e20e71117335813
+
+* Issue 3593: huawei-ups2000: shutdown commands never power off the UPS on firmware V2R1C1SPC50 - the unit ignores Modbus function 0x10 writes; only 0x06 works
+https://github.com/networkupstools/nut/issues/3593
+
+* Issue 3603: huawei-ups2000: bypass.start, shutdown.return, shutdown.reboot, and shutdown.reboot.graceful are broken in all NUT versions
+https://github.com/networkupstools/nut/issues/3603
 
 Windows Drivers:
 ~~~~~~~~~~~~~~~~
--- a/drivers/huawei-ups2000.c
+++ b/drivers/huawei-ups2000.c
@@ -162,7 +162,6 @@ static void ups2000_device_identification(void);
 static size_t ups2000_read_serial(uint8_t *buf, size_t buf_len);
 static int ups2000_read_registers(modbus_t *ctx, int addr, int nb, uint16_t *dest);
 static int ups2000_write_register(modbus_t *ctx, int addr, uint16_t val);
-static int ups2000_write_registers(modbus_t *ctx, int addr, int nb, uint16_t *src);
 static uint16_t crc16(uint8_t *buffer, size_t buffer_length);
 static time_t time_seek(time_t t, int seconds);
 
@@ -1731,8 +1730,12 @@ static int ups2000_shutdown_guaranteed_return(uint16_t offdelay, uint16_t ondela
 	val[0] = (offdelay * 10) / 60;
 	val[1] = ondelay / 60;
 
-	r = ups2000_write_registers(modbus_ctx, 1047 + 10000, 2, val);
-	if (r != 2)
+	r = ups2000_write_register(modbus_ctx, 1047 + 10000, val[0]);
+	if (r != 1)
+		return STAT_INSTCMD_FAILED;
+
+	r = ups2000_write_register(modbus_ctx, 1048 + 10000, val[1]);
+	if (r != 1)
 		return STAT_INSTCMD_FAILED;
 
 	return STAT_INSTCMD_HANDLED;
@@ -2061,7 +2064,14 @@ static int ups2000_read_registers(modbus_t *ctx, int addr, int nb, uint16_t *des
 }
 
 
-static int ups2000_write_registers(modbus_t *ctx, int addr, int nb, uint16_t *src)
+/*
+ * Huawei UPS2000 doesn't officially support multiple-register writes (Modbus
+ * command 0x10), the official datasheet only supports single-register writes
+ * (Modbus command 0x06). Multiple-register writes worked on some models, but
+ * this turned out to be undefined behavior, it doesn't work with all models.
+ * This is why this function is not symmetric to ups2000_read_registers().
+ */
+static int ups2000_write_register(modbus_t *ctx, int addr, uint16_t val)
 {
 	int i;
 	int r = -1;
@@ -2072,34 +2082,28 @@ static int ups2000_write_registers(modbus_t *ctx, int addr, int nb, uint16_t *sr
 			"Please file a bug report!", addr);
 
 	for (i = 0; i < 3; i++) {
-		r = modbus_write_registers(ctx, addr, nb, src);
+		r = modbus_write_register(ctx, addr, val);
 
 		/* generic retry for modbus write failures. */
-		if (retry_status == RETRY_ENABLE && r != nb) {
-			upslogx(LOG_WARNING, "modbus_write_registers() failed (%d, errno %d): %s",
+		if (retry_status == RETRY_ENABLE && r != 1) {
+			upslogx(LOG_WARNING, "modbus_write_register() failed (%d, errno %d): %s",
 				r, errno, modbus_strerror(errno));
 			upslogx(LOG_WARNING, "Register %04d has a write failure. Retrying...", addr);
 			sleep(1);
 			continue;
 		}
-		else if (r == nb)
+		else if (r == 1)
 			retry_status = RETRY_ENABLE;
 
 		return r;
 	}
 
 	/* Give up */
-	upslogx(LOG_ERR, "modbus_write_registers() failed (%d, errno %d): %s",
+	upslogx(LOG_ERR, "modbus_write_register() failed (%d, errno %d): %s",
 		r, errno, modbus_strerror(errno));
 	upslogx(LOG_ERR, "Register %04d has a fatal write failure.", addr);
 	retry_status = RETRY_DISABLE_TEMPORARY;
 	return r;
-}
-
-
-static int ups2000_write_register(modbus_t *ctx, int addr, uint16_t val)
-{
-	return ups2000_write_registers(ctx, addr, 1, &val);
 }
 
 
--- a/drivers/huawei-ups2000.c
+++ b/drivers/huawei-ups2000.c
@@ -55,7 +55,7 @@
 #endif
 
 #define DRIVER_NAME	"NUT Huawei UPS2000 (1kVA-3kVA) RS-232 Modbus driver (libmodbus link type: " NUT_MODBUS_LINKTYPE_STR ")"
-#define DRIVER_VERSION	"0.13"
+#define DRIVER_VERSION	"0.14"
 
 #define CHECK_BIT(var,pos) ((var) & (1<<(pos)))
 #define MODBUS_SLAVE_ID 1
--- a/drivers/huawei-ups2000.c
+++ b/drivers/huawei-ups2000.c
@@ -1446,14 +1446,17 @@ static int ups2000_delay_set(const char *var, const char *string)
  *
  * 3. Calling "*handler_func" and passing "reg1". This is
  * used to handle commands that needs additional processing.
- * If "reg1" is not necessary or unsuitable, "-1" is used.
+ * If "reg1" is not necessary or unsuitable, "0" is used.
  */
-#define REG_NULL  -1, -1
+#define REG_NULL  0, -1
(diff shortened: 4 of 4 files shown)


jimklimov (approved):
LGTM, but a NEWS.adoc entry would be welcome.

Maybe even add a note in UPGRADING.adoc to announce that all earlier versions were broken in this regard. That file usually is about changes that can adversely impact previous release users/packagers, but here users of the driver have all been adversely impacted already and can get fixed now, so maybe it belongs there too.

biergaizi:
Documentation files `NEWS.adoc` and `UPGRADING.adoc` updated. Let's merge it. I've tested it with a simple command (the server is too important to shut down right now), without seeing any breakage, so these changes should be at least neutral. Let's depend on user report for the final words.

biergaizi:
Don't wait for any test report, since it's not arriving. Let's merge it.

My best guess is that the LLM agent took over the reporter's computer, registered a GitHub account for them, generated a Python workaround for the NUT bug, generated a bug report, and the task was thus deemed "completed". The bug report was simply a bonus item suggested by the agent, it was not even required. In both the LLM Agent and the reporter's personal workflows, "monitor GitHub Issues for a week, and cooperate with the developers who want to help you" simply doesn't exist.

After all, when LLM agents can rewrite the whole driver within a day, what is even the point of spending any personal time to upstream anything these days? (In this era, the most efficient solution might be "fight fire with fire", by allowing upstream developers to chat with LLM agents directly on a end-users computer to do the troubleshooting - the security would be non-existent, but it's a good thought experiment). 

Please merge this PR. While I can't guarantee the change fixed the "undefined write" problem on the reporter's particular firmware version, I can guarantee that these changes fix a real undefined behavior, so it's at least a neutral change.

biergaizi:
Don't wait for any test report, since it's not arriving. Let's merge it.

My best guess is that the LLM agent (Claude Code, OpenClaw, Codex, whatever) took over the reporter's computer (with permissions), registered a GitHub account for them, generated a Python workaround for the bug, generated a bug report, and the task was thus deemed "completed".

My conclusion is from the fact that the GitHub account is completely new, and the only activity was a bug report and a repo that contains a Python script workaround marked LLM generated with the statement:

> This project - the monitor code, scripts, documentation, debugging, and the register-level reverse engineering - was developed with the assistance of an AI coding agent. Every technical claim was validated on real hardware (power-cut drills, direct Modbus probing) and reviewed by a human before publishing.

The bug report was simply a bonus item suggested by the agent, it was not even required. In both the LLM Agent and the reporter's personal workflows, "monitor GitHub Issues for a week, and cooperate with the developers who want to help you" simply doesn't exist. I originally thought LLM agents would speed up the conversation, since they can answer difficult question, so I intentionally asked a difficult (optional) question. which was to check the return code of `modbus_write_registers()`. I was mistaken, nothing will be accelerated when people are simply not interested in restarting their agents to answer the next question. In the past, users report problems, developers say "it works on my machine!", users say "how?". In the future, users report fixes and say "it works on my machine!", developers say "how?".

After all, when LLM agents can rewrite the whole driver within a day, what is even the point of spending any personal time to upstream anything these days?

In this era when nobody has the patience to cooperate, the most efficient solution that re-enables it might be "fight fire with fire", by allowing upstream developers to chat with LLM agents running on a end-user's computer to do the troubleshooting - security exists to a certain extent, because developers can only express a high-level goal such as "attach a debugger to `func()`", not raw commands. It would be easy to jailbreak, so it remains a thought experiment but I will not be surprised if someone actually deploys it, especially in "pure software" projects when it's easy to jail the whole environment with containers or VMs.

Please merge this PR. While I can't guarantee the change fixed the "undefined write" problem on the reporter's particular firmware version, I can guarantee that these changes fix a real undefined behavior, so it's at least a neutral change.

P.S: Another thing to highlight is the fact that the so-called *"register-level reverse engineering"* is completely unnecessary, because it's documented in the datasheet cited in the driver comments, but since it's much easier for LLM agents to read source code than a PDF, the agent decided to "reverse engineer" the meaning of these registers by guessing using the source code, and generate a Markdown document that describes what the registers do. Converting the PDF to text would probably use fewer tokens, it's not done probably because a "Skill" file called "check datasheets" does not exist by default in general-purpose agents.

jimklimov:
@biergaizi : Thanks, will do, just fixing something with the CI farm before merging the long train that has collected in the queue...

PR #3609, opened 2026-09-01
common: warn when become_user() leaves root credentials

### LINKED ISSUE #3471: Security: daemons should warn if they remain `root` for the bulk of work
On systems where we can `fork` and `setuid` (non-WIN32 at this time) but still remain a `root` after the daemonization milestone, we should print a WARNING into the logs.

Maybe we do so already, but if this is missed somewhere - make sure we do.

user01010111:
Aims to close #3471, pending maintainer review and approval.

This makes the shared POSIX `become_user()` path warn once when a daemon's real or effective UID remains zero after a successful credential transition.

The previous code reported the resulting UID only in a debug-level success message after `initgroups()`, `setgid()` and `setuid()`, and did not warn when the resulting credentials remained root. Checking both resulting UIDs in the shared helper gives daemon callers one consistent invariant without duplicating warnings. Non-root starts retain their existing behaviour, and the intentionally privileged `upsmon -p` path retains its existing warning without receiving another one.

Validation:

- GCC 16.2.1 built `common.o`, `upslog`, `upsmon`, `upsd` and `dummy-ups` successfully. Clang 22.1.8 built the same targets with `-Weverything -Werror`.
- Before the change, a disposable root namespace running `upslog -u root` completed with no root warning. The candidate emitted exactly one `Warning: running as root (UID=0 EUID=0)`.
- The equivalent non-root run emitted no root warning. A bounded `upsmon -p` run retained exactly one existing intentional-root warning and emitted no new shared warning.
- The focused root regression passed, and `make -C tests check` passed all 9 tests.
- `git diff --check`, the shell syntax check and `make stylecheck` passed. A supplementary non-ASCII scan found only an unchanged character outside the diff.
- `distcheck-light` created and built the clean source archive, including the new regression, then failed at the final spellcheck on unchanged `docs/man/apcmicrolink.txt` text.
- The behavioural checks cover the POSIX implementation. The WIN32 branch is unchanged, and no physical UPS hardware was required or tested.

AI assistance: OpenAI Codex gpt-5.6-sol at high reasoning was used for investigation, implementation and validation. The human contributor remains responsible for the change.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -293,6 +293,8 @@ https://github.com/networkupstools/nut/milestone/13
       not do so for single-driver runs -- addressed with this release. [#3302]
 
  - common code:
+    * POSIX daemons now warn if their real or effective UID remains zero after
+      the common credential switch. [issue #3471, PR #3609]
     * Refactored `common::background()` method used by numerous NUT daemons
       to handle parent and child code paths by separately addressable methods.
       This got used in the shared drivers `main` code file to preclude losing
--- a/common/common.c
+++ b/common/common.c
@@ -983,6 +983,8 @@ void become_user(struct passwd *pw)
 	/* if we can't switch users, then don't even try */
 	intmax_t initial_uid = getuid();
 	intmax_t initial_euid = geteuid();
+	intmax_t final_uid;
+	intmax_t final_euid;
 
 	if (!pw) {
 		upsdebugx(1, "Can not become_user(<null>), skipped");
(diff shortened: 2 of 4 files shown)


jimklimov (approved):
LGTM, but a NEWS.adoc entry would be welcome.

user01010111:
Thanks Jim — I’ve added a concise NEWS entry under common code. I also corrected the new regression’s build-tree wiring to honour `$(EXEEXT)` and pass the exact `upslog` path, and made it use the current UID-0 account name rather than assuming it is `root`. The focused GCC build/test, UID-0 namespace regression, Automake regeneration and distribution-membership checks pass.

PR #3610, opened 2026-09-01
snmp-ups: prefer exact outlet mappings

### LINKED ISSUE #3360: snmp-ups: not checking for possibly NULL returns before dereference and use
https://github.com/networkupstools/nut/blob/c6fc9b27300d19e7109796cf696fa5d4c8fe9aa0/drivers/snmp-ups.c#L4116-L4124
* https://github.com/networkupstools/nut/security/code-scanning/1608 partially uncovers the issue, but not all implications

user01010111:
## Summary

Fixes #3360.

`snmp-ups` supports mapping tables which combine exact `outlet.*` entries
with numbered outlet templates. `su_setOID()` previously sent every outlet
name through template parsing before consulting exact table entries. A
registered outlet-shaped name with no matching template could therefore
leave `su_find_info()` returning `NULL` and pass that result to
`instantiate_info()`.

The same ordering bypassed exact server-side writable values and commands,
while cleanup inferred allocation ownership from the variable-name prefix.
This changes lookup to prefer exact mappings, returns the existing unknown
status when a numbered template is absent, handles writable server-side
values without an SNMP OID through `dstate`, and records whether an entry was
instantiated so only dynamically owned copies are freed.

`upsd` normally forwards only registered writable variables and commands;
arbitrary unpublished outlet-name parser hardening is outside this patch.

## Validation

- A focused hardware-free setvar/instcmd regression links the real `snmp-ups`
  implementation with the existing mock driver and a minimal outlet mapping.
- On the unpatched base, the supported exact `outlet.desc` reproduction exits
  139. GDB stops at the NULL `su_info_p->dfl` dereference after the template
  lookup fails. The candidate returns `STAT_SET_HANDLED` and the new value is
  available through `dstate`.
- A missing numbered outlet template returns `STAT_SET_UNKNOWN` without a
  NULL dereference.
- A valid numbered outlet template and an exact outlet instant command reach
  the SNMP set path and return the expected failure for the deliberately
  invalid test OIDs. An unavailable exact mapping returns the existing
  unknown status, and cleanup completes normally.
- GCC 16.2.1 and Clang 22.1.8 warning-as-error builds passed for the real
  `snmp-ups` driver and focused test.
- The focused test passed under GCC, Clang, AddressSanitizer and
  UndefinedBehaviorSanitizer with no candidate findings.
- LeakSanitizer could not run under the host's ptrace restrictions, so leak
  detection was disabled for the ASan/UBSan run rather than claimed as
  accepted.
- The configured driver test suite passed 1/1.
- The non-ASCII source check, trailing-whitespace and final-newline checks,
  and `git diff --check` passed.
- The normal `make distcheck-light` packaged the candidate and included all
  changed files, then stopped on pre-existing spellcheck findings in unchanged
  `NEWS.adoc` and `docs/man/apcmicrolink.txt`.
- The `make SPELLCHECK_ERROR_FATAL=no distcheck-light` run built and extracted
  the distribution, and the packaged `snmp-ups-setvar-test` passed 1/1.
  Overall distcheck remained blocked by unchanged `cppunittest` because the
  sandbox denied `socket(AF_INET, SOCK_STREAM, 0)` with `EPERM`. No candidate
  file was implicated.

No physical SNMP or UPS hardware is required for this regression.

## AI assistance

OpenAI Codex `gpt-5.6-sol` at high reasoning was used for repository
analysis, implementation, review, drafting and validation. The human
contributor remains responsible for reviewing and validating the change.


jimklimov (approved):
Looks reasonable, but a NEWS.adoc entry would be welcome.

CC @arnaudquette-eaton - would you be able to test this against some SNMP ePDUs?

jimklimov:
The mentioned "exits 139" I suppose means a SIGSEGV (128 + 11) which is usually linked to a NULL pointer dereference.

biergaizi:
This PR commits `drivers/snmp-ups-setvar-test.c` into the tree, is that expected?

NUT doesn't have a central location for unit tests, and most drivers are not designed for unit tests, the build system doesn't do unit tests for individual regressions, so an ad-hoc unit test `drivers/snmp-ups-setvar-test.c` for temporarily testing a single bug doesn't fit to the existing code organization.

jimklimov:
Thanks @biergaizi - I somehow missed the location. There is precedent for piecemeal parser checks under `tests`, so the file and Makefile change could make better sense in that directory.

user01010111:
Thanks @biergaizi and @jimklimov. This was an agent oversight: it concentrated on proving the runtime behaviour and regression, but did not compare the nearest repository-placement and release-note precedents before preparing the original change. The test should have been placed under `tests` from the outset.

The branch has now been updated to:

- move the unchanged `snmp-ups-setvar-test.c` regression to `tests/`;
- move its Automake integration to `tests/Makefile.am`, following the existing linked-driver-source pattern;
- remove the driver-local test wiring from `drivers/Makefile.am`; and
- add the requested `NEWS.adoc` entry for issue #3360 and this PR.

The regression remains focused on the affected `su_setOID()` paths: exact server-side outlet variables and commands, numbered outlet templates, missing templates, and unavailable mappings.

Validation after the relocation:

- focused Automake regression: 1/1 passed with GCC;
- direct regression builds passed with GCC and strict Clang;
- GCC ASan/UBSan run passed;
- the real `snmp-ups` target built with GCC and Clang;
- distribution generation includes the regression under `tests/`, with no copy under `drivers/`; and
- whitespace, non-ASCII and final-newline checks passed.

The regression source itself is byte-for-byte unchanged. Thank you for catching the organisational mismatch.

user01010111:
I have prepared a focused follow-up to strengthen the regression and reduce
its compilation scope.

The test now verifies the exact OID, Net-SNMP type, value and call count sent
for successful and failed setters. It also checks successful dstate updates
and confirms that rejected mappings make no device-set call.

The focused target now compiles only `snmp-ups.c` and
`snmp-ups-helpers.c`, supplying `nut_snmp_set()` at the existing public link
boundary. Production builds retain the real setter and complete MIB registry.
This removes 34 unrelated MIB translation units from the focused target. I did
not measure a timing improvement, and this change does not claim to resolve
the pending AppVeyor timeout.

Validation passed with the focused GCC Automake test (1/1), a strict Clang
build and run, sanitizer runs, production-driver builds, and a freshly
extracted distribution archive. Source whitespace and non-ASCII checks also
passed. No physical SNMP/ePDU hardware was available, and upstream CI has not
run against this local follow-up.

jimklimov:
Curious, this one would be fun to merge with DMF code base, it taps into similar `ifdef` locations ;)

user01010111:
> Curious, this one would be fun to merge with DMF code base, it taps into similar `ifdef` locations ;)

Hey Jim.. Yes, I can see the overlap around the mapping registry guards. Unless you disagree, I'd probably prefer to keep this PR focused and handle the FTY integration separately. Would you like me to draft a follow-up PR to reconcile the guards and verify the mapping ownership changes with the DMF build?

jimklimov:
No, thanks, I'll tackle that later.

Issue #3509, opened 2026-06-27
NUT-Monitor-py3qt5: Add D-Bus sleep handler for automatic reconnect after system wake-up


developersorli:
Issue Description
Problem Description

When the NUT-Monitor-py3qt5 client loses connection to the NUT server (e.g., after system suspend/resume, network interruption, or USB reconnection), it does not automatically reconnect. The client continues to show errors in the status bar and notifications, but the user must manually restart the application to restore functionality.

Affected component: NUT-Monitor-py3qt5 (Python 3 + Qt5 version of the GUI client)

Expected behavior: The client should automatically detect connection loss and attempt to reconnect to the NUT server without user intervention.
Proposed Solution

I have implemented a patch that adds an automatic reconnection mechanism to NUT-Monitor-py3qt5. When a connection error is detected during polling, the client:

    Stops the polling timer

    Waits 2 seconds (to allow network/USB to stabilize)

    Attempts to reconnect using the current connection parameters (host, port, login, password)

    Verifies that the UPS is still available on the server

    Restores the GUI state and resumes polling

The patch adds a new method __attempt_reconnect() and a flag __reconnect_attempted to prevent multiple simultaneous reconnection attempts.
The Patch
```python
diff
--- scripts/python/app/NUT-Monitor-py3qt5.in
+++ scripts/python/app/NUT-Monitor-py3qt5.new
@@ -812,6 +812,7 @@
 
     __parent_class = None
     __stop_thread  = False
+    __reconnect_attempted = False
 
     def __init__( self, parent_class ) :
         threading.Thread.__init__( self )
@@ -939,8 +940,82 @@
                 self.__parent_class._interface__widgets["status_icon"].setToolTip( status_text )
 
             except :
-                self.__parent_class.gui_status_message( _("Error from '{0}' ({1})").format( ups, sys.exc_info()[1] ) )
-                self.__parent_class.gui_status_notification( _("Error from '{0}'\n{1}").format( ups, sys.exc_info()[1] ), "warning.png" )
+                # Connection error detected — attempt automatic reconnection
+                if not self.__reconnect_attempted and self.__parent_class._interface__connected:
+                    self.__reconnect_attempted = True
+                    self.__parent_class.gui_status_message( _("Connection lost — attempting to reconnect...") )
+                    self.__parent_class.gui_status_notification( _("Connection lost"), "warning.png" )
+                    
+                    # Stop the current timer
+                    self.__timer.stop()
+                    
[...]

PR #3616, opened 2026-09-04
NUT-Monitor: fix AttributeError (and abort) in automatic reconnection [#3509, #3530]

executorrr86:
## Summary

The automatic reconnection added in #3530 crashes instead of reconnecting.
`gui_updater.__attempt_reconnect()` reads `self.__widgets[...]`, but that
attribute belongs to the `interface` class; inside `gui_updater` Python
name-mangles it to `_gui_updater__widgets`, which does not exist. The
surrounding lines already use `self.__parent_class._interface__widgets`,
so this PR does the same on the one remaining line, in the py3qt6, py3qt5
and py2gtk2 variants.

## How to reproduce (master, 59fad7a97)

1. Start `NUT-Monitor-py3qt6` connected to a local `upsd`.
2. `systemctl restart nut-server` (or any way of closing the client socket).
3. The reconnection timer fires and the process aborts:

```
  File ".../NUT-Monitor-py3qt6", line 1381, in __attempt_reconnect
    if self.__widgets["ups_authentication_check"].isChecked() :
AttributeError: 'gui_updater' object has no attribute '_gui_updater__widgets'
```

With PyQt an exception escaping a slot becomes `qFatal()`, so the GUI dies
with SIGABRT (core dump present) rather than showing an error.

With the fix, the same steps reconnect within ~2 s (new client socket on
`upsd`, status icon back to on-line, tooltip refreshing).

## Tested on

Arch Linux, Python 3.14, PyQt6 6.11 / Qt 6.11.2, `nut` 2.8.5 server,
CyberPower-based (Salicru) UPS over `usbhid-ups`. Only the py3qt6 variant
was run; py3qt5 and py2gtk2 carry the identical line and get the identical
change.

## Checklist

- [x] Described the changes in the PR submission
- [x] Single functional commit, no style changes
- [x] AI tooling disclosed: the diagnosis, fix and test were done with
      Claude Code (Anthropic) and verified by me on real hardware
- [x] Commit is Signed-off-by (DCO)

### DIFF
--- a/scripts/python/app/NUT-Monitor-py2gtk2.in
+++ b/scripts/python/app/NUT-Monitor-py2gtk2.in
@@ -1301,7 +1301,7 @@ class gui_updater( threading.Thread ) :
             ac = PyNUT.AuthConf.getAuthConf(host=host, port=port, user=login)
             if ac:
                 ac = ac.clone()
-                if self.__widgets["ups_authentication_check"].get_active() :
+                if self.__parent_class._interface__widgets["ups_authentication_check"].get_active() :
                     # If the user entered something, honor that.
                     # Otherwise, re-use what we suggested.
                     ac.user = login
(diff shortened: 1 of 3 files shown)


jimklimov:
Thanks! Not sure how this slipped by - probably had a number of manual test runs that succeeded, and then assumed something stays that way as I (and AI) iterated...

jimklimov:
Bug and fix verified interactively. Thanks again!

PR #3627, opened 2026-09-07
PyNUT: restore socket cleanup in disconnect()

### LINKED ISSUE #3619: PyNUT disconnect() leaves the socket open on a retained client
At revision `59fad7a97cc67b35f2ae28b4228f92203c25c5d3`, calling `disconnect()` on a client with a socket does not send LOGOUT, close the socket or clear its reference.

The [guard in disconnect()](https://github.com/networkupstools/nut/blob/59fad7a97cc67b35f2ae28b4228f92203c25c5d3/scripts/python/module/PyNUT.py.in#L816-L827) checks `hasattr(self, '__srv_handler')`. Python stores the private attribute as `_PyNUTClient__srv_handler`, so the guard does not recognise it.

This can be checked without connecting to a server. Run from the source root:

```python
from pathlib import Path
import types

class FakeSocket:
    def __init__(self):
        self.sent = []
        self.closed = False

    def sendall(self, data):
        self.sent.append(data)

    def close(self):
        self.closed = True

sock = FakeSocket()
client._PyNUTClient__srv_handler = sock
client.disconnect()
print(sock.closed, sock.sent, client._PyNUTClient__srv_handler is sock)
```

Actual output:

```text
False [] True
```

Expected: LOGOUT is attempted, the socket is closed and the client clears its reference. Checking the actual private attribute should allow the existing cleanup to run; repeated disconnects should remain harmless.

This was checked with an in-memory socket stub. The finding concerns explicit disconnection while the client object remains alive; it does not establish that every object destruction leaks a socket.

user01010111:
Calling `disconnect()` on a retained `PyNUTClient` leaves its socket open:
the literal name passed to `hasattr()` does not undergo Python's private
name mangling. Read the existing class attribute directly so the existing
LOGOUT, close and reference cleanup code runs. Its class-level `None`
default covers deferred and partial initialization; send and close errors
remain independently suppressed.

Add 12 regression tests to the existing Python test script and run them
from `make check`, without requiring an external NUT server. They cover
cleanup order, repeat calls, explicit destructor calls, subclasses,
connection/authentication failures, send/close failures, and real loopback
transmission and closure. The same script continues to run the existing
server integration tests. Add a release-note entry; no public API,
dependency, version or hardware-support change is intended.

Closes: #3619

Validation:

- The final regression suite fails 8/12 tests on unmodified upstream
  `ab687a162ad99d2c687c48f02b498845f69c4d91`, including absent LOGOUT/close
  events and the retained open socket. All 12 pass with the fix.
- Linux ARM64: Python 2.6.9, 2.7.18, 3.5.10, 3.6.15, 3.7.17, 3.8.20,
  3.9.25, 3.10.21, 3.11.16, 3.12.14, 3.13.15 and 3.14.7 pass.
  macOS ARM64: Python 3.9.6 and 3.14.7 pass.
- Supplementary IPv6, verified localhost TLS, failed STARTTLS setup and
  cipher-fallback cleanup tests pass on Linux and macOS.
- Locally built `upsd` and `dummy-ups`: retained plain/TLS clients release
  their sockets and logged-in sessions after explicit/destructor cleanup;
  replacement clients reconnect. The existing Python script passes on
  Python 2.6.9 and 3.11.15. The NUT Python integration harness passes
  4 groups with no failures or skips.
- Generated modules, locally installed wheel and source distribution,
  and staged NUT install pass. Source-package installation and its installed
  tests also pass on Python 2.6.9 using setuptools 36.8.0.
- The broad Linux build, `make check` (including 12 Automake tests),
  `make stylecheck`, `make spellcheck` and `make distcheck-light` pass.
  The final distribution check generates real man pages and HTML and
  exercises the extracted archive's build, tests, install/uninstall and
  cleanup. The four changed source files match that archive exactly.

deterministic socket tests, real localhost TCP/TLS connections, and locally
built `upsd` with simulated `dummy-ups` devices.
Physical UPS models, firmware versions, and hardware combinations were
not tested.

Windows was not executed. Python 2.6 uses the plain socket path because
this module disables SSL on runtimes without `ssl.create_default_context`.

Relevant checklist:

- [x] Concrete behavior, compatibility expectations and limitations described.
- [x] Focused change; existing templates and test/distribution infrastructure used.
- [x] New source comments and release-note text use ASCII.
- [x] NEWS entry added; no driver, manual/API or UPGRADING change required.
- [x] Build, tests, package/install, spelling/style and distcheck-light pass.
- [x] AI use and model disclosed.
- [x] Human contributor review and acceptance of responsibility completed.

AI assistance: OpenAI Codex with gpt-6-astra (x-high reasoning).
the change.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -22,6 +22,10 @@ PLANNED: Release notes for NUT 2.8.6 - what's new since 2.8.5
 
 https://github.com/networkupstools/nut/milestone/13
 
+ - `PyNUTClient.disconnect()` now attempts to send `LOGOUT`, closes the
+   socket and clears its reference while the client remains alive. Cleanup
+   remains harmless when repeated or when socket operations fail. [issue #3619]
+
  - (expected) Dynamic Mapping Files (DMF) feature supported, to allow
    the driver binaries to be built once and data mappings to be loaded
    and modernized on the fly (porting from 42ITy project)
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -22,10 +22,6 @@ PLANNED: Release notes for NUT 2.8.6 - what's new since 2.8.5
 
 https://github.com/networkupstools/nut/milestone/13
 
- - `PyNUTClient.disconnect()` now attempts to send `LOGOUT`, closes the
-   socket and clears its reference while the client remains alive. Cleanup
-   remains harmless when repeated or when socket operations fail. [issue #3619]
-
  - (expected) Dynamic Mapping Files (DMF) feature supported, to allow
    the driver binaries to be built once and data mappings to be loaded
    and modernized on the fly (porting from 42ITy project)
@@ -509,6 +505,10 @@ https://github.com/networkupstools/nut/milestone/13
       query (for protocol version) to verify that handshake succeeded.
       This change impacted also the classic C `libupsclient` library.
       [issue #3387, PR #3402]
+    * `PyNUTClient.disconnect()` now attempts to send `LOGOUT`, closes the
+      socket and clears its reference while the client remains alive. Such
+      cleanup remains harmless when repeated or when socket operations fail.
+      [issue #3619, PR #3627]
     * Updated OpenSSL code paths in the `libupsclient` C library to support
       features earlier only available with NSS builds, like specifying the
       client certificate+key, optionally with password, and pinning expected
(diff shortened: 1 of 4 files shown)


jimklimov:
Regarding failed "slowbuild-run/MATRIX_TAG=gnu99-gnu++98-debian11" scenario: same issue as https://github.com/networkupstools/nut/pull/3628#issuecomment-5603973964

UPDATE: I'm on it now

PR #3628, opened 2026-09-07
PyNUT: use command identifiers in description requests

### LINKED ISSUE #3621: PyNUT GetUPSCommands() requests CMDDESC using Python's bytes representation
At revision `59fad7a97cc67b35f2ae28b4228f92203c25c5d3`, `GetUPSCommands()` on Python 3 sends the bytes representation of each command identifier in its description request.

The [command name is parsed as bytes and interpolated into a text string](https://github.com/networkupstools/nut/blob/59fad7a97cc67b35f2ae28b4228f92203c25c5d3/scripts/python/module/PyNUT.py.in#L1298-L1313). For the advertised command `test.panel.start`, this produces:

```text
GET CMDDESC dummy b'test.panel.start'
```

The expected request is:

```text
GET CMDDESC dummy test.panel.start
```

Run this check from the source root:

```python
from pathlib import Path
import types

class FakeSocket:
    def __init__(self, data):
        self.data = data
        self.sent = []

    def sendall(self, data):
        self.sent.append(data)

    def recv(self, size):
        result, self.data = self.data[:size], self.data[size:]
        return result

sock = FakeSocket(
    b"BEGIN LIST CMD dummy\n"
    b"CMD dummy test.panel.start\n"
    b"END LIST CMD dummy\n"
    b"ERR CMD-NOT-SUPPORTED\n"
)
client._PyNUTClient__srv_handler = sock
print(client.GetUPSCommands("dummy"))
print(sock.sent)
```

Actual output:

```text
{b'test.panel.start': b'test.panel.start'}
[b'LIST CMD dummy\n', b"GET CMDDESC dummy b'test.panel.start'\n"]
```

The in-memory transport records the incorrect command identifier and supplies an error response. The exception handler then substitutes the command name for its description. This fallback is specific to the supplied error response.

Current `upsd` can instead return a `CMDDESC` response containing `"Description unavailable"`, as its [GET CMDDESC implementation](https://github.com/networkupstools/nut/blob/59fad7a97cc67b35f2ae28b4228f92203c25c5d3/server/netget.c#L101-L131) shows. The defect is the incorrect identifier in the request; no live-server response was tested.

The request and response-prefix handling should use the command identifier consistently, without its Python bytes representation. The public dictionary can retain its existing bytes keys and values.

user01010111:
Closes: #3621

On Python 3, `GetUPSCommands()` formats command names parsed from `LIST CMD`
as bytes representations, sending requests such as `GET CMDDESC dummy
b'load.off'`. Current `upsd` consequently returns "Description unavailable"
even though `load.off` has a description.

Decode the command identifier as ASCII for both the description request and
the response-prefix length. The returned dictionary retains its bytes keys
and values, and description errors retain the existing identifier fallback.
Add a request-sensitive regression test under `tests`, run it through
`make check`, and record the fix in `NEWS.adoc`.

Validation:

- The regression fails on unmodified Python 3 source and on a request-only
  fix; the complete fix passes on Python 2.6.9, 2.7.18, and 3.4 through 3.14.
- A fresh localhost `upsd` with `dummy-ups` advertises `load.off`. Against the
  same server, baseline returns "Description unavailable" and the patched
  client returns "Turn off the load immediately". Exact request bytes and
  separate unavailable/error controls were checked; no UPS commands ran.
- Generated-module, explicit-path install/uninstall, wheel-install and
  source-distribution-install checks passed. The existing suite passed
  12/12, plus all six new regression tests.
- `make spellcheck`, changed-line ASCII/whitespace checks, and normal
  `make distcheck-light` passed. The latter exercised the extracted archive's
  build, tests, installation, uninstallation and redistribution, using its
  standard placeholder handling for unselected manual pages.

request-sensitive transport tests, Python runtime compatibility checks,
generated and packaged modules, build and distribution checks, and simulated
localhost integration with upsd and dummy-ups. Physical UPS models, firmware
versions, and hardware combinations were not tested.

AI assistance: OpenAI Codex with gpt-6-astra (x-high reasoning).
the change.

Relevant contribution checklist:

- [x] Described the concrete behavior, compatibility and testing limitations.
- [x] Disclosed AI use and model.
- [x] Followed affected-file style; added source text is ASCII.
- [x] Registered the new test in distribution metadata; distcheck-light passed.
- [x] Added a NEWS entry and passed spelling checks.
- [x] Human contributor has reviewed and accepted responsibility for the patch.
- [x] DCO sign-off included in each contribution commit.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -42,6 +42,10 @@ https://github.com/networkupstools/nut/milestone/13
    to patterns defined in link:docs/nut-names.txt[]
 
  - Fix fallout of development in NUT v2.8.0 through v2.8.5:
+    * PyNUT: `GetUPSCommands()` now requests command descriptions with the
+      advertised identifiers on Python 3 and uses those identifiers when
+      extracting the response. Returned keys and descriptions remain byte
+      sequences. [issue #3621]
     * `nut-scanner` tool updates:
       - Mutexes used in parallelized scans were not properly released on
         failure code paths (impacts likely all 2.8.x until now). [PR #3551]
--- a/configure.ac
+++ b/configure.ac
@@ -3800,6 +3800,7 @@ if test x"${nut_with_nut_monitor}" != xno ; then
     fi
 fi
 
+dnl Options regarding NUT-Monitor GUI app:
 AC_MSG_CHECKING([if we can and should install NUT-Monitor desktop application])
 AS_IF([test x"${nut_enable_configure_debug}" = xyes], [
     AC_MSG_NOTICE([(CONFIGURE-DEVEL-DEBUG) nut_with_nut_monitor-4: ${nut_with_nut_monitor}])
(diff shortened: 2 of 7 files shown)


jimklimov:
You are on a roll again, now that the previous PR merge train has left the station, and NUT CI farm is more available to queries :)

jimklimov:
Curious: this failed on just one build combo, for a fairly valid reason:
```
  File "../../../tests/pynut-commands-test.py", line 5, in <module>
    import unittest
ModuleNotFoundError: No module named 'unittest'
make[5]: *** [Makefile:2508: pynut-commands-test] Error 1
```

On the CI side, I can add the module so we would test the feature everywhere, but on the project source side - the module should be listed in build dependencies I think (`setup.py.in` or `tox.ini`? maybe with some keyword for test-not-runtime dependency?)

And perhaps the presence of the module should be checked in `configure.ac` and drive decisions in `Makefile.am`, similar to how Qt5/6 are detected, etc.
* UPDATE: This change landed into existing `scripts/python/module/test_nutclient.py.in`, at least in another PR with same problem, so probably the check should be in it, not the autoconf/automake recipes. Just `try` to import the module, and run/skip test cases based on the outcome. Or mixed, differently handled here and there.

So maybe until this is handled, I would NOT update that build agent, so we would know builds/checks no longer crash because of this.

UPDATE: I'm on it now...

Issue #1306, opened 2022-02-18
Driver "discovery/dumping" mode escapes some chars that upsc client does not


jimklimov:
This causes needless differences in data walks collected by different methods; not sure if there are more issues other than cosmetic inconvenience:
````
/tmp/nutdev13-master.dump.txt:device.1.outlet.7.desc: Server input\#1
/tmp/nutdev13-master.dump.txt:device.2.outlet.7.desc: Server input\#2
/tmp/nutdev13-master.txt:device.1.outlet.7.desc: Server input#1
/tmp/nutdev13-master.txt:device.2.outlet.7.desc: Server input#2
````

@aquette : any thoughts, please? :)

**Comment by jimklimov** [comment]
Trying to reproduce the issue, I added such outlet names into a dump-file:

```
~/nut $ grep Outlet data/epdu-managed.dev
outlet.1.desc: Outlet #1
outlet.2.desc: "Outlet \#2"
outlet.3.desc: "Outlet #3"
outlet.4.desc: Outlet \#4
```

Dumping is odd:
```
~/nut $ ./drivers/dummy-ups -a UPS2 -d1 | grep Outlet
Running as foreground process, not saving a PID file
outlet.1.desc: Outlet
outlet.2.desc: Outlet \#2
outlet.4.desc: Outlet \#4
```
* Reasonably the trailing `#1` is lost as a comment
* Less understandable why the properly quoted `"Outlet #3"` is lost (also in proper driver run, below)

Via `upsc` (hashes not escaped, quotes not added):
```
~/nut $ ./clients/upsc UPS2@localhost:12345 | grep Outlet
outlet.1.desc: Outlet
outlet.2.desc: Outlet #2
outlet.4.desc: Outlet #4
```

jimklimov:
Possibly the correct solution to this is adding an option for `upsc` to provide output in a format unambiguously useful for DDL and/or feeding back into `dummy-ups`, and updating docs/scripts to suggest that. Any added escaped markup may be a surprise for existing consumers (scripts etc.), so `upsc` should not change default behavior.

Also revise parser use in `dummy-ups`: quoted `"Outlet #3"` looks like it should work. Is such spelling ok for configs?..

Issue #607, opened 2018-09-04
Clarify documentation on conf file parser with respect to quotes and comments


clepple:
https://github.com/networkupstools/nut/blob/v2.7.4/common/parseconf.c#L319-L327

We should probably clarify that even inside double-quoted strings, a `#` needs to be preceded by a backslash (`\`).

This may also affect GUI tools that generate `*.conf` files from arbitrary user input.

https://alioth-lists.debian.net/pipermail/nut-upsuser/2018-September/011227.html

PR #3629, opened 2026-09-07
PyNUTClient: decode NUT values and escape authentication arguments

### LINKED ISSUE #3620: PyNUT GetUPSList() rejects descriptions containing escaped quotes
At revision `59fad7a97cc67b35f2ae28b4228f92203c25c5d3`, `GetUPSList()` raises `ValueError` when an ordinary UPS description contains an escaped quote.

The [network protocol documentation](https://github.com/networkupstools/nut/blob/59fad7a97cc67b35f2ae28b4228f92203c25c5d3/docs/net-protocol.txt#L42-L45) permits embedded quotes escaped with backslashes. However, [GetUPSList()](https://github.com/networkupstools/nut/blob/59fad7a97cc67b35f2ae28b4228f92203c25c5d3/scripts/python/module/PyNUT.py.in#L1213-L1216) splits the response on every quote and expects exactly two parts.

The server's [LIST UPS implementation](https://github.com/networkupstools/nut/blob/59fad7a97cc67b35f2ae28b4228f92203c25c5d3/server/netlist.c#L239-L242) also encodes descriptions with `pconf_encode()`, so the example follows its output formatting.

For example, this is a valid response:

```text
BEGIN LIST UPS
UPS dummy "Office \"Main\" UPS"
END LIST UPS
```

Run this check from the source root:

```python
from pathlib import Path
import types

class FakeSocket:
    def __init__(self, data):
        self.data = data

    def sendall(self, data):
        pass

    def recv(self, size):
        result, self.data = self.data[:size], self.data[size:]
        return result

response = (
    b'BEGIN LIST UPS\n'
    b'UPS dummy "Office \\"Main\\" UPS"\n'
    b'END LIST UPS\n'
)
client._PyNUTClient__srv_handler = FakeSocket(response)
print(client.GetUPSList())
```

Actual result:

```text
ValueError: too many values to unpack (expected 2, got 4)
```

Expected:

```python
{b"dummy": b'Office "Main" UPS'}
```

The control response with `"Office UPS"` succeeds. This was checked using an in-memory transport, without a running server or UPS.

The description needs parsing according to the documented escaping rules while preserving the existing bytes keys and values.

user01010111:
Descriptions containing escaped quotes can make `GetUPSList()` fail, while
other Python readers truncate quoted values or retain wire escapes. Decode
NUT tokens once in a shared private parser, preserving existing byte/string
return types and the command-description error fallback. Use the same
grammar for `nutauth.conf` values and INCLUDE filenames, and escape literal
usernames and passwords before sending them. Double quotes group tokens;
single quotes remain literal characters.

Encode `upsd` variable and command descriptions with the existing
`pconf_encode()` helper before returning DESC and CMDDESC. Keep the existing
response-size limit. Add deterministic tests and a localhost integration
case to the existing test infrastructure. Update NEWS in its client-library
and data-server sections, UPGRADING, the authentication manual and Python
README; add the documented API names to the spelling dictionary.

Closes: #3620

This implements the agreed Python/server scope from the discussion in this
PR. The C quoted-hash correction (#3633), Perl parsing (#3634)
and NDE configuration fixes (#3635) are separate contributions. Command-name formatting reuses the existing
#3628 change; its ownership remains with that PR.

Validation:

- The new 13-test protocol/configuration suite produces six failures and
  one error on the pre-feedback module. Existing ordinary controls pass.
  The patched module passes all 35 tests: 16 UPS-list tests, 13 shared
  protocol/configuration tests and six existing command tests.
- Linux ARM64: Python 2.6.9, 2.7.18, 3.4.10, 3.5.10, 3.6.15, 3.7.17,
  3.8.20, 3.9.25, 3.10.21, 3.11.16, 3.12.14, 3.13.15 and 3.14.7 pass.
  Native macOS ARM64 tests pass on Python 3.9.6 and 3.14.7.
- The Python tokenizer matches the corrected C parser token-for-token on
  352 valid ASCII cases, including 3,304 split-boundary checks and text
  input checks. Python deliberately preserves bytes where C configuration
  parsing filters high bytes; no new Unicode conversion policy is added.
- Locally built upsd and dummy-ups pass the Python integration group on
  Linux and macOS: eight passed, zero failed or skipped on each. Existing
  TLS checks and the new plaintext punctuation fixture cover decoded
  values/descriptions, INCLUDE credentials, ordinary controls, successful
  secondary LOGIN and incorrect-password rejection. No power commands are
  issued by the new fixture.
- Configured wheel, source distribution and staged module each pass all
  35 tests on Linux Python 3.11.15; installed module bytes match the
  configured source. The direct CI packaging recipe also builds successfully.
- Native build and eight Automake tests pass. Linux build, 12 Automake
  tests, documentation, stylecheck, spellcheck and distcheck-light pass.
  The final distribution check uses real manual pages and exercises the
  archive's build/test/install/uninstall/cleanup lifecycle. Changed source
  files are verified against that archive.

deterministic protocol/configuration tests, corrected-C parser comparisons,
package/install tests and localhost upsd with simulated dummy-ups, including
the existing TLS integration checks.
Physical UPS models, firmware versions, and hardware combinations were
not tested.

Windows and other operating systems were not executed. Python 2.6 uses
the plain socket path. Simulated tests do not establish physical-device
compatibility. The existing server response-size ceiling and configuration
include lookup/precedence remain unchanged. Users should remove manual
unescaping workarounds and replace single quotes previously used to group
configuration values with double quotes, as described in UPGRADING.

Relevant checklist:

- [x] Concrete behavior, compatibility expectations and limitations described.
- [x] Focused changes; maintained templates and existing test infrastructure used.
- [x] New source and documentation text uses ASCII.
- [x] NEWS placement, UPGRADING, authentication manual and Python README reviewed and updated.
- [x] Build, tests, package/install, spelling/style and distcheck-light pass.
- [x] AI use and models disclosed.
- [x] Human review of this revised patch and acceptance of responsibility completed.
- [x] Human DCO authorization supplied for this revised contribution.

AI assistance: OpenAI Codex with gpt-6-astra (x-high and high reasoning)
and gpt-daybreak-blue-latest (high reasoning).
the change.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -494,6 +494,10 @@ https://github.com/networkupstools/nut/milestone/13
       the overflows it addressed) in normal device interactions. [#3588]
 
  - NUT client libraries:
+    * `PyNUTClient`: fixed `GetUPSList()` rejecting UPS descriptions containing
+      escaped quotes. Description escapes are decoded once, preserving byte
+      keys and values and allowing `GetUPSNames()` to work with these entries.
+      [issue #3620]
     * Complete support for actions documented in `docs/net-protocol.txt`
       was implemented in C++, Python and PERL bindings in-tree, and for Java
       in link:https://github.com/networkupstools/jNut[jNut] nearby. Among
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -498,10 +498,12 @@ https://github.com/networkupstools/nut/milestone/13
       the overflows it addressed) in normal device interactions. [#3588]
 
  - NUT client libraries:
-    * `PyNUTClient`: fixed `GetUPSList()` rejecting UPS descriptions containing
-      escaped quotes. Description escapes are decoded once, preserving byte
-      keys and values and allowing `GetUPSNames()` to work with these entries.
-      [issue #3620]
+    * `PyNUTClient`: fixed parsing of escaped descriptions and values in
+      server replies, including `GetUPSList()` and its consumers. Shared
+      parsing also handles `nutauth.conf` values, comments and included
+      filenames using NUT syntax. Authentication arguments are escaped
+      before transmission. Existing response types are preserved; single
+      quotes in configuration values are now literal. [issue #3620, PR #3629]
     * Complete support for actions documented in `docs/net-protocol.txt`
       was implemented in C++, Python and PERL bindings in-tree, and for Java
       in link:https://github.com/networkupstools/jNut[jNut] nearby. Among
@@ -614,6 +616,9 @@ https://github.com/networkupstools/nut/milestone/13
     * Added a `clean_exit()` handler similar to that in `upsmon`. [PR #3499]
 
  - `upsd` data server updates:
+    * Escape variable and command descriptions from `cmdvartab` when sending
+      `DESC` and `CMDDESC` responses, so embedded quotes, backslashes and
+      hashes are decoded correctly by clients. [PR #3629]
     * If we hit "Too many open files" during configuration reload, close
       the oldest client connection and retry. [issue #3365]
     * If the `MAXCONN` requested in the configuration file exceeds the OS
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -42,6 +42,10 @@ https://github.com/networkupstools/nut/milestone/13
    to patterns defined in link:docs/nut-names.txt[]
 
  - Fix fallout of development in NUT v2.8.0 through v2.8.5:
+    * PyNUT: `GetUPSCommands()` now requests command descriptions with the
+      advertised identifiers on Python 3 and uses those identifiers when
+      extracting the response. Returned keys and descriptions remain byte
+      sequences. [issue #3621]
     * `nut-scanner` tool updates:
       - Mutexes used in parallelized scans were not properly released on
         failure code paths (impacts likely all 2.8.x until now). [PR #3551]
--- a/UPGRADING.adoc
+++ b/UPGRADING.adoc
@@ -31,6 +31,19 @@ command line, in order to quickly pick up any other removed option flags.
 Changes from 2.8.5 to 2.8.6
 ---------------------------
 
+- `PyNUTClient` now decodes NUT escapes in server descriptions and values
+  exactly once. Applications which worked around the old behavior by
+  removing escape characters themselves should stop doing so. The existing
+  bytes/string return types are unchanged. [PR #3629]
+
+- Python's `nutauth.conf` reader now follows NUT quoting rules: single
+  quotes are literal characters, not string delimiters. Replace single
+  quotes used for grouping with double quotes, and quote or escape spaces
+  within values. Backslashes introduce literal characters rather than
+  being retained, and unquoted hashes begin comments. Pass unescaped
+  usernames and passwords to `PyNUTClient`; the client handles their wire
+  escaping. [PR #3629]
+
 - PLANNED: Keep track of any further API clean-up?
 
 - Potentially a breaking change for C++ clients that rushed to use the new
--- a/docs/man/nutauth.conf.txt
+++ b/docs/man/nutauth.conf.txt
@@ -86,6 +86,13 @@ should be double-quoted and/or use escape sequences, like in other NUT files.
 
 Blank lines and characters after an un-quoted hash (`#`) are ignored.
 
+Single quotes are literal characters; only double quotes group a value
+containing spaces. A backslash escapes the following character once,
+including a space, quote, backslash or hash. It does not introduce C or
+Python escape sequences. These rules also apply to `INCLUDE` filenames.
+For compatibility with older NUT parsers, escape a hash with a backslash
+even inside double quotes.
+
 Example:
 
 	# Global defaults
--- a/docs/nut.dict
+++ b/docs/nut.dict
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 3827 utf-8
+personal_ws-1.1 en 3830 utf-8
 AAC
 AAS
 ABI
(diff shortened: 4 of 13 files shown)


jimklimov:
Thanks! I think this parser code makes sense as a separate helper method, which should be applied to all strings returned by `upsd` or read from NUT configuration files which are expected to follow similar escaping semantics (e.g. `nutauth.conf`).

user01010111:
> Thanks! I think this parser code makes sense as a separate helper method, which should be applied to all strings returned by `upsd` or read from NUT configuration files which are expected to follow similar escaping semantics (e.g. `nutauth.conf`).

Thanks, Jim. Happy to extract a shared helper and reuse it across the relevant Python readers. Before expanding this PR, could you clarify a few scope points?

- Would you prefer all affected PyNUT response readers and `nutauth.conf` parsing addressed in this PR, or configuration parsing as a follow-up? I’m assuming other language bindings are outside this change.
- Should Python’s configuration parsing match C `parseconf` exactly? Python currently strips single quotes, while C treats them literally. The manual’s wording also suggests `#` inside double quotes is permitted, whereas C requires it to be escaped.
- If end-to-end validation exposes related outbound credential or server-response escaping problems, would you prefer those fixed here or in separate PRs?

To clarify, I’d preserve the existing return types and keep the changes focused on NUT’s escaping semantics.

jimklimov:
Fixes for Python (and C if there is indeed some inconsistency) may well be in this PR. Regarding C, I think there was a bug logged about `#` maybe, and about different outputs of driver internal dumps vs `upsc` #1306 certainly - maybe that is related to this same pain point.

I think PERL binding may suffer similarly, but can be addressed separately based on lessons learned here.
Wondering now if e.g. shell parsing in NDE is impacted (if it at all touches on values that might be so escaped)?..

user01010111:
Thanks, Jim. I traced the related paths and checked them with deterministic parser probes, NDE self-test mode, and a localhost `upsd`/`dummy-ups` setup. The findings are:

- Several Python response readers truncate escaped descriptions or values. `GetUPSCommands()` also formats command names as `b'...'` under Python 3, so fixing escape parsing alone would leave that path broken.
- Python's `nutauth.conf` reader mishandles escapes and trailing comments. `INCLUDE` filenames need escape handling too. Decoded credentials also need correct outbound encoding to reach the server unchanged.
- `upsd` does not re-escape descriptions loaded from `cmdvartab` when sending `DESC` and `CMDDESC`. The localhost test confirmed malformed responses and incorrectly parsed descriptions.
- Perl retains escape characters in returned values, confirming the separate follow-up you suggested.
- NDE is affected: a trailing comment on `port = "auto"` prevents the expected USB classification, and backslash handling can corrupt paths. This is in its `ups.conf` parsing; it does not consume `upsd` descriptions.
- #607 appears to be the earlier hash-escaping issue. The `nutauth.conf` manual's wording about quoted hashes differs from current C behavior, so the documentation should reflect whichever rules we agree on.

I propose keeping the shared Python parsing helper and affected readers, configuration/include parsing, necessary outbound encoding, the Python 3 command-name correction, and the two C response-encoding fixes in this PR, with the relevant tests and documentation updates. Existing Python return types would be preserved. Perl and NDE would follow separately. Does that scope split work for you?

Two compatibility points before implementing:

1. For Python's `nutauth.conf` reader, are you happy to retain its existing acceptance of single-quoted values and literal hashes inside quotes while correcting escapes and comments, rather than tightening those forms to current C syntax?
2. Would you support a separate, focused follow-up allowing literal `#` inside double quotes in the shared C parser, while retaining `\#` support and escaped output for older consumers? That would address the surprising quoted-hash behavior you noted in #1306. I would leave `upsc`'s default output unchanged and keep any optional export mode outside this PR.

No physical UPS hardware was available for testing. These checks covered software behavior and simulated/local integration; physical UPS models, firmware versions, and hardware combinations were not tested.

jimklimov:
Yes, splitting the PRs makes sense.

The C parser breaking for a hash inside quotes (the `"Outlet #3"` example from #1306) is a bug. IIRC that is same code for configs and for values. An escaped character, whether inside or outside quotes, should be handled properly, so `"Outlet \#2"` and `Outlet \#4` should both be internally treated as a string which has a space-hash-number sequence, no backslashes added.
* It is a separate matter that `upsc` historically reports values for easier scripted consumption, so the `KEY: VALUE` lines are not quoted and may include multiple tokens in the VALUE part.
* The recently added `upsc` JSON output mode should handle string escape sequences properly.
* I *think* the configuration parser insists on use of quotes and other escaping for multi-word single tokens, such as descriptions (maybe it accepts unquoted lines with an escaped space as well?) and that is a correct predictable way to go about it for inputs.
* For Python/PERL, follow the syntax used by C. Single quotes may be an overlooked artifact of AI generation of the larger-stroke picture for having something working as the first shot, doesn't make it a non-bug still :)

jimklimov:
In Appveyor build, this still complains:
```
================================
Wed Sep  9 02:24:04 UTC 2026 [INFO] [testcase_sandbox_python_escaping] Check escaped data and credentials with PyNUT
  File "<stdin>", line 19, in <module>
AssertionError: Description unavailable
Wed Sep  9 02:24:04 UTC 2026 [ERROR] [testcase_sandbox_python_escaping] Error: PyNUT regression check failed
================================
Wed Sep  9 02:24:05 UTC 2026 [INFO] Stopping test daemons
Read error : 109
================================
Wed Sep  9 02:24:05 UTC 2026 [INFO] OVERALL: PASSED=26 FAILED=1 SKIPPED=0
      1 [F] testcase_sandbox_python_escaping
Wed Sep  9 02:24:05 UTC 2026 [FATAL] Error: Some test scenarios failed!
make[3]: *** [Makefile:782: check-NIT] Error 1
make[2]: *** [Makefile:1790: check-recursive] Error 1
make[1]: *** [Makefile:2216: check] Error 2
make: *** [Makefile:1163: check-recursive] Error 1
```

jimklimov (approved):

PR #3633, opened 2026-09-08
parseconf: accept literal hashes inside double quotes

user01010111:
A description such as `desc = "Rack #3"`, or a dummy-ups value such as `outlet.3.desc: "Outlet #3"`, triggers a parse error with the original parser and loses the configured value. Accept a literal `#` inside double quotes. Unquoted hashes still start comments, and escaped characters still decode once inside or outside double quotes. Retain the historical error helper under `#if 0` for future parser error handling.

Add portable regression coverage for all three parser entry points and a named NIT testcase checking exact text and decoded JSON through real localhost `upsd`, `dummy-ups` and `upsc`. The group handles sandbox preparation and cleanup. Each query reports its start and distinguishes watchdog expiry from an ordinary process failure, retaining both output streams in failure diagnostics. `NIT_PARSECONF_TIMEOUT` controls the overall per-query watchdog, defaulting to 60 seconds, independently of the connection timeout. Per-client debug settings and missing-Python/JSON skip/failure behaviour are preserved.

Update the configuration, network protocol and dummy-ups documentation, NEWS and UPGRADING. Explain how dummy-ups joins multiple value tokens and show the escaped-space alternative. The parser API, character filters, limits and incomplete-input behaviour remain unchanged. `pconf_encode()` still escapes hashes for older peers; default `upsc` output and driver code are unchanged. No protocol, ABI or driver version bump is needed. This remains independent of #3629.

Related: #607, #1306.

Validation of the original implementation, before this review follow-up:

- Four quoted-hash fixtures failed on unmodified `2caa3c87501a1aba25a3070ad66757470bd1bf03` through line, file and character parsing. All 23 fixtures and boundary/encoder checks passed with the fix; the real NIT regression also failed before the fix.
- Native macOS ARM64 and Linux ARM64 checks covered builds, compiler/language variants, sanitizers, parser integration and distribution/install paths. The original full Linux NIT run passed 31 groups with TLS required, and the original JSON checks passed on Python 2.6.9.

Validation of the review follow-up at `fd8c0f65bcf311f89a62524a61a5d82be145f2a2`:

- Linux GCC/OpenSSL build with warnings and `-Werror`, the C parser test, and the helper self-test pass. The same two-second child succeeds with a five-second watchdog and fails with a one-second watchdog. Checks cover benign stderr, ordinary nonzero exit, captured output, invalid limits and the before/after timeout diagnostics.
- The focused parser NIT group passes with plaintext and with TLS required: five passes, zero failures and zero skips in each run, including all 13 outlet fixtures. The TLS run verifies certificates and records TLS 1.3 with OpenSSL 3.6.4. Named-testcase counters, debug-setting isolation and missing-prerequisite paths also pass focused checks.
- `make stylecheck`, source ASCII and whitespace checks pass. The real user-manual, dummy-ups manual and UPGRADING render, and the changed documents pass their spelling targets.
- `make distcheck-light` with `AM_DISTCHECK_CONFIGURE_FLAGS=--disable-spellcheck` passes the distribution build, all 12 tests and install/uninstall checks. All six follow-up source files match the generated archive. The unmodified default spelling gate rejects `kernel's` and `firmware's` in unchanged NEWS and apcmicrolink documentation; that baseline dictionary failure is retained as a limitation.

The successful socket-dependent checks ran with localhost access enabled. Earlier isolated attempts were denied ordinary local sockets. Python 2 was unavailable for the follow-up runtime checks. No physical UPS hardware, live deployment or native NetBSD/NSS reproduction was exercised in the follow-up. The reported NetBSD delay still needs a native rerun; the improved watchdog diagnostics do not establish its underlying cause. EOF and multiline semantics remain a separate discussion.

AI assistance: the original implementation used OpenAI Codex with `gpt-6-astra` and `gpt-daybreak-blue-latest`. This review follow-up used OpenAI Codex with `gpt-6-astra` at high reasoning. The human contributor authorised the DCO sign-off and remains responsible for the contribution.


jimklimov:
Something not right in new test script (expecting empty `stderr`? Did not drill into it yet):
```
...
Fri Sep 11 02:26:48 UTC 2026 [INFO] [testcase_sandbox_start_drivers_after_upsd] Got output for UPS2:
OL BOOST
SSL handshake done successfully with client ::1
Init SSL with certificate database located at /var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3633/obj/tests/NIT/tmp/etc/cert/upsmon
Connecting in SSL to 'localhost' and look at certificate called 'NIT data server'
Intended to authenticate server localhost:35052 : SUCCESS
SSL handshake done successfully with server localhost
Connected to NUT server localhost in SSL
Fri Sep 11 02:26:58 UTC 2026 [INFO] [testcase_sandbox_start_drivers_after_upsd] Got JSON output for UPS2:
"OL BOOST"
Fri Sep 11 02:26:58 UTC 2026 [INFO] [testcase_sandbox_start_drivers_after_upsd] PASSED: Expected drivers are now responding via UPSD
  File "/var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3633/tests/NIT/parseconf-test.py", line 112, in <module>
    check(sys.argv[1], sys.argv[2])
    ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3633/tests/NIT/parseconf-test.py", line 78, in check
    plain = query(upsc, ups).splitlines()
            ~~~~~^^^^^^^^^^^
  File "/var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3633/tests/NIT/parseconf-test.py", line 64, in query
    raise AssertionError('upsc %r returned %d: %r' %
                         (args, process.returncode, error))
AssertionError: upsc ('dummy@127.0.0.1:35052',) returned -9: b'Init SSL with certificate database located at /var/shm/jenkins-nutci-netbsd-92-amd64/nut_nut_PR-3633/obj/tests/NIT/tmp/etc/cert/upsmon\n'
Fri Sep 11 02:27:56 UTC 2026 [INFO] Stopping test daemons
mainloop: Interrupted system call
Signal 15: exiting
Signal 15: exiting
Signal 15: exiting
Signal 15: exiting

================================
Fri Sep 11 02:27:56 UTC 2026 [INFO] OVERALL: PASSED=26 FAILED=1 SKIPPED=4
   1 [F] testgroup_sandbox_parseconf
   1 [S] testcase_sandbox_cppnit_simple_admin
   1 [S] testcase_sandbox_cppnit_upsmon_master
   1 [S] testcase_sandbox_cppnit_upsmon_primary
   1 [S] testcase_sandbox_cppnit_without_creds
Fri Sep 11 02:27:56 UTC 2026 [FATAL] Error: Some test scenarios failed!
```

UPDATE: Reading the script code, I think the fault was not due to parsing an unexpected "Init SSL..." as was my original theory, but `process.communicate()` for `upsc` returning `-9` (as in `kill -9`?) and then reported in a somehow clumsy manner :)

It would help troubleshooting if the test case reported itself starting (and BTW `testgroup_sandbox_parseconf` should really be a sandbox preparation and a caller of test cases in that box - even if one case so far).

common/parseconf.c
```diff
@@ -283,26 +283,9 @@ static int findeol(PCONF_CTX_t *ctx)
...
 }
 
-/* set up the error reporting details */
-static void pconf_seterr(PCONF_CTX_t *ctx, const char *errmsg)
-{
-	snprintf(ctx->errmsg, PCONF_ERR_LEN, "%s", errmsg);
-
-	ctx->error = 1;
-}
-
```
jimklimov, line 293: I feel uneasy about dropping this method, it was with us for 20+ years. This drop also makes `STATE_PARSEERR` value un-used. There are a couple of other uses for `ctx->errmsg` at least (although those sanity-checkers probably don't want to change `ctx->error` value nor their return codes).

I'd rather hide these removed codes into `#if 0` fences as an example how to revive this functionality should our parser report errors again, or find actual use for them in error handling. We should have some other cases, right?

docs/man/dummy-ups.txt
```diff
@@ -141,8 +141,32 @@ This definition file, specified by the `port` argument in the example above,
...
+	outlet.4.desc: Outlet \#4
+
+These values become `Outlet`, `Outlet #2`, `Outlet #3`, and `Outlet #4`,
+respectively. Older versions of the parser require `\#` even inside
+double quotes. Backslashes quote the next permitted character once, so
+`\n` becomes `n`, while `\\` becomes a single literal backslash.
+A backslash followed by a physical newline continues the logical line.
+An unescaped `=` outside double quotes is a separate token; quote or
+escape it when it is part of a value.
+
```
jimklimov, line 165: Can you please double-check and clarify whether "Outlet #4" (upon reading from such data in e.g. config files) becomes two tokens, while others would be one token. When reading from a `.dev/.seq` file in `dummy-ups`, do we actually inject only the first value token for the key token?

Maybe add an example with a backslash-escaped space as mentioned in `config-notes.txt` now?

common/parseconf.c
```diff
@@ -283,26 +283,9 @@ static int findeol(PCONF_CTX_t *ctx)
...
 {
-	/* user is trying to break us */
-	if (ctx->ch == '#') {
-		pconf_seterr(ctx, "Unbalanced word due to unescaped # in quotes");
-		endofword(ctx);
-
-		/* this makes us drop all the way out of the caller */
-		return STATE_PARSEERR;
-	}
-
```
jimklimov, line 305: CC @aquette @arjendekorte @clepple : do you remember any rationale for rejecting an unescaped hash sign INSIDE double-quoted values (e.g. passwords, descriptions like `"Outlet #4"` etc.)?

How here a "user is trying to break us" ?

common/parseconf.c
jimklimov: One of these discussions about the parser mentioned handling multi-line strings where properly quoted, and `qc_literal()` confirms this is intentional (with removal of the actual newline character from the collected string). I first thought about reporting these as errors in a quoted context, but not sure about that now (the code probably meant to handle lines ending with a backslash+newline escape, no idea if it was about a simple or escaped newline inside the quotes too).

But we could certainly `pconf_seterr()` if there is an EOF while we are inside the quoted string context.

Not sure about other non-ASCII values (unprintable < 32 other than 10==newline, or locale-specific > 127, or potentially unicode) - it is probably a bit too late to start rejecting outright or handling them somehow specially :)

tests/NIT/parseconf-test.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
+# Exact values through configuration, dummy-ups, upsd and upsc.
```
jimklimov, line 2: Similarly to changes in some other PR of this bunch, makes sense to add a comment here that:
* normally the caller (`nit.sh`) selects which `$PYTHON` interpreter runs this script, so we do not bother with `.in` templating
* expected to be, and should remain, compatible with Py2 and Py3

jimklimov (changes requested):

common/parseconf.c
jimklimov: @user01010111 : if we do go through with this, it is a significant enough change to mention also in `UPGRADING.adoc`.

user01010111:
Following up on the [NetBSD test result](https://github.com/networkupstools/nut/pull/3633#issuecomment-5630969369), [fd8c0f65](https://github.com/networkupstools/nut/commit/fd8c0f65bcf311f89a62524a61a5d82be145f2a2) moves execution and result reporting into `testcase_sandbox_parseconf`, with the group handling sandbox preparation and cleanup. Queries now announce and flush their start, distinguish watchdog expiry from ordinary failure, and retain stdout/stderr in failure diagnostics. `NIT_PARSECONF_TIMEOUT` provides a bounded overall query limit, defaulting to 60 seconds; the per-client debug setting is preserved.

The old ten-second watchdog can produce the reported `-9`, but the log does not prove it caused this particular kill or identify the NetBSD/NSS delay. The updated helper checks pass, and the real parser group passes locally with TLS 1.3 and certificate verification: five passes, zero failures and zero skips. A native NetBSD rerun is still needed.

The distribution build/install/uninstall check and all 12 tests also pass with spellcheck disabled; the default gate rejects existing `kernel's`/`firmware's` entries in unchanged files. Changed-document spelling checks pass. The PR description records this exception and separates the original platform checks from the follow-up validation.

Review and implementation assistance: OpenAI Codex (`gpt-6-astra`).

docs/man/dummy-ups.txt
user01010111: Confirmed: unquoted `Outlet \#4` produces two value tokens, `Outlet` and `#4`. `dummy-ups` joins all value tokens with spaces, so it stores `Outlet #4`; it does not discard the second token.

[fd8c0f65](https://github.com/networkupstools/nut/commit/fd8c0f65bcf311f89a62524a61a5d82be145f2a2) explains this and adds `Outlet\ \#4` as the single-token equivalent, plus a real integration fixture for that form. The configuration notes also show escaped spaces in `SHUTDOWNCMD`. The updated examples render correctly and the parser integration checks pass.

common/parseconf.c
user01010111: I cannot establish the original rationale from the code alone, so I would welcome the historical context. The change permits literal hashes inside double quotes while preserving unquoted comments. `pconf_encode()` continues escaping hashes for compatibility with older peers. I have left this question open for the maintainers who know the history.

common/parseconf.c
user01010111: On the [EOF and multiline point](https://github.com/networkupstools/nut/pull/3633#discussion_r3987133624), [fd8c0f65](https://github.com/networkupstools/nut/commit/fd8c0f65bcf311f89a62524a61a5d82be145f2a2) preserves the existing behaviour. The current tests deliberately retain partial-input acceptance, and the character-at-a-time entry point has no EOF signal. I suggest agreeing EOF handling across the parser entry points in a separate follow-up, keeping multiline and character handling unchanged here.

tests/NIT/parseconf-test.py
user01010111: Added beside the shebang in [fd8c0f65](https://github.com/networkupstools/nut/commit/fd8c0f65bcf311f89a62524a61a5d82be145f2a2): `nit.sh` selects `$PYTHON`, and the script must remain compatible with Python 2.6+ and Python 3. The watchdog continues using the older-compatible timer and subprocess APIs. Python 3 helper/integration checks pass.

I also built an isolated CPython 2.6.9 runtime and ran the published helper self-test plus the real parser integration group with that interpreter. The self-test passes; plaintext and certificate-verified TLS each pass with five passes, zero failures and zero skips, covering all 13 fixtures. The TLS run negotiated TLS 1.3. No source changes were needed.

Validation assistance: OpenAI Codex (`gpt-6-astra`, high reasoning).

common/parseconf.c
user01010111: Agreed. [fd8c0f65](https://github.com/networkupstools/nut/commit/fd8c0f65bcf311f89a62524a61a5d82be145f2a2) adds an `UPGRADING.adoc` entry under 2.8.5 to 2.8.6. It explains literal hashes inside double quotes, unchanged unquoted comments, continued `\#` support and the need to retain escaped forms when sharing configuration with older versions. Encoded protocol output still escapes hashes for older peers. The PR description is updated accordingly.

common/parseconf.c
user01010111: Agreed. [fd8c0f65](https://github.com/networkupstools/nut/commit/fd8c0f65bcf311f89a62524a61a5d82be145f2a2) restores `pconf_seterr()` under `#if 0`, preserving the example alongside the existing `STATE_PARSEERR` handling. Active parser behaviour is unchanged by this follow-up.

jimklimov:
For now, deferring the merge so our busy wise men have time to chime in with historic context.

On the technical side, with recent changes - LGTM :)

jimklimov (approved):

common/parseconf.c
clepple: I have no additional information on why parsing works that way - IIRC it predates even the CVS import of the source, which is before I was doing anything other than driver work. It sounds like the new behavior is documented, so I have no specific objections.

jimklimov: Thanks!

PR #3640, opened 2026-09-11
CI: Test NUT_SSL_CLIENT_VALIDATION_VARIANTS in the build matrix

jimklimov:
Follow-up for #1711 and #3329

Now that client validation is possible with both NSS and OpenSSL, and NIT can set up client and server certificates for testing, extend `BUILD_TYPE=default-all-errors*` handling with both variants here.

When this is quick-matrixed, it should sub-variate for non-trivial SSL backends (openssl/nss specifically, or yes/auto lacking those specifics).


jimklimov:
Ubuntu Trusty (14.04) - uncommon OpenSSL antiqueness that did not backfire elsewhere:
```
../../server/netssl.c: In function 'openssl_cert_verify_callback':
../../server/netssl.c:406:24: error: 'X509_V_ERR_HOSTNAME_MISMATCH' undeclared (first use in this function)
   if (san_ok && err == X509_V_ERR_HOSTNAME_MISMATCH) {
                        ^
../../server/netssl.c:406:24: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [upsd-netssl.o] Error 1
```

Probably warrants a `configure.ac` test and/or a label-gate for `ci_build.sh`...

UPDATE: nah, an old `ifdef` may cut it here.

PR #3657, opened 2026-09-17
nut-scanner: record the libupsclient fallback result in nutscan_avail_nut

com6056:
When `get_libname()` can't resolve `libupsclient`, `nutscan_init()` falls back to `lt_dlopen()` by name, and two of those attempts (`tools/nut-scanner/nutscan-init.c:743` and `:757` on current master) store their result in `nutscan_avail_xml_http` instead of `nutscan_avail_nut`. They came in with 1d0a13210, so every release since 2.8.3 has them. The XML/HTTP flag then just follows whatever the `libupsclient` load did. If that load fails, `-M` goes away even though `libneon` loaded fine. If it succeeds after `libneon` failed, `-M` is offered with none of the neon functions loaded. Going by `scan_xml_http.c` a NetXML reply would then call through NULL pointers, though that part isn't tested.

This also wraps the unversioned `libupsclient` attempt in the same `if (!nutscan_avail_nut)` as its neighbours, which changes nothing today since the loader hands back its cached result, and fixes two `#endif` comments naming the wrong macro (the `SOPATH_LIBUPSCLIENT` and `SOPATH_LIBFREEIPMI` blocks).

Tested on Alpine edge with master at fe4e6fec3, with `libupsclient.so*` moved into `/opt/bundle` and that dir listed only in `/etc/ld-musl-x86_64.path`, so `get_libname()` falls through to libtool. Without the `libupsclient.so.7` link there, the old code logs `succeeded to load the library for LibNeon` and then `nut-scanner -a` doesn't list XML. With the link there and `libneon.so*` moved away, the old code logs `XML search disabled.` and lists XML anyway. With this change XML is listed in the first case and not in the second, and the client library still loads through the fallback whenever it can. With everything in its normal place the library loading and `-a` output are the same before and after.

`make spellcheck` and `make distcheck-light` don't go green in an Alpine edge container even on pristine master, since aspell there rejects `kernel's` and `firmware's` in `NEWS.adoc` and `docs/man/apcmicrolink.txt`. This branch fails at the same spot with exactly the same rejected words, and nothing new from the added NEWS entry.

NUT search itself still ends up disabled in the first setup, because the loader caches the first failure and never tries the next name. That's what I raised on #3654 and it needs its own change.

Disclosure per the template, this was written with Claude Code (Anthropic) using Claude Opus 5, which found the bug, made the change, ran the tests above and drafted this text, with extra review passes from Claude Opus 5 and Claude Fable 5.1 subagents. I reviewed all of it.

## General points

  comments or text documents.

- [x] Use of coding helper tools and AI should be disclosed in the commit
  or PR comments (it is interesting to know which ones do a decent job).
  As with other contributions, a human is responsible and thanked for the
  quality and content of the change, and is presumed to have the right to
  post that code to be published further under the project's license terms.

- [x] Especially with involvement of AI, including modern IDE coding aid,
  please be sure to revise that proposed code and documentation changes
  follow NUT code style guide -- this helps portability across the decades
  worth of supported systems. Notably, avoid Unicode characters where ASCII
  text is expected (C sources and headers, manual pages and other `acsiidoc`
  inputs). Particularly AI is keen on adding `mdash` characters instead of
  plain ASCII double-dash (which renders into the long dash where applicable).

- [x] Added a bullet point into `NEWS.adoc`, possibly also `UPGRADING.adoc`
  if there is something packagers or custom-build users should take into
  account (new driver categories, configuration options, dependencies...)


jimklimov (approved):

jimklimov:
My bad, catch good, thanks! :)

Issue #2576, opened 2024-08-02
Thread count in parallel `nut-scanner` should scale down in case of "Too many open files"


jimklimov:
As slightly noted in issue #2575 and in PRs that dealt with parallelized scans in `nut-scanner`, depending on platform defaults and particular OS deployment and third-party library specifics, `nut-scanner` may run out of file descriptors despite already trying to adapt the maximums to `ulimit` information where available.

As seen recently and culminating in commit 2c3a09ef0cbc845d53f603fdf9316c6f0f901979 of PR #2539 (issue #2511), certain libnetsnmp builds can consume FD's for network sockets, local filesystem looking for per-host configuration files or MIB files, for directory scanning during those searches, etc. This is a variable beyond our control, different implementations and versions of third-party code can behave as they please. Example staged with that commit reverted and a scan of a large network range:
````
...
   0.321562     [D5] nutscan_ip_ranges_iter_inc: got IP from range: 172.28.67.254
   0.321597     [D4] nutscan_scan_ip_range_snmp: max_threads_scantype=0 curr_threads=1022 thread_count=1022 stwST=-1 stwS=0 pass=1
   0.321573     [D2] Entering try_SysOID_thready for 172.28.67.253
   0.321667     [D5] nutscan_ip_ranges_iter_inc: got IP from range: 172.28.67.255
   0.321703     [D4] nutscan_scan_ip_range_snmp: max_threads_scantype=0 curr_threads=1023 thread_count=1023 stwST=-1 stwS=0 pass=1
   0.321677     [D2] Entering try_SysOID_thready for 172.28.67.254
   0.321782     [D5] nutscan_ip_ranges_iter_inc: got IP from range: 172.28.68.0
   0.321817     [D4] nutscan_scan_ip_range_snmp: max_threads_scantype=0 curr_threads=1024 thread_count=1024 stwST=-1 stwS=-1 pass=0
   0.321851     [D2] nutscan_scan_ip_range_snmp: Running too many scanning threads (1024), waiting until older ones would finish
   0.321796     [D2] Entering try_SysOID_thready for 172.28.67.255
   0.475060     [D2] Failed to open SNMP session for 172.28.67.147
/var/lib/snmp/hosts/172.28.66.252.local.conf: Too many open files
/var/lib/snmp/hosts/172.28.65.208.local.conf: Too many open files

<blocks on "too many threads" anyway, but skips a number of hosts> 
````

[...]

PR #3673, opened 2026-09-20
nut-scanner: retry SNMP session opens after descriptor exhaustion

user01010111:
Depends on #3649. I am opening this as a draft while that PR is unmerged.
The branch includes its four existing commits followed by this four-file
SNMP recovery change.

I retain SNMP targets whose session open fails with `EMFILE` and retry them one
at a time after the original SNMP scans finish. If a serial attempt still runs
out of descriptors, I report that host and stop retrying it. Other session-open
errors keep their existing handling.

This follows PR #3649 and partially addresses #2576. It uses the existing
admission helpers and does not adjust shared thread limits or other scan types.

I tested the actual scanner worker in an instrumented harness with NetSNMP
5.9.5.2 and a loopback SNMP responder. The baseline missed all five targets under
controlled descriptor exhaustion; this change found all five exactly once after
descriptors were released. Persistent exhaustion terminated with a warning per
host and balanced descriptor and permit counts.

I also ran 124 focused scenarios across semaphore, counter-only and no-thread
Linux builds with address, undefined-behaviour and leak sanitizers. The GNU89
build, project checks and `distcheck-light` passed, including 16/16 test
programmes and staged install/uninstall. Documentation rendering was disabled;
the light distribution check used its standard placeholder manual pages.

AI assistance: OpenAI Codex, including a focused review of worker ownership and
admission accounting.


tools/nut-scanner/scan_ipmi.c
jimklimov, line 1043: Was there a particular reason to remove this (opportunistic) cleanup? Did it break something? Do we potentially lose something with it dropped?

tools/nut-scanner/scan_ipmi.c
```diff
@@ -1051,6 +1032,8 @@ nutscan_device_t * nutscan_scan_ip_range_ipmi(nutscan_ip_range_list_t * irl, nut
 			} /* if: could we "pass" or not? */
 		} /* while */
 
+		free(ip_str);
```
jimklimov, line 1045: Shouldn't we set `ip_str=NULL` here too, erring on the safe side?

tools/nut-scanner/scan_nut.c
```diff
@@ -913,8 +921,13 @@ nutscan_device_t * nutscan_scan_ip_range_nut_authconf(nutscan_ip_range_list_t *
...
 
-			if ((nut_arg = (struct scan_nut_arg*)malloc(sizeof(struct scan_nut_arg))) == NULL) {
+			if (ip_dest == NULL
+			|| (nut_arg = (struct scan_nut_arg*)malloc(sizeof(struct scan_nut_arg))) == NULL
+			) {
 				upsdebugx(0, "%s: Memory allocation error", __func__);
+#if defined HAVE_PTHREAD && (defined HAVE_SEMAPHORE_UNNAMED || defined HAVE_SEMAPHORE_NAMED)
+				nut_scanner_semaphore_release(semaphore, semaphore_scantype, max_threads_scantype);
+#endif
 				free(ip_dest);
```
jimklimov, line 943: Probably should safely set `ip_dest=NULL` here...

tools/nut-scanner/scan_nut.c
```diff
@@ -943,53 +956,24 @@ nutscan_device_t * nutscan_scan_ip_range_nut_authconf(nutscan_ip_range_list_t *
...
-# endif /* HAVE_PTHREAD_TRYJOIN */
-					break;
-				}
-				else {
-					thread_array = new_thread_array;
+			{
+				int ret = nut_scanner_thread_create(&thread_array, &thread_count,
+					list_nut_devices_thready, (void *)nut_arg);
+				if (ret != 0) {
+					free_nut_arg(nut_arg);
```
jimklimov, line 976: Set `nut_arg=NULL` here?

tools/nut-scanner/scan_nut.c
```diff
@@ -530,12 +544,12 @@ static void * list_nut_devices_thready(void * arg)
...
-		free(target_hostname);
-	if (hostname)
+	if (hostname) {
 		free(hostname);
-	if (nut_arg)
-		free(nut_arg);
+	}
+	if (nut_arg) {
+		free_nut_arg(nut_arg);
+	}
```
jimklimov, line 555: Set `nut_arg=NULL` here to err on the safe side? Maybe propagate it as `**` pointer so all owners/observers end up with NULL when they should?

tools/nut-scanner/scan_snmp.c
```diff
@@ -1402,24 +1435,27 @@ nutscan_device_t * nutscan_scan_ip_range_snmp(
 		} /* if: could we "pass" or not? */
 	} /* while */
 
+	free(ip_str);
```
jimklimov, line 1448: Shouldn't we set `ip_str=NULL` here too, erring on the safe side?

tools/nut-scanner/scan_xml_http.c
```diff
@@ -794,6 +775,8 @@ nutscan_device_t * nutscan_scan_ip_range_xml_http(nutscan_ip_range_list_t * irl,
 			} /* if: could we "pass" or not? */
 		} /* while */
 
+		free(ip_str);
```
jimklimov, line 788: Shouldn't we set `ip_str=NULL` here too, erring on the safe side?

jimklimov (changes requested):
Posted a few specific questions here, mostly about erring on the safe side while working with memory, and will post some casual fixes to the code too.

General questions:
* Why is this refactor not applied to serial port scanning? IIRC it used semaphores too?..
* Why is the fallback (to follow up with overflowing IP addresses) only applied to SNMP and not other protocols? The idea does seem sane...

Issue #1518, opened 2022-07-27
Clean up configure.ac - move more tests to m4 files


jimklimov:
This is generally desirable, but more so on Windows branch that added many more tests which clutter general configure logic

Issue #3108, opened 2025-10-02
CI: Allow to re-use findings cached by earlier `configure` script runs


jimklimov:
Currently all local and CI builds safely default to not using cached findings about method availabilities, etc.

We do however suffer long build times in general, and prohibitively long on AppVeyor and sometimes CircleCI (time-limited VMs) in particular.

It was checked locally that running with different `configure` options and trusting older `config.cache` is counter-productive (e.g. libusb 1.0 vs 0.1 method availability records are misleading). However, for repetitive runs of same configuration in same environment this feature can be helpful.

* Detecting configuration/packaging changes vs. a previous run, in order to invalidate an old `config.cache` file when it is obsolete, could be a good bonus feature.
* Alternately, add optional caching-or-not to the autotools methods used in our m4 sub-scripts for specific dependencies. so some builders can re-evaluate those items always and others would use them from cache, while standard libc method availability there etc. would be always cached if possible.

PR #3675, opened 2026-09-20
configure: move semaphore checks into an M4 macro

user01010111:
Move the semaphore checks from `configure.ac` into `NUT_CHECK_SEMAPHORES`, preserving the probe body and execution order. Add a macro-reference entry and `SEMLIBS` to the spelling dictionary. This is an incremental cleanup towards #1518.

Validation against `a66c00901a31` on Linux and native Windows (MSYS2/MinGW64):

- Autoconf 2.73 produces byte-identical `configure` and `include/config.h.in` files before and after the extraction on both platforms.
- Seven configuration comparisons on each platform match, covering threading `auto`/`no`/`yes`, cached results and a cache-forced missing semaphore header.
- Project builds and `make check` pass on both platforms. Windows uses `DRIVER_BUILD_LIST=dummy-ups.exe`; the unsuffixed-target failure reproduces on the unchanged baseline. The root-dependent `become_user()` subcheck skips on Windows.
- The changed documentation passes rendering and spelling checks. The release archive includes the new macro, and regeneration and building from the extracted archive pass.

Default `distcheck-light` encounters existing spelling failures in `NEWS.adoc` and `docs/man/apcmicrolink.txt`, reproduced on the unchanged baseline. With only `--enable-spellcheck=no` added to its default light flags, the complete distribution build, tests, staged installation, uninstallation and cleanup pass.

Prepared and validated with assistance from OpenAI Codex, based on GPT-6.


jimklimov:
Thanks! Now adding a layer of autoconf caching support on top of this, in style of #3108 solutions applied earlier to other m4 scripts.

jimklimov (approved):

Issue #2293, opened 2024-01-29
Implement missing WIN32 bits for `nutconf` (library, test and tool)


jimklimov:
Follow up from #2290 and #2283 which provided initial integration of `nutconf` related codebase from 2013 into modern NUT of 2024. 

To complete a milestone in finite time, this effort had to cut some corners - notably about the WIN32 platform that was also recently added to NUT, notably about use of signals and Unix pipes to other processes (something NUT WIN32 builds do with named pipes), as well as some hard-coded assumptions like the temporary file path.

This is considered a low-hanging fruit, because by now NUT has several implementations of functionally equivalent code in this area, the effort would be just to identify and transplant applicable solutions.

PR #3676, opened 2026-09-20
libnutconf: fix character reads from Windows sockets

user01010111:
I reproduced `NutSocket::getChar()` returning `NUTS_ERROR` with `EBADF` on a connected Windows TCP socket while `getString()` read the same payload successfully. I changed the character read to use the existing `sktread()` wrapper, which selects `recv()` on Windows and retains `read()` on POSIX.

I replaced the socket test's fork/signal setup with a connected loopback pair and a pthread reader so it also runs on Windows. The character case checks the complete payload, repeated peeks, consumption and EOF; a separate string case provides a control. I also added a NEWS entry.

This is partial progress on #2293. Process and signal support remain separate work.

I validated the change with C++11 and warnings treated as errors:

- Native Windows 11/MSYS2 MinGW64, GCC 16.2.0: the final regression fails only the character case with the original production source, then passes with the correction. CppUnit reports 36 tests; four existing process/signal bodies remain skipped and both socket bodies execute. The scoped build, `make check` and three concurrent suite runs pass. I used `WINDRES='windres --use-temp-file'` for the resource preprocessor and `DRIVER_BUILD_LIST=dummy-ups.exe` to select the generated Windows driver target.
- Linux, GCC 16.2.1: the scoped build, `make check`, all 36 CppUnit cases and three concurrent suite runs pass.
- `make distcheck-light` passes with nutconf, libnutconf, CppUnit and dummy-ups enabled, using the same C++11/warnings settings and documentation rendering disabled.
- I checked the C++ changes against the coding and portability guide; whitespace, added-line ASCII and project style checks pass. Whole-project spelling passes with a local dictionary copy adding only `kernel's` and `firmware's`, which also fail on unchanged upstream text. The repository dictionary is unchanged.

I used OpenAI Codex (`gpt-6-astra`) to investigate, prepare and test this change.


jimklimov (approved):
Seems to make sense, thanks :)

PR #3681, opened 2026-09-20
Correct source retrieval and configure documentation

user01010111:
The source retrieval examples now use HTTPS. The configure documentation drops the unsupported `--with-ipv6` option and gives `${libdir}/pkgconfig` as the pkg-config installation directory.

The option and generated path were checked against the current build. Read-only checks of the public HTTPS URLs succeeded.


jimklimov (approved):

jimklimov:
Further context:
* https://blog.readthedocs.com/github-git-protocol-deprecation/ (2022) - should use `https://` or `ssh://` URLs, not `git://` (plain and insecure)
* IPv6 is "just supported" by NUT where available, without special `configure` options. The checks did exist in SVN-era code, per commit b22485bac82745afc1459660737c55b0b4cffd91 (2007) - but apparently were removed some time later.

PR #3684, opened 2026-09-20
Correct scanner simulation options and scan examples

user01010111:
The scanner documentation and help now describe both simulation-file extensions and the configuration-directory override. The USB-only scan example no longer shows SNMP results, and the CIDR example now includes the address shown in its output.

The built `nut-scanner` help shows the corrected text. Source checks confirmed extension and directory handling, and an address-membership check confirmed the corrected CIDR. No discovery scan was run.


docs/man/nut-scanner.txt
```diff
@@ -336,11 +331,11 @@ The same using CIDR notation:
         port = "192.168.0.42"
 ----
 
-To scan NUT servers with a timeout of '10' seconds on IP range '192.168.0.0
-to 192.168.0.127' using CIDR notation:
+To scan NUT servers with a timeout of '10' seconds on IP range '192.168.1.0
+to 192.168.1.127' using CIDR notation:
 
 ----
-:; nut-scanner -O -t 10 -m 192.168.0.0/25
```
jimklimov, line 343: Not sure why this change - a `192.168.0.0/25` is a perfectly valid local network range. Might be a bit uncommon, so okay :)

docs/man/nut-scanner.txt
```diff
@@ -309,10 +308,6 @@ To scan USB devices only:
 
 ----
 :; nut-scanner -U
-
-[nutdev-usb1]
-        driver = "snmp-ups"
-        port = "192.168.0.42"
```
jimklimov, line 315: Oops, the example output was wrong. Does not mean the example should have been dropped though, will add a suitable one.

jimklimov (approved):

docs/man/nut-scanner.txt
user01010111: The third octet changed to match the `dummy-test@192.168.1.28` result shown below the command; the original subnet itself is valid. Agreed that the USB output should have been corrected rather than removed—thanks for restoring a suitable example.

jimklimov: Ah, right, that was not seen in original GitHub diff context. Thanks for the correct fix then :)
Also there is a `192.168.0...` example below for IPMI as a valid range, so okay again :)

jimklimov (approved):

PR #3687, opened 2026-09-20
Correct client connection defaults and remote query example

user01010111:
Five client manuals still give an obsolete connection-timeout default. They now document the ten-second built-in default and the environment/command-line override order. The `upsc` loop example also preserves the remote host and port on each query.

Source checks covered all five defaults and their override order. Running the exact loop with an `upsc` stub confirmed that every query keeps the remote host and port.


jimklimov (approved):

jimklimov:
Cross-link: follow-up to #2847

PR #3688, opened 2026-09-20
Clarify client authentication configuration and unattended commands

user01010111:
The `upsmon` overview now uses the correct certificate-setting name and documents both its direct `upsmon.conf` directives and `AUTHCONF` file selector. The `upslog` and `upsstats.cgi` overviews now refer to their configuration manuals, including the `AUTHCONF` selector used by `upsstats.cgi`. The unattended `upscmd`/`upsrw` instructions cover complete authentication-file credentials and the restriction on partial command-line credentials; the `upscmd` diagnostics also name the required `instcmds` permission.

The descriptions were checked against the configuration parsers, selector handling, credential branches and instant-command permission check. The revised `upsmon` page also passes HTML/man rendering, link readback, spelling and ASCII checks. Live authentication and remote commands were not exercised.

The original combined documentation changes passed the project build, man/HTML/PDF generation, spelling and `distcheck-light` on native Linux. The follow-up validation covers the revised `upsmon` paragraph.


docs/man/upsmon.txt
```diff
@@ -142,8 +142,7 @@ SSL/TLS CONFIGURATION
 
 *upsmon* can connect to linkman:upsd[8] servers using SSL/TLS with per-server
 certificate verification and client certificate authentication. SSL/TLS
-configuration is defined in linkman:upsmon.conf[5] using CERTPATH, CERTVERIFY,
-CERTNAME, and CERTHOST directives.
+configuration is described in linkman:upsmon.conf[5].
```
jimklimov: This replacement does not seem right, the directives do exist in `upsmon.conf`. Since recently (#3607), it should also accept an `AUTHCONF /path/name` keyword to specify same nutauth.conf file as other clients.

docs/man/upslog.txt
```diff
@@ -234,8 +234,7 @@ supports per-connection SSL contexts for independent certificate verification
 and client certificate identities. This allows *upslog* to securely log data
 from multiple UPS servers managed by different entities (different CAs or
 security realms). Configuration is specified via linkman:nutauth.conf[5]
-using CERTPATH, CERTVERIFY, CERTNAME, and CERTHOST directives, and is
-automatically applied per-connection.
+and is automatically applied per-connection.
```
jimklimov, line 238: This replacement does not seem right, the directives do exist in `nutauth.conf`, or close to that (`CERTNAME` => `CERTIDENT_NAME`, niche of `CERTHOST` is replaced by section name).

Although I agree in principle that it may be a bit tautological to list them here, and such mention will bit-rot in the future if the list of SSL/TLS keywords would evolve further - as the difference between replaced doc and reality shows already.

Oh well, I guess no action is needed for this one (and the likes of it in non-`upsmon` clients).

jimklimov (changes requested):
I don't fully agree with the change for `upsmon.txt` (as commented there), others look okay - thanks!

docs/man/upsmon.txt
user01010111: The list contained the wrong `CERTNAME` spelling, but removing it also lost useful upsmon-specific guidance. The revised paragraph retains the direct directives with `CERTIDENT` and adds `AUTHCONF` with a link to `nutauth.conf(5)`. The names and file selection match the parser and existing configuration manual; rendering and spelling checks pass.

jimklimov (approved):
Thanks for the update!

PR #3695, opened 2026-09-20
Remove unsupported Powervar quick-test configurability guidance

user01010111:
Neither Powervar driver implements the `ups.conf` setting described for quick-test duration. That instruction is removed from both interface manuals; the existing two-minute duration description is retained.

Option registration and quick-test branches were checked in both drivers. The physical test duration was not retested.


jimklimov:
Thanks! Looking at it, the text was probably copied from `oneac` driver manual when `powervar*` were introduced. The latter is revised in #3699 to spell `testtime` correctly.

PR #3697, opened 2026-09-20
Correct the upsdrvsvcctl command reference

user01010111:
The `upsdrvsvcctl` reference now lists its commands correctly and describes the existing shutdown fallback through `upsdrvctl`. The source TODO that still described this fallback as missing is removed.

The command dispatch and fallback were checked in the source. Both the source and generated scripts passed `sh -n`; no service or shutdown command was run.


jimklimov (approved):

jimklimov:
Scratching my head a decade down the road, what exactly were those comments about? Natively calling `drivername -k -a upsname` from this script?

At least, the `shutdown` passthrough to `upsdrvctl` binary and further to the configured UPS were there since initial PR #330 and possibly since the commit that added the script itself (might be squashed from more internal iterations though, with the comment coming from an older now-lost commit?)...

Issue #3702, opened 2026-09-22
Legrand Keor DK 3k (0665:5161) — HID Power Device not matched by any subdriver, and ups.status undecodable from the hid2nut table


LucVanw:
<h2 dir="ltr">Summary</h2>

<p dir="ltr">The <strong>Legrand Keor DK 3kVA</strong> (same product family as the Daker DK / Daker DK+ of #2730)
ships a Cypress-based communication board. It enumerates as <strong><code>0665:5161</code></strong> — the Cypress
vendor ID, not the Legrand <code>0x1cb0</code> used by <code>legrand-hid.c</code> — while exposing a perfectly
standard HID Power Device report descriptor.</p>

<p dir="ltr">Consequences with current code:</p>

<ul dir="ltr">
<li><code>usbhid-ups</code> rejects it: <code>failed to match a subdriver to vendor and/or product ID</code>.</li>
<li><code>nutdrv_qx</code> and <code>blazer_usb</code> match it (Cypress VID) and claim the interface, but the unit
answers <strong>no</strong> Megatec command at all. It does not speak Megatec over USB.</li>
</ul>

<p dir="ltr">After adding the VID/PID to <code>legrand_usb_device_table[]</code>, the <code>Legrand HID</code> subdriver binds
and the device works — but most values stay unmapped, because the Keor DK uses a <strong>third
HID path layout</strong>, distinct from both the Keor PDU (<code>UPS.Input.*</code> / <code>UPS.Output.*</code>) and the
Keor SP (<code>UPS.PowerConverter.*</code> / <code>UPS.PowerSummary.*</code>).</p>

<p dir="ltr">A PR implementing the device-table entry and the path mapping accompanies this issue. Two
questions are left here for the maintainers, because they cannot be settled from the
<code>hid2nut</code> table alone.</p>

<hr>
<h2 dir="ltr">Device</h2>

<div role="group" aria-label="Code" tabindex="0"><div><div></div></div><div><pre style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono);"><code style="color: rgb(20, 24, 31); background: transparent; font-family: var(--font-mono); white-space: pre-wrap;">Bus 009 Device 002: ID 0665:5161 Cypress Semiconductor USB to Serial
  idVendor           0x0665 Cypress Semiconductor
  idProduct          0x5161 USB to Serial
  iManufacturer      Legrand
  iProduct           Keor DK 3k
  iSerial            Reserve
[...]

PR #3703, opened 2026-09-22
legrand-hid: add Legrand Keor DK (0665:5161) and its HID path layout

LucVanw:
Refs #3702

Some units of the Legrand Keor DK / Daker DK range ship a Cypress-based communication board. They enumerate as 0665:5161 (Cypress vendor ID) while exposing a standard HID Power Device report descriptor, so usbhid-ups matched no subdriver, and nutdrv_qx / blazer_usb had nothing to talk to — these units do not answer Megatec commands over USB.

This PR:

adds the VID/PID to legrand_usb_device_table[], with disable_interrupt_pipe (the unit never sends anything on the interrupt endpoint);
maps the Keor DK's HID path layout, which is a third variant next to the Keor PDU (UPS.Input.* / UPS.Output.*) and the Keor SP (UPS.PowerConverter.* / UPS.PowerSummary.*): readings live under UPS.OutletSystem.*, UPS.Output.* and UPS.BatterySystem.*, with a uniform 1e7 scaling already covered by the existing legrand_times10M_info converter;
adds the HCL entry in data/driver.list.in.

Tested on a Keor DK 3kVA, NUT master (2.8.5.1-0+ga66c009), Debian 13, libusb 1.0.28. Input and output voltages were cross-checked against the unit's front panel display.

Resulting variables: input.voltage, input.voltage.nominal, input.frequency, input.frequency.nominal, input.transfer.high, input.transfer.low, output.voltage, output.frequency, output.current, output.realpower, output.power, ups.power.nominal, ups.realpower.nominal, ups.load, battery.charge, battery.runtime, battery.temperature, plus the existing identification and beeper entries.

Scope — two things deliberately left out

ups.status is still empty. The unit returns only 3 payload bytes for report 0x32 while its descriptor declares 5, and every meaningful flag lives in the third byte, inside an anonymous 24-bit block that has no usage path and so cannot be addressed from the hid2nut table. The bit assignments are measured and documented in the issue (ACPresent, Charging, Discharging, FullyCharged), but wiring them up needs a device-specific read, which I would rather agree on with you first.

battery.voltage and battery.voltage.nominal read wrong on this model (0.8 and 0, instead of 81.9 V and 72 V). UPS.BatterySystem.Battery.Voltage and .ConfigVoltage are already mapped for the Keor SP with a different exponent, and I did not want to change those lines unilaterally and risk a regression on hardware I cannot test. Options are discussed in the issue.

Everything here is additive and touches no existing mapping, so it should not affect the Keor PDU or Keor SP. I only have the one unit, so I have not extrapolated to the rest of the range — the HCL entry is scoped to the Cypress-based comm board.


LucVanw:
Thanks for the NEWS / dict commit — I rebased on top of it rather than overwriting it.

I have pushed a second commit that closes both open questions from #3702 through a
`fix_report_desc` hook, gated on `0665:5161`:

* **`ups.status`**: the named `PresentStatus` flags of report `0x32` are relocated to the
  bits the firmware actually uses (16 / 20 / 21). One more descriptor defect turned up on
  the way: those booleans inherit the global `Unit`/`UnitExp` (seconds, 10⁻²) from report
  `0x30`, so a set flag read as `0.01` and `ups.status` came out as `OB` on mains. Clearing
  `Unit`/`UnitExp` on the relocated items gives `OL CHRG` as expected.
* **`battery.voltage` / `.nominal`**: rather than touching the Keor SP entries, the DK's
  unit exponents are adjusted so the existing entries yield 81.9 V / 72 V.

I have updated the PR description accordingly. Your NEWS line is still accurate; feel
free to extend it with "including report descriptor fix-ups" if you think it is worth
mentioning.

One thing I have **not** re-checked yet: the `OB DISCHRG` transition through the driver
itself. The bit positions come from a live `hidraw` capture, but I will only be on site
tomorrow to pull the mains plug with the patched driver running. I will report back here.

A design question you may prefer to settle: since the uniform 1e7 factor on this unit is
really a wrong `UnitExp` in the descriptor, an alternative would be to correct the
exponent of every DK item in the fix-up and drop the `legrand_times10M_info` /
`legrand_times100M_info` entries from the mapping. It would be cleaner, but a larger
change; I kept the table-based approach for now. Happy to rework it either way.

jimklimov:
```
legrand-hid.c:341:28: error: format specifies type 'unsigned long' but the argument has type 'long' [-Werror,-Wformat]
  339 |                                         "0x%08x offset %u -> %u, Unit 0x%08lx -> 0, UnitExp %d -> 0",
      |                                                                         ~~~~~
      |                                                                         %08lx
  340 |                                         (unsigned int)leaf, (unsigned int)pData->Offset,
  341 |                                         (unsigned int)offset, pData->Unit, pData->UnitExp);
      |                                                               ^~~~~~~~~~~
```

